PR #9, opened 2024-04-23
8 update sampling functions to allow sample by quantile

### LINKED ISSUE #8: increase LSOA sample number
rewrite sampling functions to take more OAs to add to sample. Add mid-level populated OAs from EPC dataset

crispy-wonton:
Fixes #8 

# Description

 Update sampling functions to allow sample by quantile rather than just selecting min/max UPRNs for each country. The user can set the quantiles but default to 0, 0.25, 0.5, 0.75, 1.

# Instructions for Reviewer

Please could you:
- check for bugs/mistakes
- let me know if you know how to set the seed to get the same sample subsets each time, I tried setting it in different ways including setting a global seed in the script with the lines below but it didn't work.
```
import random
random.seed(n)
```
To test, run:
`python asf_heat_pump_suitability/pipeline/get_samples/sample_by_area.py`
This should save two sample files to the S3 bucket in the `outputs` dir. One for lsoa and one for msoa.

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/pipeline/get_samples/sample_by_area.py
+++ b/asf_heat_pump_suitability/pipeline/get_samples/sample_by_area.py
@@ -57,7 +57,7 @@ def get_dicts_quantile_sample_per_nation(
         tuple[list[dict]]: dicts containing sample information for each sample area per nation at specified quantiles
     """
     _count = df.group_by([area, "country_code"]).agg(pl.col("UPRN").count())
-    _count = _count.filter(pl.col("UPRN") >= min_count)
+    _count = _count.filter(pl.col("UPRN") >= min_count).sort(area)
 
     sample_areas = []
     for q in quantiles:
--- a/asf_heat_pump_suitability/pipeline/get_samples/sample_by_area.py
+++ b/asf_heat_pump_suitability/pipeline/get_samples/sample_by_area.py
@@ -9,51 +9,68 @@ def filter_df_area_sample_epc(
     area: str,
     country_codes: list = ["E", "S", "W"],
     min_count: int = 10,
+    quantiles: List[int] = [0, 0.25, 0.5, 0.75, 1],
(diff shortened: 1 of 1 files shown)


asf_heat_pump_suitability/pipeline/get_samples/sample_by_area.py
```diff
@@ -9,51 +9,68 @@ def filter_df_area_sample_epc(
...
         country_codes (list): country codes to sample from
+        min_count (int): minimum count of UPRNs in areas selected
+        quantiles (List[int]): quantiles to sample from. Value 0 <= x <= 1.
 
     Returns
-        tuple[list[dict]]: dicts containing sample information for two sample areas per nation: min and max count of UPRNs.
+        tuple[list[dict]]: dicts containing sample information for each sample area per nation at specified quantiles
     """
     _count = df.group_by([area, "country_code"]).agg(pl.col("UPRN").count())
     _count = _count.filter(pl.col("UPRN") >= min_count)
```
lizgzil, line 61:
```suggestion
_count = _count.sort(area).drop_nulls()
```

lizgzil: this should fix the random sample side of things, and I noticed there was a null MSOA value - perhaps this gets fixed later though

lizgzil (dismissed):
this looks great! the code ran for me!

asf_heat_pump_suitability/pipeline/get_samples/sample_by_area.py
crispy-wonton: thanks @lizgzil ! tested and it works for me too.

Only areas with a non-null country code get sampled. Country code is generated from the area column itself so nulls should never be sampled. I just left the nulls in at this point for that reason as they are excluded later :)

PR #11, opened 2024-04-30
10 test reweighting

### LINKED ISSUE #10: test reweight EPC dataset by area
conduct first pass at reweighting EPC dataset by output area using census-derived validation datasets

crispy-wonton:
Fixes #10 

# Description

Add code to test reweighting sample LSOA data to target marginals from census data for 2 features. Conduct reweighting using Iterative Proportional Fitting (IPF) algorithm, using Python package [_balance_](https://github.com/facebookresearch/balance) by Facebook Research which repackages [_ipfn_](https://github.com/Dirguis/ipfn).

Changes:

- add script from notebook to process sample LSOAs; process target data from census for tenure and property_type features; create dummy data where required; and reweight sample data
- add `notebooks/README` with information about notebook and setup instructions
- add new `base_getters` function to get data from Excel file
- add new data source containing target data for reweighting
- update `.gitignore` and `requirements.txt`

# Instructions for Reviewer

How to run code:
- Follow set up instructions in `asf_heat_pump_suitability/notebooks/README`
- You will need to edit your local copy of `ipfn`: go to `ipfn/ipfn.py` file; change lines 228 and 232.
- Replace the following phrase:
`table_current.loc[feature, weight_col]`
with:
`table_current.loc[feature, [weight_col]]`
-  Here is the [Nesta forked ipfn repo](https://github.com/nestauk/ipfn)- I'm awaiting permissions to push to it or create issues
- Run the `reweight_epc_test.ipynb` notebook
- When you run the notebook and come to the cell where `test_lsoa` variable is set, there are many `test_lsoa` options. The first one has sample size 11, so is a good demonstration of the reweighting on a small sample. The second one is a sample without missing categories and with a good existing representation of features.

Please could you:

- Review code for any bugs/mistakes
In particular, please could you:
- Check that the preprocessing of the target and sample datasets is appropriate
- Check that reweighting has been applied correctly as a method
- Check the logic of dummy row creation
- Double check that the appropriate functions from _balance_ have been used (see [API](https://import-balance.org/docs/api_reference/))

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/config/base.yaml
```diff
@@ -1,6 +1,7 @@
 data_source:
   gb_ons_postcode_dir_url: "https://www.arcgis.com/sharing/rest/content/items/487a5ba62c8b4da08f01eb3c08e304f6/data" # Aug 2023 data
   gb_ons_postcode_dir_file_path: "Data/ONSPD_AUG_2023_UK.csv" # Aug 2023 data
+  EW_housing_characteristics_target_url: "https://www.ons.gov.uk/file?uri=/peoplepopulationandcommunity/housing/datasets/numberofdwellingsbyhousingcharacteristicsinenglandandwales/2021comparedwith2011/hou01tables2021.xlsx" # 2021 census, Mar 2023 update
```
lizgzil: I think we should download and save to S3 all the datasets we use. I'm thinking if the ONS change / remove any of the data/urls it'll be pretty annoying! We can still keep details of when we downloaded them and where from in the READMEs.

requirements.txt
```diff
@@ -1,2 +1,6 @@
 polars
 s3fs>=2023.3.0
+glmnet_python @ git+https://github.com/bbalasub1/glmnet_python.git@1.0
+balance
+jupytext
+fastexcel
```
lizgzil, line 7: I guess we will have to add our `ipfn` fork to this after balance is installed?

asf_heat_pump_suitability/notebooks/reweight_epc_test.py
```diff
@@ -0,0 +1,272 @@
...
+
+# %% [markdown]
+# ## Import validation sets
+
+# %%
+# Load property_type data from census
+property_type_path = config["data_source"]["EW_housing_characteristics_target_url"]
+property_type = base_getters.get_df_from_excel_url(
+    property_type_path, sheet_name="2c", engine="calamine"
+)
```
lizgzil: Would be good to put these lines (and the rest of the validation datasets) in `get_datasets.py` so we can just run e.g. 
```
from asf_heat_pump_suitability.getters.get_datasets import EW_census_housing_characteristics
property_type  = EW_census_housing_characteristics()
```

asf_heat_pump_suitability/notebooks/reweight_epc_test.py
```diff
@@ -0,0 +1,272 @@
...
+# %%
+sample["lsoa"].value_counts().sort(by="count").to_dicts()
+
+# %% [markdown]
+# ## Prepare sample
+
+# %%
+# Filter out NA values
+sample = sample.filter(~pl.col("BUILT_FORM").is_in(["", "unknown"]))
+sample = sample.filter(~pl.col("TENURE").is_in(["", "unknown"]))
```
lizgzil, line 28: Noting that this decreases the data from 11298 rows to 6722. Since this is pretty chunky, we should remember lots of the data isn't being used

asf_heat_pump_suitability/notebooks/reweight_epc_test.py
```diff
@@ -0,0 +1,272 @@
...
+        "Area Code": "lsoa",
+        "Owned or shared ownership": "owner-occupied",
+        "Social Rented": "rental (social)",
+        "Private Rented or lives rent free": "rental (private)",
+    }
+)
+
+# %%
+# Replace censored values ("c") with 0 value and convert cols to int
+censored_vals = 0
```
lizgzil, line 136: is there a reason to replace with 0? Since it means <10, maybe a mid of 5 is better?

asf_heat_pump_suitability/notebooks/reweight_epc_test.py
```diff
@@ -0,0 +1,272 @@
...
+
+# %%
+# Generate artificial target population dataset with proportions derived from census data
+marginals = {
+    "tenure": {k: v / sum(tenure_dict.values()) for k, v in tenure_dict.items()},
+    "property_type": {
+        k: v / sum(property_dict.values()) for k, v in property_dict.items()
+    },
+}
+target_df = balance.weighting_methods.rake.prepare_marginal_dist_for_raking(marginals)
```
lizgzil, line 195: I get an error with this:
`AttributeError: module 'balance' has no attribute 'weighting_methods'`

I have balance version 0.9.0. Do you have a different one? If so, it might be worth pinning your version in the requirements?

lizgzil: ok, it worked if I did:

```
from balance.weighting_methods.rake import prepare_marginal_dist_for_raking
target_df =prepare_marginal_dist_for_raking(marginals)
```

weird!!

asf_heat_pump_suitability/notebooks/reweight_epc_test.py
```diff
@@ -0,0 +1,272 @@
...
+
+# %%
+subset = sample.filter(pl.col("lsoa") == test_lsoa)
+
+# %%
+target_features = {"tenure": tenure_dict, "property_type": property_dict}
+
+
+# %%
+def generate_df_dummies(feature_dicts):
```
lizgzil, line 209: think you'll need `subset` to be an input too going forward

asf_heat_pump_suitability/notebooks/reweight_epc_test.py
```diff
@@ -0,0 +1,272 @@
...
+)
+dummy_rows = dummy_rows.with_columns(
+    pl.Series(name="id", values=[f"dummy_{_}" for _ in range(0, len(dummy_rows))])
+)  # create dummy df
+
+
+# Fill in null values in dummy df
+for feature in ["tenure", "property_type"]:
+    missing_vals = list(dummy_rows.drop_nulls()[feature].unique())
+    if dummy_rows[feature].is_null().sum() > 0:
```
lizgzil:
```suggestion
    if dummy_rows[feature].is_null().any():
```

asf_heat_pump_suitability/notebooks/reweight_epc_test.py
```diff
@@ -0,0 +1,272 @@
...
+        if len(missing_vals) > 0:  # fill with random missing category first
+            dummy_rows = dummy_rows.with_columns(
+                pl.col(feature).fill_null(random.choices(missing_vals, k=1)[0])
+            )
```
lizgzil: I guess this is going to fill in all nulls with the same one random missing category, would be nicer to be all random?

This doesn't quite work, but something like:
```
dummy_rows.with_columns(
    pl.when(pl.col(feature).is_null())
    .then(pl.lit(random.choices(missing_vals,k=len(dummy_rows))))
    .otherwise(pl.col(feature))
    .alias('Prob')
)
```

lizgzil (commented):
This looks great! good job @crispy-wonton . I have a few comments doted about and I reckon the more substantial bits are:
- download the validation datasets and put on S3 - maybe we need a readme somewhere describing the datasets and where and when they were downloaded
- getters for all the validation data from S3 + the v minor processing (getting rid of the top lines + null rows)

The notebook all ran for me and I checked the outputs and processing as I went and have noted the minor concerns. Your use of balance looked good to me.

requirements.txt
lizgzil: our fork is now updated on the master branch too https://github.com/nestauk/ipfn

asf_heat_pump_suitability/notebooks/reweight_epc_test.py
crispy-wonton: very good point

crispy-wonton: maybe it's worth us coming up with a way to fill the unknowns as a later objective for the project to return to. what do you think?

lizgzil (dismissed):

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: great point, have added ONSPD and housing characteristics dataset to S3. I have left the `gb_ons_postcode_dir_url` and `gb_ons_postcode_dir_file_path` in for now as that will need a bigger refactor of the data getter functions. Better to do in a separate PR.

asf_heat_pump_suitability/notebooks/reweight_epc_test.py
crispy-wonton: I chose to set to zero originally to simplify the reweighting. On categories with such low representation in the census, it is less likely they will be represented in EPC so it avoided having to add dummy rows in those cases and I figured it wouldn't drastically change the weights to remove them. One thing I only think of now is that if the group is actually present in EPC and we remove it from the target dataset, we will get an error. 

I can switch it to 5 to see what happens.

asf_heat_pump_suitability/notebooks/reweight_epc_test.py
crispy-wonton: huh, weird. I checked and I also have 0.9.0.
not sure what's going on there but good to have your fix for ref!

crispy-wonton: @helloaidank can you please let me know if you get the same error when you're running this notebook (if you do run it at some point, no need to do it specifically to test this out)?

asf_heat_pump_suitability/notebooks/reweight_epc_test.py
crispy-wonton: Agree. I think I will do this when creating the reweighting pipeline rather than in this PR.

asf_heat_pump_suitability/notebooks/reweight_epc_test.py
crispy-wonton: yep good point!

asf_heat_pump_suitability/notebooks/reweight_epc_test.py
crispy-wonton: this will have to be addressed in a separate issue as i still can't get it to work at the moment

PR #18, opened 2024-05-16
17 prepare epc reweighting

crispy-wonton:
Fixes #17 

# Description

Changes:
- add `pipeline/reweight_epc/reweight_epc.py` with functions to prepare EPC columns for reweighting. Preparation involves standardising columns so that they match the categories of data in the census/CDRC. This is not required for `tenure` feature because the census data has been collapsed to match the EPC categories in the pipeline.
- update `config/base.yaml` with mapping for functions above
- add sample preparation function to `pipeline/enhance_epc/run_script.py`
- fix `save_output` param in `run_script.main` to be optional rather than required

# Instructions for Reviewer

Please could you:
- check for bugs and mistakes
- check the recategorisation of EPC data is correct for each feature
- check columns are added as expected when running code. New columns should be added: [`property_type`, `number_of_rooms`, `build_year`] with the appropriate categories. To run the code (e.g. in a notebook), run the following after pulling the branch (note this df will only contain a sample of LSOAs from EPC, not the whole dataset - this is for speed purposes in testing):
```
from asf_heat_pump_suitability.pipeline.enhance_epc import run_script
df = run_script.main(epc_path="s3://asf-heat-pump-suitability/outputs/epc_sample_lsoa.parquet")
df.head()
```

# Checklist:

- [x] I have refactored my code out from `notebooks/`
- [x] I have checked the code runs
- [ ] I have tested the code
- [x] I have run `pre-commit` and addressed any issues not automatically fixed
- [x] I have merged any new changes from `dev`
- [x] I have documented the code
  - [x] Major functions have docstrings
  - [ ] Appropriate information has been added to `README`s
- [x] I have explained this PR above
- [x] I have requested a code review

PR #26, opened 2024-06-04
12 reweight epc

### LINKED ISSUE #12: add reweighting to pipeline
operationalise reweighting of LSOA samples with Iterative Proportional Fitting and add to processing pipeline:
- create and add dummy rows to EPC dataset for missing features
- conduct reweighting per LSOA

crispy-wonton:
Fixes #12 

# Description

Create basic reweighting pipeline for 3 features: property_type, build_year, and tenure. 

Changes:
- add functions to `reweight_epc.py` sample to: generate dummy rows in sample for missing data; generate balance sample; reweight balance sample
- add new function to `target_getter.py` to get uncensored property_type target dataset. I.e. dataset without censoring small values.
- move existing functions from `reweight_epc.py` for preparing weighting features to new `prepare_sample.py` file
- add weighting process to `enhance_epc/run_script.py` and output run stats.
- add function to round proportions of target data so that proportions total 1 for each LSOA**. 
 
**This is to prevent the script from erroring out with `OutOfMemoryError` when total proportions != 1 for each feature. The error occurs because `balance` attempts to generate an artificial target population that matches the least common multiple (LCM) of the total proportions. E.g. if total proportions are 0.999, 1, and 1.001 this can lead to extremely large pandas dfs generated by `balance`. This is because each feature could generate 999, 1000, and 1001 rows respectively. The LCM is then 999 million rows which causes the error.

## Run script info

The `enhance_epc/run_script.py` script now processes the processed & deduplicated EPC dataset to add LSOA and MSOA columns and then assign weights to each row (where possible) for each LSOA & saves results to S3. A separate file with weighting run time and dropped rows per LSOA is also output and saved to S3. Weights are provided as raw outputs from the `balance` package in the resulting `weights` column and as proportions of 1 in the `proportional_weights` column.

Note, there is data loss throughout the weighting process. The resulting weights are left joined back onto the processed EPC dataset so here we will have rows that have `Null` values in the weights columns. If one property has a null weight, then it could be for any of these reasons:
- the LSOA was skipped because there was no target data found in the target datasets for the given LSOA (e.g. currently all Scottish LSOAs and some England and Wales LSOAS) - in which case all properties will be null for this LSOA
- the row was not assigned an LSOA on the join with the ONSPD data set
- the row did not have sufficient data for weighting - i.e. nulls in any of the features used for weighting
- the row had a category for one of the features that was not present in the target - we are currently dropping these rows before weighting. E.g. for LSOA X: if the EPC data shows 2 rows with 'detached houses' but the target data has 0 recordings of detached houses, the 2 rows will be dropped from the EPC data before weighting.

# Instructions for Reviewer

- The main file with the most new code is `reweight_epc.py`
- The main new function in `prepare_target.py` for review is `_get_dict_proportions_sum_one()` which is the one you wrote to ensure proportions for each feature sum to 1. I've added an additional logic to first identify eligible categories to add the diff to.
- Fyi, the `prepare_sample.py` functions are pretty much all copied directly across and have been reviewed previously in PR #18 - however I did update the `drop_nulls_feature_cols` function to try and reduce hard-coding so please could you check that one. 
- The other changes need review too - the first 2 i mentioned here are the most crucial to the reweighting pipeline

Please could you:
- Review the code and check for general bugs/ mistakes
- please could you check the logic on the key weighting functions and the rounding proportions function
- check the docstrings and in-code documentation makes sense, especially for the key weighting functions and the proportional rounding algorithm
- let me know if you have suggestions for better parameter names (e.g. lsoa_marginals, target_marginals, feature_marginals)
- double check dummy row generation logic. Note: I didnt remove the dummy rows at the end because the weights get left joined to EPC so they should be removed by that process anyway but please let me know if we should remove anyway / if my logic is wrong.
- Please could you test the script runs for you (dont need to run the whole thing but just until the first few LSOAs have been weighted - you should get lots of balance warnings and info messages + vaguely visible tqdm progress bar):
`python asf_heat_pump_suitability/pipeline/enhance_epc/run_script.py --epc_path s3://asf-daps/lakehouse/processed/epc/deduplicated/processed_dedupl-0.parquet`

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/pipeline/reweight_epc/reweight_epc.py
```diff
@@ -1,95 +1,216 @@
...
+                dummy_rows = dummy_rows.with_columns(
+                    pl.col(feature).fill_null(
+                        sample[feature].value_counts().max()[feature][0]
+                    )
+                )
+
+    return dummy_rows
+
+
+def _generate_list_missing_from_sample(
```
crispy-wonton, line 179: This function and the one below are repetitive. I chose to keep them separate because I thought it's less likely to introduce errors from us applying the function incorrectly if we merged into one function.

Let me know if you have a better way that won't have this issue.

asf_heat_pump_suitability/pipeline/reweight_epc/prepare_sample.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    df = (
+        df.with_columns(  # replace "unknown" with null in string feature columns
+            (cs.string() & cs.by_name(features)).replace(
+                {"unknown": None}, return_dtype=pl.String
+            )
+        )
+        .with_columns(
+            pl.col("build_year").fill_null("unknown")
+        )  # replace null with unknown because it's a build_year category
```
lizgzil, line 25: Why do you do something different for build_year? Why is it ok for the unknowns to stay in this category but not in the others?
Also - can build_year contain both None and 'unknown' originally?

asf_heat_pump_suitability/pipeline/reweight_epc/prepare_target.py
```diff
@@ -73,25 +112,60 @@ def convert_df_proportions(df: pl.DataFrame) -> Dict[str, pl.DataFrame]:
...
+    # Make sure rounded proportions add to 1 for each feature
+    for feature_name, feature in lsoa_marginals.items():
+        # Get remainder from rounding & round remainder, otherwise this can be e.g. 0.0010000001
+        diff = round(1 - sum(feature.values()), round_n)
+
+        if diff != 0:
+
+            # If `diff` is negative, i.e. rounded sum of values is >1, then adding `diff` to a very small proportion
+            # can create negative proportions. So here we ensure `diff` can only be added to proportions if they are
+            # larger than -diff. This prevents any proportions being converted to 0 or negative.
```
lizgzil, line 148: nice!

asf_heat_pump_suitability/pipeline/reweight_epc/reweight_epc.py
```diff
@@ -1,95 +1,216 @@
...
-            pl.col("BUILT_FORM") == "Detached",
-        )
-        .then(pl.lit("Detached whole house or bungalow"))
-        .when(
-            pl.col("PROPERTY_TYPE").is_in(["House", "Bungalow"]),
-            pl.col("BUILT_FORM") == "Semi-Detached",
+    # Get EPC sample filtered for single given LSOA
+    cols = ["lsoa", "UPRN"]
+    cols.extend(features)
+    sample = df.select(cols).filter(pl.col("lsoa") == lsoa)
```
lizgzil, line 71: I'm not sure if it's worth checking and changing (I don't think it is at this stage), but FYI that I found in the evaluation that a groupby loop was more efficient than a loop in which you subset the df per lsoa within the loop. i.e.
```
df_select = df.select(cols) # the groupby was also faster with less columns, so worth doing this first too
for lsoa, sample in tqdm(df_select.group_by("lsoa")):
      results = generate_balance_sample(sample, target_marginals, lsoa)
```

asf_heat_pump_suitability/pipeline/reweight_epc/reweight_epc.py
lizgzil: I guess you could just have one function that returns two things?

```
sample_cats = set(sample[feature].unique())
target_cats = {k for k in feature_marginals.keys() if feature_marginals.get(k) > 0}
missing_from_sample_cats = list(target_cats.difference(sample_cats))
missing_from_target_cats = list(sample_cats.difference(target_cats))

return missing_from_sample_cats, missing_from_target_cats
```

lizgzil (dismissed):
This is great 🥇 ! I had some suggestions, but no need to explore them in this PR.

The logic, parameter names and documentation all look great to me 😸  

The code ran for a bit before I cancelled it - so seems to be fine!

asf_heat_pump_suitability/pipeline/reweight_epc/prepare_sample.py
crispy-wonton: We can keep build year (I think) because the target data has an 'unknown' category, whereas for the other features this doesnt exist so we can't keep unknowns. I think the reason this one has 'unknown' in the target but the others dont is that the target data for build year is from CDRC and the others are all census data. Let me know if you think a different approach is better because I don't know if we can really assume that the 'unknown' label is given for the same reason in EPC and the CDRC data. I kept it in to make sure the resulting proportions for the weighting are correct, but maybe we should drop the rows with 'unknown' after weighting. WHat do you think?

To 2nd q: I checked and actually there are no nulls in the original `CONSTRUCTION_AGE_BAND` column in the EPC dataset we're using from which `build_year` is derived so seems like this line does nothing with the current EPC iteration. Do you think we should drop nulls instead of fill them if there ever are nulls?

asf_heat_pump_suitability/pipeline/reweight_epc/reweight_epc.py
crispy-wonton: Thanks Liz, didn't think of this! I've created a separate issue #33 for this so that we can implement and test properly in later PR.

asf_heat_pump_suitability/pipeline/reweight_epc/prepare_sample.py
lizgzil: this all makes sense! I think it's good to keep the weighting pipeline as you have it - it's a shame the others don't have an unknown category in their target datasets. I had forgotten it was due to the target data.

I don't think you should drop nulls if there ever are them... makes sense that these would mean the same thing as 'unknown'?

On removing these rows after the weighting... I can see pros and cons! I think keep them in for now. We could always think about removing these properties later down the line when calculating HP suitability, but they are a minority anyway so kind of think it won't make a difference anyway!

crispy-wonton: Thanks for checking this Liz. I agree that for EPC there's not likely to be a diff between 'unknown' and null values. Now that I think about it, it's possible that in the EPC processing pipeline all nulls get filled with 'unknown' anyway.

asf_heat_pump_suitability/pipeline/reweight_epc/reweight_epc.py
crispy-wonton: Mm I did consider this but if we put our variable names the wrong way round when calling the function, we can get an erroneous run but it will be undetectable as the code won't error out. Instead it would add the wrong dummy rows to sample and drop the wrong feature categories from target. That would be more difficult to detect in the results. I'll keep as is to avoid these errors.

PR #28, opened 2024-06-11
Error analysis pipeline

### LINKED ISSUE #27: Add error analysis to a pipeline

lizgzil:
Fixes #27

---

# Description

In this PR:
- A little refactor of some of Aidan's code [preweight_tenure_error_analysis.py](https://github.com/nestauk/asf_heat_pump_suitability/pull/28/files#diff-7f1d94c4e7a89b5e43b5b3bc0af3b3592f83652a5fa8f311c0a3a9aa904aba89)
- Run error analysis on all LSOAs for all features, before and after weighting
- Makes plots from the results

Fixes # (issue)

# Instructions for Reviewer

- Don't worry about `asf_heat_pump_suitability/getters/s3_getters.py` - it is all copied from previous projects!
- Aidan's already reviewed code has been split into `asf_heat_pump_suitability/pipeline/error_analysis/error_analysis_utils.py` and `asf_heat_pump_suitability/pipeline/error_analysis/sample/preweight_tenure_error_analysis.py`. the only thing that needs review in these is the `get_error_reduction` function in `error_analysis_utils`
- Please check for bugs etc in `asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py`
- Try running:
```
python asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py --sample
```

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/getters/s3_getters.py
+++ b/asf_heat_pump_suitability/getters/s3_getters.py
@@ -17,7 +17,7 @@ import io
 from io import BytesIO
 
 from asf_heat_pump_suitability import logger, PROJECT_DIR
-from typing import List
+from typing import List, Any, NoReturn
 import requests
 
 
@@ -41,31 +41,42 @@ def get_s3_resource():
     return s3
 
 
-def save_to_s3(bucket_name, output_var, output_file_dir):
+def save_to_s3(bucket_name: str, output_var: Any, output_file_path: str) -> NoReturn:
+    """
+    Save a variable (of any type) to S3
+
+    Args:
+        bucket_name (str): Name of the S3 bucket you are saving to
+        output_var (Any): The variable you want to save
+        output_file_path (str): Name of the output path
+
+    Returns:
+        Nothing
+    """
     s3 = get_s3_resource()
 
-    obj = s3.Object(bucket_name, output_file_dir)
+    obj = s3.Object(bucket_name, output_file_path)
 
-    if fnmatch(output_file_dir, "*.csv"):
-        output_var.to_csv("s3://" + bucket_name + "/" + output_file_dir, index=False)
-    elif fnmatch(output_file_dir, "*.parquet"):
+    if fnmatch(output_file_path, "*.csv"):
+        output_var.to_csv("s3://" + bucket_name + "/" + output_file_path, index=False)
+    elif fnmatch(output_file_path, "*.parquet"):
         output_var.to_parquet(
-            "s3://" + bucket_name + "/" + output_file_dir, index=False
+            "s3://" + bucket_name + "/" + output_file_path, index=False
         )
-    elif fnmatch(output_file_dir, "*.pkl") or fnmatch(output_file_dir, "*.pickle"):
+    elif fnmatch(output_file_path, "*.pkl") or fnmatch(output_file_path, "*.pickle"):
         obj.put(Body=pickle.dumps(output_var))
-    elif fnmatch(output_file_dir, "*.gz"):
+    elif fnmatch(output_file_path, "*.gz"):
         obj.put(Body=gzip.compress(json.dumps(output_var).encode()))
-    elif fnmatch(output_file_dir, "*.txt"):
+    elif fnmatch(output_file_path, "*.txt"):
         obj.put(Body=output_var)
     elif (
-        fnmatch(output_file_dir, "*.jpg")
-        or fnmatch(output_file_dir, "*.png")
-        or fnmatch(output_file_dir, "*.jpeg")
+        fnmatch(output_file_path, "*.jpg")
+        or fnmatch(output_file_path, "*.png")
+        or fnmatch(output_file_path, "*.jpeg")
     ):
         image_data = BytesIO(output_var)
         obj.put(Body=image_data)
-    elif fnmatch(output_file_dir, "*.json"):
+    elif fnmatch(output_file_path, "*.json"):
         obj.put(Body=json.dumps(output_var, cls=CustomJsonEncoder))
     else:
         logger.error(
@@ -73,12 +84,19 @@ def save_to_s3(bucket_name, output_var, output_file_dir):
         )
 
 
-def load_s3_data(bucket_name, file_name):
+def load_s3_data(
+    bucket_name: str,
+    file_name: str,
+) -> Any:
     """
     Load data from S3 location.
 
-    bucket_name: The S3 bucket name
-    file_name: S3 key to load
+    Args:
+        bucket_name (str): Name of the S3 bucket you are saving to
+        file_name (str): S3 key to load
+
+    Returns:
+        (Any): The file you have loaded
     """
     s3 = get_s3_resource()
 
@@ -121,14 +139,20 @@ def load_s3_data(bucket_name, file_name):
         )
 
 
-def dictionary_to_s3(data_dict: dict, s3_bucket: str, s3_folder: str, file_name: str):
+def dictionary_to_s3(
+    data_dict: dict, s3_bucket: str, s3_folder: str, file_name: str
+) -> NoReturn:
     """
     Transforms a dictionary into a json and uploads to S3.
+
     Args:
-        data_dict: dictionary with the data
-        s3_bucket: S3 bucket name where to upload the file
-        s3_folder: folder where to store the file within the S3 bucket
-        file_name: name of the file
+        data_dict (dict): Dictionary to be saved.
+        s3_bucket (str): S3 bucket name where to upload the file
+        s3_folder (str): folder where to store the file within the S3 bucket
+        file_name (str): name of the file
+
+    Returns:
+        Nothing
     """
     s3_client = boto3.client("s3")
     obj = io.BytesIO(json.dumps(data_dict).encode("utf-8"))
@@ -140,10 +164,11 @@ def read_json_from_s3(bucket: str, file_path: str) -> dict:
     Reads a json file from S3 without downloading it.
 
     Args:
-        bucket: S3 bucket name
-        file_path: file path (including file name)
+        bucket (str): S3 bucket name
+        file_path (str): file path (including file name)
+
     Returns:
-        dictionary with json file data
+        (dict): The file you have loaded
     """
     s3_resource = boto3.resource("s3")
     json_file = s3_resource.Object(bucket, file_path)
--- a/asf_heat_pump_suitability/getters/s3_getters.py
+++ b/asf_heat_pump_suitability/getters/s3_getters.py
@@ -1,0 +1,151 @@
+"""
+Data getters (and savers)
+"""
+
+from fnmatch import fnmatch
+import json
+import pickle
+import gzip
+import os
+
+import pandas as pd
+import boto3
+from decimal import Decimal
+import numpy
+import yaml
+import io
+from io import BytesIO
+
+from asf_heat_pump_suitability import logger, PROJECT_DIR
+from typing import List
+import requests
+
+
+class CustomJsonEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, Decimal):
+            return float(obj)
+        elif isinstance(obj, numpy.integer):
+            return int(obj)
+        elif isinstance(obj, numpy.floating):
+            return float(obj)
+        elif isinstance(obj, numpy.ndarray):
+            return obj.tolist()
+        elif isinstance(obj, set):
+            return list(obj)
+        return super(CustomJsonEncoder, self).default(obj)
+
+
+def get_s3_resource():
+    s3 = boto3.resource("s3")
+    return s3
+
+
+def save_to_s3(bucket_name, output_var, output_file_dir):
+    s3 = get_s3_resource()
+
+    obj = s3.Object(bucket_name, output_file_dir)
+
+    if fnmatch(output_file_dir, "*.csv"):
+        output_var.to_csv("s3://" + bucket_name + "/" + output_file_dir, index=False)
+    elif fnmatch(output_file_dir, "*.parquet"):
+        output_var.to_parquet(
+            "s3://" + bucket_name + "/" + output_file_dir, index=False
+        )
+    elif fnmatch(output_file_dir, "*.pkl") or fnmatch(output_file_dir, "*.pickle"):
+        obj.put(Body=pickle.dumps(output_var))
+    elif fnmatch(output_file_dir, "*.gz"):
+        obj.put(Body=gzip.compress(json.dumps(output_var).encode()))
+    elif fnmatch(output_file_dir, "*.txt"):
+        obj.put(Body=output_var)
+    elif (
+        fnmatch(output_file_dir, "*.jpg")
+        or fnmatch(output_file_dir, "*.png")
+        or fnmatch(output_file_dir, "*.jpeg")
+    ):
+        image_data = BytesIO(output_var)
+        obj.put(Body=image_data)
+    elif fnmatch(output_file_dir, "*.json"):
+        obj.put(Body=json.dumps(output_var, cls=CustomJsonEncoder))
+    else:
+        logger.error(
+            'Function not supported for file type other than "*.csv", "*.parquet", "*.jsonl.gz", "*.jsonl", "*.json", "*.png", "*.jpeg".'
+        )
+
+
+def load_s3_data(bucket_name, file_name):
+    """
+    Load data from S3 location.
+
+    bucket_name: The S3 bucket name
+    file_name: S3 key to load
+    """
+    s3 = get_s3_resource()
+
+    obj = s3.Object(bucket_name, file_name)
+    if fnmatch(file_name, "*.jsonl.gz"):
+        with gzip.GzipFile(fileobj=obj.get()["Body"]) as file:
+            return [json.loads(line) for line in file]
+    if fnmatch(file_name, "*.yml") or fnmatch(file_name, "*.yaml"):
+        file = obj.get()["Body"].read().decode()
+        return yaml.safe_load(file)
+    elif fnmatch(file_name, "*.jsonl"):
+        file = obj.get()["Body"].read().decode()
+        return [json.loads(line) for line in file]
+    elif fnmatch(file_name, "*.json.gz"):
+        with gzip.GzipFile(fileobj=obj.get()["Body"]) as file:
+            return json.load(file)
+    elif fnmatch(file_name, "*.json"):
+        file = obj.get()["Body"].read().decode()
+        return json.loads(file)
+    elif fnmatch(file_name, "*.csv"):
+        return pd.read_csv("s3://" + bucket_name + "/" + file_name)
+    elif fnmatch(file_name, "*.parquet"):
+        return pd.read_parquet("s3://" + bucket_name + "/" + file_name)
+    elif fnmatch(file_name, "*.pkl") or fnmatch(file_name, "*.pickle"):
+        file = obj.get()["Body"].read().decode()
+        return pickle.loads(file)
+    elif (
+        fnmatch(file_name, "*.jpg")
+        or fnmatch(file_name, "*.png")
+        or fnmatch(file_name, "*.jpeg")
+    ):
+        # Download the image from S3 into a BytesIO object
+        image_data = BytesIO()
+        obj.download_fileobj(image_data)
+        return image_data
+
+    else:
+        logger.error(
+            'Function not supported for file type other than "*.csv", "*.parquet", "*.jsonl.gz", "*.jsonl", or "*.json"'
+        )
+
+
+def dictionary_to_s3(data_dict: dict, s3_bucket: str, s3_folder: str, file_name: str):
+    """
+    Transforms a dictionary into a json and uploads to S3.
+    Args:
+        data_dict: dictionary with the data
+        s3_bucket: S3 bucket name where to upload the file
+        s3_folder: folder where to store the file within the S3 bucket
+        file_name: name of the file
+    """
+    s3_client = boto3.client("s3")
+    obj = io.BytesIO(json.dumps(data_dict).encode("utf-8"))
+    s3_client.upload_fileobj(obj, s3_bucket, os.path.join(s3_folder, file_name))
+
+
+def read_json_from_s3(bucket: str, file_path: str) -> dict:
+    """
+    Reads a json file from S3 without downloading it.
+
+    Args:
+        bucket: S3 bucket name
+        file_path: file path (including file name)
+    Returns:
+        dictionary with json file data
+    """
+    s3_resource = boto3.resource("s3")
+    json_file = s3_resource.Object(bucket, file_path)
+    json_file = json_file.get()["Body"].read().decode("utf-8")
+    return json.loads(json_file)
--- a/asf_heat_pump_suitability/pipeline/error_analysis/error_analysis_utils.py
+++ b/asf_heat_pump_suitability/pipeline/error_analysis/error_analysis_utils.py
@@ -1,86 +1,31 @@
 """
-This script performs an error analysis on the tenure data by LSOA (Lower Layer Super Output Area).
-
-First, it loads the sample and target data. Then, it processes all LSOAs to get error metrics and differences in counts and proportions. These metrics are saved to a CSV file.
-
-Next, it creates and melts dataframes for counts and proportions differences, extracts values from these melted dataframes, and converts them to pandas dataframes.
-
-Finally, it creates boxplots and difference plots for counts and proportions differences. These plots provide a visual representation of the differences in counts and proportions by tenure for each LSOA.
-
-This script is intended to be run as a standalone script.
+Functions needed to calculate errors in the proportions of features per LSOA
 """
 
 import polars as pl
 import s3fs
-from asf_heat_pump_suitability import config
 import matplotlib.pyplot as plt
 import seaborn as sns
 import pandas as pd
 import math
 import numpy as np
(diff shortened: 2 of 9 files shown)


asf_heat_pump_suitability/getters/s3_getters.py
crispy-wonton, line 1: Some of these are missing docstrings and type hinting. They look really useful :)

asf_heat_pump_suitability/pipeline/error_analysis/error_analysis_utils.py
```diff
@@ -185,14 +108,85 @@ def process_single_lsoa(
...
+
+def get_error_reduction(
+    before_proportions: dict,
+    after_proportions: dict,
+    target_proportions: dict,
+) -> dict:
+    """
+    Find the average error reduction between two sets of proportions and the target
+
+    """
```
crispy-wonton: Missing args and returns. What are the keys and values in each dict? Is this used per feature or will one dict contain all the features?

Also, it would be really helpful to briefly explain what the output metric represents and the range of possible values (is it inf: -inf?). 

Please can you confirm if my interpretation is correct for my understanding:
- a negative error_reduction = weighting improved proportions closer to target
- positive error_reduction means weighting made the error worse
- more negative numbers mean there was greater improvement and lower remaining error
- negative numbers closer to zero mean there was less improvement and higher remaining error

Also, what happens if all the error was corrected and there is no remaining error? It looks like those instances aren't recorded - should they be?

asf_heat_pump_suitability/pipeline/error_analysis/error_analysis_utils.py
```diff
@@ -185,14 +108,85 @@ def process_single_lsoa(
...
+    error_reduction = {}
+    for key in before_proportions:
+        after_diff = after_proportions.get(key, 0) - target_proportions.get(key, 0)
+        if after_diff != 0:
+            error_reduction[key] = (
+                after_proportions.get(key, 0) - before_proportions.get(key, 0)
+            ) / after_diff
+
+    if error_reduction:
+        return np.mean(list(error_reduction.values()))
```
crispy-wonton: since this function isnt returning the `error_reduction` dict, I think `error_reduction` can just be a list

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+
+Run with:
+
+python asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
+	--reweighted_dir "s3://asf-heat-pump-suitability/outputs/2023_Q2_EPC_enhanced_weights.parquet"
+	--sample
+
+[remove the --sample argument to run on full dataset]
+
+"""
```
crispy-wonton: This is helpful

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+from asf_heat_pump_suitability import logger
+
+main_error_metrics = [
+    "rmse_all_cats",
+    "mae_all_cats",
+    "rmse_missing_cats",
+    "mae_missing_cats",
+]
+
+evaluation_feature_names = ["tenure", "property_type", "build_year", "nrooms"]
```
crispy-wonton: can these variables go in if name = main and be passed as args to functions

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+def parse_arguments():
+
+    parser = ArgumentParser()
+
+    parser.add_argument(
+        "--reweighted_dir",
```
crispy-wonton:
```suggestion
        "--reweighted_path",
```

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+    epc_subset: pl.DataFrame,
+) -> dict:
+    """
+    Get errors between target and unweighted, and target and weighted, for one feature and one LSOA.
+
+    Args:
+                    feature_name (str): The feature name to calculate errors for
+                    lsoa_code (str): The LSOA
+                    target_features (dict): A nested dictionary of the counts for all property features and LSOAs
+                    target_marginals (dict): A nested dictionary of the proportions for all property features and LSOAs
```
crispy-wonton: Does this mean input is like: 
`{"LSOA_code": {"feature1a": count, "feature1b": count, "feature2a": count}}` 
etc.?

Also is the only diff between target features and target marginals the fact that it's counts v proportions?

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+                    dict: A dictionary containing various metrics for both the unweighted and weighted EPC data compared to
+                                    the target dataset for this feature.
+
+    """
+
+    target_counts = target_features[feature_name].get(lsoa_code)
+    target_proportions = target_marginals[feature_name].get(lsoa_code)
+
+    if target_counts:
+        orig_counts_list = epc_subset[feature_name].value_counts().to_dict()
```
crispy-wonton:
```suggestion
        orig_counts_dict = epc_subset[feature_name].value_counts().to_dict()
```
I got confused by the name on line 95 :D

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+
+"""
+
+import polars as pl
+import polars.selectors as cs
+from tqdm import tqdm
+
+from argparse import ArgumentParser
+
+from asf_heat_pump_suitability.pipeline.error_analysis.error_analysis_utils import *
```
crispy-wonton: I would import `error_analysis_utils` and then call functions from the module with `error_analysis_utils.function() `where required as I find it easier to locate origins of code that way. Same with modules below

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+        orig_counts = dict(
+            zip(orig_counts_list[feature_name], orig_counts_list["count"])
+        )
+        orig_proportions = calculate_proportions(orig_counts)
+        error_metrics_orig = get_error_metrics(
+            orig_counts, orig_proportions, target_counts, target_proportions
+        )
+        if not all(epc_subset["weight"].is_null()):
+            # Get the sum of the weights for each feature category (not technically a 'count')
+            reweighted_counts_list = (
```
crispy-wonton:
```suggestion
            reweighted_counts_dict = (
```

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+                epc_subset.group_by(feature_name).agg(pl.col("weight").sum()).to_dict()
+            )
+            reweighted_counts = dict(
+                zip(
+                    reweighted_counts_list[feature_name],
+                    reweighted_counts_list["weight"],
+                )
+            )
```
crispy-wonton: Fyi, I think the counts will appear very wrong in the error_metrics_reweighted because this 'weight' col is based on an artificial population of 1000, so it's basically just the proportion * 1000. Did they look very off to you when you ran it? 

However, when we come to microsimulation, we will produce actual counts so we will need this functionality. We will probably just have to update the 'weight' name in this script. I dont think you need to change it now I'm just saying for info.

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+                reweighted_proportions,
+                target_counts,
+                target_proportions,
+            )
+
+            average_error_reduction = get_error_reduction(
+                orig_proportions, reweighted_proportions, target_proportions
+            )
+
+            prop_reweighted_counts_list = (
```
crispy-wonton:
```suggestion
            prop_reweighted_counts_dict = (
```

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+                .agg(pl.col("proportional_weight").sum())
+                .to_dict()
+            )
+            prop_reweighted_counts = dict(
+                zip(
+                    prop_reweighted_counts_list[feature_name],
+                    prop_reweighted_counts_list["proportional_weight"],
+                )
+            )
+            prop_reweighted_proportions = calculate_proportions(prop_reweighted_counts)
```
crispy-wonton: I dont think you need to calculate proportions as they are already proportions. If this line is here because the proportions dont add up to 1, it would be because we have removed the dummy rows so you'll get it totalling <1 in some cases. 

This would actually change the weights in some cases.

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+            reweighted_counts_list = (
+                epc_subset.group_by(feature_name).agg(pl.col("weight").sum()).to_dict()
+            )
+            reweighted_counts = dict(
+                zip(
+                    reweighted_counts_list[feature_name],
+                    reweighted_counts_list["weight"],
+                )
+            )
+            reweighted_proportions = calculate_proportions(reweighted_counts)
```
crispy-wonton: I think we need to join the dummy rows to the EPC dataset instead of drop them (what we've currently done) because we really need them for the 'reweighted' part of this analysis considering the way this script is written - specifically the `calculate_proportions` function.

This function calculates proportions based on ‘totals’, but these totals exclude the weights assigned to dummy rows. So for some of the LSOAs where dummy rows had to be added, we are actually changing the weights and then doing the error analysis on changed weights. This will have a bigger effect on LSOAs in EPC which were missing a category that has a high proportion in the target data. E.g. if a dummy row was given a fractional weight of 0.25 then this calculation in line 112 will calculate the proportions based on a total of `1 - 0.25`.

To fix this, we would have to do an outer join of the reweighted data onto EPC rather than left join and then re-run the reweighting pipeline and this error analysis.

Fixing this also means we will get a more accurate avg error reduction metric because the avg gets slightly skewed when we remove the dummy weights from the calculation.

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+        help="Whether to test this script on a sample or not",
+        default=False,
+        action="store_true",
+    )
+    args = parser.parse_args()
+
+    return args
+
+
+def get_errors_for_lsoa(
```
crispy-wonton, line 58: I think worth adding some notes within the function about what each section is doing :)

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+
+    final_metrics = {}
+    for metric in main_error_metrics:
+        final_metrics[metric] = {
+            "unweight": error_metrics_orig.get(metric, None),
+            "reweight": error_metrics_reweighted.get(metric, None),
+            "reweight-prop": error_metrics_prop_reweighted.get(metric, None),
+            "error_reduction": average_error_reduction,
+        }
+    return final_metrics
```
crispy-wonton: Just thinking about debugging this if we got an unexpected output later, you could have info logging statements if all weights are null or no target data is found and only return `final_metrics` dict if there actually are any metrics.

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+def load_epc_reweights(
```
crispy-wonton, line 153: You can put the cleaning function in `pipeline/enhance_epc/prepare_epc.py` :)

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+            name="nrooms", values=reweighted_epc_df["nrooms"].replace(None, "unknown")
+        )
+    )
+    return reweighted_epc_df
+
+
+def filter_epc(reweighted_epc_df: pl.DataFrame, target_features: dict) -> pl.DataFrame:
+    """
+    Filter the EPC dataset to just what is needed for calculating errors on.
+    This helps with processing time.
```
crispy-wonton: SMART

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+    # Bit of cleaning to allign with target data
+    reweighted_epc_df = reweighted_epc_df.rename({"NUMBER_HABITABLE_ROOMS": "nrooms"})
+    reweighted_epc_df = reweighted_epc_df.with_columns(
+        pl.col("nrooms").map_elements(
+            lambda x: "9+" if x >= 9 else str(x), return_dtype=pl.String
+        )
+    )
+    reweighted_epc_df = reweighted_epc_df.with_columns(
+        pl.Series(
+            name="nrooms", values=reweighted_epc_df["nrooms"].replace(None, "unknown")
```
crispy-wonton: What's the reason for this line im curious?

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+    parser.add_argument(
+        "--save_output",
+        help="S3 path to save evaluation results to",
+        type=str,
+        default=None,
+    )
+
+    parser.add_argument(
+        "--sample",
+        help="Whether to test this script on a sample or not",
```
crispy-wonton:
```suggestion
        help="Whether to test this script on a non-random sample or not",
```

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+        reweighted_epc_df_with_target = reweighted_epc_df_with_target.filter(
+            pl.col("lsoa").is_in(lsoas)
+        )
+
+    logger.info(
+        f"Evaluating for {reweighted_epc_df_with_target['lsoa'].n_unique()} LSOAs"
+    )
+
+    full_results = {}
+    for lsoa_code, epc_subset in tqdm(reweighted_epc_df_with_target.group_by("lsoa")):
```
crispy-wonton, line 246: oh cool I didnt know you could return a groupby into 2 variables this way!

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,273 @@
...
+        full_results[lsoa_code] = feature_results
+
+        # Save to S3
+
+    if not args.save_output:
+        name = args.reweighted_dir.replace("s3://asf-heat-pump-suitability/", "")
+        name = f"{name.split('.parquet')[0]}_evaluation.json"
+        if args.sample:
+            name = f"{name.split('.json')[0]}_sample.json"
+        args.save_output = name
```
crispy-wonton: nice!

crispy-wonton (commented):
Hi @lizgzil ,

Thanks so much for building these functions into a script and adding the s3 getters, really helpful!

I have left some comments, mostly on code documentation. The main important comments are those on `evaluate_reweighting.py` about the way proportions are calculated and used, given that we've removed dummy rows. I think the way the proportions for error analysis are currently calculated will actually change the weights for LSOAs where dummy rows were added. Please check my comments to see if you agree, and see the simple fix about joining weights to the EPC dataset (in the [reweighting pipeline](https://github.com/nestauk/asf_heat_pump_suitability/blob/dev/asf_heat_pump_suitability/pipeline/enhance_epc/run_script.py)) with an outer rather than left join.

Lastly, I didn't get a chance to test running the code yet but I wanted to get these comments through as soon as I discovered it. I will do that before this is merged.

Thanks so much! Let me know if you have questions :)

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
lizgzil: good idea!

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
lizgzil: good thinking

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
lizgzil: will do that! I don't know why I even did that because I also dislike `import *` for the same reason!

asf_heat_pump_suitability/pipeline/error_analysis/error_analysis_utils.py
lizgzil: so true!

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
lizgzil: not quite - it looks like `{'tenure': {'E01022833': {'owner-occupied': 548, 'rental (social)': 53}, 'E01013414': {}}, 'property_type': {}}` - I'll add it to the docstring.

Yes - thats right it is just the proportions vs counts

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
lizgzil: I think because there was an 'unknown' category in the target dataset

asf_heat_pump_suitability/pipeline/error_analysis/error_analysis_utils.py
lizgzil: Great points of clarification. In considering these, I realised that the calculation is incorrect anyway.

I have changed it and gave a much more detailed docstring with the answers to the questions you've posed here.

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
lizgzil: I don't actually ever use the counts in the error metrics outputted other than to calculate the proportion of each feature category, so I'm not sure if it matters to even change the 'weight' name going forward? May have misinterpreted this though?

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
lizgzil: I think this was kept in the check that using either of 'weights' or 'proportional_weight' gave the same evaluation scores. But I think you are right and it is redundant to have it in - I will remove

lizgzil:
Thanks so much @crispy-wonton - your comments were extremely useful ! I've addressed your comments in this commit https://github.com/nestauk/asf_heat_pump_suitability/pull/28/commits/e75407168e71d3633c3ac0048929fed003571be0

I haven't rerun the run_script.py with the dummy rows included. I will try tomorrow and then rerun the evaluation.

lizgzil:
@crispy-wonton update:
- I have reran run_script.py with the outer join. There are 16788545 rows in this output. When the data is loaded we get this though:

Screenshot 2024-07-12 at 09 58 30

Which isn't what I expected (that row where UPRN= `10024356313.0` but UPRN_right = null?). I'm not quite sure if this matters, but it may represent a problem?

lizgzil:
latest evaluation is also run with changes to the script as in this PR + inclusion of dummy rows.
New results are better for property type than before.

The reweighting produces better Tenure proportions MAE 92.53% of the time
The reweighting produces the same Tenure proportions MAE 0.1% of the time
The reweighting produces better Property Type proportions MAE 93.5% of the time
The reweighting produces the same Property Type proportions MAE 0.17% of the time
The reweighting produces better Build Year proportions MAE 86.19% of the time
The reweighting produces the same Build Year proportions MAE 1.55% of the time

crispy-wonton:
> @crispy-wonton update:
> 
> * I have reran run_script.py with the outer join. There are 16788545 rows in this output. When the data is loaded we get this though:
> 
> Screenshot 2024-07-12 at 09 58 30
> 
> Which isn't what I expected (that row where UPRN= `10024356313.0` but UPRN_right = null?). I'm not quite sure if this matters, but it may represent a problem?

I think this is due to the fact that in preparation for reweighting, we have to drop some EPC rows because of our reweighting exclusion criteria (any nulls in any reweighting feature ; or a category in EPC that isn't present in target data) or the LSOA is not present in the target data so the LSOA didnt get reweighted. In this case, I checked and can see that none of the properties in that LSOA received a weight. Therefore I think it's the latter reason although I didn't check the target data specifically to see if it was missing that LSOA.

asf_heat_pump_suitability/pipeline/error_analysis/error_analysis_utils.py
```diff
@@ -185,14 +128,105 @@ def process_single_lsoa(
...
+    Returns:
+        np.float64: The average relative error reduction before and after reweighting across each category for a feature in an LSOA.
+            This value is between -inf and 1
+
+        Interpreting the output:
+            - A negative value means weighting made the errors worse.
+            - A positive value means weighting improved the results.
+            - A very negative value (-inf) means weighting was much worse than no weighting.
+            - A value close to 1 means weighting made the results much better.
+            - A value of zero mean the errors with or without weighting were the same.
```
crispy-wonton, line 164: This docstring is perfect!

asf_heat_pump_suitability/pipeline/error_analysis/error_analysis_utils.py
```diff
@@ -185,14 +128,105 @@ def process_single_lsoa(
...
+        )
+        error_reduction = before_diff - after_diff
+        if before_diff == 0:
+            before_diff = 0.0000001  # To avoid division by 0
+        # The percentage reduction in error when switching between before and after weighting
+        error_reduction_list.append(error_reduction / before_diff)
+    if error_reduction_list:
+        return np.mean(error_reduction_list)
+    else:
+        return None
```
crispy-wonton, line 183: I think this is a good logic and works really well!

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,271 @@
...
+    ],
+) -> dict:
+    """
+    Get errors between target and unweighted, and target and weighted, for one feature and one LSOA.
+
+    Args:
+                    feature_name (str): The feature name to calculate errors for
+                    lsoa_code (str): The LSOA
+                    target_features (dict): A nested dictionary of the counts for all property features and LSOAs
+                        For example, {'tenure': {'E01022833': {'owner-occupied': 548, 'rental (social)': 53}, 'E01013414': {}}, 'property_type': {}}
```
crispy-wonton, line 78: This example is really helpful! I think I should go back and add these type of things to the reweighting pipeline code that i wrote

lizgzil:
@crispy-wonton I just saw your comment - although I don't think you were concerned. So am I ok to merge this?

asf_heat_pump_suitability/pipeline/reweight_epc/evaluate_reweighting.py
```diff
@@ -0,0 +1,271 @@
...
+                                    the target dataset for this feature.
+
+    """
+
+    # Find target values for this feature and LSOA
+    target_counts = target_features[feature_name].get(lsoa_code)
+    target_proportions = target_marginals[feature_name].get(lsoa_code)
+
+    if target_counts:
+        # If we have target information then calculate the errors
```
crispy-wonton, line 95: Just to check because I can't remember - will LSOAs with no target have an empty dict or will they be missing from `target_features` and `target_marginals` completely?

crispy-wonton (commented):
The updates look great - I think the error reduction calculation function is really clear now and all looks good. Thank you!

crispy-wonton:
> @crispy-wonton I just saw your comment - although I don't think you were concerned. So am I ok to merge this?

Yes I don't think it's a problem because I think it's an expected behaviour of the pipeline. I believe `UPRN` comes from the original EPC dataset and `UPRN_right` comes from the reweighted EPC rows. So any rows that were excluded from the reweighting because of our reweighting criteria would not be present in the reweighted EPC dataset, therefore they will appear as null UPRNs in the outer join.

I think this is good to merge :)

Issue #40, opened 2024-07-17
Estimate garden size for EPC properties


crispy-wonton:
Estimate garden size using Microsoft building footprint files and INSPIRE land extent files.

PR #41, opened 2024-07-17
36 add garden size

### LINKED ISSUE #36: Get file boundaries for all land parcel files
INSPIRE polygons showing the extent of registered properties in England and Wales can be found on the [gov UK website](https://use-land-property-data.service.gov.uk/datasets/inspire/download).

These land parcel polygons are organised into files by council. To speed up search when matching UPRN and building footprint to land parcel, we need to find the bounding box of each file.

crispy-wonton:
Fixes #36 and #40

# Description

Estimate garden size / outdoor area for each EPC property where possible using INSPIRE land extent files and Microsoft building footprint files. The script will save a separate file of all EPC properties with available garden size estimates which can then be joined to the full EPC on UPRN.

Steps:
1. Match building footprint and land extent files
2. Find intersections of building footprint and land extent polygons
3. Calculate garden size (m2)
4. Garden size to EPC data

Summary of changes:

Key files with geospatial manipulations:
- Add `asf_heat_pump_suitability/pipeline/prepare_features/building_footprint.py` containing functions to prepare Microsoft building footprint geodataframes for spatial joins
- Add `asf_heat_pump_suitability/pipeline/prepare_features/land_extent.py` containing functions to prepare INSPIRE land extent geodataframes for spatial joins
- Add `asf_heat_pump_suitability/pipeline/prepare_features/garden_size.py` containing functions to find intersection of land polygons and building polygons to calculate outdoor space / garden size
- Add `asf_heat_pump_suitability/pipeline/run_scripts/calculate_garden_size.py` containing run script to estimate garden size per property

Additional files:
- Add `save_utils.py` and `geo_utils.py` with generic util functions
- Add data sources to `config/README.md`
- Update `asf_heat_pump_suitability/getters/base_getters.py` and `asf_heat_pump_suitability/getters/get_datasets.py` with new geo-getters to load geospatial data
- Add new function to `asf_heat_pump_suitability/pipeline/prepare_features/lat_lon.py` to generate point geometries from x, y coordinates
- Update `config/base.yaml` with new data sources
- Update `schemas.py` with new schemas

# Instructions for Reviewer

In order to test the code in this PR you need to run:
`python -i asf_heat_pump_suitability/pipeline/run_scripts/calculate_garden_size.py --use_mapping s3://asf-heat-pump-suitability/source_data/2023_land_parcels_with_file_polygons_sample_N3.geojson --save_epc_gardens s3://asf-heat-pump-suitability/outputs/2023_Q2_EPC_garden_size_estimates_N3.parquet --epc_path s3://asf-heat-pump-suitability/outputs/2023_Q2_EPC_enhanced_weights_and_features.parquet`

This will run the script on a sample of 3 land extent : building footprint file matches, rather than running through all the matches as this takes 16+ hours. You can then check the outputs are as expected in your terminal in interactive python.

Please could you:
- pay special attention to all geospatial joins and any coordinate conversions to ensure they are achieving the expected result and done correctly. The coordinate conversions and spatial joins were the things I found most difficult.
- let me know if there are better / more efficient and, especially, less error-prone ways to do any of the geospatial (or other) manipulations. I think there might be some built-in functionalities in geopandas that i could use instead for some of the conversions but I couldnt find them - especially for converting bounds to a polygon?
- feel free to respond to comments marked #TODO if you have ideas/answers :)
- check docstrings are clear and correct, especially with respect to coordinate reference systems
- check for general mistakes / bugs

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/config/README.md
crispy-wonton, line 1: To review: click three dots in right hand corner of file and select 'view file'

asf_heat_pump_suitability/utils/geo_utils.py
```diff
@@ -0,0 +1,65 @@
+import logging
+import geopandas as gpd
+import shapely
+
+
+def transform_gdf_drop_duplicates(gdf: gpd.GeoDataFrame) -> gpd.GeoDataFrame:
```
crispy-wonton, line 6: The reason we don't just drop duplicate geometries here and instead drop duplicate representative points is because there were some geometries in the files which were nearly identical but not exactly the same. We want to drop identical and almost identical overlapping land parcels/ building footprints.

See comment below - this assumes identical polygons will always generate the same representative points. Is this correct?

asf_heat_pump_suitability/pipeline/prepare_features/building_footprint.py
```diff
@@ -0,0 +1,152 @@
...
+    ]
+
+    gdf = gpd.GeoDataFrame(df, geometry="geometry", crs="EPSG:27700").rename(
+        columns={"Url": "ms_url"}
+    )
+
+    return gdf
+
+
+def convert_quadkey_to_bounds(quadkey: str, transformer: pyproj.Transformer) -> tuple:
```
crispy-wonton: This function and the one below need special attention in review to check conversions are correct and lat / lon used correctly.

asf_heat_pump_suitability/config/README.md
crispy-wonton, line 1: To review: click three dots in right hand corner of file and select 'view file'

requirements.txt
```diff
@@ -6,4 +6,11 @@ balance
 jupytext
 fastexcel
 tqdm
-sklearn
\ No newline at end of file
+sklearn
```
danlewis85: This should be `scikit-learn`, `pip` throws an error with `sklearn` as this is outdated.

danlewis85: Also needs `regex`

danlewis85: Also needs `pyogrio`

danlewis85:
**A note on Microsoft Buildings data format**

The Microsoft data is in a format called geojson lines (or geojson sequence). This is a version of the geojson format that is useful in cloud and distributed/parallel processing contexts. The standard geojson format has a single root, making it hard to split into multiple process, geojson sequence is essentially a sequence of individual geojsons (one for each feature) making it simple to parallelise.

However, the geojson lines here have been saved arbitrarily as a csv, even though it's just a geojson followed by a newline for as many features as exist. This means that geopandas/pyogrio/gdal incorrectly reads the format when trying to do it automatically. However, with a bit of gdal magic you can read the file directly in one line.

`filepath = 'https://minedbuildings.blob.core.windows.net/global-buildings/2024-05-28/global-buildings.geojsonl/RegionName%3DUnitedKingdom/quadkey%3D031311231/part-00128-f7c7c3eb-e50e-4937-b0e3-071fe0699cfb.c000.csv.gz'`

With your method (takes ~46s for me)
`data = load_gdf_csv_gz(filepath)`

With Fiona (slower ~55s for me):
`data = geopandas.read_file(f"GeoJSONSeq:/vsigzip//vsicurl/{filepath}", engine='fiona')`

With pyogrio (faster ~32s for me):
`data = geopandas.read_file(f"GeoJSONSeq:/vsigzip//vsicurl/{filepath}", engine='pyogrio')`

NB the `use_arrow` flag for pyogrio made no difference on speed for me.

**What's going on?**
`<driver>:` prefix forces gdal/ogr to use a specific driver, avoiding automatically using the csv driver.
`vsi...` is gdal's [virtual file system notation](https://gdal.org/user/virtual_file_systems.html), this means gdal can directly handle the gzip compression (vsigzip) and the source (vsicurl).

I don't think there's a case for changing what you're currently doing, in terms of using requests to grab the file, ungzipping and loading the data, but it's always nice to know you don't have to.

However, if you think that data loading is a big bottleneck, then it might be worth looking at pyogrio directly, I note it can also load directly to an arrow table, which might be faster still that the geopandas approach above.

sqr00t:
@danlewis85 I wonder if you've installed OpenMP `brew install libomp`. PyArrow uses OpenMP to manage multithreading contexts (parallelising things). Furthermore, setting an environment variable `OMP_NUM_THREADS=8` for an M1 Pro Macbook (has 8 cores) would make sure that all cores are being used, but the default if unset is to use what the system allows, accounting for other running tasks. Without `libomp` the behaviour should revert to single threaded loading.

danlewis85:
Nice one @sqr00t - I installed openmp using conda:

Running with arrow took my system ~8s, a huge difference.
`data = geopandas.read_file(f"GeoJSONSeq:/vsigzip//vsicurl/{filepath}", engine='pyogrio', use_arrow=True)`

asf_heat_pump_suitability/pipeline/prepare_features/land_extent.py
```diff
@@ -0,0 +1,163 @@
+import pandas as pd
+import geopandas as gpd
+import regex as re
+from typing import List
+from asf_heat_pump_suitability import config
```
danlewis85, line 7: You're importing a module object without anything in it (except yamls) - this means the code will fail at line 36 (module objects are not subscriptable).

Instead, you need to load the config yaml file.

asf_heat_pump_suitability/getters/get_datasets.py
```diff
@@ -1,4 +1,7 @@
 import polars as pl
+import pandas as pd
+import geopandas as gpd
+import logging
 from asf_heat_pump_suitability import config
```
danlewis85, line 5: This appears to be an empty module (with yaml files in) so you can't import it, rather you need to load the yaml from this location.

asf_heat_pump_suitability/pipeline/prepare_features/land_extent.py
crispy-wonton: This should work because [asf_heat_pump_suitability/asf_heat_pump_suitability/__init__.py](https://github.com/nestauk/asf_heat_pump_suitability/blob/dev/asf_heat_pump_suitability/__init__.py) returns config/base.yaml as a dict called `config` when you import 'config' which can then be subscripted. It's part of the DS cookiecutter I think?

asf_heat_pump_suitability/getters/get_datasets.py
danlewis85: Apologies, I must have broken this somehow. I restarted my kernel and it works fine!

asf_heat_pump_suitability/pipeline/prepare_features/land_extent.py
danlewis85: Apologies, I must have broken this somehow. I restarted my kernel and it works fine!

danlewis85:
**Converting bounds to a polygon**

Note that shapely has a `box` class which takes: xmin, ymin, xmax, ymax. This is slightly easier than using the `Polygon` class to make a box.

This enables:
``` python
from shapely import box

box(*data.total_bounds)
```

danlewis85:
**land_extent.generate_gdf_map_file_to_bounds()**

I think this function would benefit from a small change. It's definitely a good thing to match data in the first instance on name, as otherwise the process would take a long time. However, I think you should 1. treat nulls and any matches > 1 as null, and 2. use spatial relationships to fill those nulls.

Here is so example code on how I would see this working:
```python
def generate_gdf_map_file_to_bounds(
    land_extent_location: str = None,
    ladnm_col: str = "LAD23NM",
    keep_cols: list = ["LAD23NM", "LAD23CD", "geometry"],
    save_as: str = None,
) -> gpd.GeoDataFrame:

    council_bounds = transform_gdf_council_bounds(ladnm_col, keep_cols)
    land_extent_files = base_getters.list_files_s3_location(land_extent_location)

    matches = _match_list_file_to_name(
        land_extent_files=land_extent_files,
        council_names=council_bounds["council_name_std"],
    )

    # NEW!! Single out matches
    matches = [None if not match else match if len(match) == 1 else None for match in matches]

    matches = pd.DataFrame(
        {"inspire_file_name": land_extent_files, "council_bounds_matches": matches}
    ).explode("council_bounds_matches")

    file_to_bounds = matches.merge(
        council_bounds,
        how="left",
        left_on="council_bounds_matches",
        right_on="council_name_std",
    )

    file_to_bounds = gpd.GeoDataFrame(
        file_to_bounds, crs="EPSG:27700", geometry="geometry"
    )
    # UPDATED!!!
    file_to_bounds = fill_nulls_file_bounds(file_to_bounds, council_bounds)

    if save_as:
        file_to_bounds.to_file(save_as, crs="EPSG:27700")

    return file_to_bounds
```
Where the `fill_nulls_file_bounds()` function now looks something like:

```python
def fill_nulls_file_bounds(gdf: gpd.GeoDataFrame, boundaries: gpd.GeoDataFrame) -> gpd.GeoDataFrame:
    """
    Fill file bounds polygons for INSPIRE land extent files with no file polygons.

    Args:
        gdf (gpd.GeoDataFrame): GeoDataFrame of INSPIRE land extent files and file polygons
        boundaries (gpd.GeoDataFrame): GeoDataFrame of boundaries that provide mapping.

    Returns:
        gpd.GeoDataFrame: land extent files with file polygons
    """
    missing_bbox = gdf[gdf["council_bounds_matches"].isnull()][
        "inspire_file_name"
    ].to_list()

 # NB move hard coded column names to function arguments.
    for file in missing_bbox:
        # Load data
        land_parcels_gdf = get_datasets.load_gdf_inspire_land_parcels(f"s3://{file}")
        # Get the majority boundary name for a sample of polygon centres
        candidate_nm = (geopandas.sjoin(land_parcels_gdf.sample(500).centroid.to_frame('geometry'),
                                    boundaries)['LAD23NM']
                                    .value_counts()
                                    .index[0])
        # Select county that 'contains' centroid
        candidate = boundaries.loc[boundaries['LAD23NM']==candidate_nm]
        # update file_to_bounds
        gdf.loc[gdf['inspire_file_name']==file,
                ['LAD23NM', 'LAD23CD', 'geometry', 'council_name_std']] = candidate.to_numpy()[0]
        print(f"file no. {missing_bbox.index(file)+1} of {len(missing_bbox)}")

    return gdf
```
This means that you can assign the actual boundary geometries and identifiers.

This is a bit time consuming though as it appears there is no way to speed up the read on gml files. Using `pyogio` just reading the info takes the same time as reading the full file.

This may fail if the underlying boundary is a weird shape, so a more robust (but computationally expensive) approach could be to randomly sample a few plots, join them to the boundaries and then take the modal boundary.

I've updated the code to use the sampling approach, it doesn't seem to take much long and it effectively deals with a couple of edge cases (e.g. where a larger district is oddly shaped due to the existence of a city council).

danlewis85:
**Replace pyproj WGS84 -> BNG Projections to convertbng in building_footprint.py**

In asf_heat_pump_suitability/pipeline/prepare_features/building_footprint.py renew code to use convertbng over pyproj for accuracy.

requirements.txt
crispy-wonton: Ah great thank you for flagging all these!

I thought regex is a built-in module with python. Did you need to install it separately?

crispy-wonton:
> **Converting bounds to a polygon**
> 
> Note that shapely has a `box` class which takes: xmin, ymin, xmax, ymax. This is slightly easier than using the `Polygon` class to make a box.
> 
> This enables:
> 
> ```python
> from shapely import box
> 
> box(*data.total_bounds)
> ```

Nice! I was looking for something like this but couldn't find it. Thank you!

crispy-wonton:
> **land_extent.generate_gdf_map_file_to_bounds()**
> 
> I think this function would benefit from a small change. It's definitely a good thing to match data in the first instance on name, as otherwise the process would take a long time. However, I think you should 1. treat nulls and any matches > 1 as null, and 2. use spatial relationships to fill those nulls.
> 
> Here is so example code on how I would see this working:
> 
> ```python
> def generate_gdf_map_file_to_bounds(
>     land_extent_location: str = None,
>     ladnm_col: str = "LAD23NM",
>     keep_cols: list = ["LAD23NM", "LAD23CD", "geometry"],
>     save_as: str = None,
> ) -> gpd.GeoDataFrame:
> 
>     council_bounds = transform_gdf_council_bounds(ladnm_col, keep_cols)
>     land_extent_files = base_getters.list_files_s3_location(land_extent_location)
> 
>     matches = _match_list_file_to_name(
>         land_extent_files=land_extent_files,
>         council_names=council_bounds["council_name_std"],
>     )
> 
>     # NEW!! Single out matches
>     matches = [None if not match else match if len(match) == 1 else None for match in matches]
> 
>     matches = pd.DataFrame(
>         {"inspire_file_name": land_extent_files, "council_bounds_matches": matches}
>     ).explode("council_bounds_matches")
> 
>     file_to_bounds = matches.merge(
>         council_bounds,
>         how="left",
>         left_on="council_bounds_matches",
>         right_on="council_name_std",
>     )
> 
>     file_to_bounds = gpd.GeoDataFrame(
>         file_to_bounds, crs="EPSG:27700", geometry="geometry"
>     )
>     # UPDATED!!!
>     file_to_bounds = fill_nulls_file_bounds(file_to_bounds, council_bounds)
> 
>     if save_as:
>         file_to_bounds.to_file(save_as, crs="EPSG:27700")
> 
>     return file_to_bounds
> ```
> 
> Where the `fill_nulls_file_bounds()` function now looks something like:
> 
> ```python
> def fill_nulls_file_bounds(gdf: gpd.GeoDataFrame, boundaries: gpd.GeoDataFrame) -> gpd.GeoDataFrame:
>     """
>     Fill file bounds polygons for INSPIRE land extent files with no file polygons.
> 
>     Args:
>         gdf (gpd.GeoDataFrame): GeoDataFrame of INSPIRE land extent files and file polygons
>         boundaries (gpd.GeoDataFrame): GeoDataFrame of boundaries that provide mapping.
> 
>     Returns:
>         gpd.GeoDataFrame: land extent files with file polygons
>     """
>     missing_bbox = gdf[gdf["council_bounds_matches"].isnull()][
>         "inspire_file_name"
>     ].to_list()
> 
>     # NB move hard coded column names to function arguments.
    for file in missing_bbox:
        # Load data
        land_parcels_gdf = get_datasets.load_gdf_inspire_land_parcels(f"s3://{file}")
        # Get the majority boundary name for a sample of polygon centres
        candidate_nm = (geopandas.sjoin(land_parcels_gdf.sample(500).centroid.to_frame('geometry'),
                                    boundaries)['LAD23NM']
                                    .value_counts()
                                    .index[0])
        # Select county that 'contains' centroid
        candidate = boundaries.loc[boundaries['LAD23NM']==candidate_nm]
        # update file_to_bounds
        gdf.loc[gdf['inspire_file_name']==file,
                ['LAD23NM', 'LAD23CD', 'geometry', 'council_name_std']] = candidate.to_numpy()[0]
        print(f"file no. {missing_bbox.index(file)+1} of {len(missing_bbox)}")
> 
>     return gdf
> ```
> 
> This means that you can assign the actual boundary geometries and identifiers.
> 
> This is a bit time consuming though as it appears there is no way to speed up the read on gml files. Using `pyogio` just reading the info takes the same time as reading the full file.
> 
> This may fail if the underlying boundary is a weird shape, so a more robust (but computationally expensive) approach could be to randomly sample a few plots, join them to the boundaries and then take the modal boundary.

Thanks @danlewis85 ! Great point to set those with multiple matches to null and fill them. That will definitely save some loading time later.

In terms of being time consuming - the previous fill_nulls method had to load the files as well so it shouldnt be slower than it is currently, but with the above note about setting multiple matches to null, there will be a few more files to pair (although I think not many).

Instead of using centroid, I think I can use representative_point. This should still allow matches for boundaries that have an unusual/multipolygon shape from my understanding - rather than sampling. What do you think?

danlewis85:
> Instead of using centroid, I think I can use representative_point. This should still allow matches for boundaries that have an unusual/multipolygon shape from my understanding - rather than sampling. What do you think?

After testing, I changed this to use the centroid of the centroids of the land registry polygons. This is essentially the centre of mass of the data. It seemed in a couple of cases that using the centre of the bounding box was too crude given the underlying boundaries. I've updated the code snippet to reflect this.

I don't think representative point will make a significant difference compared to centroid, in the case of a box they're likely to be very similar.

After further testing, I've implemented the sampling-based approach and updated the code snippet. This now seems to work in all cases.

asf_heat_pump_suitability/pipeline/prepare_features/garden_size.py
```diff
@@ -0,0 +1,125 @@
...
+    Returns:
+        Mapping: mapping where keys are INSPIRE land parcel file names and values are Microsoft building footprint file
+        names
+    """
+    logging.info("Mapping building footprint files to INSPIRE land registry files")
+    gdf = gpd.overlay(land_files_gdf, building_files_gdf, how="intersection")
+    gdf = gdf.sort_values(by="inspire_file_name")
+    file_dict = pd.Series(
+        gdf["ms_url"].values, index=gdf["inspire_file_name"]
+    ).to_dict()
```
danlewis85: **This is a bug.**

The overlay operation produces a many-many relationship - a land registry dataset can intersect with 1 or more building datasets and vice versa. This means the index in your series is non-unique. When you cast to a dictionary, dictionaries cannot have non-unique keys. So this process silently gets rid of >50% of your matches (dict is length 316, while match Series object is length 711.

This means you're ultimately processing a lot less data than you should be.

danlewis85: The fix is easy though, you don't need to cast the Series to a dictionary as the Series (for your purposes) will behave like a dictionary anyway. e.g.
```python
 for land_file, building_file in tqdm(file_matches.items()):
    # do something
```
Should work just fine as a series as `.items()` on a Series produces a generator of index:value pairs.

asf_heat_pump_suitability/pipeline/run_scripts/calculate_garden_size.py
```diff
@@ -0,0 +1,140 @@
...
+    prev = None
+    total_gardens = 0
+    for land_file, building_file in tqdm(file_matches.items()):
+
+        # Only load land extent gdf if we haven't loaded already
+        if land_file != prev:
+            # Prepare land parcel data
+            land_parcels_gdf = land_extent.transform_gdf_land_parcels(
+                f"s3://{land_file}"
+            )  # TODO is there a cleaner way to generate the file path here without f-string
```
danlewis85: I don't really see an issue with the f string, but the alternative would be to build it using `pathlib`

asf_heat_pump_suitability/pipeline/prepare_features/garden_size.py
```diff
@@ -0,0 +1,125 @@
...
+    Args:
+        land_files_gdf (gpd.GeoDataFrame): INSPIRE land file names and file bounding polygons
+        building_files_gdf (gpd.GeoDataFrame): Microsoft building footprint file names and file bounding polygons
+
+    Returns:
+        Mapping: mapping where keys are INSPIRE land parcel file names and values are Microsoft building footprint file
+        names
+    """
+    logging.info("Mapping building footprint files to INSPIRE land registry files")
+    gdf = gpd.overlay(land_files_gdf, building_files_gdf, how="intersection")
```
danlewis85: This should be:
```python
gdf = gpd.sjoin(land_files_gdf, building_files_gdf, predicate="intersects")
```
It produces the same result as overlay much more cheaply.

Basically, spatial joins (sjoins) join tables on the basis of a spatial relationship, but they don't change any of the actual geometries. The overlay operation does this, plus it create a new geometry based on the spatial operation.

So, sjoin will return matching records for an intersection and the _original_ geometries. Whereas, overlay will return matching records for an intersection and a new geometry that describe that intersection only - like with sets.

In this case you only want to understand the spatial relationships, not the actual intersection, so overlay is overkill.

asf_heat_pump_suitability/pipeline/run_scripts/calculate_garden_size.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # Check where building footprint files and land extent files overlap
+    file_matches = garden_size.match_dict_files_land_building(
+        land_files_gdf=land_file_bounds, building_files_gdf=microsoft_file_bounds
+    )
+
+    epc_gardens = []
+    prev = None
+    total_gardens = 0
+    for land_file, building_file in tqdm(file_matches.items()):
```
danlewis85, line 91: This loop could probably be abstracted to a pipeline function rather than being part of the script.

asf_heat_pump_suitability/utils/geo_utils.py
danlewis85: Yes, I traced where representative point comes from (geopandas representative Point -> shapely point on surface -> geos interior point area) and found a description of the algorithm: https://libgeos.org/doxygen/classgeos_1_1algorithm_1_1InteriorPointArea.html
My reading of this is that it is deterministic and you should get the same result every time for the same input polygon.

asf_heat_pump_suitability/utils/geo_utils.py
```diff
@@ -0,0 +1,65 @@
...
+    Drop polygons with the same representative point. This drops both duplicate and nearly identical geometries.
+
+    Args:
+        gdf (gpd.GeoDataFrame): GeoDataFrame with polygon geometries
+
+    Returns:
+        gpd.GeoDataFrame: GeoDataFrame with duplicate polygon geometries dropped
+    """
+    # TODO: this assumes duplicate polygons will always generate the same representative points. Is this correct?
+    gdf["rep_point"] = gdf.representative_point()
```
danlewis85: There's a margin speed up if you cast this to a well known binary instead of leaving it as a shapely point object:
```python
gdf["rep_point"] = gdf.representative_point().to_wkb()
```

asf_heat_pump_suitability/utils/geo_utils.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Get bounding polygon of GeoDataFrame.
+
+    Args:
+        gdf (gpd.GeoDataFrame): GeoDataFrame
+
+    Returns:
+        shapely.Polygon: bounding polygon of GeoDataFrame
+    """
+    gdf_bounds = gdf["geometry"].total_bounds
```
danlewis85: Can replace with:
```python
return shapely.box(*gdf.total_bounds)
```

asf_heat_pump_suitability/pipeline/prepare_features/garden_size.py
```diff
@@ -0,0 +1,125 @@
...
+                >= gdf["min_size_of_small_building"]
+            )
+        )
+        | (
+            (gdf["building_intersection_area_m2"] > outbuilding_size)
+            & (
+                gdf["building_intersection_area_m2"]
+                >= gdf["min_size_of_large_building"]
+            )
+        )
```
danlewis85, line 78: I'm not 100% confident in this logic as I find it quite confusing to make sense of.

To be included as a valid intersection between a building polygon and land parcel, the following conditions must be met, either:

The candidate intersection is less than or equal to the maximum out building size and greater than or equal to 45% of the original building size.

or,

The candidate intersection is larger than the maximum size of an outbuilding and greater than or equal to 5% of the original building size.

I think this makes sense, but I think it'd be worthwhile revisiting once you've adjusted the building footprint coordinate conversions to make sure it's still performing as you expect.

requirements.txt
danlewis85: Ignore `regex`, that seems to have been because I failed to import properly when testing.

danlewis85 (changes requested):
This is an impressive PR with a lot of work in it. My recommendations are as follows:

1. Replace any ONS/OS geojsons (e.g. local authority areas) with geopackages (or another format of your choice). This will ensure you get authoritative data in British National Grid format. The geojson standard requires data to be in WGS84, hence why they're best avoided for this use case.
2. Update code to use `convertbng` instead of `pyproj` for converting building polygons to BNG to ensure greater accuracy.
3. Update code for faster loading of building footprints as per mine and @sqr00t recommendations (gdal drivers, virtual files systems, arrow and openmp).
    - Unfortunately this won't apply to loading land registry gml files, so they can stay as is.
4. Revise the `generate_gdf_map_file_to_bounds()` function as suggested to better fill nas.
    - the working approach I've outlined uses a sample of plot centroids to find the modal boundary intersected. This should be robust to edge cases unlike earlier attempts.
5. Fix bug in `garden_size.match_dict_files_land_building()` where casting to a dictionary loses information.
6. Having made changes, check logic for building-plot intersections still makes sense.

There are a few additional suggests in the comments that may also be worth your attention, but aren't super important.

Happy to re-review a revised version of this PR.

asf_heat_pump_suitability/utils/geo_utils.py
crispy-wonton: Do you mean this whole function can be replaced with that one line?

asf_heat_pump_suitability/pipeline/prepare_features/garden_size.py
crispy-wonton: Thanks so much @danlewis85 for catching this! I had expected the many to many relationship but I didn't think it through that turning to a dict would lose values because of requiring unique keys!

This also explains a behaviour that I saw but didn't understand in the pipeline earlier. I had set the pipeline to only load a new land polygon file if it was different from the previous one, but had noticed it was actually loading a file in every iteration of the loop. Now I understand it's because the dict I feed into the loop only contains unique land files.

asf_heat_pump_suitability/pipeline/run_scripts/calculate_garden_size.py
crispy-wonton: The reason I left this as a script is because when we run it in interactive mode with `python -i`, we can more easily address errors if they come up without having to re-run the whole pipeline (as it took 24hr+ before the suggested speed improvements were implemented). I'll keep as is for now until we're sure it's robust and/or the pipeline is significantly sped up.

asf_heat_pump_suitability/pipeline/prepare_features/building_footprint.py
```diff
@@ -0,0 +1,161 @@
...
+    Args:
+        geos (gpd.GeoSeries): shapely polygons in CRS WGS84
+
+    Returns:
+        gpd.GeoSeries: shapely polygons in CRS EPSG:27700 (British National Grid)
+    """
+    coords = geos.get_coordinates()
+    coords["x"], coords["y"] = convert_bng(coords["x"], coords["y"])
+    # TODO: conversion back to polygons is rate-limiting step
+    s = coords.groupby(coords.index).apply(lambda l: shapely.Polygon(zip(l.x, l.y)))
```
crispy-wonton: any ideas on speeding this up?

asf_heat_pump_suitability/pipeline/prepare_features/building_footprint.py
```diff
@@ -0,0 +1,161 @@
...
+    df = df[df["Location"] == "UnitedKingdom"]
+
+    transformer = _set_crs_transformer()
+
+    # Use convertbng to convert QuadKeys where possible, otherwise use pyproj transformer
+    df["geometry"] = [
+        (
+            polygons
+            if (polygons := convertbng_quadkey_to_polygon(qk))
+            else convertpyproj_quadkey_to_polygon(qk, transformer)
```
danlewis85, line 30: Love this use of the walrus operator.

crispy-wonton: heheh i knew you would :D

asf_heat_pump_suitability/pipeline/prepare_features/building_footprint.py
danlewis85: import `numpy` and replace lines 143 and 144 with:
```
splits = numpy.split(coords.to_numpy(), coords.index.value_counts().sort_index().to_numpy().cumsum())[:-1]
return gpd.GeoSeries([shapely.Polygon(pnts) for pnts in splits]).set_crs(epsg=27700)
```

Basically, the optimisation I hit on was splitting the x,y coords by their index, rather than using the expensive groupby.

To get the split indices: `coords.index` is the index from `get_coordinates()`, `.value_counts()` is the number of vertices per building shape, `.sort_index()` returns it to index order (rather than ascending on count) `.to_numpy()` gets the numpy array representation and `.cumsum()` produces the cumulative sum of the geometry lengths which are the points at which to split the list of coords to recreate the individual building polygons. Finally, you need to index to `-1` as otherwise you get a spare empty array on the end.

danlewis85 (dismissed):
This looks good. I've tested and it's working as expected.

Have suggested a simple speed up for the coordinate conversion.

Happy for you to push the speed up and a couple of cosmetic things we discussed (wrapping warnings and retry for loading from s3). Then for you to merge.

This one's done!

PR #55, opened 2024-08-09
52 add wales data for listed building feature

### LINKED ISSUE #52: Add Wales data for listed buildings
We have only added the Historic England data for the listed buildings and have not included the Wales dataset for listed buildings.

helloaidank:
---

# Description

This is a PR which adds the listed buildings from the Wales dataset and merges them with the dataset from England and then adds them as a feature to the EPC dataset. This is done via spatial joining.
Unfortunately I had some issues with previous PR (listed buildings PR 48) therefore changes suggested from that have been implemented in this PR (therefore please ignore functions such as `spatial_join_epc_with_listed_buildings` in `listed_buildings.py`.  I guess something to think about is surrounding whether there is a better way to do the merging of nations datasets, as currently it doubles the time with regards to the spatial join when running over the whole dataset. 

Fixes #52 

# Instructions for Reviewer
I've made changes to / added the following files:
- `pipeline/run_scripts/add_features.py`
- `pipeline/prepare_features/listed_buildings.py`
- `getters/get_datasets.py`
- `getters/base_getters.py`
- `config/base.yaml`
- `config/README.md`
- `requirements.txt`
- `environment.yaml`

In order to test the code in this PR you need to run the following:
`python pipeline/run_scripts/add_features.py --epc_path "s3://asf-heat-pump-suitability/outputs/2023_Q2_EPC_enhanced_weights_sample_10k.parquet" --save_output "s3://asf-heat-pump-suitability/outputs/2023_Q2_EPC_enhanced_weights_listed_buildings_sample_test_10k.parquet"`, this will take a few minutes.

Please pay special attention to any bugs, sense test the functions and any tips on how we could make this more optimised in terms of the merging of the listed buildings features for Wales and England and then have a quick look over the outputs. 

# Checklist:

- [X] I have checked the code runs
- [ ] I have tested the code
- [X] I have run `pre-commit` and addressed any issues not automatically fixed
- [X] I have merged any new changes from `dev`
- [X] I have documented the code
  - [X] Major functions have docstrings
  - [X] Appropriate information has been added to `README`s
- [X] I have explained this PR above
- [X] I have requested a code review


asf_heat_pump_suitability/getters/base_getters.py
```diff
@@ -107,6 +107,20 @@ def get_content_from_s3_path(path: str) -> bytes:
...
+    Load GeoDataFrame from Geopackage on S3.
+    Args:
+        s3_uri (str): URI to S3 Geopackage
+    Returns:
+        gpd.GeoDataFrame
+    """
+    fs = s3fs.S3FileSystem()
+    with fs.open(s3_uri, "rb") as f:
+        gdf = gpd.read_file(f)
+    return gdf
```
crispy-wonton: have you tested just using `gpd.read_file()` without s3fs file system? it might work but not sure

asf_heat_pump_suitability/getters/get_datasets.py
```diff
@@ -131,13 +131,15 @@ def get_df_spa_offgasgrid() -> pl.DataFrame:
     return df
 
 
-def get_df_historicengland_listedbuildings() -> pl.DataFrame:
+def get_df_listedbuildings(data_source_key: str) -> pl.DataFrame:
     """
-    Get raw Historic England 'Listed Buildings' dataset.
+    Get raw 'Listed Buildings' dataset.
```
crispy-wonton: Add crs of resulting dataset to docstring :)

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,56 +29,95 @@ def transform_df_EPC_X_and_Y_to_point(
     return gdf
 
 
-def get_filtered_df_listed_buildings() -> gpd.GeoDataFrame:
+def get_filtered_df_listed_buildings(data_source_key: str) -> gpd.GeoDataFrame:
     """
     Get and filter out listed buildings from Historic England dataset.
```
crispy-wonton: What do you mean by filter here? It looks like it loads the data and then just select relevant cols.

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,56 +29,95 @@ def transform_df_EPC_X_and_Y_to_point(
...
 def spatial_join_epc_with_listed_buildings(
-    enhanced_epc_df: pl.DataFrame, listed_buildings_df: gpd.GeoDataFrame
-) -> gpd.GeoDataFrame:
+    enhanced_epc_df: pl.DataFrame,
+    listed_buildings_df: gpd.GeoDataFrame,
+    chunk_size: int = 100000,
+) -> pl.DataFrame:
     """
     Spatial join EPC dataset with listed buildings dataset.
     Args:
```
crispy-wonton, line 59: Missing chunk_size in args

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,56 +29,95 @@ def transform_df_EPC_X_and_Y_to_point(
...
         enhanced_epc_df (pl.DataFrame): Enhanced EPC dataset
         listed_buildings_df (gpd.GeoDataFrame): Filtered Historic England dataset with only listed buildings grade and geometry.
     Returns:
-        gpd.GeoDataFrame: EPC dataset with listed buildings grade without geometry.
+        pl.DataFrame: EPC dataset with listed buildings grade without geometry.
```
crispy-wonton:
```suggestion
        pl.DataFrame: EPC dataset with listed buildings grade
```

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,56 +29,95 @@ def transform_df_EPC_X_and_Y_to_point(
...
     """
-    epc_gdf = transform_df_EPC_X_and_Y_to_point(enhanced_epc_df)
-    epc_gdf_temp = epc_gdf[["geometry"]].copy()
-    epc_gdf_temp["index"] = epc_gdf.index
-    joined_gdf = gpd.sjoin(
-        epc_gdf_temp, listed_buildings_df, how="left", predicate="intersects"
+
+    # Add a unique index column for merging later
+    enhanced_epc_df = enhanced_epc_df.with_row_index("index")
+    partitions = enhanced_epc_df.with_row_count("chunk_id").select(
```
crispy-wonton: I didnt know what `with_row_count` did so I looked it up in the docs. [Polars API](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.with_row_count.html) says it's deprecated and to use `with_row_index` instead so maybe you could try replacing it. Not sure if it will work for your purpose but worth testing.

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,56 +29,95 @@ def transform_df_EPC_X_and_Y_to_point(
...
-    epc_gdf["index"] = epc_gdf.index
-    result_gdf = epc_gdf.merge(
-        joined_gdf[["Grade", "index"]], on="index", how="left"
-    ).drop(columns=["geometry", "index"])
-    return result_gdf
+    # group based on the created index, resulting chunk_size partitons
+    data_partitioned = enhanced_epc_df.with_columns(partitions).partition_by("chunk_id")
+    logging.info(f"Adding listed buildings to EPC in {len(data_partitioned)} chunks")
+    for i, enhanced_epc_df_chunk in tqdm(enumerate(data_partitioned)):
+        epc_gdf = transform_df_EPC_X_and_Y_to_point(enhanced_epc_df_chunk)
```
crispy-wonton: Btw, this same function exists in `prepare_features/lat_lon.py`. It's not on dev yet as it was created in [branch 36](https://github.com/nestauk/asf_heat_pump_suitability/blob/03b5fa76826322796ff9e1761c750e936952ed16/asf_heat_pump_suitability/pipeline/prepare_features/lat_lon.py) which hasnt been merged. When everything is merged, we can refactor our code to just keep one function. Nothing to change for now but just keep in mind :)

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,56 +29,95 @@ def transform_df_EPC_X_and_Y_to_point(
...
+        epc_gdf_temp = epc_gdf[["geometry", "index"]].copy()
```
crispy-wonton:
```suggestion
        epc_gdf = transform_df_EPC_X_and_Y_to_point(enhanced_epc_df_chunk)[["geometry", "index"]]
```
This might save some memory, instead of copying the df each time and retaining the full df in memory.
If you use this, you would also need to update the code below to use `epc_gdf` instead of `epc_gdf_temp`.

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,56 +29,95 @@ def transform_df_EPC_X_and_Y_to_point(
...
+        joined_gdf_chunk = gpd.sjoin(
+            epc_gdf_temp, listed_buildings_df, how="inner", predicate="intersects"
+        )
+        # Drop the geometry column
+        joined_gdf_chunk = joined_gdf_chunk.drop(columns=["geometry", "index_right"])
```
crispy-wonton:
```suggestion
        joined_gdf_chunk = gpd.sjoin(
            epc_gdf_temp, listed_buildings_df, how="inner", predicate="intersects"
        ).drop(columns=["geometry", "index_right"])
```
Minor but I think you could chain this if you wanted to.

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,56 +29,95 @@ def transform_df_EPC_X_and_Y_to_point(
...
+        if i == 0:
+            joined_gdf = joined_gdf_chunk
+        else:
+            joined_gdf = pd.concat([joined_gdf, joined_gdf_chunk])
```
crispy-wonton: Alternatively, you can append them into a list and then concatenate all at the end outside the loop.

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,56 +29,95 @@ def transform_df_EPC_X_and_Y_to_point(
...
+    enhanced_epc_nationtwo_listed_buildings_df: pl.DataFrame,
+) -> pl.DataFrame:
+    """
+    Merge listed buildings datasets from two nations into a single dataset.
+    Args:
+        enhanced_epc_nationone_listed_buildings_df (pl.DataFrame): EPC dataset with listed buildings from nation one
+        enhanced_epc_nationtwo_listed_buildings_df (pl.DataFrame): EPC dataset with listed buildings from nation two
     Returns:
-        pl.DataFrame: Polars DataFrame
+        enhanced_epc_with_all_listed_buildings (pl.DataFrame): EPC dataset with listed buildings from both nations
```
crispy-wonton:
```suggestion
        pl.DataFrame: EPC dataset with listed buildings from both nations
```
Can just have the return type without the variable name :)

asf_heat_pump_suitability/pipeline/run_scripts/add_features.py
```diff
@@ -69,11 +69,9 @@ def main(epc_path: str, save_output: Optional[str] = None) -> pl.DataFrame:
     # Add feature: lat/long
     logging.info("Adding lat/lon data to EPC")
     uprn_latlon_df = lat_lon.transform_df_osopen_uprn_latlon()
-    epc_latlon_df = epc_df.select(["UPRN"])
 
     # Join enhanced datasets together
     enhanced_epc_df = enhanced_epc_df.join(uprn_latlon_df, how="left", on="UPRN")
-    enhanced_epc_df = epc_df.join(uprn_latlon_df, how="left", on="UPRN")
```
crispy-wonton, line 76: Woops! not sure why i had that in there - thanks for cleaning it!

environment.yaml
```diff
@@ -4,5 +4,6 @@ channels:
 dependencies:
   - pip
   - python=3.9
```
crispy-wonton: Just fyi, I've had to update this to `python=3.10` in branch 36 for functionalities. Might be worth updating and testing to make sure your branch works with it now :)

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,56 +29,95 @@ def transform_df_EPC_X_and_Y_to_point(
...
 
+def merge_listed_buildings_nations(
+    enhanced_epc_nationone_listed_buildings_df: pl.DataFrame,
+    enhanced_epc_nationtwo_listed_buildings_df: pl.DataFrame,
+) -> pl.DataFrame:
+    """
+    Merge listed buildings datasets from two nations into a single dataset.
+    Args:
+        enhanced_epc_nationone_listed_buildings_df (pl.DataFrame): EPC dataset with listed buildings from nation one
+        enhanced_epc_nationtwo_listed_buildings_df (pl.DataFrame): EPC dataset with listed buildings from nation two
```
crispy-wonton: to make the code more readable, it's fine to give them shorthand names because you have the full explanation in the docstring :)

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,56 +29,95 @@ def transform_df_EPC_X_and_Y_to_point(
...
+                pl.col("*")
+                .exclude("Grade")
+                .first()
+                .name.keep(),  # Keep the first occurrence of all columns except 'Grade'
+                pl.col("Grade")
+                .map_elements(
+                    lambda x: ", ".join(set(str(i) for i in x if i is not None)),
+                    return_dtype=pl.Utf8,
+                )
+                .alias("Grade"),  # Merge 'Grade' values into a single column
```
crispy-wonton: this is some smooth code my dude! nice! can add a lil more notation as theres a lot happening here :)

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,56 +29,95 @@ def transform_df_EPC_X_and_Y_to_point(
...
+        # Drop the geometry column
+        joined_gdf_chunk = joined_gdf_chunk.drop(columns=["geometry", "index_right"])
+        if i == 0:
+            joined_gdf = joined_gdf_chunk
+        else:
+            joined_gdf = pd.concat([joined_gdf, joined_gdf_chunk])
 
+    enhanced_epc_df = enhanced_epc_df.join(
+        pl.from_pandas(joined_gdf), how="left", on="index"
+    )
```
crispy-wonton: Since we need to join England and Wales data together later, it might be quicker just to return the `pl.from_pandas(joined_gdf)` here. Then England and Wales dfs can be merged later with your other function, and finally, the fully merged dataset can be joined to the full EPC dataset in the `add_features.py` script.

crispy-wonton (commented):
Looks good thanks @helloaidank ! 

I ran the code snippet for the 10k sample set and it works great :) I didn't test it on the full EPC - have you? Just so we know if the chunking works well.

A couple of questions on the output. There is a blank value for 'Grade' for many EPC properties. Is this a null value? i.e., not a listed building? 

Secondly, it looks like 'Grade' is the only new column added to the dataset. Is that meant to be the case? If so, why do we need [lines 109-113 in this function](https://github.com/nestauk/asf_heat_pump_suitability/blob/6bda23ed670274cbc1dc2c03690a50be779fdff1/asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py#L88)? Could we just transform and add the 'Grade' series to each UPRN on its own and then join back to the full EPC dataset later? Sorry if I misunderstood.

I didn't find any bugs, just some suggestions for cleaning the code a little bit and a couple of questions. Have a look through the comments and let me know if there's anything you disagree with and then I'll make the changes :)

asf_heat_pump_suitability/getters/base_getters.py
helloaidank: I haven't tested it out, but I think s3fs is needed to read from the S3 bucket? Unless there is some internal functionality or have I misunderstood that?

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
helloaidank: What I mean is that it filters the dataframe for the relevant columns, i.e. it selects the geometry and grade columns!

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
helloaidank: Yeah, I'd imagine there is a few of these cases of re-usable functions!

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
helloaidank: Yep this is a good suggestion!

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
helloaidank: Cool, happy with this!

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
helloaidank: Mmmm yes, I think this might make it quicker! Great suggestion 👯

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
helloaidank: Agreed with you here, that I need to document this better.

helloaidank:
```suggestion
                # For all columns except 'Grade':
                # Retain the first occurrence of the data associated with each 
                 unique UPRN.
                # The column names remain unchanged after this operation.
                pl.col("*")
                .exclude("Grade")
                .first()
                .name.keep(),  # Keep the first occurrence of all columns except 'Grade'
                # For the 'Grade' column:
                # Combine all non-null values associated with each UPRN into a single string, 
                # with values separated by commas. Duplicates are removed in the process.
                pl.col("Grade")
                .map_elements(
                    lambda x: ", ".join(set(str(i) for i in x if i is not None)),
                    return_dtype=pl.Utf8,
                )
                .alias("Grade"),  # Merge 'Grade' values into a single column
```

environment.yaml
helloaidank: Ohhhhhh exciting, let's see if it collapses

helloaidank:
> Looks good thanks @helloaidank !
> 
> I ran the code snippet for the 10k sample set and it works great :) I didn't test it on the full EPC - have you? Just so we know if the chunking works well.
> 
> A couple of questions on the output. There is a blank value for 'Grade' for many EPC properties. Is this a null value? i.e., not a listed building?
> 
> Secondly, it looks like 'Grade' is the only new column added to the dataset. Is that meant to be the case? If so, why do we need [lines 109-113 in this function](https://github.com/nestauk/asf_heat_pump_suitability/blob/6bda23ed670274cbc1dc2c03690a50be779fdff1/asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py#L88)? Could we just transform and add the 'Grade' series to each UPRN on its own and then join back to the full EPC dataset later? Sorry if I misunderstood.
> 
> I didn't find any bugs, just some suggestions for cleaning the code a little bit and a couple of questions. Have a look through the comments and let me know if there's anything you disagree with and then I'll make the changes :)
Thank you for taking the time to review, it is much appreciated!
I have tested it out on the whole dataset, and it works!

The blank value is just an empty string, but in retrospect I think it is better if we fill this with a 'None' type, I don't think that it is difficult, probably just an extra/changed step.

With regards to your second point, we are merging EPC datasets with the same values elsewhere but just a different Grade Column (one has the Grade column for Wales and the other has the Grade column for England). I think we could probably just filter the 'enhanced dataset' here so that we are only reading in the UPRN + Grade and then match it with the enhanced EPC dataset afterwards if that makes sense and I think that's what you suggested. 

Lines 109-113 were there to ensure that we didn't have duplicate UPRNs of all the other columns apart from the 'Grade' column. 

Hope that makes sense, bit frustrated that I can't make these changes as it's always great to improve the code you've made but alas, public discourse calls!

Good luck and thanks for the great review 🕺 .

asf_heat_pump_suitability/getters/base_getters.py
crispy-wonton: I think mostly file system is required to **write** to s3 and most of the time s3 URIs can be **read** directly.

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,95 +30,58 @@ def transform_df_EPC_X_and_Y_to_point(
...
+        df = epc_gdf.sjoin(listed_buildings_gdf, how="inner", predicate="intersects")[
+            ["UPRN", "listed_building_grade"]
+        ].drop_duplicates(subset="UPRN")
 
-    return enhanced_epc_df
+        dfs.append(df)
 
+    df = pl.from_pandas(pd.concat(dfs))
 
-def merge_listed_buildings_nations(
```
crispy-wonton: Function removed and replaced with simpler concat and join in `add_features.py`

asf_heat_pump_suitability/getters/get_datasets.py
```diff
@@ -131,15 +133,23 @@ def get_df_spa_offgasgrid() -> pl.DataFrame:
     return df
 
 
-def get_df_listedbuildings(data_source_key: str) -> pl.DataFrame:
+def load_gdf_listed_buildings(nation: str, **kwargs) -> gpd.GeoDataFrame:
```
crispy-wonton, line 136: I made the function slightly less generic. My thinking: With the data source key, you can technically put any key for any data source in the config in here. However, because the function is loading a geospatial file, this won't work for most of the datasets. Even when it does work, it would then not be loading listed buildings data as the function describes. The change here to add the `nation` keyword makes the function more specific.

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,95 +30,58 @@ def transform_df_EPC_X_and_Y_to_point(
     return gdf
 
 
-def get_filtered_df_listed_buildings(data_source_key: str) -> gpd.GeoDataFrame:
+def transform_gdf_listed_buildings(nation: str) -> gpd.GeoDataFrame:
```
crispy-wonton, line 33: Here I cleaned the name as it's not a getter function, and also added some additional preprocessing steps to clean and prepare the geometry data for joining to EPC. I also updated the docstring so it makes sense if you load England or Wales data

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,95 +30,58 @@ def transform_df_EPC_X_and_Y_to_point(
...
+    )
 
-    return filtered_listed_building_gdf
+    return gdf
 
 
-def spatial_join_epc_with_listed_buildings(
-    enhanced_epc_df: pl.DataFrame,
-    listed_buildings_df: gpd.GeoDataFrame,
+def sjoin_df_epc_with_listed_buildings(
```
crispy-wonton, line 51: Here I changed the function return value. It now returns the inner join of EPC with listed buildings. I.e. you will only get UPRNs with listed building grade if they are a listed building. We will left join this to the full EPC later in the script. The purpose of this was to increase speed and to simplify the function.

asf_heat_pump_suitability/pipeline/run_scripts/add_features.py
```diff
@@ -96,34 +96,30 @@ def main(epc_path: str, save_output: Optional[str] = None) -> pl.DataFrame:
     logging.info("Adding off gas grid column to EPC")
     off_gas_postcodes = off_gas.process_off_gas_data()
     enhanced_epc_df = off_gas.add_off_gas_feature(enhanced_epc_df, off_gas_postcodes)
-    logging.info("Adding Historic England listed buildings to EPC")
```
crispy-wonton: Here the logic remains the same but I shortened some of the variable names to increase readability of the code.

asf_heat_pump_suitability/pipeline/run_scripts/add_features.py
```diff
@@ -96,34 +96,30 @@ def main(epc_path: str, save_output: Optional[str] = None) -> pl.DataFrame:
...
+        enhanced_epc_df, w_listed_buildings_df
     )
-    enhanced_epc_df = listed_buildings.merge_listed_buildings_nations(
-        enhanced_epc_england_listed_buildings_df, enhanced_epc_wales_listed_buildings_df
+
+    listed_buildings_df = pl.concat(
+        [e_listed_buildings_df, w_listed_buildings_df], how="vertical"
     )
+    enhanced_epc_df = enhanced_epc_df.join(listed_buildings_df, how="left", on="UPRN")
+
```
crispy-wonton, line 117: Here we concat the inner joins of the listed buildings for England and Wales and then left join to EPC in one go. This avoids two bigger joins to the full EPC.

PR #56, opened 2024-08-09
44 add conservation area

### LINKED ISSUE #44: Add conservation area flag
Add flag to indicate whether or not property is in a [conservation area](https://www.planning.data.gov.uk/dataset/conservation-area)

crispy-wonton:
Fixes #44 

# Description

Add 2 flags to each EPC row indicating 1) whether the property is within a known building conservation area or 2) whether the council / LAD (local authority districts) in which the property is located has building conservation area data available.

Changes:

- Add `pipeline/prepare_features/conservation_areas.py` with new functions to transform the historic England building conservation area data in preparation for joining with EPC.
- Refactor `pipeline/prepare_features/output_areas.py` function to include local authority data in the ONSPD dataset and update function name.
- Update `pipeline/run_scripts/add_features.py` to join building conservation area data to EPC
- Add historic England building conservation area data to `config/base.yaml` and `config/README.md`
- Add new getter to `get_datasets.py` to load historic England building conservation area dataset
- DELETE `pipeline/enhance_epc/enhance_epc.py` and refactor `pipeline/enhance_epc/run_script.py` as a result. I deleted this file as it only held one function in it which was a legacy function from an earlier version of the pipeline and was never developed into it's original planned form. It was not a broadly useful function that could be applied to multiple use cases. I removed the function and refactored it into the run script to clean up the repo. I don't think it's used in another unmerged branch although let me know if it is and I can add back in.

# Instructions for Reviewer

To run the code, you can run the following line:
`python -i asf_heat_pump_suitability/pipeline/run_scripts/add_features.py --epc_path s3://asf-heat-pump-suitability/outputs/2023_Q2_EPC_enhanced_weights_and_features.parquet --save_output s3://asf-heat-pump-suitability/outputs/test.parquet`

- Function deletions and refactoring. Please could you double check the refactoring has been done correctly and also let me know if you have used any deleted/refactored functions in an unmerged branch.
- Logic of the 2 new columns added to EPC. 'in_conservation_area' column values can either be True or Null. 'lad_conservation_area_data_available' values can either be True, False, or Null. The issue is, because the conservation area data does is not complete, we can't really confirm if properties which are not labelled as being in a conservation area are actually not in a conservation area, or if they are in a conservation area that is missing from the dataset. The purpose of the 'lad_conservation_area_data_available' column is to differentiate this. Does this make sense or would you categorise things differently?
- General checks for mistakes/ bugs and that joins/merges have been conducted correctly

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/config/README.md
```diff
@@ -16,6 +16,7 @@
...
+| `["data_source"]["E_historic_england_conservation_areas"]`     | `source_data/Aug2024_historic_england_conservation_areas_E.geojson`                            | [Historic England](https://www.planning.data.gov.uk/dataset/conservation-area)                                                                      | June 2024   | Conservation areas in England (places of special architectural and historic interest).                                                                                                                                                                                                                                                     |
```
helloaidank: Just for my understanding, are we supposed to use the geojson file for the data source in this case as it uses the WGS84 projected coordinate system and not the BNG boundaries?

asf_heat_pump_suitability/getters/get_datasets.py
```diff
@@ -132,6 +132,21 @@ def get_df_osopen_uprn_latlon(**kwargs) -> pl.DataFrame:
...
+def load_gdf_historic_england_conservation_areas() -> gpd.GeoDataFrame:
+    """
+    Load GeoDataFrame with conservation area polygons from Historic England (CRS: EPSG:4326).
+
+    Returns:
+        gpd.GeoDataFrame: polygons of conservation areas in England
+    """
+    gdf = gpd.read_file(
+        config["data_source"]["E_historic_england_conservation_areas"],
+        engine="pyogrio",
```
helloaidank, line 151: Ah is `"pyogrio"` the lightweight and faster version of `fiona`? How much faster is it? Could definitely use this in different functions within the code base!

asf_heat_pump_suitability/pipeline/prepare_features/conservation_areas.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    gdf = get_datasets.load_gdf_historic_england_conservation_areas()[
+        ["name", "geometry"]
+    ].to_crs("EPSG:27700")
+
+    return gdf
+
+
+def generate_df_conservation_area_data_availability(
+    ladcd_col: str = "LAD23CD",
```
helloaidank, line 24: What is "LAD23CD" ? :)

asf_heat_pump_suitability/pipeline/prepare_features/output_areas.py
```diff
@@ -3,24 +3,36 @@
...
-    Process and clean ONS postcode directory dataset: standardise postcode; clean output area columns; add new
-    `country_code` column.
+    Load and transform ONS postcode directory dataset: standardise postcode; clean output area columns; add new
+    `country_code` column; map rural-urban indicators.
 
     Args
         pcd_col (str): name of column containing postcodes. Default `"pcd"`.
         ruc_col (str): name of column containing rural-urban classification codes. Default `"ru11ind"`.
+        lad_col (str): name of column containing Local Authority District (LAD) codes. Default `"oslaua"`.
         use_cols (list): columns to import. Default `["pcd", "lsoa11", "msoa11", "lsoa21", "msoa21", "ru11ind"]`.
```
helloaidank:
```suggestion
        use_cols (list): columns to import. Default `["pcd", "lsoa11", "msoa11", "lsoa21", "msoa21", "ru11ind", "oslaua]"`.
```

helloaidank: Also I cannot comment where it is located as it's not a change but in the `standardise_col_postcode` function, you use `pl.col(pcd_col).str.to_uppercase().str.replace(" ", "").alias("POSTCODE")
    )` and I just remembered Liz making the suggestion in the off-gas feature about using `pl.col(pcd_col).str.to_uppercase().str.replace(r'\s+', '').alias("POSTCODE")
    )`

asf_heat_pump_suitability/pipeline/run_scripts/add_features.py
```diff
@@ -35,32 +44,28 @@ def run():
...
-        pl.DataFrame: enhanced EPC dataset with additional features
-    """
     # Import processed EPC
-    logging.info(f"Loading EPC file from path: {epc_path}")
-    epc_df = pl.read_parquet(epc_path)
+    logging.info(f"Loading EPC file from path: {_args.epc_path}")
+    epc_df = pl.read_parquet(_args.epc_path)
+    # Join LAD code to EPC
+    # TODO this join and the preceding 2 lines can be removed once enhance_epc/run_script.py has been re-run
+    # TODO because the updated run_script.py will join the lad_code already
```
helloaidank, line 58: keep on laughing to myself when reading `lad_code`

helloaidank (commented):
Hey Roisín!

Thanks for this 🚀 , I have looked through the code and ran it, everything looks like it is in working order! I've just made a couple of minor comments and asked a couple of questions. With regards to the refactoring and deleted script, I don't think any of this affects the newer added features, so that's not an issue.

The code runs fine 👍 .

asf_heat_pump_suitability/config/README.md
crispy-wonton: Here we just have to use the geojson because I could only find this file format (or CSV or JSON) for [the latest version of the cons area data](https://www.planning.data.gov.uk/dataset/conservation-area) that I could find. I couldnt find a .gpkg or .shp file for this version.

They do have geopackage and shapefile versions on the [Historic England website](https://opendata-historicengland.hub.arcgis.com/datasets/historicengland::conservation-areas/explore?location=52.734863%2C-2.491828%2C6.62), but this seems to be a slightly earlier version of the data (July 2024 instead of August 2024) and as I understand from the description, the gov data combines Historic England data with a couple of other sources. Ultimately I went with the gov data because I thought it had a few more cons areas. :)

crispy-wonton: So the only reason is the file type availability, nothing to do with CRS in this case.

asf_heat_pump_suitability/pipeline/prepare_features/conservation_areas.py
crispy-wonton: This is just the name of the column in the dataset with the Local Authority District codes (like "E1000001"). These codes will be used in the join with EPC later.

asf_heat_pump_suitability/getters/get_datasets.py
crispy-wonton: "For Geopandas versions <1.0, GeoPandas defaulted to use Fiona. Starting from GeoPandas version 1.0, the global default has changed from Fiona to Pyogrio.

The main reason for this change is performance. Pyogrio is optimized for the use case relevant for GeoPandas: reading and writing in bulk. Because of this, in many cases speedups >5-20x can be observed." 

^ taken from the [geopandas docs ](https://geopandas.org/en/latest/docs/user_guide/fiona_to_pyogrio.html):)

asf_heat_pump_suitability/pipeline/prepare_features/output_areas.py
crispy-wonton: Great shout, well remembered! Thank yoU!

asf_heat_pump_suitability/config/README.md
helloaidank: Ok great to know, thanks for that! Just good to check I haven't misunderstood something and happy with your reasoning for going with the geojson.

asf_heat_pump_suitability/getters/get_datasets.py
helloaidank: Ah ok, so if it's the default is there a reason we explicitly set out `engine="pyogrio"`? We are using Geopandas version more recent than 1.0 right?

asf_heat_pump_suitability/pipeline/prepare_features/conservation_areas.py
helloaidank: Ah ok, but what does the 23CD stand for?

asf_heat_pump_suitability/getters/get_datasets.py
crispy-wonton: You're right, it's not necessary to explicitly add the engine here because it should be used by default. Sometimes I put default params in explicitly anyway so it's clearer how the function works, especially for debugging :)

asf_heat_pump_suitability/pipeline/prepare_features/conservation_areas.py
crispy-wonton: I think 23 comes from the fact that the data is from the year 2023 and I think "CD" stands for code? Not sure, this was just my assumption. It's the column name in the council boundaries file which is downloaded from [data gov uk](https://www.data.gov.uk/dataset/288458f7-7789-47d0-80d4-ffdf746c6b75/local-authority-districts-december-2023-boundaries-uk-bfe), not a column I created :)

PR #58, opened 2024-08-14
57 add building conservation areas wales

### LINKED ISSUE #57: Add Wales building conservation area data
Add conservation area data for Welsh EPC addresses. Source: https://datamap.gov.wales/layers/inspire-wg:conservation_areas

crispy-wonton:
Fixes #57 

# Description

Add building conservation area data for Wales to `add_features.py` pipeline so that 'in_conservation_area' flag on EPC dataset will flag if EPC property is located in a building conservation area in England or Wales.

Summary of changes:
- Add new getter to `get_datasets.py` for Wales building conservation area data
- Add source of Wales data to `config/base.yaml` and `config/README`
- Update `prepare_features/conservation_areas.py` to load and transform Wales building cons area data as well as England data
- Minor changes to logging and notation in `run_scripts/add_features.py`

# Instructions for Reviewer

To test the code run:
`python -i asf_heat_pump_suitability/pipeline/run_scripts/add_features.py --epc_path s3://asf-heat-pump-suitability/outputs/2023_Q2_EPC_enhanced.parquet --save_output s3://asf-heat-pump-suitability/outputs/test.parquet`

- Please could you double check I've used the correct dataset ([source](https://datamap.gov.wales/layers/inspire-wg:conservation_areas)). The most recent version I could find was from Sept 2022. This dataset also has a relatively low number of cons areas - around 500.
- Please could you let me know if there are any obvious speed improvements you can see
- Please could you check for general bugs and mistakes

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/pipeline/prepare_features/conservation_areas.py
```diff
@@ -24,56 +34,55 @@ def generate_df_conservation_area_data_availability(
...
+    cons_areas_gdf = transform_gdf_building_cons_areas()
 
     # Join EPC UPRNs within or on boundaries of conservation areas
-    epc_df = epc_df.sjoin(cons_areas_gdf, how="inner", predicate="intersects").drop(
-        columns=["index_right", "geometry"]
+    df = (
+        df.sjoin(cons_areas_gdf, how="inner", predicate="intersects")
+        .drop(columns=["index_right", "geometry"])
+        .drop_duplicates(subset="UPRN")
     )
```
lizgzil, line 82: I think this can be sped up a lot by joining it in chunks (it helped on Aidan's listed building code anyway). See how he did it in [this function](https://github.com/nestauk/asf_heat_pump_suitability/blob/018ae286c97bb9900ea3cc0c3d6c20540a0c6129/asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py#L51)

lizgzil: but I'd leave that for future changes really as it didn't take that long anyway! seems like the biggest time is taken in `df = lat_lon.generate_gdf_uprn_coords(df)[["UPRN", "geometry"]]`

lizgzil (commented):
thanks @crispy-wonton ! This looks good and the code ran for me (I just ran it until line 99).

500 conservation zones seems about right (from [this](https://cadw.gov.wales/advice-support/placemaking/legislation-guidance/conservation-areas#:~:text=Over%20500%20conservation%20areas%20in,and%20Conservation%20Areas)%20Act%201990.)). and I can't see anything more recent than that 2022 dataset either.

I have left a comment for a speed up idea that helped on the `sjoin` in Aidan's listed building functions -but I don't think it's worth adding since it's not a huge bottleneck in the timing anyway!

I had a look at the counts:

```
in_conservation_area  lad_conservation_area_data_available
NaN                   True                                    15879367
                      False                                    2389646
                      NaN                                        15956
True                  True                                     1549246
                      NaN                                          333
                      False                                          1
```
...so there is one property (UPRN 41104898.0) that is in a conservation zone, but it's LA doesn't have any conservation zones? Not sure whats going on there?

asf_heat_pump_suitability/pipeline/prepare_features/conservation_areas.py
crispy-wonton: Thanks Liz! Great idea :) 
Yeh loading the lat lon coordinates dataset takes ages because it is an absolutely huge dataset. It has all UPRNs for whole of GB with their latitude and longitude...

crispy-wonton: I'll see if we're loading additional unnecessary columns

crispy-wonton:
> ...so there is one property (UPRN 41104898.0) that is in a conservation zone, but it's LA doesn't have any conservation zones? Not sure whats going on there?

I looked into this. For that property, I looked the address up on Google maps. It looks like the cause is that the postcode is incorrect in the EPC dataset by one character(it's XXX 4XX in EPC dataset and XXX 9XX in google maps). The rest of the address matches with a property in the different postcode. What has happened is this:
- df1 = EPC UPRNs are spatial joined to conservation areas using the geometry of the UPRN's lat lon and the cons area polygons. This gives us the 'in_conservation_area' flag.
- df2 = Conservation areas are spatial joined to the LAD boundaries to generate a separate df that has a flag for each LAD to tell us whether "lad_conservation_area_data_available".
- df1 is left joined to the full EPC dataset on "UPRN"
**- df2 is left joined to the full EPC on "LAD_code"**. LAD code is a feature we add to EPC earlier. It's matched to each row in EPC using the postcode. Therefore, if the postcode is wrong, the EPC row will be matched to an incorrect LAD code, and when we left join df2 in this step, it could give us the wrong "lad_conservation_area_data_available" flag.

See image below:
Blue point marker = location of UPRN given from coordinates matched to EPC row
Blue circle = location of UPRN given from (likely incorrect) postcode of address given in EPC row
Green area = conservation area spatial match
Orange area = LAD boundary spatial match

As you can see, the incorrect postcode which we were joining on is outside the LAD boundaries. Hence, the property is matched to a cons area but it's receiving a flag to say cons data is not available.

I think we can rectify this issue by left joining df2 with a spatial join rather than join on LAD code.

image

crispy-wonton:
@lizgzil I've updated the code to conduct a spatial join with LAD boundaries to fix the weird bug found.

Here's the updated run code. I wouldnt recommend running because it takes an hour but feel free to load and look at the results:
`python -i asf_heat_pump_suitability/pipeline/run_scripts/add_features.py --epc_path s3://asf-heat-pump-suitability/outputs/20240824_2023_Q4_EPC_weighted.parquet --save_as s3://asf-heat-pump-suitability/outputs/20240827_2023_Q4_EPC_weighted_features.parquet)`

Here are the new counts:
(I know it doesnt add up to the original counts but that's because of using the updated weighted EPC dataset which I produced last week. I think since running it originally, we added some lines to drop rows with null. I need to double check why there are 2 million fewer rows in this new dataset but I think that might be why. I've checked, and we don't lose any rows in this script so it happens in the reweighting.).
```
 in_conservation_area   ┆ lad_conservation_area_data_available │
│ null                                ┆ {null,18924}                    │
│ null                                 ┆ {false,2283524}                 │
│ null                                 ┆ {true,14026042}                 │
│ true                                 ┆ {true,1423664}
```

lizgzil:
thanks @crispy-wonton - good debugging! Feel free to merge when you like, I guess the 2 million row disappearance can be investigated elsewhere at another point (and perhaps as you say it is because of the null row deletion)

PR #71, opened 2024-10-02
68 Add grid capacity feature

### LINKED ISSUE #68: Add grid capacity feature

shabrf:
---

# Description

This pull request introduces a new feature to calculate the grid capacity for heat pump installations per LSOA. The main changes include:

1. Implementation of data processing functions for each Distribution Network Operator (DNO): ENW, NPg, SPEN, SSEN, UKPN, and WPD.
2. Creation of a unified substation dataset combining data from all DNOs.
3. Development of functions to distribute substation headroom to LSOAs and calculate headroom per household.
4. Implementation of a heat pump suitability assessment function that calculates the percentage of households in each LSOA that could potentially install heat pumps based on available grid capacity.
5. Integration of all components into a main `calculate_grid_capacity()` function.

Fixes #68 

# Instructions for Reviewer

In order to test the code in this PR, you need to:

1. Run the main script containing the `calculate_grid_capacity()` function.
2. Check the output DataFrame for correct column names and data types, especially the 'heatpump_installation_percentage' column.
3. Check the incorporation of this output in the wider pipeline

Please pay special attention to:

1. The data processing functions for each DNO (ENW, NPg, SPEN, SSEN, UKPN, WPD) to ensure they correctly handle the specific data formats and structures for each operator.
2. The `distribute_substation_headroom()` function, which allocates substation capacity to LSOAs based on household count. Verify that this allocation is performed correctly and fairly.
4. The `assess_heatpump_suitability()` function, particularly the calculation of 'max_heatpumps' and 'heatpump_installation_percentage'. Ensure these calculations are accurate and properly consider the power requirements of heat pumps.
5. The main `calculate_grid_capacity()` function to confirm it correctly integrates all components and produces a coherent final dataset.
6. Error handling and edge cases, such as LSOAs with very low household counts or unusually high/low headroom values.

The expected output should be a DataFrame containing LSOA-level data, including the percentage of households that could install heat pumps in each LSOA. This percentage should never exceed 100% and should accurately reflect the relationship between available grid capacity and household count.

# Checklist:

- [x] I have tested the code


asf_heat_pump_suitability/getters/s3_getters.py
```diff
@@ -123,6 +125,12 @@ def load_s3_data(
     elif fnmatch(file_name, "*.pkl") or fnmatch(file_name, "*.pickle"):
         file = obj.get()["Body"].read().decode()
         return pickle.loads(file)
+    elif fnmatch(file_name, "*.gpkg"):
+        with BytesIO(obj.get()["Body"].read()) as file:
+            return gpd.read_file(file)
+    elif fnmatch(file_name, "*.geojson"):
+        with BytesIO(obj.get()["Body"].read()) as file:
+            return gpd.read_file(file)
```
crispy-wonton, line 134: Nice, thanks for adding this :)

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+
+    if match:
+        count = float(match.group(1))
+        individual_capacity = float(match.group(2))
+        return count * individual_capacity
+
+    try:
+        return float(value)
+    except ValueError:
+        print(f"Warning: Could not parse capacity value: {value}")
```
crispy-wonton:
```suggestion
        logging.warning(f"Could not parse capacity value: {value}")
```
need to import logging at top of script as well

asf_heat_pump_suitability/getters/s3_getters.py
```diff
@@ -174,3 +182,33 @@ def read_json_from_s3(bucket: str, file_path: str) -> dict:
     json_file = s3_resource.Object(bucket, file_path)
     json_file = json_file.get()["Body"].read().decode("utf-8")
     return json.loads(json_file)
+
+
+def get_shapefile_from_s3(bucket_name: str, directory_path: str):
```
crispy-wonton: What's the reason for doing this instead of just passing `file.shp` (e.g. saved in `config/base.yaml` under `data_source` key) to `gpd.read_file()`?

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+        return np.nan
+
+
+def process_shapefile(shapefile_directory: str) -> gpd.GeoDataFrame:
+    """
+    Process a shapefile from S3 and return it as a GeoDataFrame.
+
+    Args:
+        shapefile_directory: S3 directory containing the shapefile.
```
crispy-wonton: Is the directory that is input meant to be a subdir of `asf-heat-pump-suitability` bucket's `source_data` directory?

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+    """
+    bucket_name = "asf-heat-pump-suitability"
+    temp_dir = get_shapefile_from_s3(bucket_name, shapefile_directory)
+
+    try:
+        shp_file = next(f for f in os.listdir(temp_dir) if f.endswith(".shp"))
+        shp_path = os.path.join(temp_dir, shp_file)
+        gdf = gpd.read_file(shp_path)
+        gdf = gdf.to_crs(epsg=27700)  # Convert to British National Grid
+        return gdf
```
crispy-wonton: You should be able to just read from s3 file path directly. You can save the s3 path to the `.shp` filename in the `config/base.yaml` under `data_source` key and load the files with a simpler function where you need them.

Let me know if i'm missing something :)

asf_heat_pump_suitability/pipeline/suitability/calculate_suitability.py
```diff
@@ -143,6 +143,18 @@
     "HN_N": 2,
 }
 
+# to refine
```
crispy-wonton:
```suggestion
# TODO to refine
```
just so it's easier to search for later :)

asf_heat_pump_suitability/pipeline/suitability/calculate_suitability.py
```diff
@@ -283,6 +297,11 @@ def compute_df_total_score_per_epc(
         pl.when(pl.col("CURRENT_ENERGY_RATING").is_in(["A", "B", "C"]))
         .then(epc_threshold_scores.get(tech_type))
         .alias("epc_rating_score"),
+        pl.when(
+            pl.col("heatpump_installation_percentage") > grid_installation_threshold
+        )
```
lizgzil, line 294: This is fine for now, but just adding a thought that I wonder whether the score added should just be the continuous value in `pl.col("heatpump_installation_percentage")/100`, rather than a score based off whether this value is greater than 30% or not.

Can't work out whether that would make `compute_df_max_score_per_row` a bit hellish or not! e.g. would lines 359-362 be

```
pl.when(pl.col("heatpump_installation_percentage").is_not_null())
        .then(1)
        .otherwise(0)
        .alias("grid_capacity_max"),
```
or `.then(pl.col("heatpump_installation_percentage").max/100)` (which might be 1 anyway)

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+CRS = "EPSG:4326"  # Geometry coordinate reference system - standard longitude/latitude projection
+TAKEUP_RATE = 30  # Percentage of households in each LSOA to install heat pumps
+POWER_PER_HEATPUMP = 8  # Power rating per heat pump in kW
+COEFFICIENT_OF_PERFORMANCE = (
+    2.5  # Heat output (heat pump rating) / electricity consumed
+)
+SUBSTATION_SCALING_FACTOR = 1.0  # Factor to scale substation power rating (MVA) by
+
+
+def _parse_binary_geometry(
```
lizgzil: this function could live in `utils/geo_utils.py`

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+        return count * individual_capacity
+
+    try:
+        return float(value)
+    except ValueError:
+        print(f"Warning: Could not parse capacity value: {value}")
+        return np.nan
+
+
+def process_shapefile(shapefile_directory: str) -> gpd.GeoDataFrame:
```
lizgzil: this might be better in `getters/s3_getters.py`

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+        shp_file = next(f for f in os.listdir(temp_dir) if f.endswith(".shp"))
+        shp_path = os.path.join(temp_dir, shp_file)
+        gdf = gpd.read_file(shp_path)
+        gdf = gdf.to_crs(epsg=27700)  # Convert to British National Grid
+        return gdf
+    finally:
+        shutil.rmtree(temp_dir)
+
+
+def generate_enw_df() -> gpd.GeoDataFrame:
```
lizgzil: I think all these substation generate functions could all be in the `getters/` folder in a new file. That way all our reading data files (and moderate processing) are in the same folder. Also, the hardcoded s3 locations that you read from should all be in the `config/base.yaml` file - with a description of each (sorry it's a pain) in the `config/README.md`. I think that's right but @crispy-wonton knows more about the structure!

I also think this is a nice to have situation, and can be addressed in another PR if you don't have time.

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+            "Longitude",
+            "Grid Reference",
+        ]
+    ]
+    ssen_shape = pd.concat([shepd_shape, sepd_shape])
+    ssen_df = ssen.merge(
+        ssen_shape, left_on="Primary Substation Name", right_on="Primary"
+    )
+
+    ssen_df["operator"] = "UKPN"
```
lizgzil: should this be "SSEN"

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+if __name__ == "__main__":
+    grid_capacity_results = calculate_grid_capacity()
+    print(grid_capacity_results.head())
+    print(f"Total LSOAs assessed: {len(grid_capacity_results)}")
+    print(
+        f"LSOAs with sufficient grid capacity: {grid_capacity_results['has_grid_capacity'].sum()}"
+    )
+    print(
+        f"Percentage of LSOAs with sufficient capacity: {(grid_capacity_results['has_grid_capacity'].mean() * 100):.2f}%"
+    )
```
lizgzil: think this is old and can be deleted since 'has_grid_capacity' doesn't exist anymore

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+    substations["headroom_mva"] = (
+        substations["firm_capacity_mva"] - substations["peak_demand_mva"]
+    )
+
+    # Load and process LSOA boundary data
+    lsoa_boundaries = load_s3_data(
+        "asf-heat-pump-suitability",
+        "source_data/Output_Areas_2021_EW_BGC_V2_4299916833741807639.geojson",
+    )
+    lsoa_gdf = aggregate_oa_to_lsoa(lsoa_boundaries)
```
lizgzil: would be useful in the `getters/get_datasets.py` file

lizgzil: @crispy-wonton is this file consistent with other LSOA geometries used?

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+
+    # Load and process household data
+    households_df = get_datasets.get_df_ons_number_of_households()
+    households_df = households_df.rename(
+        mapping={"mnemonic": "LSOA21CD", "2021": "household_count"}
+    ).to_pandas()
+
+    # Process substation and LSOA data
+    headroom_df = process_substation_lsoa_data(substations, lsoa_gdf, households_df)
```
lizgzil, line 284: I'm not sure how much of an issue this is but `len(substations) = 4225` and `substations['id'].nunique() = 4115`, and when I look at some of the duplicated IDs, sometimes it's due to multiple duplicates of the entire row and sometimes there are lots of Nones.
Screenshot 2024-10-09 at 17 42 43

So, it might be worth deleting duplicates? I think this effects the 
```
joined = gpd.sjoin(
        lsoa_with_households, substations_df, how="inner", predicate="intersects"
    )
```
line in the`distribute_substation_headroom` function.

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+    )
+    substation_total_households = substation_total_households.rename(
+        columns={"household_count": "total_households"}
+    )
+
+    # Merge total households back to joined dataframe
+    joined = joined.merge(substation_total_households, on="index_right")
+
+    # Calculate the fraction of substation's load for each LSOA
+    joined["load_fraction"] = joined["household_count"] / joined["total_households"]
```
lizgzil, line 147: I checked that `joined.groupby('index_right')['load_fraction'].sum()` all equalled 1, but noticed that `joined.groupby('id')['load_fraction'].sum()` didn't. Then I realised the substation ID column isn't unique which made me notice some duplicates which might be better removed?

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
lizgzil: do you think this is an issue that needs fixing or did you expect it?

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+    ).to_pandas()
+
+    # Process substation and LSOA data
+    headroom_df = process_substation_lsoa_data(substations, lsoa_gdf, households_df)
+
+    # Assess heat pump suitability
+    consumption_per_heatpump = POWER_PER_HEATPUMP / COEFFICIENT_OF_PERFORMANCE
+    result = assess_heatpump_suitability(
+        headroom_df, consumption_per_heatpump, TAKEUP_RATE
+    )
```
lizgzil: It doesn't look like this argument should be TAKEUP_RATE as the function says its voltage_factor?

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+    Returns:
+        DataFrame with processed data including distributed headroom per household.
+    """
+    lsoa_headroom = distribute_substation_headroom(
+        substations_df, lsoa_gdf, households_df
+    )
+    return calculate_headroom_per_household(lsoa_headroom, households_df)
+
+
+def assess_heatpump_suitability(
```
lizgzil, line 209: I don't have a good sense of capacity/voltage etc to feel like I can do a good job reviewing this function! @danlewis85 @crispy-wonton may be better placed

lizgzil (commented):
Thanks so much @shabrf 🎉  This is great - so thorough and easy to review :)

I made a few comments. My suggestion of moving some of the functions to other places in the repo is a nice to have - if you don't think you'll have time to do it then don't worry - we can create an issue and address it later.

I noticed a potential bug about duplicate substation data - no idea how big this is / whether it is to be expected.

I've tagged @crispy-wonton and @danlewis85 in a few specific places to check too.

Thanks so much!

asf_heat_pump_suitability/pipeline/run_scripts/run_add_features.py
```diff
@@ -162,6 +163,10 @@ def parse_arguments() -> argparse.Namespace:
     )
     epc_df = epc_df.join(listed_buildings_df, how="left", on="UPRN")
 
+    logging.info("Adding grid capacity column to EPC")
+    grid_capacities = grid_capacity.calculate_grid_capacity()
+    epc_df = epc_df.join(grid_capacity, how="left", on="lsoa21")
```
crispy-wonton: we need to use the `lsoa` column from `epc_df` for joins. `lsoa` is composed of `lsoa21` filled with `lsoa11` where missing.

asf_heat_pump_suitability/pipeline/suitability/calculate_suitability.py
```diff
@@ -240,6 +253,7 @@ def compute_df_total_score_per_epc(
     density_threshold: int = 60,
     garden_threshold: int = 10,
     external_space_threshold: int = 2,
+    grid_installation_threshold: int = 30,
```
crispy-wonton, line 248: Just a note that we need to record somewhere in our internal documentation (and generally publish as well) why we set the threshold at 30.

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+    enw_substations = load_s3_data(
+        "asf-heat-pump-suitability",
+        "source_data/grid_operators/ENW/ndp-pry-bsp-headroom.parquet",
+    )
+    enw_shape = load_s3_data(
+        "asf-heat-pump-suitability",
+        "source_data/grid_operators/ENW/ndp-pry-voronoi.parquet",
+    )
+
+    # Filter data for primary substations, current year, and best view scenario
```
crispy-wonton: What is 'best view scenario'?

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+    )
+    enw["operator"] = "ENW"
+    enw["geopoint"] = enw["geopoint"].apply(_parse_binary_geometry)
+
+    # Create GeoDataFrame and perform spatial join with shape data
+    enw_gdf = gpd.GeoDataFrame(enw, geometry="geopoint", crs=CRS)
+    enw_shape["geo_shape"] = enw_shape["geo_shape"].apply(_parse_binary_geometry)
+    enw_shape["geo_point_2d"] = enw_shape["geo_point_2d"].apply(_parse_binary_geometry)
+    enw_shape_gdf = gpd.GeoDataFrame(enw_shape, geometry="geo_shape", crs=CRS)
+    enw_df = gpd.sjoin(enw_gdf, enw_shape_gdf, how="right", predicate="within")
```
crispy-wonton:
```suggestion
    enw_df = gpd.sjoin(enw_gdf, enw_shape_gdf, how="right", predicate="intersects")
```
I have been using the `intersects` predicate because it returns `True` if the boundary or interior of the object intersect in any way with those of the other, so it will match points which are on the boundaries as well as within the polygons. https://shapely.readthedocs.io/en/latest/manual.html#object.intersects

I see you used intersects in the function below so was there a specific reason to use `within` here?

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+        gdf = gpd.read_file(shp_path)
+        gdf = gdf.to_crs(epsg=27700)  # Convert to British National Grid
+        return gdf
+    finally:
+        shutil.rmtree(temp_dir)
+
+
+def generate_enw_df() -> gpd.GeoDataFrame:
+    """
+    Generate a GeoDataFrame for Electricity North West (ENW) substations.
```
crispy-wonton: Please can you add a bit more explanation of what the generated geodataframe contains? Is it the point geometries within a certain area? :)

Same with docstrings of functions below

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+    ].rename(
+        columns={
+            "substation_id": "id",
+            "maximum_demand_mva_per_primary_substation": "peak_demand_mva",
+            "geo_shape": "geo_shape",
+        }
+    )
+
+
+def generate_npg_df() -> gpd.GeoDataFrame:
```
crispy-wonton:
```suggestion
def generate_npg_gdf() -> gpd.GeoDataFrame:
```
These should all be gdf i think

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+        enw_substations[["substation", "headroom_mva", "geopoint"]],
+        enw_demand[["number", "maximum_demand_mva_per_primary_substation"]],
+        left_on="substation",
+        right_on="number",
+    )
+
+    # Calculate firm capacity and add operator information
+    enw["firm_capacity_mva"] = (
+        enw["maximum_demand_mva_per_primary_substation"] + enw["headroom_mva"]
+    )
```
crispy-wonton: Do you know that max demand doesnt already include headroom?

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+    finally:
+        shutil.rmtree(temp_dir)
+
+
+def generate_enw_df() -> gpd.GeoDataFrame:
+    """
+    Generate a GeoDataFrame for Electricity North West (ENW) substations.
+
+    Returns:
+        GeoDataFrame with ENW substation data.
```
crispy-wonton: Would be useful to add the CRS of the output geo dataframe as we have multiple CRS in this repo

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+    enw["firm_capacity_mva"] = (
+        enw["maximum_demand_mva_per_primary_substation"] + enw["headroom_mva"]
+    )
+    enw["operator"] = "ENW"
+    enw["geopoint"] = enw["geopoint"].apply(_parse_binary_geometry)
+
+    # Create GeoDataFrame and perform spatial join with shape data
+    enw_gdf = gpd.GeoDataFrame(enw, geometry="geopoint", crs=CRS)
+    enw_shape["geo_shape"] = enw_shape["geo_shape"].apply(_parse_binary_geometry)
+    enw_shape["geo_point_2d"] = enw_shape["geo_point_2d"].apply(_parse_binary_geometry)
```
crispy-wonton: What are 'geo_shape' and 'geo_point_2d'? polygon and point geoms?

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+
+
+def generate_spen_df() -> gpd.GeoDataFrame:
+    """
+    Generate a GeoDataFrame for Scottish Power Energy Networks (SPEN) substations.
+
+    Returns:
+        GeoDataFrame with SPEN substation data.
+    """
+    spen_spd_substations = load_s3_data(
```
crispy-wonton: what's spd vs spm? can you add a note to the code to explain what they mean

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+
+    # Process UKPN data
+    ukpn_primary_substations["firm_capacity_mva"] = ukpn_primary_substations.apply(
+        lambda x: (
+            x["firmcapacitywinter"]
+            if x["seasonofconstraint"] == "Winter"
+            else x["firmcapacitysummer"]
+        ),
+        axis=1,
+    )
```
crispy-wonton: cool

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+    ukpn_primary_substations["headroom_mva"] = (
+        ukpn_primary_substations["firm_capacity_mva"]
+        * ukpn_primary_substations["demand"]
+        / 100
```
crispy-wonton: why is this the calculation to get headroom? please can you add notation in the code :)

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+    return pd.concat(
+        [
+            generate_enw_df(),
+            generate_npg_df(),
+            generate_spen_df(),
+            generate_ssen_df(),
+            generate_ukpn_df(),
+            generate_wpd_df(),
+        ]
+    )
```
crispy-wonton: nice, i like how clean this is

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+        crs="EPSG:27700",
+    ).to_crs(CRS)
+
+
+def generate_substations_df() -> gpd.GeoDataFrame:
+    """
+    Generate a combined GeoDataFrame for all grid operators' substations.
+
+    Returns:
+        GeoDataFrame with combined substation data from all operators.
```
crispy-wonton: add CRS

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+        ]
+    )
+
+
+def aggregate_oa_to_lsoa(oa_gdf: gpd.GeoDataFrame) -> gpd.GeoDataFrame:
+    """
+    Aggregate Output Area (OA) geometries to Lower Layer Super Output Area (LSOA) level.
+
+    Args:
+        oa_gdf: GeoDataFrame containing OA-level data with LSOA codes.
```
crispy-wonton: so is this input meant to be OA polygons with LSOA codes?

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+        raise ValueError(
+            "Input GeoDataFrame must have a defined coordinate reference system (CRS)"
+        )
+
+    lsoa_gdf = oa_gdf.dissolve(by="LSOA21CD", as_index=False)
+    return lsoa_gdf.reset_index(drop=True)
+
+
+def distribute_substation_headroom(
+    substations_df: gpd.GeoDataFrame,
```
crispy-wonton:
```suggestion
    substations_gdf: gpd.GeoDataFrame,
```

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+
+    Returns:
+        GeoDataFrame aggregated to LSOA level.
+    """
+    if oa_gdf.crs is None:
+        raise ValueError(
+            "Input GeoDataFrame must have a defined coordinate reference system (CRS)"
+        )
+
+    lsoa_gdf = oa_gdf.dissolve(by="LSOA21CD", as_index=False)
```
crispy-wonton: something like `dissolve_by` could be an arg of the function with `LSOA21CD` as default to make it more flexible

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+
+    # Perform spatial join between LSOAs and substations
+    joined = gpd.sjoin(
+        lsoa_with_households, substations_df, how="inner", predicate="intersects"
+    )
+
+    # Calculate total households served by each substation
+    substation_total_households = (
+        joined.groupby("index_right")["household_count"].sum().reset_index()
+    )
```
crispy-wonton: i think it could be more explicit that `index_right` is the substation IDs, maybe this col can be renamed especially as it's used again below.

also, could this double count households, e.g. if an LSOA gets joined to two substations?

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+
+    # Calculate percentage of households that could install heat pumps
+    lsoa_data["heatpump_installation_percentage"] = (
+        lsoa_data["max_heatpumps"] / lsoa_data["household_count"] * 100
+    ).clip(upper=100, lower=0)
+
+    # Calculate excess or deficit capacity
+    lsoa_data["capacity_difference_kw"] = lsoa_data["headroom_kw"] - (
+        lsoa_data["household_count"] * power_per_heatpump
+    )
```
crispy-wonton, line 237: nice

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
crispy-wonton: I would suggest posting the numbers you've used in the team slack channel to get feedback on assumptions related to power for HPs :) they'll know best!

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+
+def calculate_grid_capacity() -> pd.DataFrame:
+    """
+    Calculate the grid capacity for heat pump installations across all LSOAs.
+
+    This function performs the following steps:
+    1. Generate and process substation data from all grid operators
+    2. Load and process LSOA and household data
+    3. Distribute substation headroom to LSOAs
+    4. Assess heat pump suitability based on available capacity
```
crispy-wonton, line 250: this is really helpful!

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+from asf_heat_pump_suitability.getters import get_datasets
+
+# Constants
+CRS = "EPSG:4326"  # Geometry coordinate reference system - standard longitude/latitude projection
+TAKEUP_RATE = 30  # Percentage of households in each LSOA to install heat pumps
+POWER_PER_HEATPUMP = 8  # Power rating per heat pump in kW
+COEFFICIENT_OF_PERFORMANCE = (
+    2.5  # Heat output (heat pump rating) / electricity consumed
+)
+SUBSTATION_SCALING_FACTOR = 1.0  # Factor to scale substation power rating (MVA) by
```
crispy-wonton, line 28: Would be great to get ASF team colleagues inputs on these constants to check they're appropriate

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+    # Generate and process substation data
+    substations = generate_substations_df()
+    substations["firm_capacity_mva"] = substations["firm_capacity_mva"].apply(
+        _parse_capacity
+    )
+    substations["peak_demand_mva"] = substations["peak_demand_mva"].apply(
+        _parse_capacity
+    )
+
+    # Apply substation rating scaling factor
```
crispy-wonton, line 264: What is the scaling factor for?

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
crispy-wonton: dont think we've used LSOA geometries except for plotting in Flourish, where I think we used these boundaries: s3://asf-heat-pump-suitability/source_data/Lower_layer_Super_Output_Areas_2021_EW_BFE_V9_-9107090204806789093/

It looks like you've used BGC boundaries (from the file name) and we've used BFE which are the most detailed. I'm not sure it matters much for this use case. @danlewis85 do you have an opinion?

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+    headroom_df = process_substation_lsoa_data(substations, lsoa_gdf, households_df)
+
+    # Assess heat pump suitability
+    consumption_per_heatpump = POWER_PER_HEATPUMP / COEFFICIENT_OF_PERFORMANCE
+    result = assess_heatpump_suitability(
+        headroom_df, consumption_per_heatpump, TAKEUP_RATE
+    )
+
+    # Rename LSOA column for consistency
+    result = result.rename(columns={"LSOA21CD": "lsoa21"})
```
crispy-wonton:
```suggestion
    result = result.rename(columns={"LSOA21CD": "lsoa"})
```

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,758 @@
...
+    print(f"Total LSOAs assessed: {len(grid_capacity_results)}")
+    print(
+        f"LSOAs with sufficient grid capacity: {grid_capacity_results['has_grid_capacity'].sum()}"
+    )
+    print(
+        f"Percentage of LSOAs with sufficient capacity: {(grid_capacity_results['has_grid_capacity'].mean() * 100):.2f}%"
+    )
+
+    # Optionally, save results to a file
+    # grid_capacity_results.to_csv("grid_capacity_assessment_results.csv", index=False)
```
crispy-wonton: you can put this as an arg to the script. here's my arg parser you can use as a baseline if you wanted to add it to save you the time:

```
def parse_arguments() -> argparse.Namespace:
    """
    Create ArgumentParser and parse.

    Returns:
        argparse.Namespace: populated `Namespace`
    """
    parser = ArgumentParser()
    parser.add_argument(
        "--save_as",
        help="Path to save grid capacity results to. If unspecified, save with default filename.",
        type=str,
        required=False,
        default=None,
    )

    return parser.parse_args()
```
Then in script you can set the default name.

crispy-wonton (commented):
Thanks so much for working on this @shabrf ! You've clearly done your research and put a lot of thought into it.

I've made some suggestions on documentation and left some questions for my understanding (and others if we need to come back to update this at some point in the future).

If you haven't already, it would be good to take some of the assumptions about constants and the grid to ASF colleagues to confirm they're appropriate.

Note, in the repo we're using `lsoa` instead of `lsoa21` as the LSOA code column. This is made up of LSOA21 and LSOA11 codes.

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
shabrf: The substation data for this provider was only available in a forecasting dataset which had different estimated per year on grid enhancements, there were different forecasts, including "Leading the way" (optimistic forecast) and "Falling short" (pessimistic). "Best view" was the middle ground scenario. The forecast was  based on 2023 data so there isn't actually that much variation between the views since it's only a 1 year difference.

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
shabrf: This scales the substation capacities used in the calculation by a uniform factor e.g. to calculate the suitability for a 10% increase in capacity for all substations, the substation scaling factor would be 1.1.

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
shabrf: Thanks for picking up on this Liz! It wasn't expected and fixed now by dropping duplicates

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
shabrf: Yes that's what they are, that was the terminology used in the source datasets

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -20,6 +20,24 @@ data_source:
   E_historicengland_listed_buildings: "s3://asf-heat-pump-suitability/source_data/Jun2024_vJul2024_HistoricEngland_listedbuilding_E.gpkg"
```
crispy-wonton: Some of these are missing the geocoverage prefix of the config source data naming convention:
_geocoverage_source_name_ - EWS (England / Wales/Scotland); GB; UK; or global.
The last 4 of WPD sources and also need a name. It's not immediately obvious what data they contain.
Btw, to confirm, 'primaries' refer to areas a substation serves?

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -77,490 +57,54 @@ def _parse_capacity(value: Any) -> float:
...
-
-    Returns:
-        GeoDataFrame with combined substation data from all operators.
+        gpd.GeoDataFrame: DataFrame with EPSG:4326 (WGS84) projection containing:
+            - id: Unique substation identifier (format varies by operator)
+            - firm_capacity_mva: Maximum power the substation can safely deliver (MVA)
+            - peak_demand_mva: Maximum observed power demand at the substation (MVA)
+            - geo_shape: Polygon geometry representing the substation's service area
+            - operator: Distribution network operator code
+                       (one of: "ENW", "NPg", "SPEN", "SSEN", "UKPN", "WPD")
```
crispy-wonton, line 85: Love this

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -715,11 +259,9 @@ def calculate_grid_capacity() -> pd.DataFrame:
     )
 
     # Load and process LSOA boundary data
-    lsoa_boundaries = load_s3_data(
-        "asf-heat-pump-suitability",
-        "source_data/Output_Areas_2021_EW_BGC_V2_4299916833741807639.geojson",
-    )
-    lsoa_gdf = aggregate_oa_to_lsoa(lsoa_boundaries)
+    lsoa_gdf = gpd.read_file(
+        "s3://asf-heat-pump-suitability/source_data/Lower_layer_Super_Output_Areas_2021_EW_BFE_V9_-9107090204806789093/LSOA_2021_EW_BFE_V9.shp"
```
crispy-wonton: This dataset can go in the config

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -732,27 +274,44 @@ def calculate_grid_capacity() -> pd.DataFrame:
...
-    print(
-        f"Percentage of LSOAs with sufficient capacity: {(grid_capacity_results['has_grid_capacity'].mean() * 100):.2f}%"
+        "Average LSOA installation percentage: "
+        + str(grid_capacity_results["heatpump_installation_percentage"].mean())
     )
 
-    # Optionally, save results to a file
-    # grid_capacity_results.to_csv("grid_capacity_assessment_results.csv", index=False)
+    if args.save_as:
+        grid_capacity_results.to_csv(args.save_as, index=False)
```
crispy-wonton, line 328: What dir is this saving to?

asf_heat_pump_suitability/getters/get_dno_datasets.py
```diff
@@ -0,0 +1,467 @@
...
+CRS = "EPSG:4326"  # Geometry coordinate reference system - standard longitude/latitude projection
+
+
+def generate_enw_gdf() -> gpd.GeoDataFrame:
+    """
+    Generate a GeoDataFrame for Electricity North West (ENW) primary substations and their service areas.
+
+    The function processes substation data from ENW's Distribution Future Energy Scenarios (DFES)
+    and Network Development Plan (NDP) datasets. It combines point locations of substations with
+    their corresponding service area polygons (Voronoi polygons) and capacity/demand information.
```
crispy-wonton, line 26: This is super helpful thank you!

asf_heat_pump_suitability/getters/get_dno_datasets.py
```diff
@@ -0,0 +1,467 @@
...
+        config["data_source"]["ENW_dfes_primaries"]
+    ).to_pandas()
+    enw_substations = base_getters.get_df_from_parquet_s3_path(
+        config["data_source"]["ENW_ndp_headroom"]
+    ).to_pandas()
+    enw_shape = base_getters.get_df_from_parquet_s3_path(
+        config["data_source"]["ENW_ndp_voronoi"]
+    ).to_pandas()
+
+    # Filter data for primary substations, current year, and best view scenario
```
crispy-wonton, line 46:
```suggestion
    # Filter data for primary substations, current year, and best view scenario
    # Best view scenario represents middle ground future projections of demand on substation
```
Feel free to correct it as appropriate but something like this will help me understand the code later!

asf_heat_pump_suitability/getters/get_dno_datasets.py
```diff
@@ -0,0 +1,467 @@
...
+    enw["geopoint"] = enw["geopoint"].apply(parse_binary_geometry)
+
+    # Create GeoDataFrame and perform spatial join with shape data
+    enw_gdf = gpd.GeoDataFrame(enw, geometry="geopoint", crs=CRS)
+    # polygon geometry of substation distribution area
+    enw_shape["geo_shape"] = enw_shape["geo_shape"].apply(parse_binary_geometry)
+    # point geometry of substation location
+    enw_shape["geo_point_2d"] = enw_shape["geo_point_2d"].apply(parse_binary_geometry)
+    enw_shape_gdf = gpd.GeoDataFrame(enw_shape, geometry="geo_shape", crs=CRS)
+    enw_df = gpd.sjoin(enw_gdf, enw_shape_gdf, how="right", predicate="intersects")
```
crispy-wonton: Do you know if we get any duplicates here? E.g. if a substation is on the boundary of 2 distribution areas then it could get joined to both.

If that's the case then I think you were right before to use `within` rather than `intersects`, sorry about that. But i would expect that to be edge cases anyway.

asf_heat_pump_suitability/getters/get_dno_datasets.py
```diff
@@ -0,0 +1,467 @@
...
+    ).to_pandas()
+    spen_spd_shape = base_getters.get_df_from_parquet_s3_path(
+        config["data_source"]["SPEN_spd_polygons"]
+    ).to_pandas()
+    spen_spm_shape = base_getters.get_df_from_parquet_s3_path(
+        config["data_source"]["SPEN_spm_polygons"]
+    ).to_pandas()
+
+    # Process SPM data
+    spm_df = spen_spm_substations.merge(spen_spm_shape, how="left", on="primary_group")
```
crispy-wonton, line 221: To check, what does "primary_group" mean? (sorry if I already asked, I couldnt find it in the existing comments). Please could you add a note?

asf_heat_pump_suitability/getters/get_dno_datasets.py
```diff
@@ -0,0 +1,467 @@
...
+    )
+    enw["operator"] = "ENW"
+    enw["geopoint"] = enw["geopoint"].apply(parse_binary_geometry)
+
+    # Create GeoDataFrame and perform spatial join with shape data
+    enw_gdf = gpd.GeoDataFrame(enw, geometry="geopoint", crs=CRS)
+    # polygon geometry of substation distribution area
+    enw_shape["geo_shape"] = enw_shape["geo_shape"].apply(parse_binary_geometry)
+    # point geometry of substation location
+    enw_shape["geo_point_2d"] = enw_shape["geo_point_2d"].apply(parse_binary_geometry)
```
crispy-wonton: Minor, but why do we parse and keep 'geo_point_2d' here? It doesnt look like it gets used

asf_heat_pump_suitability/getters/get_dno_datasets.py
```diff
@@ -0,0 +1,467 @@
...
+            x["firmcapacitywinter"]
+            if x["seasonofconstraint"] == "Winter"
+            else x["firmcapacitysummer"]
+        ),
+        axis=1,
+    )
+
+    # Calculate the headroom
+    # The "demand" column is the percentage of total capacity that the substation has available at peak load
+    # So to calculate the actual value in MVA, we need to multiply demand headroom by capacity
```
crispy-wonton, line 366: Nice! Thanks!

asf_heat_pump_suitability/getters/get_dno_datasets.py
```diff
@@ -0,0 +1,467 @@
...
+            "primary",
+            "primarysubstationname",
+            "firm_capacity_mva",
+            "maximum_load_mva",
+            "headroom_mva",
+            "geo_point_2d",
+            "geo_shape",
+        ]
+    ]
+    ukpn["geo_point_2d"] = ukpn["geo_point_2d"].apply(parse_binary_geometry)
```
crispy-wonton: Same as above, doesn't look like geo_point_2d gets used

crispy-wonton (commented):
Hi @shabrf 
Thanks so much for adding the new documentation. It's much easier to follow the code and understand what's going on so thank you for that! I've just left a couple more clarification questions and places to annotate further , to make future maintenance and update of this code easier.
There is a minor point about `geo_point_2d` seemingly not being used in some places and I think you were right to use `within` on your spatial joins initially although `intersects` could still be appropriate if it doesnt generate duplicates.

Thank you!!

asf_heat_pump_suitability/getters/get_dno_datasets.py
shabrf: Primary groups are groups of 1-3 geographically close primary substations. This grouping type only seems to be used by SPEN specifically for the SPM region. It's also used by ENW. 

Thanks for raising this, it's made me notice that the substation capacity data for SPEN-SPM and ENW hadn't been aggregated to the the primary group level before the spatial join so all the primary substations within each group have the same geometry data which effects the headroom distribution calculation later on. I've fixed by aggregating by primary group prior to the spatial join.

asf_heat_pump_suitability/config/base.yaml
shabrf: Primaries refers to primary substations as opposed to secondary (lower level which only 1 DNO has data on) or grid (higher level) substations.

asf_heat_pump_suitability/getters/get_dno_datasets.py
shabrf: The point geometry should be close to the centre of the distribution area since the dataframe containing the distribution areas has a row per substation. I've tested with both joins and confirmed that there isn't any difference in the resulting dataframes so we're free of these edge cases.

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: Great thank you!

asf_heat_pump_suitability/getters/get_dno_datasets.py
crispy-wonton: Thanks for checking with the different joins! Sounds good.

asf_heat_pump_suitability/getters/get_dno_datasets.py
```diff
@@ -68,35 +69,44 @@ def generate_enw_gdf() -> gpd.GeoDataFrame:
...
+                "firm_capacity_mva",
+                "maximum_demand_mva_per_primary_substation",
+                "geo_shape",
+                "operator",
+            ]
+        ].rename(
+            columns={
+                "substation_id": "id",
+                "maximum_demand_mva_per_primary_substation": "peak_demand_mva",
+                "geo_shape": "geo_shape",
```
crispy-wonton: Should `"substation_id"` in this dict be `"pry_group"` now? Also, i dont think you need to rename `geo_shape`

asf_heat_pump_suitability/getters/get_dno_datasets.py
```diff
@@ -146,14 +156,13 @@ def generate_npg_gdf() -> gpd.GeoDataFrame:
...
-    npg_demand["geo_point_2d"] = npg_demand["geo_point_2d"].apply(parse_binary_geometry)
     npg_demand["geo_shape"] = npg_demand["geo_shape"].apply(parse_binary_geometry)
-    npg_shape = npg_demand[["geo_point_2d", "geo_shape"]].drop_duplicates()
+    npg_shape = npg_demand["geo_shape"].drop_duplicates()
 
     # Create GeoDataFrames and perform spatial join
     npg_gdf = gpd.GeoDataFrame(npg, geometry="location", crs=CRS)
     npg_shape_gdf = gpd.GeoDataFrame(npg_shape, geometry="geo_shape", crs=CRS)
-    npg_df = gpd.sjoin(npg_shape_gdf, npg_gdf, how="left", predicate="intersects")
+    npg_df = gpd.sjoin(npg_shape_gdf, npg_gdf, how="left", predicate="within")
```
crispy-wonton:
```suggestion
    npg_df = gpd.sjoin(npg_shape_gdf, npg_gdf, how="left", predicate="contains")
```
There is a bug here which can be fixed by changing the predicate I think. Currently, the function returns a df of len 0. I think the current predicate is trying to polygons from `npg_shape_gdf` located _within_ point geoms in `npg_gdf` which is not matching anything.

crispy-wonton: This also makes me think we could add an AssertionError in `grid_capacity.generate_substations_gdf` (or something similar) to raise an error if any of the expected DNOs are missing from the `operator` column in the final `gdf`. Otherwise this could be a silent fail.

asf_heat_pump_suitability/getters/get_dno_datasets.py
crispy-wonton: Later, when the DNO gdfs are concatenated in `grid_capacity.generate_substations_gdf` we end up with `pry_group` col and no `id` col. The `pry_group` col has nan values for the other DNOs.

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
```diff
@@ -0,0 +1,316 @@
...
+    lsoa_with_households = lsoa_gdf.merge(households_df, on="LSOA21CD", how="left")
+
+    # Perform spatial join between LSOAs and substations
+    joined = gpd.sjoin(
+        lsoa_with_households, substations_gdf, how="inner", predicate="intersects"
+    ).rename(columns={"index_right": "substation_id"})
+
+    # Calculate total households served by each substation
+    substation_total_households = (
+        joined.groupby("substation_id")["household_count"].sum().reset_index()
```
crispy-wonton, line 137: Should this use the existing `id` column from the `substations_gdf` instead of the index? I guess it should be the same and maybe it's better (if there are null ids). Just wanted to check though.

crispy-wonton (commented):
There's a bug in `generate_npg_gdf` and in `generate_enw_gdf` which I have flagged.
I've also left a final question in `grid_capacity.py`.

Once these are fixed, it looks good to merge.

asf_heat_pump_suitability/pipeline/prepare_features/grid_capacity.py
shabrf: Yup that's probably more descriptive, there's no null values for the id column in substation_gdf so happy to make this change.

crispy-wonton (dismissed):

PR #81, opened 2024-11-06
adding file for calculating comparison of hn zones as well as plottin…

### LINKED ISSUE #80: Comparing the HN zones selected by DESNZ to the HP suitability dataset
We want to compare the DESNZ HN zones to the HN scores from our HP suitability dataset as validation.

helloaidank:
---

# Description
All files are contained within the `asf_heat_pump_suitability/analysis/hn_zones` directory.

The script `comparison_of_hn_zones.py` processes and compares DESNZ heat network (HN) to Nesta's heat pump suitability data for Liverpool, and outputs some files for plotting.

Steps taken:
1. Load DESNZ HN GeoData and perform a spatial join with LSOA polygons.
2. Process Nesta heat pump suitability scores for Liverpool.
3. Identify LSOAs in DESNZ HN list not in HP suitability data.
4. Calculate and log average scores and Mean Absolute Error (MAE).
5. Save processed data to files.

Outputs:
- 'liverpool_with_desnz_hn_lsoa.gpkg': GeoDataFrame with LSOA codes added to HN zones.
- 'liverpool_hp_suitability_lsoas.json': List of unique LSOA codes in Liverpool.
- 'liverpool_hp_suitability_scores_with_desnz.parquet': Processed suitability scores with DESNZ pilot scores and Mean Absolute Error (MAE).
- 'script_output.log': Log file with metrics such as avg HN scores and MAE 

The script `comparison_of_hn_zones.py` visualises heat network (HN) and heat pump suitability data for Liverpool.
Steps taken:
1. Load and preprocess data:
   - Read Parquet and GeoPackage files.
   - Extract unique LSOA geometries.
   - Load LSOA geometries and convert CRS if needed.
   - Filter Liverpool LSOAs.
2. Plot Liverpool LSOA geometries and save as PNG.
3. Merge suitability scores with geometries and convert to GeoDataFrame.
4. Plot and save overlay of DESNZ Pilot Score = 1 on Liverpool heat network zones.
5. Define and use a function to plot absolute error maps by DESNZ Pilot Score, saving the plots in the output plots directory. 

Fixes #80 

# Instructions for Reviewer

In order to test the code in this PR you need to run the following commands:
`python asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py`

This should output a few data files to a directory called `output_data`, please check that they are there. If you could see how long this script will take as we have to scale it to ~20 other local authorities. For me it takes 30s-1 minute to run, which seems OK as we likely won't be running this repeatedly (i.e. it's good enough) but any major efficiency improvements are always welcome. 

Then to visualise some of this data, run the following:
`python asf_heat_pump_suitability/analysis/hn_zones/plot_comparison_of_hn_zones.py`
This will output some visuals to `output_plots` for the Liverpool LA. 

Please pay special attention to whether the outputs/data we are getting seem sensible and that the MAE checks out, this is the primary motivation for this PR. 

Two things to be aware of regarding the methodology and visuals:

- For the DESNZ Liverpool HN zones, a couple of the LSOAs are actually outside the LA of Liverpool and at the moment these are not taken into account in the script. There could be a case to include them in an exception but for the moment I have left them out.
- The shape file with the UK LSOAs geometries that I do read in include land and water, whereas the DESNZ HN Zones are only for land. In terms of the absolute error, this does not make a difference (I don't think), but it does mean that the visuals for the DESNZ HN Zones look different to the Liverpool LSOA map as shown by the overlay plot.

For the next evolution of this script, we will scale it to the other 20 LAs we have and I think it's probably a good call to put these file paths into a config file. I am also curious as to where you think this should live in the repo!

Thanks so much for taking the time to review! It's much appreciated. 

# Checklist:

- [X] I have refactored my code out from `notebooks/`
- [X] I have checked the code runs
- [ ] I have tested the code
- [X] I have run `pre-commit` and addressed any issues not automatically fixed
- [X] I have documented the code
  - [X] Major functions have docstrings
- [X] I have explained this PR above
- [X] I have requested a code review


asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,350 @@
...
+
+
+# Define file paths
+LIVERPOOL_GPKG_PATH = "s3://asf-heat-pump-suitability/heat_network_desnz_data/heat-network-zone-map-Liverpool.gpkg"
+LSOA_SHP_PATH = "s3://asf-heat-pump-suitability/source_data/Lower_layer_Super_Output_Areas_2021_EW_BFE_V9_-9107090204806789093/LSOA_2021_EW_BFE_V9.shp"
+NESTA_HP_SUITABILITY_PARQUET_PATH = "s3://nesta-open-data/asf_heat_pump_suitability/2023Q4/20240925_2023_Q4_EPC_heat_pump_suitability_per_lsoa.parquet"
+LA_TO_ANALYSE = "Liverpool"
+
+
+def setup_logging_and_file_path(
```
crispy-wonton, line 66: Nice, this is really useful!

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,350 @@
...
+    """
+    # Ensure the output directory exists
+    os.makedirs(output_dir, exist_ok=True)
+
+    # Define the full path to the log file
+    log_file_path = os.path.join(output_dir, log_filename)
+
+    # Clear existing logging handlers if any
+    for handler in logging.root.handlers[:]:
+        logging.root.removeHandler(handler)
```
crispy-wonton: According to this thread on stackoverflow, it's better to use `logger.handlers.clear()` https://stackoverflow.com/questions/7484454/removing-handlers-from-pythons-logging-loggers

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,350 @@
...
+    if desnz_hn_gdf.crs != desnz_hn_gdf.crs:
+        lsoa_gdf = lsoa_gdf.to_crs(desnz_hn_gdf.crs)
+
+    # Perform spatial join to match Liverpool polygons with LSOA codes
+    joined_gdf = gpd.sjoin(
+        desnz_hn_gdf,
+        lsoa_gdf[["geometry", "LSOA21CD"]],
+        how="left",
+        predicate="intersects",
+    )
```
crispy-wonton:
```suggestion
    joined_gdf = desnz_hn_gdf.sjoin(
        lsoa_gdf[["geometry", "LSOA21CD"]],
        how="left",
        predicate="intersects",
    )
```

This is buried in the `geopandas` documentation but it's suggested to use `sjoin` as a method rather than function: https://geopandas.org/en/stable/docs/user_guide/mergingdata.html#spatial-joins

> For historical reasons, both methods are also available as top-level functions [sjoin()](https://geopandas.org/en/stable/docs/reference/api/geopandas.sjoin.html#geopandas.sjoin) and [sjoin_nearest()](https://geopandas.org/en/stable/docs/reference/api/geopandas.sjoin_nearest.html#geopandas.sjoin_nearest). It is recommended to use methods as the functions may be deprecated in the future.

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,350 @@
...
+    la_unique_lsoas = la_filtered_scores["LSOA21CD"].unique().to_list()
+
+    return la_filtered_scores, la_unique_lsoas
+
+
+def add_desnz_pilot_score(
+    la_hp_suitability_scores: pl.DataFrame, list_of_desnz_hn_lsoas: List[str]
+) -> Tuple[pl.DataFrame, float, float]:
+    """
+    Add a 'DESNZ_pilot_score' column to the Nesta heat pump suitability data and compute average scores.
```
crispy-wonton: What are the average scores computed here? Is it average Nesta HN score per local authority? Please can you add it to the documentation :)

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,350 @@
...
+        predicate="intersects",
+    )
+
+    # Extract unique LSOA codes from the joined GeoDataFrame
+    desnz_hn_unique_lsoas = joined_gdf["LSOA21CD"].dropna().unique().tolist()
+
+    return joined_gdf, desnz_hn_unique_lsoas
+
+
+def process_nesta_hp_suitability_scores(
```
crispy-wonton: Might be worth renaming with something to indicate that this function filters by LA

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,350 @@
...
+
+
+def add_desnz_pilot_score(
+    la_hp_suitability_scores: pl.DataFrame, list_of_desnz_hn_lsoas: List[str]
+) -> Tuple[pl.DataFrame, float, float]:
+    """
+    Add a 'DESNZ_pilot_score' column to the Nesta heat pump suitability data and compute average scores.
+
+    Args:
+        la_hp_suitability_scores (pl.DataFrame): Data containing 'LSOA21CD' and 'HN_N_avg_score_weighted' columns.
```
crispy-wonton: To clarify, this df should be for a single LA?

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,350 @@
...
+            - Average 'HN_N_avg_score_weighted' for rows where 'DESNZ_pilot_score' is 0.
+    """
+    # Add 'DESNZ_pilot_score' column: 1 if LSOA is in DESNZ pilot zone, else 0
+    la_hp_suitability_scores = la_hp_suitability_scores.with_columns(
+        pl.col("LSOA21CD")
+        .is_in(list_of_desnz_hn_lsoas)
+        .cast(pl.Int8)
+        .alias("DESNZ_pilot_score")
+    )
+    # Calculate average scores where 'DESNZ_pilot_score' == 1 and == 0
```
crispy-wonton: To clarify, here we are calculating our average Nesta HN score across all LSOAs in all DESNZ pilot zones and separately for all non-pilot zones?

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,350 @@
...
+        la_hp_suitability_scores, pilot_score=0
+    )
+    return la_hp_suitability_scores, avg_hn_score_pilot_1, avg_hn_score_pilot_0
+
+
+def _calculate_hn_pilot_average_score(
+    la_hp_suitability_scores: pl.DataFrame, pilot_score: int
+) -> float:
+    """
+    Calculate the average 'HN_N_avg_score_weighted' for a specified 'DESNZ_pilot_score' value.
```
crispy-wonton: I would maybe reword this to make it a bit clearer. Something like:
```suggestion
    Calculate the average Nesta Heat Network score for LSOAs in (`DESNZ_pilot_score == 1`) or not in  (`DESNZ_pilot_score == 0`) DESNZ heat network pilot areas. 
```

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,350 @@
...
+
+    # Filter and calculate the average score
+    filtered_la_hp_suitability_scores = la_hp_suitability_scores.filter(
+        pl.col("DESNZ_pilot_score") == pilot_score
+    )
+    avg_hn_score_pilot = (
+        filtered_la_hp_suitability_scores.select(
+            pl.col("HN_N_avg_score_weighted").mean()
+        ).item()
+        if filtered_la_hp_suitability_scores.height > 0
```
crispy-wonton: nice , I didn't know dfs had a `height` attribute!

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,350 @@
...
+    # Calculate the mean of the absolute error
+    mae = hp_suitability_scores_with_error["absolute_error"].mean()
+    logging.info(
+        f"Mean Absolute Error (MAE) for {desnz_col} vs {nesta_hn_score_col}: {mae}"
+    )
+
+    return hp_suitability_scores_with_error, mae
+
+
+def main():
```
crispy-wonton: I was using `main()` and `run()` functions previously. I like how clean they are but the downside is they don't work well with `python -i` when debugging or experiencing other issues. This is because if the `main()` function is interrupted then there are no variables saved to memory to investigate.

You could just replace the `def main():` line with `if __name__ == "__main__"` if you want to remove it.

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,350 @@
...
+    # Define the output directory and set up logging
+    output_dir = os.path.join(
+        PROJECT_DIR, "asf_heat_pump_suitability/analysis/hn_zones/output_data/"
```
crispy-wonton: I would suggest moving the hn_zone dir to `outputs/hn_zones` and then just save the results directly into that dir :)

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,350 @@
...
+    )
+    setup_logging_and_file_path(output_dir)
+
+    # Load the data and perform spatial join
+    liverpool_with_desnz_hn_lsoa, list_of_liverpool_desnz_hn_lsoas = load_hn_geodata(
+        LIVERPOOL_GPKG_PATH, LSOA_SHP_PATH
+    )
```
crispy-wonton: I think it's good practice to name your arguments in custom functions as an extra check that params are input in the correct order :) I would add those here and below.

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,350 @@
+"""
```
crispy-wonton, line 1: First of all, this is great work. The code is clear and neat and I really like the logging. Great job!

Below are the logs from running the script. The results of average HN score for DESNZ pilot 0 and 1 seem very similar and with really high mean error rates. At least 0 has a lower score on avg than 1, but they are so close together and also both very central values at 0.52-0.61. If I saw those results out of context I don't think I would interpret that 0.52 is unsuitable for a HN but 0.61 really is.

It might be worth looking at the underlying data for the suitability features for these LSOAs and try to identify if they have anything in common that we seem to be misclassifying.

[Looking at the relevant features for Heat Networks in the model](https://github.com/nestauk/asf_heat_pump_suitability?tab=readme-ov-file#heat-pump-suitability-scores), they are: not a listed building; not in building conservation zone (bear in mind a lot of properties will have a NAN flag here because of the patchiness of the conservation zone data. We assigned those not in a cons zone as NAN because we couldn't know if they were in one or if there was just cons data missing). ; off-gas; property in building with other properties (i.e. a flat); and in an urban area.

One reason that they have such similar scores could be that a lot of properties will score very similarly in a city even with variation between them. E.g. property density doesn't count for HN so there's no scale to density except urban/rural binary, and presumably all the LSOAs in this sample are urban. I'm guessing most properties in Liverpool will be not listed buildings and not off-gas(?). Those 3 features alone give a score of 0.47 (2.25/4.75). But still we would expect those on the higher scores to be around 0.9 if they are flats as well, which would put the score at 0.89 (4.25/4.75).

I think it might be true what you said in one of the stand ups, that small parts of LSOAs that have a tiny overlap with the DESNZ zones might be affecting the results. Maybe we need to set a threshold for the amount of overlap with DESNZ zone required to be considered in the analysis. We can start by setting it at 50% area (or 50% of OAs in an LSOA).

NFO: Loaded Liverpool with DESNZ HN LSOA data.
INFO: Created 133 records
INFO: Processed Nesta HP suitability scores for Liverpool.
INFO: number of LSOAs in DESNZ list not in HP suitability data: 2
INFO: Average HN_N_avg_score_weighted: 0.5608973509933775
INFO: Avg HN_N_avg_score_weighted for DESNZ_pilot_score = 1: 0.6132439024390244
INFO: Avg HN_N_avg_score_weighted for DESNZ_pilot_score = 0: 0.524927374301676
INFO: Mean Absolute Error (MAE) for DESNZ_pilot_score vs HN_N_avg_score_weighted: 0.46865231788079464
INFO: Mean Absolute Error (MAE) for DESNZ_pilot_score = 1: 0.3867560975609756
INFO: Mean Absolute Error (MAE) for DESNZ_pilot_score = 0: 0.524927374301676

crispy-wonton (changes requested):
Hey @helloaidank. 
I've reviewed your `comparison_of_hn_zones.py` script and left comments. I need to come back and review the plotting script but wanted to publish these comments already so you can start scaling up :)

The code looks great! I do think we need to change the methodology slightly to exclude any small bits of LSOAs that might be skewing the error results because they look so similar for DESNZ pilot and non-pilot zones. And we want to make sure our results are representative of the same/similar locations so that we can assess them. If they still come out like this then maybe the features for HN model need to be improved!
Let me know if you have questions !

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
helloaidank: Ah ok! This is because it's a list and therefore iterating and mutating over the list is a no no as it may introduce a bug with e.g. indexing

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
helloaidank: Good spot!

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
helloaidank: Yas queen! It's the average Nesta HN per LA. I've updated the doc string to reflect that.

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
helloaidank: renamed to `filter_la_hp_scores`

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
helloaidank: Yep, it's for a single LA. I've added that now!

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
helloaidank: Yep that is correct, I've edited the comment to make that clearer!

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
helloaidank: Thanks for this, great suggestion for the doc string!

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
helloaidank: Ah ok, I haven't gotten into the habit of using `python -i` yet so didn't run into this as an issue. I've now changed it, likely I'll have to define these lines of code as a function (with a different name to main) anyway when scaling.

asf_heat_pump_suitability/analysis/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,225 @@
...
+import os
+import logging
+from asf_heat_pump_suitability import PROJECT_DIR
+
+# Set up logging configuration
+logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
+
+# Define constants for file paths
+INPUT_DIR = os.path.join(
+    PROJECT_DIR, "asf_heat_pump_suitability/analysis/hn_zones/output_data/"
```
crispy-wonton: same as above - can move this to the main outputs dir :)

asf_heat_pump_suitability/analysis/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,225 @@
...
+    except Exception as e:
+        logging.error(f"Error loading data: {e}")
+        raise
+
+
+def preprocess_data(
+    liverpool_hp_suitability_lsoas: list, lsoa_shp_path: str = LSOA_SHP_PATH
+) -> gpd.GeoDataFrame:
+    """
+    Preprocess the data by extracting unique LSOA geometries and filtering Liverpool LSOAs.
```
crispy-wonton:
```suggestion
    Load and preprocess LSOA geospatial data by extracting unique LSOA geometries and filtering Liverpool LSOAs.
```

asf_heat_pump_suitability/analysis/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,225 @@
...
+    fig, ax = plt.subplots(figsize=(12, 6))
+    liverpool_lsoa_geometries_gdf.plot(ax=ax)
+    plt.title("Liverpool LSOA Geometries")
+    plt.axis("off")
+    plt.savefig(os.path.join(output_dir, "liverpool_data_lsoas.png"))
+    plt.show()
+
+
+def merge_data(
+    liverpool_hp_suitability_scores_pd: pd.DataFrame,
```
crispy-wonton, line 207: In general, you could shorten argument names and in-function variable names if you wanted. If you document what they represent well in the docstring, then you can use names like `df` and `gdf` here. I tend to do that in my code because I think it makes the code easier to read, because the lines then become a lot shorter. 

Of course it's a matter of preference because your descriptive names can make it easier to understand what's going on in more complex pieces of code. So do whatever you prefer :)

asf_heat_pump_suitability/analysis/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,225 @@
...
+
+    Returns:
+        gpd.GeoDataFrame: GeoDataFrame with merged suitability scores and geometries.
+    """
+    liverpool_hp_suitability_with_geometry = liverpool_hp_suitability_scores_pd.merge(
+        liverpool_lsoa_geometries_gdf, on="LSOA21CD", how="left"
+    )
+    liverpool_hp_suitability_gdf = gpd.GeoDataFrame(
+        liverpool_hp_suitability_with_geometry, geometry="geometry"
+    )
```
crispy-wonton: I believe it's good practice to set the CRS when generating a GeoDataFrame with the `crs` arg. If you do add it, you can add it in the docstring too so we know which CRS is output :)

asf_heat_pump_suitability/analysis/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,225 @@
...
+    return liverpool_hp_suitability_gdf
+
+
+def plot_overlay(
+    liverpool_hp_suitability_gdf: gpd.GeoDataFrame,
+    liverpool_heat_network_zones_filepath: str = LIVERPOOL_GPKG_PATH,
+    output_dir: str = OUTPUT_DIR,
+):
+    """
+    Plot and save overlay of DESNZ Pilot Score = 1 on Liverpool heat network zones.
```
crispy-wonton: I'm slightly confused by the docstring and plot title. I thought the DESNZ pilot areas are the same as the Liverpool HN zones. So what are the Liverpool HN zones?

Also, it would be great to add a legend to the plot :)

EDIT: based on our slack convo, maybe we can rename as "overlay of DESNZ pilot heat network zones on LSOAs in Liverpool".
Legend can then also help make that clearer.

asf_heat_pump_suitability/analysis/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,225 @@
...
+def plot_absolute_error_map(
+    hp_suitability_gdf: gpd.GeoDataFrame,
+    score: int,
+    title: str,
+    filename: str,
+    output_dir: str = OUTPUT_DIR,
+):
+    """
+    Plot an absolute error map for areas in Liverpool based on the DESNZ pilot score,
+    overlaying regions with a different score as boundaries.
```
crispy-wonton: I'm confused by what this means: "overlaying regions with a different score as boundaries"?

asf_heat_pump_suitability/analysis/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,225 @@
...
+    liverpool_lsoa_geometries_gdf = preprocess_data(liverpool_hp_suitability_lsoas)
+    plot_lsoa_geometries(liverpool_lsoa_geometries_gdf)
+    liverpool_hp_suitability_gdf = merge_data(
+        liverpool_hp_suitability_scores_pd, liverpool_lsoa_geometries_gdf
+    )
+    plot_overlay(liverpool_hp_suitability_gdf)
+    plot_absolute_error_map(
+        liverpool_hp_suitability_gdf,
+        score=1,
+        title="Absolute Error Map for DESNZ Pilot Score = 1 for Liverpool",
```
crispy-wonton: I think this map and below need a slightly more descriptive plot title / legend to be clear what the error is and what DESNZ pilot score means.

crispy-wonton (commented):
I've reviewed the plot comparison code. It looks good and I don't see any issues or bugs. Mainly just a couple of suggestions for documentation.

I got 4 maps output from the script and they look the same as the ones you showed in the sprint review (see file names below). Great job with the heat mapping, it looks really good and the code is really clean and readable and simple! Really nice work!!

image

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
helloaidank: Just on this, the outputs directory is in the `.gitignore`, so I am unsure if this is the best place to put the `hn_zones` directory?

asf_heat_pump_suitability/analysis/hn_zones/plot_comparison_of_hn_zones.py
helloaidank: I think at least initially I like keeping the names descriptive especially on first iteration as I have a tendency to forget what each variable means! I'm also sure we can factor this down, as it will be for every LA rather than just Liverpool. But thinks like heat network (can abbreviate to hn as I'm very familiar with that by now).

asf_heat_pump_suitability/analysis/hn_zones/plot_comparison_of_hn_zones.py
helloaidank: Added :)

asf_heat_pump_suitability/analysis/hn_zones/plot_comparison_of_hn_zones.py
helloaidank: Yes, good point, it's not super clear to those who aren't me!

asf_heat_pump_suitability/analysis/hn_zones/plot_comparison_of_hn_zones.py
helloaidank: Probably trying to over explain here, I've deleted this line now haha. I was trying to say we are including regions here which don't have an MAE in the plot for aesthetics.

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
helloaidank, line 1: Hi @crispy-wonton!

Thank you for your comments, I’ve tweaked the methodology so that we are looking at the fraction of area covered in an LSOA by the DESNZ HN zones. We have taken this fraction to be the ‘DESNZ HN pilot score’ and this is what we compare to the Nesta HN score per LSOA. 

We also now output a parquet file with the avg Nesta HN s as a function of the DESNZ HN pilot score (fraction of the LSOA that is covered by a DESNZ HN zone).
I’ve also moved this all to the ‘output’ directory as I mentioned, but we probably should put it somewhere else because I keep on getting the git .ignore warning. 

In the plotting script, I’ve added a plot of avg Nesta HN vs DESNZ HN pilot zone and also perform and plot a linear regression and show the R^{2} on it. 

Thanks for taking another look and for your really useful comments!

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,447 @@
...
+    """
+    Add a 'DESNZ_pilot_fraction' column to the Nesta heat pump suitability data using the fraction of area covered by HN zones and compute the average Nesta HN score per LA.
+
+    Args:
+        la_hp_suitability_scores (pl.DataFrame): Data containing 'LSOA21CD' and 'HN_N_avg_score_weighted' columns for a single LA.
+        joined_gdf (gpd.GeoDataFrame): GeoDataFrame containing 'LSOA21CD' and 'fraction_covered' columns.
+
+    Returns:
+        Tuple[pl.DataFrame, float, float]:
+            - Updated DataFrame with 'DESNZ_pilot_fraction' column added (fraction of area covered by HN zones).
```
crispy-wonton:
```suggestion
            - Updated DataFrame with 'DESNZ_pilot_fraction' column added (fraction of local authority area covered by HN zones).
```
Is this correct?

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
crispy-wonton: Ah okay that makes sense, I hadn't thought of that!

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
crispy-wonton: I think it's fine that the outputs go into a folder that's ignored by git. I don't think we want to push the results to github.

I can see you've change it in the new script. To clarify, I just meant to save the output files to the `outputs` dir. I think the scripts themselves should stay in the `analysis` folder. :) But perhaps I misunderstood your reasoning. What do you think?

Lastly, you could also save your outputs directly to S3 if you wanted. At least the `.json` files, if not the plots. We could make a new `evaluation` dir in the `asf-heat-pump-suitability` bucket.

asf_heat_pump_suitability/analysis/hn_zones/plot_comparison_of_hn_zones.py
crispy-wonton: As above, I meant moving the `output_data` and `output_plots` dirs only, not the scripts :)

asf_heat_pump_suitability/analysis/hn_zones/plot_comparison_of_hn_zones.py
crispy-wonton: No problem, go with what you find easier to use. As you've said we can shorten them later if necessary.

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,447 @@
+"""
```
crispy-wonton, line 1: Nice script documentation!

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,447 @@
...
+        ],
+    )
+    logging.info(f"Logging setup complete. Logs are saved to {log_file_path}.")
+
+
+def load_hn_geodata(
+    desnz_hn_gpkg_path: str, lsoa_shp_path: str
+) -> Tuple[gpd.GeoDataFrame, List[str]]:
+    """
+    Load the HN GeoPackage file and perform a spatial join with LSOA polygons to add a column of LSOA codes.
```
crispy-wonton:
```suggestion
    Load the Heat Networks GeoPackage file and perform a spatial join with LSOA polygons to add a column of LSOA codes.
```
We are also calculating the intersecting area so I think we should add that to the docstring :)

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,447 @@
...
+
+    Args:
+        desnz_hn_gpkg_path (str): Path to the GeoPackage file with the HN zones.
```
crispy-wonton:
```suggestion
        desnz_hn_gpkg_path (str): Path to the GeoPackage file with the heat network zones.
```

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,447 @@
...
+        format="%(levelname)s: %(message)s",
+        handlers=[
+            logging.FileHandler(log_file_path, mode="w"),  # Overwrite log file each run
+            logging.StreamHandler(),
+        ],
+    )
+    logging.info(f"Logging setup complete. Logs are saved to {log_file_path}.")
+
+
+def load_hn_geodata(
```
crispy-wonton: Should we call this `load_transform_hn_geodata` because we are doing some processing?

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,447 @@
...
+    """
+    # Load DESNZ HN polygons GeoPackage
+    desnz_hn_gdf = pyogrio.read_dataframe(
+        desnz_hn_gpkg_path, layer="heat-network-zone-map-Liverpool"
+    )
+
+    # Load the LSOA polygons file with 'LSOA21CD' codes
+    lsoa_gdf = gpd.read_file(lsoa_shp_path)
+
+    # Remove 'index_right' column if it exists
```
crispy-wonton, line 169: When does it exist vs not exist?

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,447 @@
...
+    joined_gdf["intersection_area"] = joined_gdf.apply(
+        lambda row: row.geometry.intersection(
+            lsoa_gdf.loc[row.index_right, "geometry"]
+        ).area,
+        axis=1,
+    )
+    lsoa_gdf["total_area"] = lsoa_gdf.geometry.area
+
+    # Merge the LSOA geometries into the joined GeoDataFrame
+    joined_gdf = joined_gdf.merge(lsoa_gdf[["LSOA21CD", "total_area"]], on="LSOA21CD")
```
crispy-wonton: I think it might be easier to perform an `overlay` here with `how=intersection`. The resulting GeoDataFrame should have the intersecting area between the DESNZ heat network zone polygons and the LSOA polygons. You can then take the area of this geometry.

I think if you calculate the LSOA total area first, it should retain that column so you only need to do one overlay function.

Here's the documentation for `overlay`: https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.overlay.html#geopandas.GeoDataFrame.overlay

Here it is for overlay types:
https://geopandas.org/en/latest/docs/user_guide/set_operations.html

So i think the code would look something like this (although I haven't tested):
```
lsoa_gdf["total_area"] = lsoa_gdf.geometry.area
# Get all intersections between DESNZ heat network zones and LSOAs
joined_gdf = gpd.overlay(desnz_hn_gdf, lsoa_gdf, how="intersection")
# Calculate area of intersections
joined_gdf["fraction_covered"] = joined_gdf["geometry"].area / joined_gdf["total_area"]
```

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,447 @@
...
+# Define file paths
+# LIVERPOOL_GPKG_PATH = "s3://asf-heat-pump-suitability/heat_network_desnz_data/heat-network-zone-map-Liverpool.gpkg"
+# LSOA_SHP_PATH = "s3://asf-heat-pump-suitability/source_data/Lower_layer_Super_Output_Areas_2021_EW_BFE_V9_-9107090204806789093/LSOA_2021_EW_BFE_V9.shp"
+# NESTA_HP_SUITABILITY_PARQUET_PATH = "s3://nesta-open-data/asf_heat_pump_suitability/2023Q4/20240925_2023_Q4_EPC_heat_pump_suitability_per_lsoa.parquet"
+LA_TO_ANALYSE = "Liverpool"
+LIVERPOOL_GPKG_PATH = "heat-network-zone-map-Liverpool.gpkg"
+LSOA_SHP_PATH = "LSOA_2021_EW_BFE_V9.shp"
+NESTA_HP_SUITABILITY_PARQUET_PATH = (
+    "20240925_2023_Q4_EPC_heat_pump_suitability_per_lsoa.parquet"
+)
```
crispy-wonton: Will we be reading the paths from S3? I think it makes it easier than reading from local dir.

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,447 @@
...
+    # Extract unique LSOA codes for LA
+    la_unique_lsoas = la_filtered_scores["LSOA21CD"].unique().to_list()
+
+    return la_filtered_scores, la_unique_lsoas
+
+
+def add_DESNZ_pilot_fraction(
+    la_hp_suitability_scores: pl.DataFrame,
+    joined_gdf: gpd.GeoDataFrame,
+    optional_threshold: Optional[float] = 0,
```
crispy-wonton, line 230: `optional_threshold` is missing from docstring

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,447 @@
...
+        pl.col("fraction_covered").fill_null(0).alias("DESNZ_pilot_fraction")
+    ).drop("fraction_covered")
+
+    # Calculating the average Nesta HN score for non-zero and zero DESNZ pilot scores
+    avg_hn_score_pilot_nonzero = _calculate_hn_pilot_average_score(
+        la_hp_suitability_scores,
+        pilot_score_condition=lambda col: col > optional_threshold,
+    )
+    avg_hn_score_pilot_zero = _calculate_hn_pilot_average_score(
+        la_hp_suitability_scores, pilot_score_condition=lambda col: col == 0
```
crispy-wonton: Should this set `lambda col: col <= optional_threshold`? E.g. if we set `optional_threshold` to 0.5 for example, do you want to include anything below 0.5 in the analysis or exclude it entirely?

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,447 @@
...
+def _calculate_hn_pilot_average_score(
+    la_hp_suitability_scores: pl.DataFrame,
+    pilot_score_condition: Callable[[pl.Expr], pl.Expr],
+) -> float:
+    """
+    Calculate the average Nesta Heat Network score for LSOAs in (`DESNZ_pilot_fraction > 0`) or not in (`DESNZ_pilot_fraction == 0`) DESNZ heat network pilot areas.
+
+    Args:
+        la_hp_suitability_scores (pl.DataFrame): DataFrame containing 'DESNZ_pilot_fraction' and 'HN_N_avg_score_weighted' columns for a LA.
+        pilot_score_condition (Callable[[pl.Expr], pl.Expr]): The value of 'DESNZ_pilot_fraction' to filter by (must be >0 or 0).
```
crispy-wonton: I think we could change this to a boolean input and then use an `if` / `else` condition. I think it would be clearer and easier to read. The arg could be something like:
`hn_zones (bool): if True, calculate average Nesta heat network score for LSOAs in DESNZ heat network zones. Set to False to calculate the average score for LSOAs not in heat network zones.`

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,447 @@
...
+def calculate_mae_for_pilot_score(
+    hp_suitability_scores_with_desnz: pl.DataFrame,
+    condition_score_value: Callable[[pl.Expr], pl.Expr],
+) -> float:
+    """
+    Calculate the Mean Absolute Error (MAE) for entries satisfying a specified condition on 'DESNZ_pilot_fraction'.
+
+    Args:
+        hp_suitability_scores_with_desnz (pl.DataFrame): DataFrame containing 'DESNZ_pilot_fraction' and 'absolute_error' columns.
+        condition_score_value (Callable[[pl.Expr], pl.Expr]): A function defining the condition to filter 'DESNZ_pilot_fraction'.
```
crispy-wonton: Again, it's not super clear to me what the input should be here. I would suggest changing to maybe a numerical/boolean input if possible.

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,447 @@
...
+if __name__ == "__main__":
+    # Argument parser for optional threshold
+    parser = argparse.ArgumentParser(
+        description="Process heat network zones and calculate scores."
+    )
+    parser.add_argument(
+        "--optional_threshold",
+        type=float,
+        default=0.0,
+        help="Optional threshold for DESNZ pilot fraction for calculating average Nesta HN score.",
```
crispy-wonton: Can you give the range this value can be in. I'm assuming 0-1. Also would be good to do this in the docstrings of the functions above where they take the `optional_threshold` argument.

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+OUTPUT_DIR = os.path.join(PROJECT_DIR, "outputs/hn_zones/output_plots/")
+
+# Ensure the output directory exists
+os.makedirs(OUTPUT_DIR, exist_ok=True)
+
+# Define a variable for target CRS to avoid hardcoding
+TARGET_CRS = "EPSG:27700"
+
+
+def load_data():
```
crispy-wonton: missing type hinting for return type

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+    """
+    Load the necessary data files and return them.
+
+    Returns:
+        Tuple:
+            - pd.DataFrame: DataFrame with Liverpool heat pump suitability scores.
+            - List[str]: List of LSOA codes for Liverpool.
+            - pl.DataFrame: DataFrame with average HN scores coverage.
```
crispy-wonton: Not sure what this last dataset is?

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
+"""
+This script processes and visualises heat network (HN) and heat pump suitability data for Liverpool.
```
crispy-wonton:
```suggestion
This script processes and visualises DESNZ heat network (HN) and Nesta heat pump suitability data for Liverpool.
```

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+
+# Set up logging configuration
+logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
+
+# Define constants for file paths (LEAVE UNCHANGED)
+INPUT_DIR = os.path.join(PROJECT_DIR, "outputs/hn_zones/output_data/")
+
+LIVERPOOL_HP_SUITABILITY_PARQUET = os.path.join(
+    INPUT_DIR, "liverpool_hp_suitability_scores_with_desnz.parquet"
+)
```
crispy-wonton: I think this can go in an `evaluation` folder in S3.

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+
+AVERAGE_HN_SCORES_COVERAGE_PARQUET = os.path.join(
+    INPUT_DIR, "average_scores_by_threshold.parquet"
+)
```
crispy-wonton: Same here

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+print(AVERAGE_HN_SCORES_COVERAGE_PARQUET)
```
crispy-wonton: Is this print statement meant to still be here, or should it be a logging statement?

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+
+LSOA_SHP_PATH = "s3://asf-heat-pump-suitability/source_data/Lower_layer_Super_Output_Areas_2021_EW_BFE_V9_-9107090204806789093/LSOA_2021_EW_BFE_V9.shp"
+LIVERPOOL_GPKG_PATH = "s3://asf-heat-pump-suitability/heat_network_desnz_data/heat-network-zone-map-Liverpool.gpkg"
+LSOA_JSON_PATH = os.path.join(INPUT_DIR, "liverpool_hp_suitability_lsoas.json")
```
crispy-wonton: Same here, can be saved to S3

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+def preprocess_data(
+    liverpool_hp_suitability_lsoas: list,
+    lsoa_shp_path: str = LSOA_SHP_PATH,
+    target_crs: str = TARGET_CRS,
+) -> gpd.GeoDataFrame:
+    """
+    Load and preprocess LSOA geospatial data by extracting unique LSOA geometries and filtering Liverpool LSOAs.
+
+    Args:
+        liverpool_hp_suitability_lsoas (list): List of LSOA codes for Liverpool.
```
crispy-wonton: To clarify, is this only the LSOAs that we have HP suitability data for, or all LSOAs in Liverpool? The docstring makes it sound like the latter, but the variable name makes it sound like the former.

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+        liverpool_lsoa_geometries_gdf (gpd.GeoDataFrame): GeoDataFrame with Liverpool LSOA geometries.
+        output_dir (str): Output directory to save the plot image.
+    """
+    logging.info("Plotting LSOA geometries...")
+    fig, ax = plt.subplots(figsize=(12, 6))
+    liverpool_lsoa_geometries_gdf.plot(ax=ax)
+    plt.title("Liverpool LSOA Geometries")
+    plt.axis("off")
+    plt.savefig(os.path.join(output_dir, "liverpool_data_lsoas.png"))
+    plt.savefig(os.path.join(output_dir, "liverpool_data_lsoas.pdf"))
```
crispy-wonton, line 202: Is it meant to save as `.pdf` also? If so, we should add it to the docstring :) although i think `.png` is enough

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+        lsoa_geometries_gdf["LSOA21CD"].isin(liverpool_hp_suitability_lsoas)
+    ]
+
+    if liverpool_lsoa_geometries_gdf.empty:
+        raise ValueError("No LSOAs found for Liverpool in the provided LSOA shapefile.")
+
+    return liverpool_lsoa_geometries_gdf
+
+
+def plot_lsoa_geometries(liverpool_lsoa_geometries_gdf, output_dir: str = OUTPUT_DIR):
```
crispy-wonton: missing type hinting for `liverpool_lsoa_geometries_gdf`

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+    plt.savefig(os.path.join(output_dir, "liverpool_data_lsoas.pdf"))
+    plt.show()
+
+
+def merge_data(
+    liverpool_hp_suitability_scores_pd: pd.DataFrame,
+    liverpool_lsoa_geometries_gdf: gpd.GeoDataFrame,
+) -> gpd.GeoDataFrame:
+    """
+    Merge suitability scores with geometries and convert to GeoDataFrame.
```
crispy-wonton:
```suggestion
    Merge LSOA suitability scores with geometries and convert to GeoDataFrame.
```

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+                       If > 0.0, plot regions where DESNZ_pilot_fraction >= score.
+                       If 0, plot regions where DESNZ_pilot_fraction == 0.
+        title (str): Title for the plot.
+        filename (str): Filename (PNG) to save the plot.
+        output_dir (str): Output directory to save the plot image.
+
+    Raises:
+        ValueError: If score < 0.
+    """
+    logging.info(f"Plotting absolute error map for score {score}...")
```
crispy-wonton:
```suggestion
    logging.info(f"Plotting absolute error map for score threshold {score}...")
```

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+    output_dir: str = OUTPUT_DIR,
+):
+    """
+    Plot an absolute error map for areas in Liverpool based on the DESNZ pilot score.
+
+    Args:
+        hp_suitability_gdf (gpd.GeoDataFrame): GeoDataFrame containing Liverpool's LSOA geometries and
+                                               heat pump suitability scores, including 'absolute_error' and
+                                               'DESNZ_pilot_fraction' columns.
+        score (float): DESNZ pilot score threshold.
```
crispy-wonton, line 283: missing the possible range of values for `score`. I'm guessing it's 0-1.

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+    filename: str,
+    output_dir: str = OUTPUT_DIR,
+):
+    """
+    Plot the average Nesta HN score against DESNZ Pilot Fraction Threshold, fit a regression line,
+    and display R^2.
+
+    Args:
+        average_hn_scores_coverage_df (pl.DataFrame): DataFrame with average Nesta HN scores and DESNZ Pilot Fraction.
+        fraction_thresholds_col (str): Column name for DESNZ Pilot Fraction Threshold.
```
crispy-wonton, line 339: Is 'Pilot Fraction Threshold' column different from 'Pilot Fraction' column?

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+        hn_avg_scores_col (str): Column name for HN_N Avg Score Weighted.
```
crispy-wonton:
```suggestion
        hn_avg_scores_col (str): Column name for Nesta Heat Network Avg Score Weighted.
```

outputs/hn_zones/plot_comparison_of_hn_zones.py
crispy-wonton: PS This description is missing running instructions :)
but I did run it all and it worked great!

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+    hn_avg_scores = average_hn_scores_coverage_df[hn_avg_scores_col].to_list()
+
+    x = np.array(fraction_thresholds)
+    y = np.array(hn_avg_scores)
+
+    # Fit a linear model
+    coefficients = np.polyfit(x, y, 1)
+    model = np.poly1d(coefficients)
+    predicted_y = model(x)
+    r_squared = r2_score(y, predicted_y)
```
crispy-wonton, line 355: Haven't used `numpy` for linear models before, cool!

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+
+    ax.set_ylim(y_min, y_max)
+    ax.spines["top"].set_visible(False)
+    ax.spines["right"].set_visible(False)
+
+    # Break mark on the y-axis
+    d = 0.015
+    kwargs = dict(transform=ax.transAxes, color="k", clip_on=False)
+    ax.plot((-d, +d), (-d, +d), **kwargs)
+    ax.plot((-d, +d), (1 - d, 1 + d), **kwargs)
```
crispy-wonton, line 386: Very nice!

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+
+    ax.set_title(
+        "Average Nesta HN Score vs Fraction of DESNZ HN zone in LSOA", fontsize=14
```
crispy-wonton, line 389:
```suggestion
        "Average Nesta HN Score vs Fraction of LSOA in DESNZ Heat Network zone", fontsize=14
```
Should it be this way round? Otherwise lmk because I got confused!

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -0,0 +1,387 @@
...
+    )
+    ax.set_xlabel("Fraction of DESNZ HN zone in LSOA", fontsize=12)
+    ax.set_ylabel("Average Nesta HN Score", fontsize=12)
+    ax.grid(True)
+    ax.text(
+        0.05,
+        0.95,
+        f"$R^2 = {r_squared:.2f}$",
```
crispy-wonton, line 397: For my understanding, what does the dollar sign do here?

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -0,0 +1,447 @@
...
+
+def calculate_mae_for_all(
+    hp_suitability_scores: pl.DataFrame, desnz_col: str, nesta_hn_score_col: str
+) -> Tuple[pl.DataFrame, float]:
+    """
+    Calculate the Mean Absolute Error (MAE) between the DESNZ pilot score and Nesta's HN suitability score, add the absolute error column to the DataFrame,
+    and log the result.
+
+    Args:
+        hp_suitability_scores (pl.DataFrame): DataFrame containing the DESNZ pilot (actual) and Nesta HN suitability score (predicted) columns.
```
crispy-wonton, line 376: It's really helpful to label them as 'actual' and 'predicted' here! :)

crispy-wonton (commented):
Hi @helloaidank ,

This looks great thank you!
I have mainly suggested some cosmetic changes and updates to documentation and a couple of places to simplify the inputs of functions.

There is one point on `outputs/hn_zones/comparison_of_hn_zones.py` that suggests to use `overlay` instead of `sjoin` with an intersection. I think it will be easier to apply. Let me know if you have questions about this.

Secondly, sorry my point in the previous review was not clear. I had meant for you to save the outputs of the script to the `outputs` directory in the repo, but I think the scripts themselves should be in the `analysis` folder where you originally had them.

Having said this, to make the final review easier and quicker, I would suggest keeping the files in the same place for now as you make further edits to the code. This means when I do my final review, I can review only the changes since this review, rather than the entire file showing up as diffs (due to it being moved to a new folder). If there's a way to do this on Github anyway that I don't know about , please let me know! After the final review, we can then move the files back to the `analysis` folder.

Lastly, I think you could create a new folder in S3, called `evaluation/liverpool_heat_networks` or something, and save your intermediate (and also final!) outputs to there if you wanted. I think this is better than saving locally because then the whole project team has direct access to the results :)

Great work! I ran your code and it produced the expected outputs and was clearly documented.

Let me know if you have any questions :)

outputs/hn_zones/comparison_of_hn_zones.py
helloaidank: Yep that's correct!

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
helloaidank: Ah ok, I misunderstood, I thought you meant moving the whole directory to the outputs folder. I will move it over after we've finished this review :).

helloaidank: After weighing it up in my head, I do think it probably makes sense to have a version saved to s3 (so people can play around with the plotting scripts without needing to run the comparison script) but if you are iterating locally, then make sure you save the files locally because it takes ages otherwise hehe and is a bit annoying.

outputs/hn_zones/comparison_of_hn_zones.py
helloaidank: Thank you Roisín!

outputs/hn_zones/comparison_of_hn_zones.py
helloaidank: good point!

outputs/hn_zones/comparison_of_hn_zones.py
helloaidank: just wondering if we want to replace all HN with heat network zones? It appears in quite a few places. It's just that at the beginning of the documentation I have the following: "This script analyses DESNZ heat network (HN) zones..."

helloaidank: I just want to be consistent, if you think it's maybe clearer to replace HN with heat network then maybe that's the way to go! But just checking there isn't another reason?

outputs/hn_zones/comparison_of_hn_zones.py
helloaidank: Yep, good suggestion!

outputs/hn_zones/comparison_of_hn_zones.py
helloaidank: It exists when a spatial join has been performed on the geodataframe object and doesn't when it hasn't I think!

outputs/hn_zones/comparison_of_hn_zones.py
helloaidank: Ah brilliant, this is much simpler than what I had and gives the same results! Nice one Roisín! 🥇

outputs/hn_zones/comparison_of_hn_zones.py
helloaidank: So, it's much quicker to read from local (from my experience). I should've commented this out. For iterating over the code, it's quite a nice option! I might add a flag here to make it smoother 👍

outputs/hn_zones/comparison_of_hn_zones.py
helloaidank: I think calculating the zero value of the HN score is a nice baseline to have and compare to. We could in addition calculate this value, just understanding if this is useful for us to know?

outputs/hn_zones/comparison_of_hn_zones.py
helloaidank: I completely agree that this change improves readability, especially for someone quickly scanning through the code. I initially designed it this way for flexibility and to make the function reusable in different contexts! Will make the change. It's a pretty cool functionality 👍

outputs/hn_zones/comparison_of_hn_zones.py
helloaidank: just to say on this parameter, you basically define the function in the parameter argument e.g. `pilot_score_condition(pl.col("DESNZ_pilot_fraction"))` where you can set pilot_score_condition as the function that will be performed on the selected data (in this case the "DESNZ pilot fraction" column), e.g. be bigger than threshold ">threshold".

outputs/hn_zones/comparison_of_hn_zones.py
helloaidank: Yep, it's 0-1, I've added that in here in the help string as well as in the parameters doc strings within the relevant functions!

outputs/hn_zones/plot_comparison_of_hn_zones.py
helloaidank: We looped through a bunch of different fraction of area thresholds (for DESNZ HN coverage) to see how this parameter affects the average Nesta heat network score.

outputs/hn_zones/plot_comparison_of_hn_zones.py
helloaidank: Good point!

outputs/hn_zones/plot_comparison_of_hn_zones.py
helloaidank: I've kept a local option too!

outputs/hn_zones/plot_comparison_of_hn_zones.py
helloaidank: My bad, meant to delete this!

outputs/hn_zones/plot_comparison_of_hn_zones.py
helloaidank: It's all LSOAs in Liverpool!

outputs/hn_zones/plot_comparison_of_hn_zones.py
helloaidank: I like having the PDF outputs as well as you can scale them quite nicely and makes the text look sharp!

outputs/hn_zones/plot_comparison_of_hn_zones.py
helloaidank: The Pilot Fraction Threshold will be what we use to exclude LSOAs with DESNZ HN coverage below a certain fraction and then calculate the average HN Nesta score based on those still included. The Pilot Fraction column is the DESNZ HN coverage per LSOA I think!

helloaidank: I've changed the doc string at the beginning of this file to reflect that!

outputs/hn_zones/plot_comparison_of_hn_zones.py
helloaidank: It's the fraction of the DESNZ HN zone in the LSOA. We are increasing the threshold, so as we increase the amount of DESNZ HN zone in the LSOA, our average Nesta HN score. This is what we expect intuitively!

outputs/hn_zones/plot_comparison_of_hn_zones.py
helloaidank: It's LaTeX notation, entering math mode!

helloaidank:
Hi there, I've added in some of the changes that you suggested!

The code for evaluation has become quite lengthy, so I think in the next PR I will likely scale to the other LAs but also modularise the code a bit across a few different scripts. Hopefully this will improve things in terms of readability! 

Thanks for taking the time to go through the code again and making suggestions, very useful ones, especially the overlay one.

I am happy to merge the PR and then make the changes I mentioned above in another one, once we're obviously happy with the code.

asf_heat_pump_suitability/analysis/hn_zones/comparison_of_hn_zones.py
crispy-wonton: This all sounds great, thanks! Good tip for saving files locally!

outputs/hn_zones/plot_comparison_of_hn_zones.py
crispy-wonton: ohhh, that's cool! didn't know you could do that, thanks for explaining !

outputs/hn_zones/plot_comparison_of_hn_zones.py
crispy-wonton: thanks for explaining, that makes sense :D

outputs/hn_zones/comparison_of_hn_zones.py
crispy-wonton: yeh it was just because it's clearer if you read one function on its own :)

outputs/hn_zones/comparison_of_hn_zones.py
crispy-wonton: great, thanks for checking it's the same!

outputs/hn_zones/comparison_of_hn_zones.py
crispy-wonton: Cool , happy to keep it at 0 :)

outputs/hn_zones/comparison_of_hn_zones.py
crispy-wonton: Ah that makes sense why you designed it like this in the first place. Let's prioritise readability for now and then we can come back to it if we need it to be more flexible. :)

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -26,31 +26,41 @@
...
-# LSOA_SHP_PATH = "s3://asf-heat-pump-suitability/source_data/Lower_layer_Super_Output_Areas_2021_EW_BFE_V9_-9107090204806789093/LSOA_2021_EW_BFE_V9.shp"
-# NESTA_HP_SUITABILITY_PARQUET_PATH = "s3://nesta-open-data/asf_heat_pump_suitability/2023Q4/20240925_2023_Q4_EPC_heat_pump_suitability_per_lsoa.parquet"
 LA_TO_ANALYSE = "Liverpool"
-LIVERPOOL_GPKG_PATH = "heat-network-zone-map-Liverpool.gpkg"
-LSOA_SHP_PATH = "LSOA_2021_EW_BFE_V9.shp"
-NESTA_HP_SUITABILITY_PARQUET_PATH = (
-    "20240925_2023_Q4_EPC_heat_pump_suitability_per_lsoa.parquet"
-)
+S3_BUCKET = "asf-heat-pump-suitability"
+S3_KEY_DIR = "evaluation/desnz_hn_zone_scores/"
```
crispy-wonton: chef's kiss for the cleaning here

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -85,14 +95,54 @@ def setup_logging_and_file_path(
     logging.info(f"Logging setup complete. Logs are saved to {log_file_path}.")
 
 
-def load_hn_geodata(
+def setup_paths(read_in_s3: bool):
```
crispy-wonton: Great idea :) !

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -85,14 +95,54 @@ def setup_logging_and_file_path(
...
+        LIVERPOOL_GPKG_PATH = "s3://asf-heat-pump-suitability/heat_network_desnz_data/heat-network-zone-map-Liverpool.gpkg"
+        LSOA_SHP_PATH = "s3://asf-heat-pump-suitability/source_data/Lower_layer_Super_Output_Areas_2021_EW_BFE_V9_-9107090204806789093/LSOA_2021_EW_BFE_V9.shp"
+        NESTA_HP_SUITABILITY_PARQUET_PATH = "s3://nesta-open-data/asf_heat_pump_suitability/2023Q4/20240925_2023_Q4_EPC_heat_pump_suitability_per_lsoa.parquet"
+    else:
+        LIVERPOOL_GPKG_PATH = "heat-network-zone-map-Liverpool.gpkg"
+        LSOA_SHP_PATH = "LSOA_2021_EW_BFE_V9.shp"
+        NESTA_HP_SUITABILITY_PARQUET_PATH = (
+            "20240925_2023_Q4_EPC_heat_pump_suitability_per_lsoa.parquet"
+        )
+    return LIVERPOOL_GPKG_PATH, LSOA_SHP_PATH, NESTA_HP_SUITABILITY_PARQUET_PATH
```
crispy-wonton: This could be a dict so the paths don't get mixed up, but not essential

outputs/hn_zones/comparison_of_hn_zones.py
crispy-wonton: missing returns type

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -122,22 +172,12 @@ def load_hn_geodata(
         how="left",
         predicate="intersects",
     )
```
crispy-wonton, line 182: I think lines 170-174 are no longer needed

outputs/hn_zones/comparison_of_hn_zones.py
```diff
@@ -253,22 +295,30 @@ def calculate_average_scores_for_thresholds(
...
+        optional_threshold (Optional[float]): The threshold value for 'DESNZ_pilot_fraction'. Defaults to 0. Range: 0-1.
 
     Returns:
         float: Average 'HN_N_avg_score_weighted' for the filtered rows.
     """
-    # Apply the condition function to filter the DataFrame
-    filtered_la_hp_suitability_scores = la_hp_suitability_scores.filter(
-        pilot_score_condition(pl.col("DESNZ_pilot_fraction"))
-    )
+    if hn_zones:
```
crispy-wonton, line 320: Thanks! The readability is great :D

outputs/hn_zones/plot_comparison_of_hn_zones.py
```diff
@@ -30,56 +43,92 @@
...
+# Define S3 variables and paths
+S3_BUCKET = "asf-heat-pump-suitability"
+S3_KEY_DIR = "evaluation/desnz_hn_zone_scores/"
+S3_KEY_LIVERPOOL_HP_SUITABILITY_PARQUET = (
+    f"{S3_KEY_DIR}liverpool_hp_suitability_scores_with_desnz.parquet"
+)
+S3_KEY_AVERAGE_HN_SCORES_COVERAGE_PARQUET = (
+    f"{S3_KEY_DIR}average_scores_by_threshold.parquet"
+)
+S3_KEY_LSOA_JSON_PATH = f"{S3_KEY_DIR}liverpool_hp_suitability_lsoas.json"
```
crispy-wonton, line 73: We don't need to change it in this PR, but for future ref, these file paths are used across multiple files which makes them great candidates to define once in a config rather than across multiple scripts. That way, if we wanted to change them, we just need to update them in one place :)

outputs/hn_zones/plot_comparison_of_hn_zones.py
crispy-wonton: kewl, ty!

crispy-wonton (commented):
Thank you so much @helloaidank for the great work! This looks really good.

A couple of very minor changes to make before merge. See comments. Mainly:
- there are some lines of redundant code which can be deleted in `outputs/hn_zones/comparison_of_hn_zones.py`
- missing returns type in one function in the same file.

We also need to move the updated `comparison_of_hn_zones.py` and `plot_comparison_of_hn_zones.py` back to the `analysis` folder. I would suggest doing a final run of the scripts before merge to double check they still work after moving.

I ran both scripts and they worked well. Here is what I ran and the results:

**File 1:**
`python -i outputs/hn_zones/comparison_of_hn_zones.py --optional_threshold 0.5 --read_in_s3 True --save_to_s3 True`

```
INFO: number of LSOAs in DESNZ list not in HP suitability data: 2
INFO: Average HN_N_avg_score_weighted: 0.5608973509933775
INFO: Avg HN_N_avg_score_weighted for DESNZ_pilot_fraction when > 0.5: 0.6906888888888889
INFO: Avg HN_N_avg_score_weighted for DESNZ_pilot_fraction = 0: 0.524927374301676

INFO: Mean Absolute Error (MAE) for DESNZ_pilot_fraction vs HN_N_avg_score_weighted: 0.4500689346396515
INFO: Mean Absolute Error (MAE) for DESNZ_pilot_fraction > 0: 0.34778144838390823
INFO: Mean Absolute Error (MAE) for DESNZ_pilot_fraction = 0: 0.524927374301676
```

**File 2:**
`python -i outputs/hn_zones/plot_comparison_of_hn_zones.py`

The plots looked as expected! :D 

Great work, thank you!

outputs/hn_zones/comparison_of_hn_zones.py
helloaidank: Good suggestion!

outputs/hn_zones/comparison_of_hn_zones.py
helloaidank: Good spot

outputs/hn_zones/plot_comparison_of_hn_zones.py
helloaidank: 100%, let's do this in the scale PR.

outputs/hn_zones/comparison_of_hn_zones.py
helloaidank: Thanks 👯‍♂️

helloaidank:
> Thank you so much @helloaidank for the great work! This looks really good.
> 
> A couple of very minor changes to make before merge. See comments. Mainly:
> 
> * there are some lines of redundant code which can be deleted in `outputs/hn_zones/comparison_of_hn_zones.py`
> * missing returns type in one function in the same file.
> 
> We also need to move the updated `comparison_of_hn_zones.py` and `plot_comparison_of_hn_zones.py` back to the `analysis` folder. I would suggest doing a final run of the scripts before merge to double check they still work after moving.
> 
> I ran both scripts and they worked well. Here is what I ran and the results:
> 
> **File 1:** `python -i outputs/hn_zones/comparison_of_hn_zones.py --optional_threshold 0.5 --read_in_s3 True --save_to_s3 True`
> 
> ```
> INFO: number of LSOAs in DESNZ list not in HP suitability data: 2
> INFO: Average HN_N_avg_score_weighted: 0.5608973509933775
> INFO: Avg HN_N_avg_score_weighted for DESNZ_pilot_fraction when > 0.5: 0.6906888888888889
> INFO: Avg HN_N_avg_score_weighted for DESNZ_pilot_fraction = 0: 0.524927374301676
> 
> INFO: Mean Absolute Error (MAE) for DESNZ_pilot_fraction vs HN_N_avg_score_weighted: 0.4500689346396515
> INFO: Mean Absolute Error (MAE) for DESNZ_pilot_fraction > 0: 0.34778144838390823
> INFO: Mean Absolute Error (MAE) for DESNZ_pilot_fraction = 0: 0.524927374301676
> ```
> 
> **File 2:** `python -i outputs/hn_zones/plot_comparison_of_hn_zones.py`
> 
> The plots looked as expected! :D
> 
> Great work, thank you!

Hi Roisín!

Thanks for going through the code again, I've made those changes and will take on the config suggestion to the scaling PR!

I've moved the code and run it, seems to be all OK!  

Happy to merge now!

crispy-wonton (approved):
Looks great, thank you! Merge away!

PR #83, opened 2024-11-08
Add anchor properties feature

### LINKED ISSUE #72: Add anchor properties feature

shabrf:
---

# Description

Added anchor property identification feature. Identifies LSOAs containing large facilities (hospitals, universities etc.) using POI data.

# Instructions for Reviewer

In order to test the code in this PR you need to:
- Run identify_anchor_properties()

# Checklist:

- [x ] I have refactored my code out from `notebooks/`
- [x] I have tested the code
- [] I have merged any new changes from `dev`


asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
```diff
@@ -0,0 +1,170 @@
...
+    "jail_and_prison",
+    "shopping_center",
+    "airport_terminal",
+    "airport",
+    "superstore",
+    "department_store",
+}
+
+
+def _safe_join_categories(categories: pd.Series) -> str:
```
crispy-wonton:
```suggestion
def _safe_join_str_categories(categories: pd.Series) -> str:
```
Here and in other functions, please can we add the return type to the name. The naming convention of functions for the repo is _action_returnType_descriptor_. Makes it easier to immediately know what the function does.

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
```diff
@@ -0,0 +1,170 @@
...
+    Load and process Points of Interest data.
+
+    Returns:
+        GeoDataFrame: Processed POI data containing only anchor properties
+
+    Raises:
+        ValueError: If required columns are missing
+    """
+    logger.info("Loading POI data...")
+    poi = load_s3_data("asf-heat-pump-suitability", "source_data/poi_uk.gpkg")
```
crispy-wonton: the "source_data/poi_uk.gpkg" can go in the config file and then the source information added to "config/README.md" :)
Also from looking at the dataset, it looks like it's a global dataset not UK specific (?) so the name should include `global` not `uk`.

Also, it might be worth loading only the required columns as it's quite slow to load. This might mean changing the `load_s3_data` getter or creating a new getter.

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
```diff
@@ -0,0 +1,170 @@
...
+
+    # Validate POI data
```
crispy-wonton: Good idea to add this!

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
```diff
@@ -0,0 +1,170 @@
...
+    )
+
+    # Filter anchor properties and reproject
+    anchor_properties = (
+        poi[poi.main_category.isin(ANCHOR_PROPERTIES)]
+        .to_crs(INPUT_CRS)
+        .to_crs(PROCESSING_CRS)
+    )
+
+    logger.info(f"Found {len(anchor_properties)} potential anchor properties")
```
crispy-wonton, line 101: Can we add to the logger the CRS of the output

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
```diff
@@ -0,0 +1,170 @@
...
+    valid_categories = set(str(cat) for cat in categories if pd.notna(cat))
+    return ", ".join(sorted(valid_categories)) if valid_categories else ""
+
+
+def load_and_process_poi() -> gpd.GeoDataFrame:
+    """
+    Load and process Points of Interest data.
+
+    Returns:
+        GeoDataFrame: Processed POI data containing only anchor properties
```
crispy-wonton: Asking because i dont know, is the convention to put the return type as you have here with `GeoDataFrame` or should it be `gpd.GeoDataFrame`. I've always been using the latter but maybe that's wrong.

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
```diff
@@ -0,0 +1,170 @@
...
+    Identify and analyze anchor properties within LSOAs.
+
+    Returns:
+        GeoDataFrame: Summary of anchor properties by LSOA containing columns:
+            - lsoa: Unique identifier for the LSOA
+            - lsoa_name: Name of the LSOA
+            - anchor_count: Number of anchor properties in the LSOA
+            - building_categories: List of main categories present
+            - building_subcategories: List of subcategories present
+            - has_anchor_property: Boolean indicating presence of anchor properties
```
crispy-wonton: I love your documentation. So clean and easy to understand!

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
```diff
@@ -0,0 +1,170 @@
...
+    """
+    try:
+        logger.info("Starting anchor property analysis...")
+
+        # Load required data
+        lsoa_gdf = logger.info("Loading LSOA boundaries...")
+        lsoa_gdf = gpd.read_file(
+            "s3://asf-heat-pump-suitability/source_data/Lower_layer_Super_Output_Areas_2021_EW_BFE_V9_-9107090204806789093/LSOA_2021_EW_BFE_V9.shp"
```
crispy-wonton: Think this was used in the other PR as well - we need to add this to the config in case we change the path, then we dont have to change it everywhere.

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
```diff
@@ -0,0 +1,170 @@
+"""Module for identifying and analyzing potential anchor properties in LSOAs."""
```
crispy-wonton: Please can you add a short instruction that you can run this as a script and how, and what the output is.

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
```diff
@@ -0,0 +1,170 @@
...
+        logger.error(f"Error in anchor property analysis: {str(e)}")
+        raise
+
+
+if __name__ == "__main__":
+    try:
+        results = identify_anchor_properties()
+
+        # Save results (modify as needed)
+        output_path = Path("anchor_property_analysis.csv")
```
crispy-wonton: Which dir does this save to?

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
crispy-wonton: What coverage does this POI dataset have. E.g. if an LSOA has no anchor properties, can we assume they actually have none or is it also likely there is just an absence of data?

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
crispy-wonton: Also i think we can select only certain columns to load here again

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
```diff
@@ -0,0 +1,170 @@
...
+        # Load required data
+        lsoa_gdf = logger.info("Loading LSOA boundaries...")
+        lsoa_gdf = gpd.read_file(
+            "s3://asf-heat-pump-suitability/source_data/Lower_layer_Super_Output_Areas_2021_EW_BFE_V9_-9107090204806789093/LSOA_2021_EW_BFE_V9.shp"
+        )
+        anchor_properties = load_and_process_poi()
+
+        lsoa_with_anchors = gpd.sjoin(
+            lsoa_gdf, anchor_properties, how="left", predicate="intersects"
+        )
```
crispy-wonton, line 133: I think because of the `intersects` predicate we end up with more LSOA + POI pairs than there are POIs, and I can see some POIs appear multiple times in the results. I'm guessing this is POIs on the boundary of 2+ LSOAs.

I think this is something to discuss in a stand up whether we want to allow this or not because I'm not sure. What do you think?

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
```diff
@@ -0,0 +1,170 @@
...
+
+        lsoa_anchor_summary["has_anchor_property"] = (
+            lsoa_anchor_summary["anchor_count"] > 0
+        )
+
+        logger.info(
+            f"Found {lsoa_anchor_summary['has_anchor_property'].sum()} LSOAs with suitable anchor properties"
+        )
+
+        return lsoa_anchor_summary
```
crispy-wonton, line 164: There's some LSOAs that have a really high number of anchor properties. 418 have more than 20 which is high but I guess believable in a dense city centre but might need a bit further investigation and checks if you haven't done so already. E.g. I can see one LSOA has 184. I think this LSOA contains Heathrow Airport. But I wonder if some POIs are kind of getting double counted, like should we be counting shops in the airport as separate properties? Or for example, if you have a university&NHS hospital, could it be tagged multiple times in the dataset as 'hospital', 'public_service_and_government' and 'university_college' ?

If so, maybe we need to add some deduplication into our processing of the anchor properties data, either of duplicate point geoms or ones which are within a certain threshold distance of each other.

crispy-wonton (changes requested):
Thank you so much for this @shabrf . Your code is so clean and easy to read!

I don't see any bugs. The main changes I'm requesting is to put the source data into the config yaml and document the sources in the config/README. That way, we can more easily update source data if needed. Also the function naming is great and really clear but missing the return type in the name which is part of this repo's naming convention.

There's also a question of whether we want to allow a single POI to join to multiple LSOAs.

I tested the functions and they work. There's some LSOAs that end up with a really high number of properties which I left a comment about.

Other than that, I just left a couple of clarifying questions. :)

Thanks so much this is great work!

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
shabrf: It's more explicit and clear to use the fully qualified name including the namespace as you've suggested to avoid ambiguity

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
shabrf: The [source](https://data.cdrc.ac.uk/dataset/point-interest-data-united-kingdom) labels it as a UK specific dataset and most locations are in the UK. The majority of the rest are in Ireland but there's also points randomly scattered globally.

shabrf: It's a dataset that combines location data from open street maps, bing and and other sources so I think we can assume it has good coverage.

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
shabrf: Is this possible with .shp files? I'm not sure if geopandas supports this outside of csvs.

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
crispy-wonton: Thanks for clarifying :)

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
crispy-wonton: Cool, makes sense to keep it as UK then!

crispy-wonton: Great thanks!

crispy-wonton: We can document (even just internally) this as a minor caveat. Potentially there are a couple of anchor properties missing but we assume few are.

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
crispy-wonton: Yes it is possible :) it's a keyword argument for `geopandas.read_file` specifically. https://geopandas.org/en/stable/docs/reference/api/geopandas.read_file.html

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
shabrf: I agree that such high numbers of anchor properties sounds unrealistic. I think deduplication based on point geoms is feasible but setting a threshold based on distance to each other from a point would be quite arbitrary since we don't know how much area around that point the building actually covers. For example, there would be quite a big difference between the optimal threshold for an airport compared to a hospital.

I think the best solution to this issue would be to incorporate building footprint data into this feature which could be used to deduplicate by checking overlaps or distance between outer bounds.

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
shabrf: I think we're happy to proceed with allowing POIs on boundaries to be allocated to multiple LSOAs as from a practical standpoint, it's reasonable to assume a heat network can span multiple LSOAs.

(As discussed)

asf_heat_pump_suitability/pipeline/prepare_features/anchor_properties.py
shabrf: Out of 47458 possible anchor properties, only 318 had duplicate point geoms, so we can assume that most of the high anchor counts are due to closely located buildings, likely within larger complexes such as university campuses and airports.

crispy-wonton: Sounds good, thanks for checking this.

I think the pipeline is missing deduplication of the POIs? I don't see it.

crispy-wonton (dismissed):

PR #110, opened 2025-01-15
104 save intermediate gardens

### LINKED ISSUE #104: Save intermediate results in garden size pipeline
Garden size pipeline takes a long time to run and prone to various errors. We should save out intermediate EPC gardens results every 100 iterations for this reason.

crispy-wonton:
Fixes #104 

# Description

Update script that calculates garden size to save intermediate results after every 100 file matches, and refactor deduplication process to be less memory intensive. These changes are to help run the script without killing the process, which is very memory intensive as there are ~20million gardens calculated before deduplication.

- modified `asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_garden_size.py`.

Note: this script was run successfully on a `g3s.xlarge` EC2 instance with 50GiB volume. The process was killed during deduplication after successfully saving out interim results, so I made edits to the script (which are in the final version here) and ran that part separately.

# Instructions for Reviewer

Please could you read through the code and ensure that files are saved out correctly. The main thing to pay attention to is ensuring that there is a) no unintended duplication of data (e.g. saving files [1-100, 1-200] instead of [1-100, 101-200]); and b) no unintended loss of data (e.g. resetting `epc_gardens` to empty list when not appropriate, or missing the first or final file matches).

To run:
`python -i asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_garden_size.py --epc s3://asf-daps/lakehouse/processed/epc/old/deduplicated/processed_dedupl-0.parquet -y 2023 -q 4 -n ews`

You can edit and/or test run the script if you like just to check it works, but I obviously wouldn't suggest running the whole thing!

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_garden_size.py
```diff
@@ -161,22 +165,69 @@ def parse_arguments() -> argparse.Namespace:
...
+    epc_gardens_df = pl.DataFrame()
+    for file in interim_files:
+        logging.info(f"Loading file: {file}")
+        df = pl.read_parquet(f"s3://{file}")
+        df = df.with_columns(pl.col(pl.Float64).round(2))
+        df = garden_size.deduplicate_df_garden_size(df)
+        epc_gardens_df = pl.concat([epc_gardens_df, df])
+
+    # Final round of deduplication
     epc_gardens_df = garden_size.deduplicate_df_garden_size(epc_gardens_df)
```
crispy-wonton, line 233: Note, this final round of deduplication will mean that the median values we created in the loop above for duplicates within interim files will be used to generate another median. So the final deduplication for any duplicates that remain at this point will calculate the median of the median & other individual values. This is not optimal. Do you have a suggestion for how we could improve this?

asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_garden_size.py
```diff
@@ -110,10 +112,12 @@ def parse_arguments() -> argparse.Namespace:
     epc_gardens = []
     prev = None
     total_gardens = 0
+    min = 0
```
lizgzil: really don't think this matters - but I think it should be `min=-1`?

(just saw that one of the outputs would be `20250203_2023_Q4_EPC_garden_size_estimates_EWS_1_100_INTERIM.parquet` and thought why isn't it `0_99`?

asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_garden_size.py
```diff
@@ -161,22 +165,69 @@ def parse_arguments() -> argparse.Namespace:
...
+            interim_results = [
+                df.with_columns(pl.col("NATIONALCADASTRALREFERENCE").cast(pl.String))
+                for df in epc_gardens
+                if len(df) > 0
+            ]
+            interim_results = pl.concat(interim_results)
+            save_as = os.path.join(
+                interim_dir,
+                f"{datetime.today().strftime('%Y%m%d')}_{year}_Q{q}_EPC_garden_size_estimates_{args.nations.upper()}_{min+1}_{i}_INTERIM.parquet",
+            )
```
lizgzil: really don't think this matters - but I saw that one of the outputs would be `20250203_2023_Q4_EPC_garden_size_estimates_EWS_1_100_INTERIM.parquet` and thought why isn't it `0_100` (or `0_99`)?

I don't think there is an issue with this though, but thought I'd mention it in case it is confusing at some point. I guess the first file has 101 files (and technically should be called `0_100`, and the rest have 100 (apart from the last one).

asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_garden_size.py
```diff
@@ -161,22 +165,69 @@ def parse_arguments() -> argparse.Namespace:
...
+        args.save_as = f"s3://asf-heat-pump-suitability/outputs/{year}Q{q}/gardens/{datetime.today().strftime('%Y%m%d')}_{year}_Q{q}_EPC_garden_size_estimates_{args.nations.upper()}.parquet"
     save_utils.save_to_s3(epc_gardens_df, args.save_as)
 
+    del epc_gardens_df
+
+    # We load the files again and deduplicate them as they are loaded as it's less memory intensive
     logging.info("Deduplicating UPRNs that were matched to multiple gardens")
-    epc_gardens_df = epc_gardens_df.with_columns(pl.col(pl.Float64).round(2))
+    epc_gardens_df = pl.DataFrame()
+    for file in interim_files:
```
lizgzil, line 225: I found `interim_files` contained one called `20250113_2023_Q4_EPC_garden_size_estimates_EWS_0_0_INTERIM.parquet`, which I hadn't expected after looking at the file names saved by running:

```
for i, (land_file, building_file) in tqdm(enumerate(file_matches.items())):
    if ((i % 100 == 0) & (i != 0)) or (i == (len(file_matches) - 1)):
            print(f". {datetime.today().strftime('%Y%m%d')}_{year}_Q{q}_EPC_garden_size_estimates_{args.nations.upper()}_{min+1}_{i}_INTERIM.parquet")
```
I can see this 0_0 file contains some different UPRNs to what is in 1_100, so don't think there is duplication or anything.

Is this what you'd expect?

asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_garden_size.py
lizgzil: hm, I don't think I do apart from only deduplicating after the for loop and not during - is this unfeasible memory-wise?  (I calculated that there were 18,388,401 properties in total before any deduplication).

But note, this only effects 104,668 properties `len(epc_gardens_df) =17334666` after the for loop and `len(epc_gardens_df) = 17229998` after the final deduplication, so it doesn't seem like a huge concern for the median calculation.

lizgzil (commented):
looks good! A tiny bit confused about some file names - I've asked some Qs

asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_garden_size.py
crispy-wonton: Yes, this is because I made changes to the if statement after I actually ran the pipeline (I added the `& (i != 0)) or (i == (len(file_matches) - 1))` retroactively to avoid saving a file with a single file pair (i.e. the 0_0 file). So the files in the `/interim` dir don't follow this exact naming but they still should be correct and not duplicated, just saved in a different partition to what it would now be saved as.

asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_garden_size.py
crispy-wonton: Yes I couldn't think of a very succinct way to code the name accurately because, as you have said below, the first interim file is 101 file pairs (0-100), whereas the next interim files all represent 100 file pairs (e.g. 101-200). So the first file should be named `0_100`.

However, it was annoying me and I think you are right that it can cause confusion later, so I have added an `if` statement to update it.

asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_garden_size.py
crispy-wonton: Good question - yes it was a memory problem. Even in EC2, the deduplication step was killing the process which is why I chose to deduplicate one file at a time in the `for` loop above. The final deduplication because there are likely some overlaps between different Microsoft building and / or land parcel files in the ~100k properties you have listed.

I will leave it as is for now because I also don't think this should be a big problem, but I will add to the backlog as an issue/something to be aware of.

crispy-wonton: See issue #120

lizgzil (approved):
thanks for the replies and fix! good to merge

PR #115, opened 2025-01-22
add garden size estimate evaluation notebook

### LINKED ISSUE #114: Evaluate garden size estimates

crispy-wonton:
Fixes #114 

# Description

add garden size estimate evaluation notebook

# Instructions for Reviewer

to generate the notebook, ensure `jupytext` is installed and then run the following line in terminal:
`jupytext --to notebook asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py`

Please could you provide some feedback on this analysis. Namely:
- Do you think the selection criteria are appropriate and applied in the correct order?
- Have I accounted for the appropriate corrections to make this a fair analysis (e.g. applying weights, checking weights add to at least 0.9 etc.)
- Are there any suggestions you have to make the analysis more robust or accurate?
- Have the weights been applied correctly?
- Is there anything we should add to the analysis?

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py
+++ b/asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py
@@ -21,7 +21,7 @@ avg_gardens_df = garden_space_avg.generate_df_garden_space_avg()
 
 # %%
 gardens_df = pl.read_parquet(
-    "s3://asf-heat-pump-suitability/outputs/2023Q4/gardens/20250115_2023_Q4_EPC_garden_size_estimates_EWS_deduplicated.parquet"
+    "s3://asf-heat-pump-suitability/outputs/2023Q4/gardens/20250115_2023_Q4_EPC_garden_size_estimates_EWS_deduplicated_FOR_EVALUATION_ONLY.parquet"
 )
 
 # %%
--- a/asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py
+++ b/asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py
@@ -1,12 +1,23 @@
 # %%
 import polars as pl
 import numpy as np
+import logging
 import matplotlib.pyplot as plt
+from datetime import datetime
 from sklearn import metrics
+from asf_heat_pump_suitability.getters import base_getters
+from asf_heat_pump_suitability.utils import save_utils
 from asf_heat_pump_suitability.pipeline.prepare_features import (
     garden_space_avg,
     household_count,
 )
+
+# %%
+# PARAMS
+year = 2023
+q = 4
+interim_dir = "s3://asf-heat-pump-suitability/outputs/2023Q4/gardens/interim/"
+nations = "EWS"
 
 # %% [markdown]
 # ## Compare calculated garden size to avg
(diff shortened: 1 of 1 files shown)


asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py
```diff
@@ -0,0 +1,385 @@
...
+)
+
+# %% [markdown]
+# ### 5. Compare weighted results
+
+# %%
+# Filter to overall average garden size. We need to compare all property types to the total average to be able to use the weights
+total_avg_gardens_df = avg_gardens_df.filter(
+    pl.col("msoa_avg_outdoor_space_property_type") == "unknown"
+)
```
lizgzil: How come you didn't compare all the property types?

asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py
```diff
@@ -0,0 +1,385 @@
...
+gardens_df = gardens_df.join(
+    uprn_count, how="left", on="NATIONALCADASTRALREFERENCE"
+).join(cadastral_garden_size, how="left", on="NATIONALCADASTRALREFERENCE")
+
+# Divide shared gardens equally among UPRNs sharing gardens
+gardens_df = gardens_df.with_columns(
+    (pl.col("cadastral_garden_size_m2") / pl.col("UPRN_count")).alias(
+        "divided_garden_area_m2"
+    )
+)
```
lizgzil, line 91: I wasn't expecting these steps. I thought "garden_area_m2" was the estimation for each UPRN, so surprised "divided_garden_area_m2" needed to be calculated. Where have I misunderstood?

asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py
```diff
@@ -0,0 +1,385 @@
...
+
+# %%
+# Apply selection criteria
+epc_df = epc_df.filter(
+    pl.col("msoa_avg_outdoor_space_m2").is_not_null(),
+    pl.col("divided_garden_area_m2").is_not_null(),
+    pl.col("divided_garden_area_m2") > 0,
+    pl.col("divided_garden_area_m2") < pl.col("divided_garden_area_m2").quantile(0.99),
+    pl.col("weight").is_not_null(),
```
lizgzil, line 320: I can't remember why a weight would be null? Was it when the value was from a dummy property?

asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py
```diff
@@ -0,0 +1,385 @@
...
+# %%
+# Calculate mean estimates per MSOA
+results = epc_df.group_by(["msoa", "msoa_avg_outdoor_space_property_type"]).agg(
+    pl.col("divided_garden_area_m2").mean().alias("mean_divided_garden_area_m2"),
+    pl.col("divided_garden_area_m2").count().alias("n_properties"),
+    pl.col("msoa_avg_outdoor_space_m2").first(),
+)
+
+# Filter results to MSOAs with averages calculated from 15 or more properties
+results = results.filter(pl.col("n_properties") >= 15)
```
lizgzil, line 136: I think here and in the weighted results you could increase this threshold - perhaps set to the 0.1 quantile? (which I think is 51). or was there a reason to use 15?

lizgzil (commented):
Hey @crispy-wonton this looked good to me, and I thought the filtering steps made sense (and in the right order) - I commented on a few bits for clarification though.

asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py
crispy-wonton: This is a good question. The reason is:

For properties that get joined to the same garden (i.e. they share a garden, or at least outdoor space) we decided in a previous sprint review that we would assign all properties which share a garden the overall size of the garden. E.g. if 10 houses share a 100m2 garden, they will all be assigned 100m2 garden size. The logic was that if a building has the appropriate outdoor space, we consider that an apartment within the building has appropriate space in theory.

I had to conduct this extra division step which partitions the size equally among households. E.g. the 10 houses sharing a 100m2 garden will now each get assigned 10m2. I had to do this to avoid double counting gardens when calculating the average. This is flawed, because if there are apartments from the same building missing from the EPC dataset, we will overestimate the divided garden size. E.g. if we only have 5 of the 10 houses sharing the 100m2 garden in our dataset, they would get assigned 20m2 each instead of 10m2.

asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py
crispy-wonton: There wasn't a particular reason behind 15, I think, but we chose a deliberately low number to maximise the number of LSOAs that would be weighted.

For comparing with the average I think it makes sense to increase to 51. We can try it with the weighting as well but bear in mind we will end up with more unweighted LSOA scores.

asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py
crispy-wonton: Here I didn't compare for property types because the dataset is weighted on property type, so I didn't think I can use the weights whilst also filtering for property type because the weights for each property type will add to less than 1. I think this would mean the resulting weighted means would be weighted to the proportion of property type in the LSOA.

However, now I think about it again, I think I could multiply the weights by the number of households per LSOA and calculate the average that way. I will test it out.

asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py
crispy-wonton: Weight can be null if:
- The UPRN did not get matched to an LSOA/DZ code
- There was no census data found for the LSOA/DZ code
- The EPC record did not have the required datapoints for reweighting (property type, tenure, & build year [England and Wales only])

There should not be any dummy rows in the dataset we are using here.

asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py
```diff
@@ -17,11 +28,43 @@
...
+#
+# For this analysis, we have to conduct this extra division step which partitions the garden size equally among households who share a garden. E.g. the 10 houses sharing a 100m2 garden will now each get assigned 10m2. This is required to avoid double counting gardens when calculating the average and to make the resulting garden size estimates comparable to the ONS MSOA-level averages which are calculated by counting shared gardens once in their averages.
+#
+# A caveat of our deduplication here is that if there are apartments from the same building missing from the EPC dataset or missing from our garden size estimates (e.g. because they are missing valid UPRN or there are data gaps in the Microsoft buildings / INSPIRE land parcel datasets), we will overestimate the divided garden size. E.g. if we only have records for 5 of the 10 houses sharing the 100m2 garden in our dataset, they would get assigned 20m2 each instead of 10m2.
```
lizgzil, line 37: super useful to have this here!

lizgzil (approved):
Thanks @crispy-wonton - looks good!

PR #117, opened 2025-01-24
Set null protected areas in E and W to false

### LINKED ISSUE #116: Set `in_protected_area` to True or False (no none)
In `run_add_features.py` I noticed that "in_protected_area" can be True or None for England and Wales, but True or False for Scotland.

This means in `calculate_suitability.py` that the `not_in_protected_area_score` isn't working as it should for Nulls. An example of this is below (I'd expect the nulls should get a score of 1):

Image

I noticed this in investigating why Scotland seemed to have higher scores for HN's across the board - it seems like some of the LSOAs in England and Wales aren't getting the points for not being in a protected area.

Not sure where the best fix to this would be - in `generate_df_uprn_in_cons_area` maybe?

lizgzil:
Fixes #116 (description of why in issue description)

# Instructions for Reviewer

I found that all of England and Wales protected areas were either Null or True, so I think it is correct to set the Nulls to False, but perhaps I am missing something?

I ran 
```
python -i asf_heat_pump_suitability/pipeline/run_scripts/run_add_features.py --epc s3://asf-daps/lakehouse/processed/epc/old/deduplicated/processed_dedupl-0.parquet -y 2023 -q 4
```

to create `'s3://asf-heat-pump-suitability/outputs/2023Q4/20250204_2023_Q4_EPC_features.parquet'`

and then ran

```
python -i asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_suitability.py --weights s3://asf-heat-pump-suitability/outputs/2023Q4/weights/20250102_2023_Q4_EPC_weights.parquet --features s3://asf-heat-pump-suitability/outputs/2023Q4/20250204_2023_Q4_EPC_features.parquet --gardens s3://asf-heat-pump-suitability/outputs/2023Q4/gardens/20250114_2023_Q4_EPC_garden_size_estimates_EWS_deduplicated.parquet -y 2023 -q 4
```

which saved to `s3://asf-heat-pump-suitability/outputs/2023Q4/suitability/20250206_2023_Q4_heat_pump_suitability_per_lsoa.csv`

### DIFF
--- a/asf_heat_pump_suitability/pipeline/prepare_features/protected_areas.py
+++ b/asf_heat_pump_suitability/pipeline/prepare_features/protected_areas.py
@@ -4,6 +4,7 @@ import numpy as np
 import polars as pl
 from asf_heat_pump_suitability import config
 from asf_heat_pump_suitability.getters import get_datasets
+from asf_heat_pump_suitability.pipeline.prepare_features import epc
 
 
 def load_transform_df_uprn_in_protected_area(gdf: gpd.GeoDataFrame) -> pl.DataFrame:
@@ -72,7 +73,7 @@ def generate_df_uprn_in_whs(
     Generate dataframe to flag UPRNs located within World Heritage Sites in Scotland.
 
     Args:
-        gdf (pl.DataFrame): dataframe with point geometries per UPRN in BNG
+        gdf (gpd.GeoDataFrame): dataframe with point geometries per UPRN in BNG
         country_col (str): column containing country names
 
     Returns:
@@ -84,7 +85,10 @@ def generate_df_uprn_in_whs(
     gdf = gdf.sjoin(whs_gdf, how="left", predicate="intersects").drop_duplicates(
         subset="UPRN"
     )
-    gdf["in_world_heritage_site_s"] = gdf["in_world_heritage_site_s"].fillna(False)
+    # Fill null with False if row has geometry, otherwise leave as null
+    gdf.loc[~gdf["geometry"].is_empty, "in_world_heritage_site_s"] = gdf.loc[
+        ~gdf["geometry"].is_empty, "in_world_heritage_site_s"
+    ].fillna(False)
 
     return pl.from_pandas(gdf[["UPRN", "in_world_heritage_site_s"]])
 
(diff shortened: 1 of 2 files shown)


asf_heat_pump_suitability/pipeline/run_scripts/run_add_features.py
```diff
@@ -151,6 +151,13 @@ def parse_arguments() -> argparse.Namespace:
         lad_cons_areas_df, how="left", left_on="lad_code", right_on="LAD23CD"
     )
 
+    epc_df = epc_df.with_columns(
+        pl.when(pl.col("lad_conservation_area_data_available_ew") == True)
```
crispy-wonton:
```suggestion
        pl.when((pl.col("lad_conservation_area_data_available_ew")) & pl.col("COUNTRY").is_in(["England", "Wales"]))
```
I just pushed a change to filter the LAD conservation area data availability results to England and Wales LADs only. Therefore, I'm not sure if we actually need what I've added above, but I think it does add some robustness to the pipeline.

asf_heat_pump_suitability/pipeline/prepare_features/protected_areas.py
crispy-wonton, line 1: I have made 2 corrections in this file, which I noticed when looking at the outputs of your change:

1. For Scottish EPC records, we were filling all records with `False` for `in_protected_area` if the UPRN does not get joined to a World Heritage Site. However, not all records have valid coordinates, which is required to determine whether or not they are in a WHS. Scottish EPC records without coordinates should therefore receive `null`, and only those with coordinates should receive `False` if they are not joined to a WHS.
2. The `generate_df_conservation_area_data_availability` was producing a flag for data availability for all nations in the UK, but it was only using conservation area data from England and Wales to create the flag. Therefore, NI and Scotland were receiving `False` flags (meaning conservation area data is not available) for all their LADs - simply because we haven't actually input any cons area data for them. I have therefore filtered the output to England and Wales LADs only. This should prevent confusion and also increase robustness of the pipeline.

Note: I did notice when i was updating the `generate_df_conservation_area_data_availability` that a single Scottish LSOA (`S12000026`) did have LAD conservation area data available, while all others were False. I think since it appears to be a true Scottish LAD and therefore under Scottish governance, we can safely filter it out and apply the World Heritage Site rule only, but just flagging in case you have different thoughts.

Please do check my logic is correct. Let me know if you have questions.

crispy-wonton (commented):
Your change looks good and I've tested it works. I've added a suggestion to your code snippet which should be redundant with additional changes I have pushed, but increases robustness.

I noticed when I was looking at the outputs of your updated code that we have erroneously assigned `False` to some Scottish properties for `in_protected_area` flag, and that our `generate_df_conservation_area_data_availability` function was misleadingly retaining NI and Scotland LADs even though we do not use conservation area data for them. I have pushed fixes for these to this branch for you to take a look at, as well as a more detailed explanation in my comment on the file.

lizgzil:
Good spotting and suggestions @crispy-wonton - I think these look good. I'll start running it!

PR #118, opened 2025-01-28
Plymouth analysis

lizgzil:
---

- no weighting applied for features

# Description

Please include a summary of the changes.

Fixes # (issue)

# Instructions for Reviewer

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/analysis/area_specific/combine_plymouth_data.py
+++ b/asf_heat_pump_suitability/analysis/area_specific/combine_plymouth_data.py
@@ -96,18 +96,69 @@ def get_tech_suitability_overall_type(ashp, hn, ashp_high_thresh, hn_high_thresh
 
 # Ad hoc manually created categories for Plymouth
 def get_tech_suitability_manual(ashp, hn):
-    if ashp < 0.68:
+    if (ashp < 0.68) & (hn > 0.7):
         return "HN suitable"
     else:
-        if (hn > 0.46) & (ashp < 0.82):
+        if ashp < 0.82:
             return "Multiple technologies may be feasible"
         else:
             return "ASHP suitable"
 
 
+def add_tenure_proportions(lsoa_df: pd.DataFrame) -> pd.DataFrame:
+
+    # Join with tenure data
+    census_tenure = get_target.transform_df_target_tenure()
+    lsoa_df_tenure = lsoa_df.merge(census_tenure.to_pandas(), how="left", on="lsoa")
+
+    # Get proportions of each tenure type
+    tenure_types = ["owner-occupied", "rental (social)", "rental (private)"]
+    lsoa_df_tenure["total_census_tenure_properties"] = lsoa_df_tenure[tenure_types].sum(
+        axis=1
+    )
+    for tenure_type in tenure_types:
+        lsoa_df_tenure[f"Proportion {tenure_type}"] = (
+            lsoa_df_tenure[tenure_type]
+            / lsoa_df_tenure["total_census_tenure_properties"]
+        )
+
+    return lsoa_df_tenure
+
+
+def get_lsoa_proportion_flats(per_property_df: pd.DataFrame) -> pd.DataFrame:
+
+    # Proportion of flats per LSOA
+    per_property_df["is_flat"] = (
+        per_property_df["property_type"] == "Flat, maisonette or apartment"
+    ).astype(int)
+    per_property_df_grouped = per_property_df.groupby("lsoa")
+    num_flats_per_lsoa = per_property_df_grouped["is_flat"].sum().reset_index()
+    num_props_per_lsoa = per_property_df_grouped.size().reset_index()
+    flats_per_lsoa = num_flats_per_lsoa.merge(num_props_per_lsoa, on="lsoa")
+    flats_per_lsoa["%flats"] = flats_per_lsoa["is_flat"] * 100 / flats_per_lsoa[0]
+
+    return flats_per_lsoa
+
+
 if __name__ == "__main__":
 
-    # Suitability scores + features per property
+    # Suitability per LSOA
+    suitability_data = pd.read_parquet(suitablitity_per_lsoa_file)
+
+    # Filter data for a particular place
+
+    place_name = "Plymouth"  # Word in the LSOA name
+    plymouth_lsoas_df = suitability_data[
+        suitability_data["lsoa_name"].apply(lambda x: place_name in str(x))
+    ]
+    plymouth_lsoas_list = list(plymouth_lsoas_df["lsoa"].unique())
+
+    plymouth_lsoas_df = add_tenure_proportions(plymouth_lsoas_df)
+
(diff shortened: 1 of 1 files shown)


asf_heat_pump_suitability/analysis/area_specific/combine_plymouth_data.py
```diff
@@ -0,0 +1,331 @@
...
+
+# Ad hoc manually created categories for Plymouth
+def get_tech_suitability_manual(ashp, hn):
+    if (ashp < 0.68) & (hn > 0.7):
+        return "HN suitable"
+    else:
+        if ashp < 0.82:
+            return "Multiple technologies may be feasible"
+        else:
+            return "ASHP suitable"
```
crispy-wonton, line 58: I may be misinterpreting the use case of this function but is there an outcome missing where none is suitable?

E.g. it looks like if `ashp == 0` and `hn == 0` for example, the function would return `"Multiple technologies may be feasible"` even though none are.

asf_heat_pump_suitability/analysis/area_specific/combine_plymouth_data.py
```diff
@@ -0,0 +1,331 @@
...
+    # Features per property
+
+    per_prop_data = pl.read_parquet(
+        suitability_per_property_file,
+        columns=[
+            "UPRN",
+            "lsoa",
+            "LATITUDE",
+            "LONGITUDE",
+            "UPRN_duplicated",
```
crispy-wonton, line 123: This needs to be removed from the suitability per property file. Not sure why it's still in there, I thought I had fixed the issue.

asf_heat_pump_suitability/analysis/area_specific/combine_plymouth_data.py
```diff
@@ -0,0 +1,331 @@
...
+        census_prop_type_plymouth.to_pandas()[["lsoa", "census_%flats"]],
+        how="left",
+        on="lsoa",
+    )
+
+    # Get categorisations of HN/ASHP zones
+
+    plymouth_lsoas_gdf = plymouth_lsoas_gdf.round(3).rename(columns=column_rename_dict)
+
+    quantile_thresh = 0.6
```
crispy-wonton: What's the reason for choosing 0.6?

asf_heat_pump_suitability/analysis/area_specific/combine_plymouth_data.py
```diff
@@ -0,0 +1,331 @@
...
+    ).with_columns(
+        (pl.col("Flat, maisonette or apartment") * 100 / pl.col("sum")).alias(
+            "census_%flats"
+        )
+    )
+    plymouth_lsoas_gdf = plymouth_lsoas_gdf.merge(
+        census_prop_type_plymouth.to_pandas()[["lsoa", "census_%flats"]],
+        how="left",
+        on="lsoa",
+    )
```
crispy-wonton, line 193: I would use these census proportions when it comes to using them, rather than proportions calculated from the EPC data.

asf_heat_pump_suitability/analysis/area_specific/combine_plymouth_data.py
```diff
@@ -0,0 +1,331 @@
...
+                "SGL - Conventional",
+                "SGL - Nesta",
+                "HN - Conventional",
+                "HN - Nesta",
+                "garden_area_m2",
+            ]
+        ).round(3)
+    )
+
+    plymouth_per_prop_data_extra.write_csv("plymouth_per_prop_data_extra.csv")
```
crispy-wonton: This path could be updated to save to the `/outputs` dir instead.

crispy-wonton (commented):
Thanks @lizgzil for working on this analysis!

I ran the code and it worked. The outputs seem ok individually, but the different suitability types seem to be contradictory to me, although maybe this is intended - let me know if so. 

I've joined the results for the following columns in the following order, below and printed their value counts:
Overall suitability type - shared/not -> Overall suitability type -> Manual suitability type
```
Individual suitable only -> ASHP suitable only -> ASHP suitable                                 50
Shared suitable only -> HN suitable only -> Multiple technologies may be feasible               26
Neither suitable -> Neither ASHP or HN suitable -> ASHP suitable                                24
Both individual and shared suitable -> Both ASHP and HN suitable -> ASHP suitable               16
Shared suitable only -> Neither ASHP or HN suitable -> Multiple technologies may be feasible    16
Shared suitable only -> HN suitable only -> ASHP suitable                                       13
Shared suitable only -> HN suitable only -> HN suitable                                         10
Shared suitable only -> Neither ASHP or HN suitable -> ASHP suitable                             7
Individual suitable only -> Neither ASHP or HN suitable -> ASHP suitable                         1
Both individual and shared suitable -> HN suitable only -> ASHP suitable                         1
```

Beyond that, the other functions look correct and appropriate and the code is easy to read. Just needs some refactoring and docstrings.

asf_heat_pump_suitability/analysis/area_specific/combine_plymouth_data.py
lizgzil: I know what you mean. I think the language is because we don't want to say that HPs are never suitable. And either way, for Plymouth (and I think most places) these values are never that low anyway.

asf_heat_pump_suitability/analysis/area_specific/combine_plymouth_data.py
lizgzil: Good to know! I think we did in the maps - but will make sure going forward

asf_heat_pump_suitability/analysis/area_specific/combine_plymouth_data.py
lizgzil: Yeh it feels a bit random, it was because a more normal quantile like 0.75 gave no values for one of the categories, so I felt like I needed to lower it in order to get at least a few LSOAs in each category. Either way, we didn't actually use this categorisation in the mapping, so this could be deleted.

asf_heat_pump_suitability/analysis/area_specific/combine_plymouth_data.py
lizgzil: very true!

lizgzil:
Thanks a lot for looking at this @crispy-wonton !! 

I think the 3 types of categorisations are confusing and unnecessary anyway since we only use one in the map (the other 2 are legacy).

It didn't feel like there were other bugs you noticed (apart from the UPRN_deduplicated which will probably cause problems in the future), so thinking my next steps are
- clean this notebook up a bit (remove the 3 types of categorisation, save to the outputs folder)
- run with latest data

Then maybe it can be merged and refactoring steps could happen later if/when we apply this to other regions.

lizgzil:
hey @crispy-wonton - I made the changes as mentioned in the previous reply - thanks a lot! I will merge now, and do any refactoring when/if parts of this code are used again.

PR #124, opened 2025-02-11
123 check building cons area data

### LINKED ISSUE #123: Check building conservation area data
We know all local authorities have at least one building conservation area. We use building conservation area polygons from Historic England and Welsh Gov websites. These are collated from the local authority level. We want to assess whether we can assume that if a local authority has any building conservation areas in these collated datasets, then the dataset is complete for that local authority, i.e. it is not missing any building conservation areas.

crispy-wonton:
Fixes #123 

# Description

- Update England building conservation area dataset with Feb 2025 data. NB: this will change our results so we need to rerun the script to ADD FEATURES and the one to CALCULATE SUITABILITY. I can do this after this has been reviewed.
- Add deduplication of geometries in England and Wales building conservation areas (this shouldn't change our results)
- Add analysis of Historic England building conservation area dataset to check completeness (described in issue and notebook)

# Instructions for Reviewer

In terminal, run the following line:
`jupytext --to notebook asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py`

Then you can run the notebook.

- Analysis is done correctly, especially the way joins are made
- Results of the analysis. Ultimately this analysis was intended to help us decide whether we can fill nulls of `in_protected_area` with `False`. I think this analysis shows we could, although we would have to accept that there may be some missing conservation areas for some LAs. Let me know what you think.
- Is there anything further you would be interested in me exploring? We could go down the route of improving the deduplication but I didn't want to get too far into that before sharing these initial results.

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
```diff
@@ -0,0 +1,319 @@
...
+    "in_conservation_area_ew"
+].astype(bool)
+cons_areas_df = pl.from_pandas(cons_areas_gdf)
+
+# %%
+# Load manually collated building conservation area count data for 51 local authorities and join to conservation areas
+counts_df = pl.read_csv(
+    "s3://asf-heat-pump-suitability/evaluation/building_conservation_area_datasets/building_conservation_area_counts_sample.csv"
+)
+cons_areas_df = cons_areas_df.join(counts_df, how="inner", on="LAD23NM")
```
caldwellst: This is dropping 3 rows where we have no conservation areas: Wakefield, Wokingham, and Leeds. Not sure if intentional, but if not, should do an `outer` join and fill in `in_conservation_area_ew` with 0 I think?

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
```diff
@@ -0,0 +1,319 @@
...
+    columns=["sitename", "geometry"]
+)
+full_cons_areas_gdf = pd.concat([e_gdf, w_gdf]).drop_duplicates(subset=["geometry"])
+full_cons_areas_gdf["in_conservation_area_ew"] = True
+
+# %%
+# Load geospatial boundaries of local authorities
+council_bounds = get_datasets.load_gdf_ons_council_bounds()
+
+# %%
```
caldwellst, line 55: What's the logic of this following section? Lose about 7% of our data here. Just that we don't trust the conservation area if it falls outside the council bounds? Would be good to add jsutification here.

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
caldwellst: These seem to be list in the initial filtering on overlay? Or at least some of them.

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
```diff
@@ -0,0 +1,319 @@
...
+
+# %% [markdown]
+# ### Compare Historic England and Wales Gov polygons to local authority polygons
+
+# %%
+cons_areas_df = epc.extend_df_country_col(cons_areas_df, lsoa_col="LAD23CD")
+
+# %%
+# LADs with far too many / few conservation areas
+sample_lads = cons_areas_df.filter((pl.col("diff") <= -30) | (pl.col("diff") >= 10))[
```
caldwellst: Any reason not to include the LAs with <= -10?

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
```diff
@@ -0,0 +1,319 @@
...
+    "LAD23CD"
+].to_list()
+
+# Add a random sample of LADs from England and Wales
```
caldwellst: Why?

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
```diff
@@ -0,0 +1,319 @@
...
+len(set(cons_areas_df.filter(pl.col("LAD23NM") == "East Hampshire")["sitename"][0]))
+
+# %%
+# Check which conservation areas are missing from Historic England
+set(eh_cons_gdf.drop_duplicates(subset=["name"])["name"]).difference(
+    set(cons_areas_df.filter(pl.col("LAD23NM") == "East Hampshire")["sitename"][0])
+)
+
+# %%
+# We can see that 'Sir George Staunton' building conservation area was joined to East Hampshire, it just has a low % overlay (23%) so was filtered out.
```
caldwellst: Interested in why dropping due to overlay in these cases then?

caldwellst (commented):
Hey Roisin, nice stuff!

I left some comments in the notebook. Overall the analysis is okay, I think one of the joins needs fixing, and a bit more commenting might've helped me track. I've put questions where I had some issues.

However, overall, the results seem to indicate we shouldn't be dropping. I thought it would be good to just go through and check the two most "off" local authorities.

Stratford-on-Avon is just missing loads of data compared in the Historic England dataset, as you note in your analysis. I looked at [their website](https://www.stratford.gov.uk/planning-building/conservation-areas-h-z.cfm) and just can't seem to find any of the conservation areas they list, when searching by name. Oxford Canal is weirdly in the HE data, but isn't on their site, but the other area in the dataset we have, Whichford, is in there. Seems not much we can do on this missingness issue?

However, I checked out as well Derbyshire Dales. This is where we have too many options compared to the Derbyshire Dales reported [conservation areas](https://www.derbyshiredales.gov.uk/planning/conservation/conservation-areas). Some manual investigation, and these extras are actually conservation areas within the [Peak District National Park](https://www.derbyshiredales.gov.uk/planning/conservation/conservation-areas)!

Could this be a common driver of differences? LAs may not be responsible for management of conservation areas that fall under a national park designation, but that isn't reflected in the HE/Wales data?

On duplicated names, I think there are a few different options that you outline! I checked a few by spot, and a lot seem to be contiguous areas stored as separate shapes. However there are some that are not next to each other. Not sure if you'd be able to fully validate these. These are probably cases where it's difficult to automatically do anything, and we wouldn't necessarily want to with more confidence. The problem case would be ones like I saw in my GitHub notifications where an entire area is classed as a conservation area.

You could check that there are no conservation area polygons fully contained in another. Could do a spatial join, where you expect each area to only contain itself, and flag others to check. Those would be ones where maybe we manually check and if agree, automatically remove, so as not to overestimate conservation area through some wild setup! Maybe @lizgzil  already checked for this in #130?

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
crispy-wonton: Sorry, you're right about the documentation, there isn't enough! I will go back and add more.

Because the analysis is focussing on conservation area data availability for Local Authority Districts (councils), the steps below are meant to join the conservation areas to their local authorities. I tried out a couple of different join methods and this one seemed to produce the best results when comparing to our count data.

E.g. `sjoin` with `intersects` predicate seemed to be too permissive and there was a high number of LADs with too many conservation areas. `sjoin` with `contains` predicate (as in, LAD contains conservation areas) seemed to be too restrictive and there was a high number of LADs with not enough conservation areas.

As a result, I chose this methodology where I calculated what percentage of each conservation area fell within the LAD and then kept those where at least 10% of the conservation area was found in the LAD. This was to avoid small intersections with neighbouring LAD conservation areas being matched because my assumption was that they wouldn't be included in the conservation area count by the LAD.

However, this is definitely an assumption so happy to change the methodology. Did you have another idea of how it could be improved?

Also, when you say we lose 7% of our data, which data do you mean?

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
crispy-wonton: I wanted to make sure I was sampling from all LADs as well as the ones with extreme errors.

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
crispy-wonton: Great spot thank you! Yes you are correct they are lost in the filtering step. Whenever I remove the filtering, I can see they are all retained. Between them they get matched to 11 conservation areas with all but 1 match being <1% of the conservation area. The other is 2.96%.

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
crispy-wonton: Not beyond reducing the number of LADs for manual checks due to time constraints, but I think it's a good idea to add.

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
crispy-wonton: See previous comment. Not sure about the methodology now though. Maybe we should just do an `sjoin` with intersection. I'm not sure if the cause of the discrepancies is that the HE and Wales Gov datasets overestimate the number of conservation areas, or that Local Authority websites don't accurately report how many conservation areas are actually in their LAD - e.g. if 0.1% of a conservation area is in their LAD, do they list it on their website? I suppose I can see a next step for me to investigate those cases specifically!

Either way, I should note that we don't apply this logic in the pipeline when joining properties to conservation areas. For that we use `sjoin` with `intersects`. This notebook is to determine if the conservation dataset we're using is accurate enough that we can label properties as `not_in_conservation_area` for LADs with conservation area data where no match with the property is found.

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
caldwellst: Ah I see!

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
caldwellst: I see now! Was slightly misunderstanding what was happening, but this is just for checking against the current setup. The approach makes sense. You could consider just taking the max for each to maintain the original row count, but after your filtration I think there's only like 20 or so rows duplicated across, so not a huge deal for just this checking. Was getting confused with this approach for analysis and how it might be applied across LSOAs, but see now what was happening.

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
caldwellst: Sorry, was confused at what we were "sampling" and doing, later on in the review was clearer after running through that piece of code.

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
caldwellst: Yeah, that's where I was getting confused, this process versus the LSOA pipeline!

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
```diff
@@ -0,0 +1,319 @@
...
+full_cons_areas_gdf = pd.concat([e_gdf, w_gdf]).drop_duplicates(subset=["geometry"])
+full_cons_areas_gdf["in_conservation_area_ew"] = True
+
+# %%
+# Load geospatial boundaries of local authorities
+council_bounds = get_datasets.load_gdf_ons_council_bounds()
+
+# %%
+full_cons_areas_gdf["cons_area_size_m2"] = full_cons_areas_gdf["geometry"].area
+full_cons_areas_gdf = gpd.overlay(
```
caldwellst: Also, just noticed that because we are doing the intersection, we are dropping over 400 rows from the original data.  A lot of this in places like Angelsey where I can see Beaumaris and the Menai Bridge, but also some villages in England as I spot check. I guess not to much to do since we are validating against the boundaries we have, but interesting as it could be another way we are losing data relative to boundaries that we do have.

crispy-wonton: I've updated the code. There are 9 conservation areas which are dropped during the join from the original data. I can see they are dropped after the initial overlay so there must not be any LAD boundary data for the areas where they are. I think this is ok to exclude from our analysis as it's such a small percentage of the conservation areas.

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
crispy-wonton: With the new condition for filtering (keeping conservation area overlays over 50m2), I can see this is no longer an issue and the three areas you mentioned are now included.
Also, I thought about instances where there are 0 conservation areas joined. I think those rows should be excluded anyway, because that's clearly erroneous - every LA should have at least 1 conservation area, so we can remove LADs with counts of 0 from this analysis.

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
crispy-wonton: Have updated the code. It doesn't make a difference to the analysis but to avoid confusion i changed it to -10.

asf_heat_pump_suitability/config/README.md
```diff
@@ -38,7 +38,7 @@ See additional data attribution statements listed below.
...
-| In building conservation area | `["E_historic_england_conservation_areas"]`                | [Historic England](https://www.planning.data.gov.uk/dataset/conservation-area)<sup>10</sup>                                                                                                                                                                                | June 2024                         | Building conservation area boundaries in England (places of special architectural and historic interest).                      |
+| In building conservation area | `["E_historic_england_conservation_areas"]`                | [Historic England](https://www.planning.data.gov.uk/dataset/conservation-area)<sup>10</sup>                                                                                                                                                                                | Feb 2025                          | Building conservation area boundaries in England (places of special architectural and historic interest).                      |
```
lizgzil: Could do this in another PR - but just flagging the issue for updating this dataset https://github.com/nestauk/asf_heat_pump_suitability/issues/167

Although I see the website says "The data may be incomplete and not yet cover all of England. We're working to improve its availability and quality."

asf_heat_pump_suitability/analysis/protected_areas/20250207_missing_protected_areas.py
```diff
@@ -0,0 +1,284 @@
...
+#
+# For this analysis, we have two key datasets:
+# 1. Geospatial polygons of building conservation areas from Historic England and the Welsh government. This is the dataset we use to create the `in_protected_area` feature. We are aiming to verify the coverage of conservation areas for each local authority.
+# 2. A manually collated building conservation area count dataset for 51 local authorities. We visited the gov websites of these 51 local authorities and recorded the number of conservation areas they report within their boundaries. This provides the ground truth to compare our counts against.
```
lizgzil, line 10: I guess this is a good reason to try to use a Historic England dataset from around the same time as you manually collated this. Adding the more recent HE dataset might mean you need to revisit this manual work in order to perform this evaluation.

lizgzil (dismissed):
think this looks great - thanks @crispy-wonton 🎉 . I didn't see any bugs. Ignore my comment about potentially updating the data - I can see how you actually need to use a version of it from the same time as your manual collated data.

I wonder whether some of the large differences in numbers of conservation zones could be contributed to if the gov websites are out of date too.

PR #131, opened 2025-02-13
Add garden size estimate distribution

crispy-wonton:
---

# Description

Add new garden size estimate distribution plot to `20250106_evaluate_garden_size_estimates.py` notebook. Also update the plot colours to use Nesta colours

# Instructions for Reviewer

Run the following line in terminal to generate the notebook:
`jupytext --to notebook asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py`

Then you can run the notebook as usual.

**I advise to skip the cell starting with the note "# We load the interim garden size files again and deduplicate them". Instead load the results of this processing directly from S3 with the cell below it. It is much faster.**

The dataset that is used to generate the plot. My intention here is to show the distribution of estimated garden sizes without double counting. E.g. each garden (including shared gardens) should only be counted once in the histogram. And they should be counted as the full garden size, not divided evenly among properties as the rest of the analysis uses.

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py
+++ b/asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py
@@ -3,6 +3,7 @@ import polars as pl
 import numpy as np
 import logging
 import matplotlib.pyplot as plt
+import matplotlib.ticker as ticker
 from datetime import datetime
 from sklearn import metrics
 from asf_heat_pump_suitability.getters import base_getters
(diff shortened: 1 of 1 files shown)


asf_heat_pump_suitability/analysis/garden_size_estimates/20250106_evaluate_garden_size_estimates.py
```diff
@@ -83,13 +88,51 @@
     uprn_count, how="left", on="NATIONALCADASTRALREFERENCE"
 ).join(cadastral_garden_size, how="left", on="NATIONALCADASTRALREFERENCE")
 
+# %%
+# Look at garden size estimate distribution
+plot_df = gardens_df.unique(subset="NATIONALCADASTRALREFERENCE")
```
crispy-wonton, line 93: This is `polars` method of deduplication.

lizgzil (approved):
All looks good to me!

PR #141, opened 2025-03-19
139 update open data

### LINKED ISSUE #94: Add names of Scottish data zones to final suitability output dataset
We need to add the common names for the Scottish Data Zones to the final suitability output so this information can be included in pop-ups in the suitability map.

### LINKED ISSUE #139: Update script to generate open dataset with new features
Add summary statistics of new features since phase 1 to the script to generate the open LSOA/DZ version of the HP suitability score dataset.

crispy-wonton:
Fixes #139 

# Description

Update processing to produce open version of the LSOA HP suitability dataset and refactor separate dataset generation script to combine it with `run_calculate_suitability.py` script.

New files:
- `pipeline/prepare_features/property_type.py` - new file with function to calculate proportions of each property type in each LSOA using census data
- `pipeline/suitability/aggregate_features.py` - new file with functions to aggregate features to LSOA-level and provide averages for each LSOA

Modified files:
- `config/README.md` - minor correction to README (a source was listed under the wrong feature), and add lookup tables used in a new 'Other datasets' table at the bottom of the README
- `config/base.yaml` - new data source added for DZ code to name lookup and move some parameters from `run_calculate_suitability.py` to the config
- `getters/get_datasets.py` - add two new getter functions to load LSOA / DZ lookup tables. this was so we can match LSOA/DZ codes to their names
- `getters/get_target.py` - correction to function made in data story branch. There was a category of census flat data missing which has now been added. I need this function to calculate the proportion of flats in each LSOA according to the census, hence copying the correction to this branch also.
- `pipeline/prepare_features/epc.py` - add new function to convert EPC rating to a binary depending on whether it scores a C or above
- `pipeline/prepare_features/output_areas.py` - add new function to concatenate all LSOA/DZ codes and their matching names into one dataframe
- `pipeline/run_scripts/run_calculate_suitability.py` - refactor to also aggregate features per LSOA, not just calculate suitability of each tech type. Also add Scottish DZ names to the dataset which were previously missing
- `pipeline/suitability/calculate_suitability.py` - add some filtering to increase speed of data processing and add new function to create dict of both HP suitability scores and aggregated features per LSOA

NOT FOR REVIEW:
- `pipeline/run_scripts/run_generate_open_dataset.py` - I started making edits to this script before I decided to refactor it and combine with `run_calculate_suitability.py`. Once we are happy with `run_calculate_suitability.py` we can delete this script. But it might be useful for reference when reviewing (or not).

# Instructions for Reviewer

To create both the suitability scoring datasets and the final open dataset, I ran the following line:
`python -i asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_suitability.py --weights s3://asf-heat-pump-suitability/outputs/2023Q4/weights/20250305_2023_Q4_EPC_weights.parquet --gardens s3://asf-heat-pump-suitability/outputs/2023Q4/gardens/20250224_2023_Q4_EPC_garden_size_estimates_EWS_deduplicated.parquet --features s3://asf-heat-pump-suitability/outputs/2023Q4/features/2023_Q4_EPC_features.parquet -q 4 -y 2023` 

It takes about 50 minutes to run (this is mainly due to 20 mins of time to save the per property dataset, so if you comment that line out it should go much quicker). The results are here for you to look at:
`s3://nesta-open-data/asf_heat_pump_suitability/2023Q4/20250319_2023_Q4_EPC_heat_pump_suitability_per_lsoa.parquet`

Once we're happy with the results, I'll create the excel version with the `README` and `key` sheets.

- the calculations to aggregate the features with the weights - I have updated the way we calculate them since the last version of the script because I think for weighting the features, we should have been using the original weights but instead we were using the recalculated weights after removing the dummies. As we discussed, the recalculated weights are fine to use for the scoring, but I think we should use the original weights for calculating the proportions of different features. Definitely want your thoughts though!
- this function: `aggregate_dict_lsoa_suitability_and_features` in `calculate_suitability.py`. It took me a while to figure out the best way to process both the features and the scoring. The weights was the main thing that made this difficult (especially because for some LSOAs we don't use the weights so I had to account for that). I think it works and I've added notation to the function to say what it's meant to be doing at each line but would be good to have you taking a careful look at it too. Thank you!

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/pipeline/suitability/calculate_suitability.py
```diff
@@ -258,3 +263,35 @@ def compute_dict_lsoa_suitability_scores(df: pl.DataFrame, lsoa: str) -> dict:
     scores_dict["n_properties"] = len(df)
 
     return scores_dict
+
+
+def aggregate_dict_lsoa_suitability_and_features(
```
crispy-wonton, line 268: This function needs careful checking.

asf_heat_pump_suitability/pipeline/suitability/aggregate_features.py
crispy-wonton, line 1: The functions in this file need careful checking as this is where the main calculations to aggregate the features are defined.

asf_heat_pump_suitability/pipeline/run_scripts/run_generate_open_dataset.py
crispy-wonton, line 1: You can skip this file because we don't use it anymore as it has been refactored into `run_calculate_suitability.py`. As mentioned in PR description, I will delete this when we're happy with the `run_calculate_suitability.py` script.

asf_heat_pump_suitability/pipeline/suitability/aggregate_features.py
crispy-wonton, line 1: The functions in this file need careful checking as this is where the main calculations to aggregate the features are defined.

asf_heat_pump_suitability/pipeline/suitability/calculate_suitability.py
```diff
@@ -226,7 +229,8 @@ def compute_df_weighted_score(df, threshold=0.5):
         .then(True)
         .otherwise(False)
         .alias("scores_weighted"),
-    )
+    ).filter(pl.col("use_weight").is_not_null())
```
lizgzil, line 232: trying to work out what difference this makes to the results. In the past were null "use_weight" rows also not included in the LSOA averages?

asf_heat_pump_suitability/pipeline/suitability/calculate_suitability.py
```diff
@@ -258,3 +263,35 @@ def compute_dict_lsoa_suitability_scores(df: pl.DataFrame, lsoa: str) -> dict:
...
+    # Filter df to specified LSOA
+    lsoa_df = df.filter(pl.col("lsoa") == lsoa_code)
+
+    # Compute weighted score for each property in one LSOA - this dataset will be one row per property in a single LSOA
+    lsoa_df = compute_df_weighted_score(lsoa_df)
+
+    # Weight and aggregate the features to LSOA-level for one LSOA
+    lsoa_df = aggregate_features.extend_df_feature_weight(lsoa_df)
+    lsoa_df = epc.convert_df_epc_rating(lsoa_df)
+    lsoa_features_dict = aggregate_features.aggregate_dict_features_per_lsoa(lsoa_df)
```
lizgzil, line 291: took me a while to get my head around the 'feature_weight' and 'total_weight' idea, but think this is a really elegant way to deal with it!!

asf_heat_pump_suitability/pipeline/suitability/calculate_suitability.py
lizgzil: I think this means `n_properties` won't include this rows, but thats ok right since they are the dummy rows and perhaps shoudn't be included anyway?

lizgzil (approved):
Screenshot 2025-03-20 at 15 36 22

I think this looks good. I've done some checks that the previous average scores are the same (which they are), the only thing that changes is the n_properties column, but I think it's more accurate now due to removing dummies (if I've got it right)?
The new columns and the weighting approach included looks good to me - can't see an issue.

asf_heat_pump_suitability/pipeline/suitability/calculate_suitability.py
crispy-wonton: Yep the reason I added this was to correct the number of `n_properties`. At this point, dummy rows are already removed. It's removing properties which didn't have sufficient features to be included in the analysis. The `n_properties` feature is described in our README as number of properties included in analysis per LSOA/DZ. I wanted this to represent the sample size of properties that the results are based on , rather than the total property count in the LSOA/DZ. We could also add a `total_properties` column with all the properties in the LSOA/DZ.

PR #149, opened 2025-04-08
Add preprocessing for flats with gas analysis

crispy-wonton:
# Description

This PR contains the preprocessing functions and scripts for an ad-hoc analysis requested by ASF team members. This PR contains the functions and scripts for processing the EPC data in preparation for the analysis, as well as 2 notebooks to explore the data and identify the best method for filling missing storey count data.

I will create a separate PR with the final notebook which uses the processed data to answer the research questions. Summary:

These modules and scripts are first in the analysis 'pipeline', used before any notebooks. I recommend reviewing them first. These conduct the general preprocessing required for the analysis:
- `building_rise.py` - functions to calculate building rise from storey count data
- `fuel_type.py` - functions to extract central heating information from EPC data including fuel type, communal heating flag, and fossil fuel heating flag
- `run_join_epc_building_footprints.py` - script to join EPC records to a corresponding building footprint from the Microsoft Global ML building dataset if available. This is how we join the building height data to the EPC records.
- `run_process_epc_flats.py` - script to preprocess EPC flats data to prepare for analysis. This is how we process the central heating information that we need for the analysis.

The notebooks use the outputs of the preprocessing. At this point in the processing, 87% of the EPC records for flats are missing flat storey count data. The notebooks both exist to identify how to fill the missing flat storey count data which is required for the final analysis to answer the research questions. 
I recommend reviewing the files in the following order:
- `analysis_explore_storey_and_height_data.py` - Here we investigate the FLAT STOREY COUNT and building height data to get a sense of their quality and how we can apply the building height data to fill in the storey count variable.
- `analysis_fill_building_rise_methods.py` - Here we test 4 different basic methodologies for converting building height data to FLAT STOREY COUNT data. We compare the methodologies at the end to select the best one.

So to recap: at the end of this mini pipeline outlined in this PR, we should have a preprocessed dataset ready to answer the final research questions. The dataset should contain EPC records of flats/maisonettes only and should be enhanced with building rise information as well as information about the central heating system.

# Instructions for Reviewer

- To join the EPC data with the building footprints, you need to run the line below. I suggest running this through a couple of iterations just to check it works but not all the way through as it takes a while:
`python -i asf_heat_pump_suitability/analysis/flats_on_fossils/run_join_epc_building_footprints.py --epc s3://asf-daps/lakehouse/2024_Q3/processed/epc/deduplicated/processed_dedupl-0.parquet -y 2024 -q 3`

- To preprocess the EPC flats data in preparation for the analysis conducted in notebooks, you need to run:
- `python -i asf_heat_pump_suitability/analysis/flats_on_fossils/run_process_epc_flats.py --epc s3://asf-daps/lakehouse/2024_Q3/processed/epc/deduplicated/processed_dedupl-0.parquet -y 2024 -q 3`

To convert the notebooks from python to ipynb:
```
pip install jupytext
jupytext --to notebook asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
jupytext --to notebook asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
```

You can then run them as normal in your IDE.

Note this analysis uses the latest EPC data (Q3 2024).

- The `analysis_fill_building_rise_methods.py` notebook. I'm keen to hear your feedback on the methodology. Do you think I've made the right methodological decisions and appropriate assumptions? Is there anything that seems wrong or weird to you?
- Have I missed any particularly useful investigative analysis that you think would add good insights to the data?
- General check for bugs and errors in any of the preprocessing modules and scripts. Anything weird you notice in them?

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
```diff
@@ -0,0 +1,91 @@
+import polars as pl
+import numpy as np
```
helloaidank: I don't think we use np in this script?

asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
```diff
@@ -0,0 +1,91 @@
...
+    Returns:
+        pl.DataFrame: dataset with all duplicated UPRNs removed
+    """
+    n = df["UPRN"].is_duplicated().sum()
+    logging.info(f"Dropping {n} rows with duplicated UPRNs")
+    return df.unique(subset="UPRN", keep="none")
+
+
+def extend_df_building_rise_from_storey_count(
+    df: pl.DataFrame, storey_col: str, building_rise_col: str
```
helloaidank, line 46: Is there a reason we want to define this as a function argument, does the `storey_col` vary?

asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
```diff
@@ -0,0 +1,91 @@
...
+    Args:
+        df (pl.DataFrame): dataset with storey count per record
+        storey_col (str): name of column containing storey count data
+        building_rise_col (str): name of new column containing building rise classification
+
+    Returns:
+        pl.DataFrame: dataframe with building rise classification
+    """
+    df = df.with_columns(
+        pl.when((pl.col(storey_col) > 0) & (pl.col(storey_col) <= 3))
```
helloaidank, line 61: Curious about the zero storey, are there places which are underground and therefore we aren't taking into account (these are likely basements).

asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
```diff
@@ -0,0 +1,91 @@
...
+
+    Args:
+        df (pl.DataFrame): dataset with storey count per record
+        storey_col (str): name of column containing storey count data
+        building_rise_col (str): name of new column containing building rise classification
+
+    Returns:
+        pl.DataFrame: dataframe with building rise classification
+    """
+    df = df.with_columns(
```
helloaidank, line 60: Love this chain of polar functions, so nice.

asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
```diff
@@ -0,0 +1,91 @@
...
+    logging.info(f"Dropping {n} rows with duplicated UPRNs")
+    return df.unique(subset="UPRN", keep="none")
+
+
+def extend_df_building_rise_from_storey_count(
+    df: pl.DataFrame, storey_col: str, building_rise_col: str
+) -> pl.DataFrame:
+    """
+    Add new column to dataframe with building rise classification derived from storey count. Buildings are classified
+    into low- (<=3 storeys), medium- (4-10 storeys) and high-rise (>10 storeys).
```
helloaidank: maybe more accurate to say between 1-3 to be consistent with the logic you have?

asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
```diff
@@ -0,0 +1,91 @@
...
+        pl.DataFrame: EPC data with cleaned `FLAT_STOREY_COUNT` column
+    """
+    logging.warning("Replacing FLAT_STOREY_COUNT values of `20+` with `20`")
+    df = df.with_columns(
+        pl.col("FLAT_STOREY_COUNT")
+        .cast(pl.String)
+        .replace(
+            "20+", "20"
+        )  # This is appropriate here because we are using storeys to determine rise
+        .replace("", "unknown")
```
helloaidank, line 89: Are there potentially any strings within this column that should be classified as unknown but aren't a empty string?

asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
```diff
@@ -0,0 +1,91 @@
...
+        .cast(pl.Float64, strict=False)
+    ).with_columns(
+        pl.when(pl.col("FLAT_STOREY_COUNT") <= 0)
```
helloaidank, line 93: Understand making negative values, but I guess building on the above comment, might be outlier buildings which actually have zero stories.

asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
```diff
@@ -0,0 +1,91 @@
+import polars as pl
```
helloaidank, line 8: personally like adding a small description at the beginning of each file to describe what's contained within there, understand we have different styles!

asf_heat_pump_suitability/analysis/flats_on_fossils/features/fuel_type.py
```diff
@@ -0,0 +1,194 @@
+import polars as pl
+
+
+def extend_df_central_heating_information(df: pl.DataFrame) -> pl.DataFrame:
+    """
+    Add `fuel_type` and `community_heating` columns to EPC dataset using data in `MAIN_FUEL` and `MAINHEAT_DESCRIPTION`
```
helloaidank, line 13: It's not really necessarily clear from the outset what the difference between the 'MAIN_FUEL' and 'fuel_type' is, pointing to the definition or short explanation of what it is could be useful.

asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    Filter out EPC rows with no building height data.
+
+    Args:
+        df (pl.DataFrame): building height data joined to EPC data
+
+    Returns:
+        pl.DataFrame: EPC records with building height data
+    """
+    n = (df["height"] == -1).sum()
```
helloaidank: Are there edge cases where df["height"] has missing values but they aren't -1? Potentially something to consider

asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
helloaidank: These are just thoughts, it might be that you've decided you aren't counting them and that's cool too. I think it's likely to be a very small %.

asf_heat_pump_suitability/analysis/flats_on_fossils/features/fuel_type.py
```diff
@@ -0,0 +1,194 @@
...
+    Args:
+        df (pl.DataFrame): dataset with `HOTWATER_DESCRIPTION` column.
+        name (str): name of new column
+
+    Returns:
+        pl.DataFrame: dataframe with fuel type of hot water feature
+    """
+    df = df.with_columns(
+        pl.when(
+            pl.col("HOTWATER_DESCRIPTION").str.to_lowercase().str.contains("recovery")
```
helloaidank: any idea what recovery means in this context?

asf_heat_pump_suitability/analysis/flats_on_fossils/features/fuel_type.py
```diff
@@ -0,0 +1,194 @@
...
+        .then(pl.lit("biomass"))
+        .when(pl.col(epc_col).str.to_lowercase().str.contains("waste"))
+        .then(pl.lit("waste"))
+        .when(pl.col(epc_col).str.to_lowercase().str.contains("wood"))
+        .then(pl.lit("wood"))
+        .when(pl.col(epc_col).str.to_lowercase().str.contains("bioethanol"))
+        .then(pl.lit("bioethanol"))
+        # Oil comes last because it's in the word 'boiler'
+        .when(pl.col(epc_col).str.to_lowercase().str.contains("oil"))
+        .then(pl.lit("oil"))
```
helloaidank, line 141: could there be an edge case here where we have "boiler, gas" for the 'MAINHEAT_DESCRIPTION' and then this gets incorrectly classed as oil?

asf_heat_pump_suitability/analysis/flats_on_fossils/features/fuel_type.py
helloaidank: Something like recovering waste heat based on what you replace it with

asf_heat_pump_suitability/analysis/flats_on_fossils/features/fuel_type.py
```diff
@@ -0,0 +1,194 @@
...
+    )
+
+    # Fill main fuel type and community heating columns
+    df = df.with_columns(
+        pl.col("fuel_type")
+        .fill_null(pl.col("fill_fuel_type"))
+        .fill_null(pl.col("fill_fuel_type_2"))
+        .fill_null(pl.col("fill_fuel_type_3")),
+        pl.col("community_heating").fill_null(pl.col("fill_community_heating")),
+    ).drop(["fill_fuel_type", "fill_community_heating"])
```
helloaidank:
```suggestion
    ).drop(["fill_fuel_type", "fill_fuel_type_2", "fill_fuel_type_3", "fill_community_heating"])
```

asf_heat_pump_suitability/analysis/flats_on_fossils/features/fuel_type.py
```diff
@@ -0,0 +1,194 @@
...
+        .when(pl.col(epc_col).str.to_lowercase().str.contains("b30"))
+        .then(pl.lit("B30"))
+        .when(pl.col(epc_col).str.to_lowercase().str.contains("coal"))
+        .then(pl.lit("coal"))
+        .when(pl.col(epc_col).str.to_lowercase().str.contains("anthracite"))
+        .then(pl.lit("coal"))
+        # RENEWABLES
+        .when(pl.col(epc_col).str.to_lowercase().str.contains("electric"))
+        .then(pl.lit("electricity"))
+        .when(pl.col(epc_col).str.to_lowercase().str.contains("heat pump"))
```
helloaidank, line 121: there will be a small % of hybrid heat pumps too, just as long as we are aware with this assumption that all heat pumps are directly powered by just electricity

asf_heat_pump_suitability/analysis/flats_on_fossils/features/fuel_type.py
```diff
@@ -0,0 +1,194 @@
...
+        .when(pl.col("fuel_type").is_not_null())
+        .then(False)
+        .otherwise(None)
+        .alias("fossil_fuel_heating")
+    )
+
+    return df
+
+
+def extend_df_fuel_type(df: pl.DataFrame, epc_col: str, name: str) -> pl.DataFrame:
```
helloaidank, line 90: Could we adopt a dictionary/pattern-based classification for `extend_df_fuel_type`.

Something like:

```
FUEL_PATTERNS = [
    # FOSSIL FUELS
    (r"\bmains gas\b", "gas"),
    (r"\blng\b", "LNG"),
    (r"\blpg\b", "LPG"),
    (r"\bb30\b", "B30"),
    (r"\banthracite\b", "coal"),
    (r"\bcoal\b", "coal"),
    (r"\boil\b", "oil"),

    # ELECTRIC (non-fossil fuels)
    (r"\belectric\b", "electricity"),
    (r"heat pump", "electricity"),

    # RENEWABLE BIOMASS / WASTE
    (r"\bbiofuel\b", "biofuel"),
    (r"\bbiogas\b", "biogas"),
    (r"\bbiomass\b", "biomass"),
    (r"\bwaste\b", "waste"),
    (r"\bwood\b", "wood"),
    (r"\bbioethanol\b", "bioethanol"),

    # NO HEATING
    (
        r"there is no heating/hot-water system",
        "no heating"
    ),
]


def classify_fuel(text: Optional[str], patterns=FUEL_PATTERNS) -> Optional[str]:
    """
    Takes a string describing the heating system/fuel.
    Returns a standardised label by matching against FUEL_PATTERNS in order.
    """
    text_l = text.lower()
    for pattern, label in patterns:
        # If the pattern is found in `text_l`, return that label
        if re.search(pattern, text_l):
            return label

    return None  # If no pattern matched

def extend_df_fuel_type_dict(
    df: pl.DataFrame, epc_col: str, name: str
) -> pl.DataFrame:
    """
    Classify the fuel type from a string column using a pattern dictionary.
    """
    return df.with_columns(
        pl.col(epc_col)
        .apply(lambda x: classify_fuel(x))  # x is a single string from epc_col
        .alias(name)
    )
```

asf_heat_pump_suitability/analysis/flats_on_fossils/run_join_epc_building_footprints.py
```diff
@@ -0,0 +1,124 @@
+import polars as pl
```
helloaidank, line 11: Short description at top of document would be great to clarify what we are running here 👯‍♂️

asf_heat_pump_suitability/analysis/flats_on_fossils/run_join_epc_building_footprints.py
```diff
@@ -0,0 +1,124 @@
...
+        help="Path to save output file with building footprint information per EPC record to. If unspecified, save with default filename.",
+        type=str,
+        required=False,
+        default=None,
+    )
+
+    return parser.parse_args()
+
+
+if __name__ == "__main__":
```
helloaidank, line 67: Potential a refactor of this, it's quite big in terms of the number of operations and not necessarily the most readable.

Something like: 
```
def load_epc_data(epc_path: str) -> pl.DataFrame:
    logging.info("Loading EPC UPRNs")
    return pl.read_parquet(epc_path, columns=["UPRN"])

def enrich_with_coordinates(epc_df: pl.DataFrame) -> pl.DataFrame:
    logging.info("Adding lat/lon data to EPC")
    uprn_coords_df = lat_lon.transform_df_osopen_uprn_latlon()
    return epc_df.join(uprn_coords_df, how="left", on="UPRN")

def generate_epc_geometries(epc_df: pl.DataFrame) -> any:
    return lat_lon.generate_gdf_uprn_coords(epc_df, usecols=["UPRN"])[["UPRN", "geometry"]]

def load_uk_uprns_geometry() -> any:
    logging.info("Loading UK UPRNs")
    uk_uprns = get_datasets.get_df_osopen_uprn_latlon()
    return lat_lon.generate_gdf_uprn_coords(uk_uprns, usecols=["UPRN", "X_COORDINATE", "Y_COORDINATE"])[["UPRN", "geometry"]]

def get_uprns_per_file(epc_gdf) -> dict:
    logging.info("Getting list of UPRNs for each building footprint file")
    bounds = building_footprint.transform_df_uk_dataset_links()
    join = bounds.sjoin(epc_gdf, how="inner", predicate="intersects")
    return join.groupby("ms_url")["UPRN"].agg(list).to_dict()

def process_building_file(building_file: str, uprns: list, epc_gdf, uk_uprns_gdf) -> pl.DataFrame:
    try:
        footprints = building_footprint.transform_gdf_building_footprints(building_file)
    except shapely.errors.GEOSException as e:
        logging.warning(f"Error loading building footprint file {building_file}: {e}. Skipping.")
        return None

    uprn_gdf = epc_gdf.loc[epc_gdf["UPRN"].isin(uprns)]
    uprns_per_building_df = (
        footprints[["building_id", "geometry"]]
        .sjoin(uk_uprns_gdf, how="left", predicate="contains")
        .drop(columns=["index_right"])
        .groupby("building_id")
        .agg({"UPRN": "count"})
        .rename(columns={"UPRN": "UPRN_count_per_building"})
    )

    joined_df = (
        footprints.sjoin(uprn_gdf, how="inner", predicate="contains")
        .drop(columns=["index_right", "geometry"])
        .join(uprns_per_building_df, how="left", on="building_id")
    )

    return pl.from_pandas(joined_df)
```

asf_heat_pump_suitability/analysis/flats_on_fossils/run_process_epc_flats.py
```diff
@@ -0,0 +1,130 @@
+import polars as pl
```
helloaidank, line 13: short description of what the code does here would be great!

Something like:
"Script to process EPC data for flats and maisonettes, enriching it with building footprint data and derived heating 
system features (e.g., fuel type, property density)."

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
```diff
@@ -0,0 +1,629 @@
...
+#
+# FLAT_STOREY_COUNT data is available from EPC data and building height estimates are available from Microsoft ML Global Building Footprints dataset.
+#
+# FLAT_STOREY_COUNT has low coverage of non-null values in EPC and Microsoft ML Global Building Footprints has decent coverage. Here we test 4 different basic methodologies for converting building height data to FLAT STOREY COUNT data. We compare the methodologies at the end to select the best one.
+
+# %%
+import polars as pl
+import numpy as np
+import matplotlib.pyplot as plt
+import sklearn
```
helloaidank, line 18: The LinearRegression didn't work by accessing this object, I had to import like the following:

```suggestion
import sklearn
from sklearn.linear_model import LinearRegression
```

asf_heat_pump_suitability/analysis/flats_on_fossils/run_process_epc_flats.py
```diff
@@ -0,0 +1,130 @@
...
+            "property_per_m2"
+        ),
+        (pl.col("building_area_m2") / pl.col("UPRN_count_per_building")).alias(
+            "avg_property_footprint_m2"
+        ),
+    )
+
+    logging.info("Adding fuel type information")
+    epc_df = fuel_type.extend_df_fuel_type(epc_df)
+
```
helloaidank, line 140: missing two positional arguments when I try to run this `extend_df_fuel_type() missing 2 required positional arguments: 'epc_col' and 'name'`

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
```diff
@@ -0,0 +1,401 @@
...
+        "FLAT_STOREY_COUNT",
+        "height",
+    ]
+).null_count() / len(flats_epc_df)
+
+# %% [markdown]
+# ## 2. View distributions of building height and flat storey count data
+
+# %%
+flats_epc_df["height"].describe()
```
helloaidank: This shows minimum as a zero height? Should we treat values as zero the same as nulls.

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
```diff
@@ -0,0 +1,401 @@
...
+
+# %%
+# Remove building heights of less than 3m
```
helloaidank: What's the rationale behind removing building heights of less than 3m?

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
```diff
@@ -0,0 +1,401 @@
...
+len(flats_epc_df.filter(pl.col("height").is_not_null())) / len(flats_epc_df)
+
+# %%
+# Coverage of flats with UPRN counts per building data
+len(flats_epc_df.filter(pl.col("UPRN_count_per_building").is_not_null())) / len(
+    flats_epc_df
+)
+
+# %%
+# Coverage of flats with height estimate and flat storey count data
```
helloaidank: should it be "or" instead of "and" in this comment?

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
```diff
@@ -0,0 +1,401 @@
...
+    )
+) / len(flats_epc_df)
+
+# %% [markdown]
+# ## 3. Compare FLAT_STOREY_COUNT and building height estimate data
+#
+# Below we plot the number of storeys against the estimated building heights. The space between the lines on the plot indicate where most points should fall if they have a storey height between 2-4m per storey plus a roof of 1.5m. The plot shows the data are likely not accurate for many records with either flat storey count or height estimates being incorrect.
+#
+# In the second plot, we show meters per storey by building height. This plot indicates there are many buildings reported to be 1 storey regardless of building height (i.e. where meters per storey equals building height). Again, this suggests there are errors in the data.
+
```
helloaidank, line 222: Plotted something like the following to get a better feel for the data, did some hex binning:
```
"""Fits a Linear Model:

Estimates the average height per storey (gradient) using linear regression.
Validates the assumed roof_constant by checking the intercept.
Visualises the Data:

Creates a hexbin plot to show the density of buildings by storey count and height.
Adds reference lines (e.g., 2m/storey, 4m/storey) for comparison.
Overlays the best-fit line from the regression model.
Provides Insights:

Helps understand the relationship between storey count and building height.
Validates assumptions about average storey height and roof height."""

import matplotlib.pyplot as plt
import matplotlib
from sklearn.linear_model import LinearRegression
import numpy as np

# Assumed roof height
roof_constant = 1.5  # adjust as needed

# Prepare X and y for regression
X = np.array(storey_height_df["FLAT_STOREY_COUNT"]).reshape(-1, 1)
y = np.array(storey_height_df["height"] - roof_constant)

# Fit linear regression
model = LinearRegression()
model.fit(X, y)
gradient = model.coef_[0]
intercept = model.intercept_

# Print the estimated average storey height
print(f"Estimated average storey height (gradient): {gradient:.2f} m")
print(f"Intercept (should be near zero if roof_constant is accurate): {intercept:.2f} m")

# Generate best-fit line to overlay
storey_range = np.arange(1, 31)
best_fit_heights = gradient * storey_range + roof_constant

# Plot
plt.figure(figsize=(10, 6), dpi=120)
hb = plt.hexbin(
    x=storey_height_df["FLAT_STOREY_COUNT"],
    y=storey_height_df["height"],
    gridsize=50,
    cmap="Blues",
    mincnt=1,
    norm=matplotlib.colors.LogNorm(),
)
plt.colorbar(hb, label="Count of Buildings")

# Reference lines
plt.plot(storey_counts, min_building_heights, color="yellow", label="2m/storey")
plt.plot(storey_counts, max_building_heights, color="red", linestyle="dashed", label="4m/storey")

# Best-fit line
plt.plot(storey_range, best_fit_heights, color="black", linestyle="dashdot", linewidth=2, label=f"Best fit: {gradient:.2f}m/storey")

plt.xlim(1, 30)
plt.ylim(0, 40)
plt.grid(True, linestyle="--", alpha=0.5)
plt.xlabel("Number of storeys")
plt.ylabel("Building height (m)")
plt.title(f"Number of storeys by building height (m), N={len(storey_height_df)}")
plt.legend()
plt.show()
```

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
```diff
@@ -0,0 +1,401 @@
...
+plt.ylabel("Building height (m)")
+plt.legend()
+plt.show()
+
+# %%
+# Plot meters per storey by building height
+plt.scatter(
+    storey_height_df["height"], storey_height_df["meters_per_storey"], alpha=0.01
+)
+plt.title(f"Meters per storey by building height (m), N={len(storey_height_df)}")
```
helloaidank, line 374: A binned density plot is quite nice to view where the different peaks are within the overall distribution for meters per storey.
There seems to peaks around ~2.3 meters/storey and ~5m/storey which is interesting in of itself.
Code for those numbers and the distribution:
```
import matplotlib.pyplot as plt
import numpy as np
from scipy.signal import find_peaks

# Get the meters per storey data
meters_per_storey = storey_height_df["meters_per_storey"].to_numpy()

# Compute histogram values
hist_vals, bin_edges = np.histogram(meters_per_storey, bins=100, density=True)

# Find peaks
peaks, properties = find_peaks(hist_vals, height=0.01)  # tweak height threshold if needed

# Convert bin indices to x-axis positions
peak_positions = bin_edges[peaks]

# Plot histogram
plt.figure(figsize=(8, 6))
plt.hist(meters_per_storey, bins=100, density=True, color="steelblue", edgecolor="white")

# Plot mean and median
mean_mps = meters_per_storey.mean()
median_mps = np.median(meters_per_storey)
plt.axvline(mean_mps, color='red', linestyle='--', linewidth=1.5, label=f"Mean = {mean_mps:.2f} m")
plt.axvline(median_mps, color='orange', linestyle=':', linewidth=1.5, label=f"Median = {median_mps:.2f} m")

# Annotate detected peaks
for pos in peak_positions:
    plt.axvline(pos, color='green', linestyle='--', linewidth=1, label=f"Peak at {pos:.2f} m")

# Remove duplicate peak labels in legend
handles, labels = plt.gca().get_legend_handles_labels()
seen = set()
new_handles_labels = [(h, l) for h, l in zip(handles, labels) if not (l in seen or seen.add(l))]
plt.legend(*zip(*new_handles_labels))

# Final plot formatting
plt.xlabel("Meters per storey")
plt.ylabel("Density")
plt.title("Distribution of estimated floor height per storey with peaks")
plt.grid(True, linestyle="--", alpha=0.4)
plt.tight_layout()
plt.show()
```

helloaidank: image

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
helloaidank: image

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
```diff
@@ -0,0 +1,401 @@
...
+        pl.col("height").is_not_null() | pl.col("FLAT_STOREY_COUNT").is_not_null()
+    )
+) / len(flats_epc_df)
+
+# %% [markdown]
+# ## 3. Compare FLAT_STOREY_COUNT and building height estimate data
+#
+# Below we plot the number of storeys against the estimated building heights. The space between the lines on the plot indicate where most points should fall if they have a storey height between 2-4m per storey plus a roof of 1.5m. The plot shows the data are likely not accurate for many records with either flat storey count or height estimates being incorrect.
+#
+# In the second plot, we show meters per storey by building height. This plot indicates there are many buildings reported to be 1 storey regardless of building height (i.e. where meters per storey equals building height). Again, this suggests there are errors in the data.
```
helloaidank, line 221: Is the suggestion that we flag these 1 storey regardless of building height as poor quality data? It suggests that the storey count data is no good for these ones right?

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
```diff
@@ -0,0 +1,401 @@
...
+
+plt.title(f"Number of storeys by building height (m), N={len(storey_height_df)}")
+plt.xlabel("Number of storeys")
+plt.ylabel("Building height (m)")
+plt.legend()
+plt.show()
+
+# %%
+# Plot meters per storey by building height
+plt.scatter(
```
helloaidank, line 369: Instead of a scatter plot, using something like hexbin helps visualise the density a bit better.
```
plt.hexbin(
    storey_height_df["height"],
    storey_height_df["meters_per_storey"],
    gridsize=100,
    cmap="viridis",
    bins='log'
)
plt.colorbar(label='log10(N)')
plt.title(f"Meters per storey by building height (m), N={len(storey_height_df):,}")
plt.xlabel("Building height (m)")
plt.ylabel("Meters per storey")
plt.show()
```

helloaidank: image

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
```diff
@@ -0,0 +1,401 @@
...
+        "meters_per_storey",
+        "manual_check_max_above_ground_storey_count",
+        "basement_visible",
+        "mixed_height",
+    ]
+)
+
+# %%
+# Accepted meters per storey - this is the min and max meters per storey we accept as valid
+min_mps = 2.3
```
helloaidank, line 628: where have we got these values from?

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
```diff
@@ -0,0 +1,629 @@
...
+# %%
+flats_epc_df = flats_epc_df.with_columns(
+    pl.col("FLAT_STOREY_COUNT").cast(pl.Float64)
+).with_columns(
+    pl.when(pl.col("height") >= 3)
+    .then(pl.col("height"))
+    .otherwise(None)
+    .alias("height"),
+    # Manual check shows the building with max storey count is incorrect
+    pl.when(pl.col("UPRN") == "100080510780.0")
```
helloaidank: manually checked this and changed the storey count for this UPRN?

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
```diff
@@ -0,0 +1,629 @@
...
+    .then(3)
+    .otherwise(pl.col("FLAT_STOREY_COUNT"))
+    .alias("FLAT_STOREY_COUNT"),
+    raw_height=pl.col("height"),
+)
+
+# %% [markdown]
+# ## 2. Add tall buildings in UK
+# It appears that high rises likely don't have the correct height / storey count data (see `analysis_explore_storey_and_height_data.py` notebook) and the models are therefore underestimating their storey counts because the heights and storey counts are too low. We will try filling some of this data in using data about the tallest buildings in the UK taken from Wikipedia (URL: https://en.wikipedia.org/wiki/List_of_tallest_buildings_in_the_United_Kingdom) and tall buildings in Manchester (URL: https://en.wikipedia.org/wiki/List_of_tallest_buildings_and_structures_in_Greater_Manchester). The dataset covers all current buildings which are over 100m tall in the UK and over 50m in Manchester. The data contains building max height in meters and storey count.
```
helloaidank: the models are underestimating their storey counts because the heights and storey counts are too low? Not sure I understand that comment

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
```diff
@@ -35,19 +36,21 @@
...
-    # Manual check shows the building with max storey count is incorrect
-    pl.when(pl.col("UPRN") == "100080510780.0")
-    .then(3)
-    .otherwise(pl.col("FLAT_STOREY_COUNT"))
-    .alias("FLAT_STOREY_COUNT"),
-    raw_height=pl.col("height"),
+flats_epc_df = (
+    flats_epc_df.with_columns(pl.col("FLAT_STOREY_COUNT").cast(pl.Float64))
+    .with_columns(
+        pl.when(pl.col("height") >= 2)
```
helloaidank, line 51: Changed from 3 m to 2 m here, what was the rationale behind this?

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
```diff
@@ -328,10 +340,12 @@ def extend_df_building_rise_from_thresholds(
 #
 # First we prepare the dataset for modelling:
 # - exclude rows with null values in features or target values
-# - exclude rows where meters per storey is >=4m. These rows have erroneous height or storey count data
+# - exclude rows where country is Scotland (due to systemic error discovered in Scottish data)
```
helloaidank: might be good to include the Scotland exclusion comment at the beginning where we filter our Scotland from the outset in `flats_epc_df`

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
```diff
@@ -328,10 +340,12 @@ def extend_df_building_rise_from_thresholds(
...
 # - exclude rows where building height is <3m
```
helloaidank: changed this to 2 m now

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
```diff
@@ -328,10 +340,12 @@ def extend_df_building_rise_from_thresholds(
...
 #
 # We don't set a minimum threshold for meters per storey because our hypothesis is that the building height data underestimates the height of taller buildings. Therefore, we are allowing meters per storey values which are smaller than is realistic to train the model to predict higher storey counts for taller buildings.
+#
+# We group buildings by building ID to ensure that the same buildings are not represented in both the training and test sets.
```
helloaidank, line 358: Both the meters per storey and the overlapping buildings would constitute label leakage! I.e. our training and test dataset would not be representative of the real dataset when we actually apply the model.

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
```diff
@@ -606,11 +567,287 @@ def extend_df_building_rise_from_thresholds(
 plt.show()
```
helloaidank, line 578: I would definitely encourage us to include the model residuals for each of the different methods (or just even LR and polynomial regression) here if possible!

Something like

```
import pandas as pd
import numpy as np
import matplotlib.pyplot as plt
pred = reg.predict(X_test)
resid = pred - y_test
# Compute residuals
resid_df = pd.DataFrame({
    "height": X_test["height"],
    "resid": resid,
})

# Bin height into 2m intervals
resid_df["height_bin"] = pd.cut(resid_df["height"], bins=np.arange(0, 60, 2))

# Group by bin, compute mean residual
grouped = (
    resid_df.groupby("height_bin")["resid"]
    .agg(["mean", "count"])
    .reset_index()
)

# Extract bin centres for plotting
grouped["bin_center"] = grouped["height_bin"].apply(lambda x: x.mid)

# Plot
plt.figure(figsize=(7, 5))
plt.plot(grouped["bin_center"], grouped["mean"], marker="o", linestyle="-")
plt.axhline(0, color='k', linestyle='--')
plt.xlabel("Height (binned)")
plt.ylabel("Mean residual (Predicted − True)")
plt.title("Mean residual by height bin")
plt.grid(True, linestyle="--", alpha=0.5)
plt.show()
```

image

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
```diff
@@ -606,11 +567,287 @@ def extend_df_building_rise_from_thresholds(
...
+uprn_sample = model_df["UPRN"].to_list()
+X = model_df.select(["height", "property_per_m2"])
+y = model_df["FLAT_STOREY_COUNT"]
+
+# %%
+model = sm.ols("FLAT_STOREY_COUNT ~ height + property_per_m2", model_df.to_pandas())
+results = model.fit()
+
+# %%
+results.summary()
```
helloaidank, line 625: For explainability, potentially changing properties per m^{2} to properties per 100 m^{2} might make things slightly more intuitive or even splitting out the UPRNs and the building footprint as two separate variables. Happy to go with what we have for the moment anyway! Just a thought.

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
```diff
@@ -606,11 +567,287 @@ def extend_df_building_rise_from_thresholds(
...
+
+# %%
+results.params
+
+# %%
+results.conf_int()
```
helloaidank, line 631: Result is interesting, obviously y-intercept doesn't mean a lot since it's unphysical but cool that as we increase the height by one metre, that indicates ~0.5 storey (so like increasing height by two metres, indicates one storey...which makes sense). The property_per_m2 gradient is harder to interpret just because the variable is so smol.

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
```diff
@@ -606,11 +567,287 @@ def extend_df_building_rise_from_thresholds(
...
+
+# %% [markdown]
+# #### Cross-validation with group ID
+#
+# We use building ID as group ID to ensure different flats in the same building are split into the same fold.
 #
-# Below we do some final exploration of the results of method 3.
+# Below, we split the model into 3 random folds 50 times and then iterate through training/testing the model on the folds. We take the average r2 values on the test sets.
+#
+# We can see that the R2 is variable. Deeper investigation (not shown in this notebook) has revealed that this is likely due to the presence of single outlier buildings in the test set that have a lot of flats in them and thus significantly affect the R2 value. E.g. we found a test set with an R2 of -0.10. After removal of a significant outlier building which contained 115 flats in the sample, the R2 increased to ~0.5.
```
helloaidank, line 643: very interesting

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
```diff
@@ -606,11 +567,287 @@ def extend_df_building_rise_from_thresholds(
...
+
+# %%
+reg.score(X, y)
+
+# %%
+# Get full data predictions of final candidate model
+candidate_preds = reg.predict(X)
+residuals = y - candidate_preds
+
+plt.hist(residuals, bins=100)
```
helloaidank, line 728: Add range to make plot clearer
```suggestion
plt.hist(residuals, bins=100)
plt.xlim(-15, 15)
```

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
```diff
@@ -606,11 +567,287 @@ def extend_df_building_rise_from_thresholds(
...
+plt.title(
+    "Distribution of residuals from selected model for predicting flat storey count"
+)
```
helloaidank, line 731:
```suggestion
)
plt.xlabel("Residuals (true - predicted storey count)")
plt.ylabel("Frequency")
plt.axvline(0, color="red", linestyle="--", label="Zero line")
```

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
```diff
@@ -606,11 +567,287 @@ def extend_df_building_rise_from_thresholds(
...
+
+results_df.to_pandas().set_index("storeys").plot(kind="bar", figsize=(10, 5))
+plt.xlabel("Storey count")
+plt.ylabel("Count of flats")
+plt.title(
+    "Count of flats per predicted storey count (after rounding) from candidate model, and models trained across 3 folds"
+)
+plt.xlim(0, 36)
+plt.show()
+
```
helloaidank, line 861: Interesting plot, we really over-predict on the first storey and then under-predict for storey count 2 and 3.

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
```diff
@@ -0,0 +1,924 @@
...
+# EHS reports 565,163 purpose-build high-rise flats in 2023 out of a total of 5,456,378 flats
+# (Total is determined by adding together 'converted flat', 'purpose built flat, low rise', and 'purpose built flat, high rise' values).
+# We can see our proportion compares quite well to the EHS proportion
+565163 / (949600 + 3941615 + 565163)
+
+# %%
+save_utils.save_to_s3(
+    flats_epc_df,
+    "s3://asf-heat-pump-suitability/outputs/2024Q3/analysis/2024_Q3_epc_flats_processed_filled_storey_count.parquet",
```
helloaidank: Potentially useful to add EW flag to the file name, just so it's clear we are only filling storey count for England and Wales.

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
```diff
@@ -0,0 +1,924 @@
...
+    )
+    .with_columns(
+        (pl.col("height") / pl.col("FLAT_STOREY_COUNT")).alias("meters_per_storey")
+    )
+    .filter(pl.col("height") >= 2)
+)
+
+# %%
+uprn_sample = model_df["UPRN"].to_list()
+X = model_df.select(["height", "property_per_m2"])
```
helloaidank, line 376: Should we not be grouping buildings by building ID here? Otherwise we are looking results where same building is in both training and test dataset right and is more liable to overfitting.

helloaidank (commented):
Hi Roisín!

Thanks for doing all this work, it's been a journey to say the least but there is light at the end of the tunnel! I've left comments throughout the scripts, happy to answer any questions you may have. Think as we've spoken about, it would be good to put together a slide deck on some of the decisions made whilst trying to keep it as concise as possible to present back to K&C. 

Nice job on pulling in height data from Microsoft ML dataset and then going through the whole process of iterating through the methods. I think we got to a good place with this, and where we're comfortable with the decisions taken! 

The evaluation done at the end is nice and robust! Really interesting finding that an outlier could pull so hard on the R^{2} but seems like otherwise we're good. 

Well done!!

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
crispy-wonton: I removed them because I thought it's unrealistic for a building that someone lives in to be smaller than 3m. However, in the final training of the model i changed it to filter out buildings smaller than 2m

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
crispy-wonton: Since they are removed anyway because they are less than 3m I don't think it matters?

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
crispy-wonton: yes!

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
crispy-wonton: Because we decided to trust the flat storey count data , I didn't filter rows with storey counts of 1 out. I don't know if we should. I think these could be maisonettes? It does seem weird that there are quite a lot of them but I don't have enough expertise on this to decide whether it's actually erroneous! Ultimately, I've left them in. Do you disagree?

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
crispy-wonton: This is a cool plot!!! :o

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
crispy-wonton: great idea, like this!

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
crispy-wonton: This is really interesting!! I would say the higher MPS peaks come from houses converted into flats which will have taller roofs and the smaller MPS peaks come from purpose-built apartments? That would be my hypothesis

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
crispy-wonton: min comes from English building regulations being 2.3m high ceiling and max comes from assuming a 3-4m ceiling with 3-2m roof (obviously this would be for a 1 storey building).

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
crispy-wonton: oh! i will test to make sure it works for me and then update. thanks!

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
crispy-wonton: was meant to change the previous notebook to 2m as well! i forgot to update it

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
crispy-wonton: I meant that the high rises appear to have too low storey counts AND too low heights in some cases which means that the model will not learn they are high rises - it will just classify them as lower rises. Poorly explained but ultimately I mean that if we correct some of the high-rise storey counts, our model will be better calibrated to the actual ground truth and therefore have improved accuracy. Does that make sense?

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
crispy-wonton: good spot! i forgot this comment was here.

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
crispy-wonton: Yes this happens in the section with further evaluation as an extra check, although I am not too worried about overfitting because it's such a simple model.

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
crispy-wonton: yes love it, this is a great plot to include - thank you for the code!

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
crispy-wonton: yeh I think that makes sense XD

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
crispy-wonton: yes! good point!! very interpretable B)

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
crispy-wonton: can show you this analysis, it's in a slidedeck

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
crispy-wonton: Yeh, makes me wonder if perhaps we should remove the 1 storey flats as you suggested. But that seems to be a big chunk of data. Perhaps, like in Scotland, those are mislabelled and should actually be 2 storey buildings? Hard to say.

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
crispy-wonton: good idea!

asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
crispy-wonton: Good question. I dont think so but I can't remember. But i guess there were some zero heights so maybe I could remove those here as well tbh.

asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
crispy-wonton: Yes, the `storey_col` name is variable. We name it when we create the column of estimates from our model :)

asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
crispy-wonton: True!

asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
crispy-wonton: hehe `when then otherwise` is so satisfying!

asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
crispy-wonton: Umm, this would mean the whole building is underground..! I dont think we need to account for those because I'm not sure many if any exist and if they do, they won't have height data ! :D 

`FLAT_STOREY_COUNT` is meant to be the storey count of the whole building the flat is situated in, not the specific storey the flat is located on.

asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
crispy-wonton: I would need to check again, I can't remember. I made this function after manually looking at the data. I think if there are strings like that, they will become nulls when casting to float in the next line.

asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
crispy-wonton: Yeh i assumed 0 storeys was an error so I'm removing.

asf_heat_pump_suitability/analysis/flats_on_fossils/features/fuel_type.py
crispy-wonton: true, maybe I should just rename it to `processed_main_fuel` cuz that's what it is!

asf_heat_pump_suitability/analysis/flats_on_fossils/features/fuel_type.py
crispy-wonton: yes, I think we need to rename the 'renewables' category to something else. perhaps 'non-fossil fuel'? or 'electric' and 'other'. what do you think?

crispy-wonton: I see you already recategorised in your comment above!

asf_heat_pump_suitability/analysis/flats_on_fossils/features/fuel_type.py
crispy-wonton: I think this is a good idea. However it will need testing (unless you already did?) which I don't have enough time to do in this PR. It's definitely an improvement!

asf_heat_pump_suitability/analysis/flats_on_fossils/features/fuel_type.py
crispy-wonton: In theory yes, this could happen. In practice, I would need to look at how many rows would have that issue. 'mains gas' seems to be an extremely common label so I hope it won't affect too many rows but good spot! I will have a look.

asf_heat_pump_suitability/analysis/flats_on_fossils/features/fuel_type.py
crispy-wonton: Yes I saw it in the context of waste heat recovery.

asf_heat_pump_suitability/analysis/flats_on_fossils/run_join_epc_building_footprints.py
crispy-wonton: This might need to be in a separate PR!

asf_heat_pump_suitability/analysis/flats_on_fossils/run_process_epc_flats.py
crispy-wonton: oh dear! thank you for flagging- will fix!

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
helloaidank: True

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
helloaidank: I think it's fine, I don't have expertise in this area either! I am guessing it might just be things like maisonettes! I just had some doubts, as I always imagine flats to be at least two storeys (anecdotally)

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
helloaidank: I think this is the Scottish data for the second peak!

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
helloaidank: Maybe good to document these assumptions here!

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
helloaidank: Yep makes sense :)

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
helloaidank: Understood, but probably good to include at this stage as you're basing the decision on metrics using this training and test dataset. The conclusion is the same but still good to do it here for rigour.

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
helloaidank: I think it just makes me think that we should not over interpret the final results when we are looking at the granularity at a per storey level. I think this would be a useful plot to show and discuss with K&C.

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
crispy-wonton: yes!

asf_heat_pump_suitability/analysis/flats_on_fossils/run_join_epc_building_footprints.py
helloaidank: depending on whether this becomes important, I think best just to leave it :)

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
crispy-wonton: Yep, that's the plan!

asf_heat_pump_suitability/analysis/flats_on_fossils/features/building_rise.py
crispy-wonton: I've checked this and there are no nulls in this column, so I think it's only -1 and 0 values to be removed.

asf_heat_pump_suitability/analysis/flats_on_fossils/features/fuel_type.py
crispy-wonton: On second thoughts, this is gonna be a lot of refactoring so I'm goin to leave it for now and go with your suggestion of explaining it in the docstring!

asf_heat_pump_suitability/analysis/flats_on_fossils/features/fuel_type.py
crispy-wonton: I filtered the dataset for rows where `MAIN_FUEL` or `MAINHEAT_DESCRIPTION` contains the word 'boiler'. See below the unique categories for each column which are returned:

`MAIN_FUEL` categories containing 'boiler': `['Community heating schemes: heat from boilers - biomass', 'heat from boilers that can use mineral oil or biodiesel (community)', 'heat from boilers using biodiesel from any biomass source (community)']`

`MAINHEAT_DESCRIPTION` categories containing 'boiler': `['Solid-fuel boiler, solid fuel', 'Electric boiler, electric', 'Community scheme, radiators, heat from boilers GÇô mains gas', 'Boiler and radiators, heat from boilers - mains gas ', 'Room heaters, heat from boilers - mains gas', 'Radiator heating, heat from boilers - gas', 'Cynllun cymunedol, heat from boilers - mains gas', 'Boiler and radiators, heat from boilers - mains gas']`

As you can see, there are no `MAIN_FUEL` categories which describe a gas boiler. In `MAINHEAT_DESCRIPTION` there is a single category which describes a gas boiler without specifying 'mains gas': 'Radiator heating, heat from boilers - gas'. This represents 5 rows in the dataset. I checked the final `fuel_type` assignment of these rows and they are assigned `gas` because of the values in the `MAIN_FUEL` column.

Because it will require further testing to change this function and make sure it will capture groups correctly if we change it, I'm going to leave it as is now for our current use case and this is a one off analysis. However, I will add a note to the code that this can happen in case we reuse the code in future.

asf_heat_pump_suitability/analysis/flats_on_fossils/run_process_epc_flats.py
crispy-wonton: I'm unable to reproduce this error. The script runs successfully for me.

Please could you try re-running after I have pushed final changes before merge?

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_explore_storey_and_height_data.py
crispy-wonton: It is! after removing the Scottish data, there is one peak at 2.4m

asf_heat_pump_suitability/analysis/flats_on_fossils/analysis_fill_building_rise_methods.py
crispy-wonton: I just realised we can divide the coefficients by 100 and it gives the same effect I think.

helloaidank (dismissed):
Happy to merge!

helloaidank (approved):
Happy to merge these new commits too! I ran over everything and looked at the code, looks good from my perspective.

PR #151, opened 2025-04-10
148 data quality checks outputs

### LINKED ISSUE #148: Data quality checks on final heat pump suitability outputs
We want to do some basic analysis of the outputs (e.g. missingness, as well as things like Garden Size is within range), one is of the heat pump suitability per lsoa as well as heat pump suitability per property.

helloaidank:
---

# Description
This PR introduces a data quality checking script, pandera_eval_hps_outputs.py, which uses [Pandera](https://pandera.readthedocs.io/) (with Polars backend) for schema validation, outlier detection, and summary reporting for the final outputs of our pipeline. 

**Key changes**:
- Adds the data quality script for validating pipeline outputs.
- Adds supporting configuration file: `config/dq_config.py`.
- Updates requirements.txt to include pandera
- All data quality validation in the pipeline will now use pandera for consistency and maintainability.
- Summary statistics and validation failure details are written to output files

_Note: Earlier consideration of alternative frameworks (Vanilla pandas, Great Expectations) have been dropped in favour of pandera as its polars compatibility is dank. Really looking forward to implementing some pandera function decorators throughout the rest of the pipeline._

Fixes #148 

# Instructions for Reviewer

Run the code using:
- `python asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py`
- A log directory will be created, where'll you'll have a text file and a csv with the result from the data quality checks. Each one should have it's own time stamp! Review these please!

Please pay attention to the following:

- The logic behind the validation checks, is there anything that we're obviously missing that we should do?
- Do the results look sensible? Should we have diagnostic plots as easier to visualise then numbers (e.g. box plots based off dataframe outputted in csv)
- Should we move this to the pipeline directory since we are directly dealing with the final outputs of the pipeline and should be something that is automatically run
- Any input on the formatting of the logs, it's quite dry at the moment but also didn't want to overengineer for the first iteration

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
+++ b/asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
@@ -10,7 +10,6 @@ Defines:
 
 from asf_heat_pump_suitability import PROJECT_DIR
 import os
-import os
 from datetime import datetime
 from urllib.parse import urlparse
 
@@ -88,7 +87,8 @@ NUMERIC_COLUMNS = [
     "census_proportion_flats",
 ]
 
-# Proportion columns in [0, 1]
+
+# Proportion columns expected to be between [0, 1]
 PROPORTION_COLUMNS = [
     "proportion_in_conservation_area",
     "proportion_listed_building",
@@ -121,11 +121,4 @@ SCORE_COLUMNS = [
     "SGL_N_avg_score_weighted",
     "HN_S_avg_score_weighted",
     "HN_N_avg_score_weighted",
-]
-
-# Columns that must be >= 0
-NON_NEGATIVE_COLUMNS = [
-    "median_garden_estimate_m2",
-    "property_density_km2",
-    "n_properties",
 ]
--- a/asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
+++ b/asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
@@ -1,0 +1,131 @@
+"""
+Configuration file for data quality checks and analysis.
+
+Defines:
+- File paths, logging setup, and timestamped output directories.
+- Expected dataset structure, including column names and types.
+- Validation rules for numeric, proportion, boolean, and categorical columns.
+- Thresholds for outlier detection (z-scores) and non-negative constraints.
+"""
+
+from asf_heat_pump_suitability import PROJECT_DIR
+import os
+import os
+from datetime import datetime
+from urllib.parse import urlparse
+
+# Path to data file
+# timestamp for tracking
+TIMESTAMP = datetime.now().strftime("%Y%m%d_%H%M%S")
+
+# S3 URI for the data file
+DATA_S3_URI = "s3://asf-heat-pump-suitability/outputs/2023Q4/suitability/20250319_2023_Q4_heat_pump_suitability_per_lsoa.csv"
+
+# parse bucket and key from the URI (if you still need them elsewhere)
+parsed_uri = urlparse(DATA_S3_URI)
+S3_BUCKET = parsed_uri.netloc
+S3_KEY = parsed_uri.path.lstrip("/")
+
+# output directory for logs (local)
+OUTPUT_DIR = os.path.join(
+    PROJECT_DIR,
+    "asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/logs",
+    TIMESTAMP,
+)
+os.makedirs(OUTPUT_DIR, exist_ok=True)
+
+# path to use in the rest of the pipeline (read straight from S3)
+DATA_PATH = DATA_S3_URI
+
+# append timestamp to log filename
+DATA_FILE_NAME = os.path.basename(S3_KEY)
+LOG_FILENAME = f"{os.path.splitext(DATA_FILE_NAME)[0]}_{TIMESTAMP}_dq.log"
+
+# Expected columns
+EXPECTED_COLUMNS = [
+    "lsoa",
+    "ASHP_S_avg_score_weighted",
+    "ASHP_N_avg_score_weighted",
+    "GSHP_S_avg_score_weighted",
+    "GSHP_N_avg_score_weighted",
+    "SGL_S_avg_score_weighted",
+    "SGL_N_avg_score_weighted",
+    "HN_S_avg_score_weighted",
+    "HN_N_avg_score_weighted",
+    "scores_weighted",
+    "n_properties",
+    "property_density_km2",
+    "rural_urban_class",
+    "has_anchor_property",
+    "heatpump_installation_percentage",
+    "median_garden_estimate_m2",
+    "proportion_in_conservation_area",
+    "proportion_listed_building",
+    "proportion_epc_c_plus",
+    "proportion_off_gas",
+    "census_proportion_flats",
+    "lsoa_name",
+]
+
+# Numeric columns
+NUMERIC_COLUMNS = [
+    "ASHP_S_avg_score_weighted",
+    "ASHP_N_avg_score_weighted",
+    "GSHP_S_avg_score_weighted",
+    "GSHP_N_avg_score_weighted",
+    "SGL_S_avg_score_weighted",
+    "SGL_N_avg_score_weighted",
+    "HN_S_avg_score_weighted",
+    "HN_N_avg_score_weighted",
+    "n_properties",
+    "property_density_km2",
+    "heatpump_installation_percentage",
+    "median_garden_estimate_m2",
+    "proportion_in_conservation_area",
+    "proportion_listed_building",
+    "proportion_epc_c_plus",
+    "proportion_off_gas",
+    "census_proportion_flats",
+]
+
+# Proportion columns in [0, 1]
+PROPORTION_COLUMNS = [
+    "proportion_in_conservation_area",
+    "proportion_listed_building",
+    "proportion_epc_c_plus",
+    "proportion_off_gas",
+    "census_proportion_flats",
+]
+
+# Boolean columns
+BOOLEAN_COLUMNS = [
+    "scores_weighted",
+    "has_anchor_property",
+]
+
+# Categorical columns and allowed values
+CATEGORICAL_COLUMNS = {
+    "rural_urban_class": ["Rural", "Urban"],
+}
+
+# Outlier detection threshold (z-score)
+OUTLIER_ZSCORE_THRESHOLD = 3.0
+
+# Score columns expected to be in [0,1]
+SCORE_COLUMNS = [
+    "ASHP_S_avg_score_weighted",
+    "ASHP_N_avg_score_weighted",
+    "GSHP_S_avg_score_weighted",
+    "GSHP_N_avg_score_weighted",
+    "SGL_S_avg_score_weighted",
+    "SGL_N_avg_score_weighted",
+    "HN_S_avg_score_weighted",
+    "HN_N_avg_score_weighted",
+]
+
+# Columns that must be >= 0
+NON_NEGATIVE_COLUMNS = [
+    "median_garden_estimate_m2",
+    "property_density_km2",
+    "n_properties",
+]
--- a/asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
+++ b/asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
@@ -15,10 +15,26 @@ import polars as pl
 import pandera.polars as pa
 import json
 import config.dq_config as cfg
+import argparse
 from pandera.polars import Check, Column, DataFrameSchema
 from asf_heat_pump_suitability.analysis.hn_zones.hnz_utils.log_utils import (
     setup_logging_and_file_path,
 )
+
+# ------------------------------------------------------------------------------
+# Argument parsing for data path
+# ------------------------------------------------------------------------------
+parser = argparse.ArgumentParser(
+    description="data quality checks for heat pump suitability data."
+)
+parser.add_argument(
+    "--data-path",
+    type=str,
+    default=cfg.DATA_PATH,
+    help="Path to the data CSV file (local or S3 URI).",
+)
+args = parser.parse_args()
+data_path = args.data_path
 
 # ------------------------------------------------------------------------------
 # Logging setup
@@ -31,14 +47,17 @@ setup_logging_and_file_path(output_dir=cfg.OUTPUT_DIR, log_filename=cfg.LOG_FILE
 # 1.  BUILD SCHEMA
 # ---------------------------------------------------------------------------
 
-schema_columns: dict[str, Column] = {}
+# schema_columns: dict[str, Column] = {}
 
 
-def add(col: str, dtype, *checks, **kw):
+def add_schema_column(
+    schema_columns: dict[str, Column], col: str, dtype: any, *checks, **kw
+):
     """
     Registers or updates a column in the `schema_columns` dictionary for schema validation.
 
     Args:
+        schema_columns (dict): The dictionary holding schema columns.
         col (str): The name of the column to register or update.
         dtype: The data type of the column (e.g., `float`, `str`, `pl.Boolean`).
         *checks: Any number of Pandera `Check` objects to validate column values.
@@ -51,27 +70,35 @@ def add(col: str, dtype, *checks, **kw):
     schema_columns[col] = Column(dtype, checks=list(checks), **kw)
 
 
+schema_col = {}
 # -- primary key ------------------------------------------------------------ #
-add("lsoa", str, unique=True, required=True, nullable=False)
+add_schema_column(
+    schema_col, col="lsoa", dtype=str, unique=True, required=True, nullable=False
+)
 
 # -- score + proportion columns in [0, 1] ----------------------------------- #
 for c in (*cfg.SCORE_COLUMNS, *cfg.PROPORTION_COLUMNS):
-    add(c, float, Check.in_range(0, 1), required=True, nullable=True)
+    add_schema_column(
+        schema_col, c, float, Check.in_range(0, 1), required=True, nullable=True
+    )
 
 # -- boolean ---------------------------------------------------------------- #
 for c in cfg.BOOLEAN_COLUMNS:
-    add(c, pl.Boolean, required=True, nullable=True)
+    add_schema_column(schema_col, col=c, dtype=pl.Boolean, required=True, nullable=True)
 
 # -- categorical ------------------------------------------------------------ #
 for c, allowed in cfg.CATEGORICAL_COLUMNS.items():
-    add(c, str, Check.isin(allowed), required=True, nullable=True)
+    add_schema_column(
+        schema_col, c, str, Check.isin(allowed), required=True, nullable=True
+    )
 
 # -- non-negative ----------------------------------------------------------- #
-for c in cfg.NON_NEGATIVE_COLUMNS:
-    add(c, float, Check.ge(0), required=True, nullable=True)
+for c in cfg.NUMERIC_COLUMNS:
+    add_schema_column(schema_col, c, float, Check.ge(0), required=True, nullable=True)
 
 # -- 0–100 percentage ------------------------------------------------------- #
-add(
+add_schema_column(
+    schema_col,
     "heatpump_installation_percentage",
     float,
     Check.in_range(0, 100),
@@ -80,44 +107,51 @@ add(
 )
 
 # -- lsoa name -------------------------------------------------------------- #
-add("lsoa_name", str, required=True, nullable=True)
+add_schema_column(schema_col, "lsoa_name", str, required=True, nullable=True)
 
 # finally create the schema
-schema = DataFrameSchema(schema_columns, strict=False, coerce=True)
+schema = DataFrameSchema(schema_col, strict=False, coerce=True)
 
 
 # ------------------------------------------------------------------------------
 # 2. LOAD WITH POLARS
 # ------------------------------------------------------------------------------
 try:
-    df_pol = pl.read_csv(cfg.DATA_PATH)
+    df_pol = pl.read_csv(data_path)
     logging.info("Data loaded successfully with Polars!")
 except FileNotFoundError:
-    logging.error(f"File not found at path: {cfg.DATA_PATH}")
+    logging.error(f"File not found at path: {data_path}")
     raise
 except Exception:
-    logging.exception(f"Could not load data from {cfg.DATA_PATH}")
+    logging.exception(f"Could not load data from {data_path}")
     raise
+
+# log any columns in the data that aren’t covered by the schema
+extra_cols = set(df_pol.columns) - set(schema.columns.keys())
+if extra_cols:
+    logging.info(
+        "Columns skipped by schema validation (strict=False): %s",
+        sorted(extra_cols),
+    )
 
 # ------------------------------------------------------------------------------
 # 3. VALIDATE WITH PANDERA (Polars backend)
 # ------------------------------------------------------------------------------
 logging.info("=== Validating with Pandera Schema ===")
 try:
-    df_valid = schema.validate(df_pol, lazy=True)
-    logging.info("Data validated successfully with Pandera!")
+    df_pol = schema.validate(df_pol, lazy=True)
+    logging.info("Data schema and ranges validated successfully with Pandera!")
 except pa.errors.SchemaErrors as err:
     logging.warning("Pandera validation failed!")
     logging.warning(json.dumps(err.message, indent=2))
     logging.info("Continuing with unvalidated data.")
-    df_valid = df_pol.clone()
 
 
 # ------------------------------------------------------------------------------
 # 4. OUTLIER DETECTION (Z-Score) IN POLARS
 # ------------------------------------------------------------------------------
 # compute z-score columns for numeric columns
-df_valid = df_valid.with_columns(
+df_pol = df_pol.with_columns(
     [
         ((pl.col(c) - pl.col(c).mean()) / pl.col(c).std()).alias(f"{c}_z")
         for c in cfg.NUMERIC_COLUMNS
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -23,3 +23,5 @@ tenacity
 numpy<2.0.0
 fiona
 bs4
+great_expectations
```
crispy-wonton: we don't need `great_expectations` now right?

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
```diff
@@ -0,0 +1,131 @@
...
+Defines:
+- File paths, logging setup, and timestamped output directories.
+- Expected dataset structure, including column names and types.
+- Validation rules for numeric, proportion, boolean, and categorical columns.
+- Thresholds for outlier detection (z-scores) and non-negative constraints.
+"""
+
+from asf_heat_pump_suitability import PROJECT_DIR
+import os
+import os
```
crispy-wonton: duplicate import

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
```diff
@@ -0,0 +1,131 @@
...
+from datetime import datetime
+from urllib.parse import urlparse
+
+# Path to data file
+# timestamp for tracking
+TIMESTAMP = datetime.now().strftime("%Y%m%d_%H%M%S")
+
+# S3 URI for the data file
+DATA_S3_URI = "s3://asf-heat-pump-suitability/outputs/2023Q4/suitability/20250319_2023_Q4_heat_pump_suitability_per_lsoa.csv"
```
crispy-wonton, line 21: we probably want to input this as a parameter when we run the code, e.g. with `argparse`

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
```diff
@@ -0,0 +1,131 @@
...
+
+# parse bucket and key from the URI (if you still need them elsewhere)
+parsed_uri = urlparse(DATA_S3_URI)
```
crispy-wonton, line 24: cool! i like `urlparse`! dont think i've seen it before

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
```diff
@@ -0,0 +1,131 @@
...
+    "heatpump_installation_percentage",
+    "median_garden_estimate_m2",
+    "proportion_in_conservation_area",
+    "proportion_listed_building",
+    "proportion_epc_c_plus",
+    "proportion_off_gas",
+    "census_proportion_flats",
+]
+
+# Proportion columns in [0, 1]
```
crispy-wonton:
```suggestion
# Proportion columns expected to be between [0, 1]
```
Is this what the note is meant to say?

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
```diff
@@ -0,0 +1,131 @@
...
+    "HN_S_avg_score_weighted",
+    "HN_N_avg_score_weighted",
+]
+
+# Columns that must be >= 0
+NON_NEGATIVE_COLUMNS = [
+    "median_garden_estimate_m2",
+    "property_density_km2",
+    "n_properties",
+]
```
crispy-wonton: I think this should just be all the numeric cols. So could we just do any tests for negativity using `NUMERIC_COLUMNS`?

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -0,0 +1,174 @@
+"""
+This script performs data quality checks and statistical analysis on heat pump suitability data.
+
+Key functionalities:
+1. Validates the dataset schema using Pandera (Polars backend).
+2. Detects outliers in numeric columns based on z-scores.
+3. Generates summary statistics for numeric columns.
+4. Logs results and saves a detailed numeric summary to a CSV file.
```
crispy-wonton: Great summary here - succinct and descriptive!

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+# ---------------------------------------------------------------------------
+# 1.  BUILD SCHEMA
+# ---------------------------------------------------------------------------
+
+schema_columns: dict[str, Column] = {}
+
+
+def add(col: str, dtype, *checks, **kw):
```
crispy-wonton: I think this function needs a more descriptive name :D 
`dtype` is missing type hinting

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+    Registers or updates a column in the `schema_columns` dictionary for schema validation.
+
+    Args:
+        col (str): The name of the column to register or update.
+        dtype: The data type of the column (e.g., `float`, `str`, `pl.Boolean`).
+        *checks: Any number of Pandera `Check` objects to validate column values.
+        **kw: Additional keyword arguments for column properties (e.g., `required`, `nullable`, `unique`).
```
crispy-wonton:
```suggestion
        **kw: Additional Pandera keyword arguments for column properties (e.g., `required`, `nullable`, `unique`).
```
Is this referring to `pandera` kwargs?

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -0,0 +1,174 @@
...
+# ------------------------------------------------------------------------------
+
+setup_logging_and_file_path(output_dir=cfg.OUTPUT_DIR, log_filename=cfg.LOG_FILENAME)
+
+
+# ---------------------------------------------------------------------------
+# 1.  BUILD SCHEMA
+# ---------------------------------------------------------------------------
+
+schema_columns: dict[str, Column] = {}
```
crispy-wonton: I recommend building this into the function below so that we're not changing a global variable which is harder to keep track of and might give us unexpected behaviours

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+    if col in schema_columns:  # merge with existing
+        prev = schema_columns[col]
+        checks = tuple(prev.checks) + tuple(checks)  # keep earlier checks
+        kw = {**{k: getattr(prev, k) for k in ("required", "nullable", "unique")}, **kw}
+    schema_columns[col] = Column(dtype, checks=list(checks), **kw)
+
+
+# -- primary key ------------------------------------------------------------ #
+add("lsoa", str, unique=True, required=True, nullable=False)
```
crispy-wonton:
```suggestion
add(col="lsoa", dtype=str, unique=True, required=True, nullable=False)
```
I recommend adding key words to the args so it's more readable :) same below

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -0,0 +1,174 @@
...
+
+# -- score + proportion columns in [0, 1] ----------------------------------- #
+for c in (*cfg.SCORE_COLUMNS, *cfg.PROPORTION_COLUMNS):
+    add(c, float, Check.in_range(0, 1), required=True, nullable=True)
```
crispy-wonton: This makes me think we should update the final dataset to ensure we dont have any rows with null scores for LSOA level data. Not for you to do in this PR, just making a note!

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -0,0 +1,174 @@
...
+# -- boolean ---------------------------------------------------------------- #
+for c in cfg.BOOLEAN_COLUMNS:
+    add(c, pl.Boolean, required=True, nullable=True)
+
+# -- categorical ------------------------------------------------------------ #
+for c, allowed in cfg.CATEGORICAL_COLUMNS.items():
+    add(c, str, Check.isin(allowed), required=True, nullable=True)
+
+# -- non-negative ----------------------------------------------------------- #
+for c in cfg.NON_NEGATIVE_COLUMNS:
```
crispy-wonton:
```suggestion
for c in cfg.NUMERIC_COLUMNS:
```
I guess perhaps you separated out the other columns to make it slightly faster because we are already checking if the others are within a given range? If it adds negligible compute time i would just check all the cols here for now because it makes it easier to update the code.

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -0,0 +1,174 @@
...
+    Check.in_range(0, 100),
+    required=True,
+    nullable=True,
+)
+
+# -- lsoa name -------------------------------------------------------------- #
+add("lsoa_name", str, required=True, nullable=True)
+
+# finally create the schema
+schema = DataFrameSchema(schema_columns, strict=False, coerce=True)
```
crispy-wonton: Since we're setting `strict=False`, perhaps we should log the columns that are not in the schema and therefore not being quality checked.

If this will take more than 5 mins to add, I suggest adding as a separate issue and adding after this is merged.

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -0,0 +1,174 @@
...
+# ------------------------------------------------------------------------------
+try:
+    df_pol = pl.read_csv(cfg.DATA_PATH)
+    logging.info("Data loaded successfully with Polars!")
+except FileNotFoundError:
+    logging.error(f"File not found at path: {cfg.DATA_PATH}")
+    raise
+except Exception:
+    logging.exception(f"Could not load data from {cfg.DATA_PATH}")
+    raise
```
crispy-wonton: is this try except block for automation purposes? i.e. we could just see the logs after it independently ran? because otherwise i would leave it to error out itself without the exception

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -0,0 +1,174 @@
...
+
+# ------------------------------------------------------------------------------
+# 3. VALIDATE WITH PANDERA (Polars backend)
+# ------------------------------------------------------------------------------
+logging.info("=== Validating with Pandera Schema ===")
+try:
+    df_valid = schema.validate(df_pol, lazy=True)
+    logging.info("Data validated successfully with Pandera!")
```
crispy-wonton: could we add a little bit more description of what successful validation means. e.g. Data schema and ranges validated successfully. or something like that :)

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -0,0 +1,174 @@
...
+except Exception:
+    logging.exception(f"Could not load data from {cfg.DATA_PATH}")
+    raise
+
+# ------------------------------------------------------------------------------
+# 3. VALIDATE WITH PANDERA (Polars backend)
+# ------------------------------------------------------------------------------
+logging.info("=== Validating with Pandera Schema ===")
+try:
+    df_valid = schema.validate(df_pol, lazy=True)
```
crispy-wonton: looks like we don't use `df_pol` again later so can just save the validated schema as `df_pol`. this means it's not misleading if we go into the exception and call an unvalidated df `df_valid`

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -0,0 +1,174 @@
...
+)
+logging.info("=== Outlier Detection (Z-Score) ===")
+
+Z = cfg.OUTLIER_ZSCORE_THRESHOLD
+num_cols = cfg.NUMERIC_COLUMNS
+
+
+# outlier counts (1-row wide → long)
+outlier_counts = df_valid.select(
+    [(pl.col(f"{c}_z") > Z).sum().alias(c) for c in num_cols]
```
crispy-wonton:
```suggestion
    [(pl.col(f"{c}_z") > Z).sum().alias(c) for c in cfg.NUMERIC_COLUMNS]
```
I would use the config directly here for consistency

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -0,0 +1,174 @@
...
+).transpose(include_header=True, header_name="variable", column_names=["n_out"])
```
crispy-wonton:
```suggestion
).transpose(include_header=True, header_name="variable", column_names=["n_outliers"])
```
I can see myself being confused about this in the logs lol

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -0,0 +1,174 @@
...
+
+outlier_counts = (
+    outlier_counts.join(nonnull_counts, on="variable")
+    .with_columns((pl.col("n_out") / pl.col("n_total") * 100).alias("pct"))
+    .filter(pl.col("n_out") > 0)
+)
+
+
+if outlier_counts.is_empty():
+    logging.info("No outliers detected (z > %.1f).", Z)
```
crispy-wonton:
```suggestion
    logging.info("No outliers detected in any columns (z > %.1f).", Z)
```

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -0,0 +1,174 @@
...
+
+# non-null counts the same way
+nonnull_counts = df_valid.select(
+    [pl.col(c).is_not_null().sum().alias(c) for c in num_cols]
+).transpose(include_header=True, header_name="variable", column_names=["n_total"])
+
+outlier_counts = (
+    outlier_counts.join(nonnull_counts, on="variable")
+    .with_columns((pl.col("n_out") / pl.col("n_total") * 100).alias("pct"))
+    .filter(pl.col("n_out") > 0)
```
crispy-wonton: I would suggest logging the results for all columns, even if the column has no outliers.

It might also be useful to save the mean and std for each column so that we can contextualise the results.

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -0,0 +1,174 @@
...
+logging.info("Column dtypes: %s", df_valid.dtypes)
+
+# --- Full numeric summary to CSV ------------------------------------------ #
+summary_df = df_valid.select(cfg.NUMERIC_COLUMNS).describe()
+summary_file = (
+    Path(cfg.OUTPUT_DIR)
+    / f"{Path(cfg.DATA_PATH).stem}_numeric_summary_{datetime.now().strftime('%Y%m%d_%H%M%S')}.csv"
+)
+summary_df.write_csv(summary_file)
+logging.info("Full numeric summary ➜ %s", summary_file)
```
crispy-wonton, line 255: this is really cool. the arrow delighted me!!

crispy-wonton (commented):
Hey @helloaidank !

Wow!! This is really great work. I think you've done an amazing job of writing some easy to read , clear, and useful data quality checks.

This is so useful. It's already highlighted some weird values for the proportion of off gas properties!!

I have made some suggestions and asked some clarification questions. I didn't find any bugs or errors. I think the main suggestion i'm interested in is logging the outlier counts/ percentages for all columns, even if the column has no outliers, and logging the mean and std for each column so that we can contextualise the outlier results.

I've listed below some additional checks which I think would be really useful for us. Let's finalise and merge the current PR first and implement these in a separate PR though: 
- Check the same script can be run for individual property level data. From looking at it, it should be. We might also need a separate script where we do these similar checks on the property level data but grouped by LSOA.
- Count & % of nulls in each column. We have the counts in the summary csv but not the %.
- Count & % of rows where scores are weighted
- How many properties are used in the analysis per LSOA. And ideally we would know what % of the true number of properties in the LSOA are represented in the data.
- potentially consider adding some plots to visualise the distributions of the data in certain columns but i dont think this is a priority.

To answer your qs in the PR description:
The logic behind the validation checks, is there anything that we're obviously missing that we should do?
- as above

Do the results look sensible? Should we have diagnostic plots as easier to visualise then numbers (e.g. box plots based off dataframe outputted in csv)
- i think they look sensible overall. I am not surprised that there are some values slightly over 1. We know this is likely to be caused by small errors on the weights introduced by rounding. We should probably cap these at 1 AFTER we have done the data qc. However, I do find the errors in the proportion of off gas weird. I am surprised they are so far over 1. Generally I wouldnt want them to go above 1.1. These probably need some further investigation. Again, not in this PR and at least there are not too many but it's weird.
- The outliers pcts look ok to me to. It is hard to know from just the counts/pct if they are concerning. E.g. an outlier on the scoring can only really be between 0-1 which is realistic, vs an outlier for garden size could be 1 billion m2, which is obviously wrong. I suppose we could consider adding some ceilings for the ones that are not proportions or 0-1 values.
- One thing that is really weird. When i look at the logs, I can see that the max outlier for proportion off gas is 1.836 (in the data validation checks section). In the csv, i can see that it is 1.934....I cant see why this has happened from looking at the code. Any idea?

Should we move this to the pipeline directory since we are directly dealing with the final outputs of the pipeline and should be something that is automatically run
- i think that sounds sensible. E.g. into a new subdir called `/pipeline/data_qc`. I think both the qc config and script can go in this subdir. I suggest to do this right before merging so that we dont lose the diffs if you make changes and want a second review!
- One thing to start considering as well is that we will probably want to refactor this into metaflow AND automate it so that it runs whenever we produce a new dataset - I would save this for a separate PR.

Any input on the formatting of the logs, it's quite dry at the moment but also didn't want to overengineer for the first iteration
- I guess if we add a lot more checks it will get harder to read so maybe we can already start thinking about if it's possible to make it easier but for now I found it readable.


A log directory will be created, where'll you'll have a text file and a csv with the result from the data quality checks. Each one should have it's own time stamp! Review these please!
- I can see these files in the log folder and they have their own time stamps. Have reviewed the results as commented above.

Overall, really excellent work. Thank you!

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
helloaidank: good steer!

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
helloaidank: Yeah! Maybe something we could implement in rest of the pipeline as it's a recurrent theme.

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
helloaidank: Yep!!

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
helloaidank: I've done this for some of them, however when I do say 'Check.in_range(0, 1)' it has to be positional and keyword arguments cannot appear before it (e.g. col=c)

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
helloaidank: good point!

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
helloaidank: Yes, I'll add these to the numeric columns!

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
helloaidank: Good point!

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
helloaidank: Think it gives extra context about the error, the script still errors out!

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
helloaidank: Yep, good point!

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
helloaidank: Good shout loool

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
helloaidank: Yep!

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
helloaidank: Agreed!

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
helloaidank: Makes sense, potentially adds a little redundancy but simplified it now

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
helloaidank: Implemented this!

helloaidank:
Thanks so much for delivering such a comprehensive review Roisín and indulging on these questions. 

I’ve implemented that main suggestions around logging the outlier counts/percentages for all columns. Please take a look when you have a moment.

Thanks for these suggestions, happy to implement them in a separate PR. 

‘Do the results look sensible?’

I am guessing once we are happy with why they are slightly over 1, then at that point we should cap them. I think some of them are well over 1, and quite a few of those seem to centre on the proportion of off gas, likely something to investigate.

I was thinking similar in terms of quality checks of parameters with not the same defined range (potentially find the largest garden in Britain and use that as the maximum). 

With regards to the proportion of off gas logs, I think this is because the failure cases are just examples rather than an exhaustive list. Basically, the maximum is not included within those. So the issue re off gas might be bigger than we think, especially given the large % of outliers for proportion of off gas. Potentially having a log of those over say 1.1 would be useful, one to add in a separate PR I think. 

Re pipeline, that sounds sensible! I’ll do this right before merging. In terms of meta flow/automating that makes sense. I’ll be going through your metaflow PRs now, so that will get me in the mindset around how we should design script going forward. 

I’ll keep the logging as is for the moment and glad to hear the logging works! 

Thanks again,

Aidan

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
```diff
@@ -0,0 +1,124 @@
...
+# output directory for logs (local)
+OUTPUT_DIR = os.path.join(
+    PROJECT_DIR,
+    "asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/logs",
+    TIMESTAMP,
+)
+os.makedirs(OUTPUT_DIR, exist_ok=True)
+
+# path to use in the rest of the pipeline (read straight from S3)
+DATA_PATH = DATA_S3_URI
```
crispy-wonton, line 37: I think you can just use the `DATA_S3_URI` directly instead of copying it here? The way you have it makes the original variable more hidden.

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
```diff
@@ -0,0 +1,124 @@
...
+# Path to data file
+# timestamp for tracking
+TIMESTAMP = datetime.now().strftime("%Y%m%d_%H%M%S")
+
+# S3 URI for the data file
+DATA_S3_URI = "s3://asf-heat-pump-suitability/outputs/2023Q4/suitability/20250319_2023_Q4_heat_pump_suitability_per_lsoa.csv"
+
+# parse bucket and key from the URI (if you still need them elsewhere)
+parsed_uri = urlparse(DATA_S3_URI)
+S3_BUCKET = parsed_uri.netloc
```
crispy-wonton, line 25: Is this variable used anywhere? I can't see it.

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
```diff
@@ -0,0 +1,124 @@
...
+    "asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/logs",
+    TIMESTAMP,
+)
+os.makedirs(OUTPUT_DIR, exist_ok=True)
+
+# path to use in the rest of the pipeline (read straight from S3)
+DATA_PATH = DATA_S3_URI
+
+# append timestamp to log filename
+DATA_FILE_NAME = os.path.basename(S3_KEY)
```
crispy-wonton, line 40: Will you need a function that updates this if the user puts a different `DATA_PATH` variable in via argparse?

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -15,11 +15,27 @@
...
+# Argument parsing for data path
+# ------------------------------------------------------------------------------
+parser = argparse.ArgumentParser(
+    description="data quality checks for heat pump suitability data."
+)
+parser.add_argument(
+    "--data-path",
+    type=str,
+    default=cfg.DATA_PATH,
+    help="Path to the data CSV file (local or S3 URI).",
```
crispy-wonton:
```suggestion
    help="Path to the heat pump suitability per LSOA data CSV file (local or S3 URI).",
```
I would put something like this for clarity. Also, you can put in the docstring at the top of the file that this script takes the heat pump suitability per LSOA data which is an output from `asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_suitability.py`

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -31,18 +47,21 @@
 # 1.  BUILD SCHEMA
 # ---------------------------------------------------------------------------
 
-schema_columns: dict[str, Column] = {}
+# schema_columns: dict[str, Column] = {}
```
crispy-wonton: should this line be deleted?

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
```diff
@@ -126,49 +160,77 @@ def add(col: str, dtype, *checks, **kw):
 logging.info("=== Outlier Detection (Z-Score) ===")
 
 Z = cfg.OUTLIER_ZSCORE_THRESHOLD
-num_cols = cfg.NUMERIC_COLUMNS
+logging.info("No outliers detected in any columns (z > %.1f).", Z)
```
crispy-wonton:
```suggestion
logging.info("Outlier Z-score threshold: (z > %.1f).", Z)
```
I think this is a mistake?

crispy-wonton (commented):
hi @helloaidank ,

looks great! just highlighted a few small things to clean up now.

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
helloaidank: True it isn't!

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
helloaidank: Yep, good spot

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/pandera_eval_hps_outputs.py
helloaidank: Not sure what happened here in my brain hahaha

asf_heat_pump_suitability/analysis/evaluate_hp_suitability_outputs/config/dq_config.py
helloaidank: Created a util function which is now in the pandera script, likely want to refactor this out to a utils file at some point!

crispy-wonton (approved):
Looks great, merge away matey 🏴‍☠️

PR #157, opened 2025-06-03
153 parallelise suitability script

### LINKED ISSUE #153: Parallelise calculate suitability run script

crispy-wonton:
Fixes #153 

# Description

Refactor EPC calculating suitability script into metaflow. Summary of changes:

New files:
- `/pipeline/flows/run_calculate_suitability_flow.py` - new flow to calculate suitability scores for individual properties and LSOAs. Refactored version of `/pipeline/run_scripts/run_calculate_suitability.py`
- `/pipeline/suitability/__init__.py` - required to import from package
- `/utils/parallel_utils.py` - new utils file with new function to assist parallelisation. This is the same function as in PR #156 

Updated files:
- `MANIFEST.in` ; `setup.cfg` ; `setup.py` - changes required to import `asf_heat_pump_suitability` as package in batch machine

# Instructions for Reviewer

I have set up the script to run on a sample of 1000 EPC records and save it to S3. The results will not make sense because there will be lots of records missing for each LSOA. So this is simply to test the flow of the whole pipeline, not the results. It should be relatively quick to run. Please could you test that the flow runs all the way through successfully.

```
export METAFLOW_USER=yournamewithnospacesorpunctuation
python asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py --datastore=s3 run --gardens s3://asf-heat-pump-suitability/outputs/2023Q4/gardens/20250224_2023_Q4_EPC_garden_size_estimates_EWS_deduplicated.parquet --epc s3://asf-heat-pump-suitability/outputs/2023Q4/features/2023_Q4_EPC_features.parquet --weights s3://asf-heat-pump-suitability/outputs/2023Q4/weights/20250305_2023_Q4_EPC_weights.parquet --year 2023 --quarter 4 --max-workers 4
```

- Please do test that the creation of the sample works for you.
- Please could you also check that the results look sensible to you by loading and viewing the full dataset that i've already created with this new flow. I've put some code below. Feel free to do other basic checks you can think of!

```
import polars as pl
from polars.testing import assert_frame_equal
# Load suitability scores from metaflow pipeline output
new_df = pl.read_parquet("s3://asf-heat-pump-suitability/outputs/2023Q4/suitability/20250603_2023_Q4_heat_pump_suitability_per_lsoa.parquet")
# Load suitability scores from original script output
old_df = pl.read_parquet("s3://asf-heat-pump-suitability/outputs/2023Q4/suitability/20250319_2023_Q4_heat_pump_suitability_per_lsoa.parquet")
# Check the outputs are the same
assert_frame_equal(new_df, old_df, check_row_order=False)
new_df.describe()
```

- the logic of the refactored flow steps where it has changed from the original script. I've indicated the parts of the script that don't need a thorough review because they are unchanged (except to refactor into metaflow syntax. Please still do review these steps for the new syntax but the logic has been reviewed. And please do review the general overall flow and the documentation.

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
+++ b/asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
@@ -168,7 +168,7 @@ class CalculateSuitabilityFlow(FlowSpec):
         print("Weighting scores and aggregating per LSOA")
         self.weighted_scores = []
 
-        for lsoa_code in tqdm(self.input["lsoa"].unique()):
+        for lsoa_code in tqdm(self.input["lsoa"].drop_nulls().unique()):
             self.weighted_scores.append(
                 calculate_suitability.aggregate_dict_lsoa_suitability_and_features(
                     self.input, lsoa_code
--- a/asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
+++ b/asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
@@ -84,9 +84,9 @@ class CalculateSuitabilityFlow(FlowSpec):
         self.epc_df = self.epc_df.join(gardens, how="left", on="UPRN")
         self.epc_df = self.epc_df.join(weights, how="left", on="UPRN")
 
-        logging.info(f"Saving augmented EPC data")
-        save_as = f"s3://asf-heat-pump-suitability/outputs/{self.year}Q{self.quarter}/augmented_epc/{datetime.today().strftime('%Y%m%d')}_{self.year}_Q{self.quarter}_epc_augmented.parquet"
-        save_utils.save_to_s3(self.epc_df, save_as)
+        # logging.info(f"Saving augmented EPC data")
+        # save_as = f"s3://asf-heat-pump-suitability/outputs/{self.year}Q{self.quarter}/augmented_epc/{datetime.today().strftime('%Y%m%d')}_{self.year}_Q{self.quarter}_epc_augmented.parquet"
+        # save_utils.save_to_s3(self.epc_df, save_as)
         self.next(self.process_features_for_suitability)
 
     @step
@@ -153,16 +153,20 @@ class CalculateSuitabilityFlow(FlowSpec):
         )
         self.next(self.weight_scores, foreach="chunks")
 
-    @batch(cpu=2, memory=4000)
+    @batch(cpu=2, memory=1000)
     @step
     def weight_scores(self):
         """
         Apply weights to scores and aggregate for each LSOA.
         """
+        import os
+
         os.system(
-            "pip install git+https://github.com/nestauk/asf_heat_pump_suitability.git"
+            "pip install git+https://github.com/nestauk/asf_heat_pump_suitability.git@153_parallelise_suitability_script"
         )
         from tqdm import tqdm
+
+        print("Attempting suitability import")
         from asf_heat_pump_suitability.pipeline.suitability import calculate_suitability
 
         print("Weighting scores and aggregating per LSOA")
--- a/asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
+++ b/asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
@@ -69,6 +69,7 @@ class CalculateSuitabilityFlow(FlowSpec):
         import polars as pl
         from datetime import datetime
         from asf_heat_pump_suitability import config
+        from asf_heat_pump_suitability.utils import save_utils
 
         self.features = config["features"]
 
@@ -83,11 +84,9 @@ class CalculateSuitabilityFlow(FlowSpec):
         self.epc_df = self.epc_df.join(gardens, how="left", on="UPRN")
         self.epc_df = self.epc_df.join(weights, how="left", on="UPRN")
 
-        self.epc_df = self.epc_df.sample(n=1000)
-
-        # logging.info(f"Saving augmented EPC data")
-        # save_as = f"s3://asf-heat-pump-suitability/outputs/{self.year}Q{self.quarter}/augmented_epc/{datetime.today().strftime('%Y%m%d')}_{self.year}_Q{self.quarter}_epc_augmented.parquet"
-        # save_utils.save_to_s3(self.epc_df, save_as)
+        logging.info(f"Saving augmented EPC data")
+        save_as = f"s3://asf-heat-pump-suitability/outputs/{self.year}Q{self.quarter}/augmented_epc/{datetime.today().strftime('%Y%m%d')}_{self.year}_Q{self.quarter}_epc_augmented.parquet"
+        save_utils.save_to_s3(self.epc_df, save_as)
         self.next(self.process_features_for_suitability)
 
     @step
@@ -119,7 +118,7 @@ class CalculateSuitabilityFlow(FlowSpec):
         """
         from datetime import datetime
         from asf_heat_pump_suitability import config
-        from asf_heat_pump_suitability.utils import parallel_utils
+        from asf_heat_pump_suitability.utils import parallel_utils, save_utils
         from asf_heat_pump_suitability.pipeline.suitability import (
             calculate_suitability,
         )
@@ -137,8 +136,8 @@ class CalculateSuitabilityFlow(FlowSpec):
         for score_df in scores:
             self.epc_df = self.epc_df.join(score_df, on="UPRN", how="left")
 
-        # save_as = f"s3://asf-heat-pump-suitability/outputs/{self.year}Q{self.quarter}/suitability/{datetime.today().strftime('%Y%m%d')}_{self.year}_Q{self.quarter}_heat_pump_suitability_per_property.parquet"
-        # save_utils.save_to_s3(self.epc_df, save_as)
+        save_as = f"s3://asf-heat-pump-suitability/outputs/{self.year}Q{self.quarter}/suitability/{datetime.today().strftime('%Y%m%d')}_{self.year}_Q{self.quarter}_heat_pump_suitability_per_property.parquet"
+        save_utils.save_to_s3(self.epc_df, save_as)
 
         # Filter to relevant columns
         use_cols = (
@@ -150,7 +149,7 @@ class CalculateSuitabilityFlow(FlowSpec):
 
         # Chunk into dfs of 1000 LSOAs
         self.chunks = parallel_utils.chunk_df_by_group(
-            self.epc_df, group_col="lsoa", n=100
+            self.epc_df, group_col="lsoa", n=1000
         )
         self.next(self.weight_scores, foreach="chunks")
 
@@ -161,12 +160,12 @@ class CalculateSuitabilityFlow(FlowSpec):
         Apply weights to scores and aggregate for each LSOA.
         """
         os.system(
-            "pip install git+https://github.com/nestauk/asf_heat_pump_suitability.git@153_parallelise_suitability_script"
+            "pip install git+https://github.com/nestauk/asf_heat_pump_suitability.git"
         )
         from tqdm import tqdm
         from asf_heat_pump_suitability.pipeline.suitability import calculate_suitability
 
-        logging.info("Weighting scores and aggregating per LSOA")
+        print("Weighting scores and aggregating per LSOA")
         self.weighted_scores = []
 
         for lsoa_code in tqdm(self.input["lsoa"].unique()):
@@ -244,10 +243,10 @@ class CalculateSuitabilityFlow(FlowSpec):
         save_utils.save_to_s3(self.suitability_df, f"{save_as}.parquet")
         save_utils.save_to_s3(self.suitability_df, f"{save_as}.csv")
 
-        # logging.info("Saving open dataset to nesta-open-data S3 bucket")
-        # save_as = f"s3://nesta-open-data/asf_heat_pump_suitability/{self.year}Q{self.quarter}/{datetime.today().strftime('%Y%m%d')}_{self.year}_Q{self.quarter}_EPC_heat_pump_suitability_per_lsoa"
-        # save_utils.save_to_s3(self.suitability_df, f"{save_as}.parquet")
-        # save_utils.save_to_s3(self.suitability_df, f"{save_as}.csv")
+        logging.info("Saving open dataset to nesta-open-data S3 bucket")
+        save_as = f"s3://nesta-open-data/asf_heat_pump_suitability/{self.year}Q{self.quarter}/{datetime.today().strftime('%Y%m%d')}_{self.year}_Q{self.quarter}_EPC_heat_pump_suitability_per_lsoa"
+        save_utils.save_to_s3(self.suitability_df, f"{save_as}.parquet")
+        save_utils.save_to_s3(self.suitability_df, f"{save_as}.csv")
         self.next(self.end)
 
     @step
--- a/asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
+++ b/asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
@@ -3,12 +3,12 @@ Calculate suitability scores of different low-carbon heating technologies in Nes
 properties and LSOAs.
 
 To run:
-python asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_suitability.py run --weights [path/to/weighted/EPC] --features [path/to/EPC/with/features] --gardens [path/to/garden/size/estimates] -y [YYYY] -q [Q] --max-workers 8
+python asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py --datastore=s3 run --weights [path/to/weighted/EPC] --features [path/to/EPC/with/features] --gardens [path/to/garden/size/estimates] --year [YYYY] --quarter [Q] --max-workers 18
 
 NB: this pipeline takes the outputs from the following scripts as inputs:
-- asf_heat_pump_suitability/pipeline/run_scripts/run_compute_epc_weights.py
-- asf_heat_pump_suitability/pipeline/run_scripts/run_add_features.py
-- asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_garden_size.py
+- asf_heat_pump_suitability/pipeline/flows/run_compute_epc_weights_flow.py
+- asf_heat_pump_suitability/pipeline/flows/run_add_features_flow.py
+- asf_heat_pump_suitability/pipeline/flows/run_calculate_garden_size_flow.py
 """
 
 from metaflow import FlowSpec, step, batch, Parameter
--- a/asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
+++ b/asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
@@ -119,7 +119,7 @@ class CalculateSuitabilityFlow(FlowSpec):
         """
         from datetime import datetime
         from asf_heat_pump_suitability import config
-        from asf_heat_pump_suitability.pipeline.utils import parallel_utils
+        from asf_heat_pump_suitability.utils import parallel_utils
         from asf_heat_pump_suitability.pipeline.suitability import (
             calculate_suitability,
         )
@@ -160,6 +160,9 @@ class CalculateSuitabilityFlow(FlowSpec):
         """
         Apply weights to scores and aggregate for each LSOA.
         """
+        os.system(
+            "pip install git+https://github.com/nestauk/asf_heat_pump_suitability.git@153_parallelise_suitability_script"
+        )
         from tqdm import tqdm
         from asf_heat_pump_suitability.pipeline.suitability import calculate_suitability
 
@@ -202,6 +205,7 @@ class CalculateSuitabilityFlow(FlowSpec):
         Calculate the proportion of flats in each LSOA from the census data and join it to the suitability scores.
         Join LSOA and DZ names to suitability scores.
         """
+        import polars as pl
         from asf_heat_pump_suitability.pipeline.prepare_features import (
             property_type,
             output_areas,
--- a/asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
+++ b/asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
@@ -5,6 +5,8 @@ properties and LSOAs.
 To run:
 python asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py --datastore=s3 run --weights [path/to/weighted/EPC] --features [path/to/EPC/with/features] --gardens [path/to/garden/size/estimates] --year [YYYY] --quarter [Q] --max-workers 18
 
+Set --sample to True to run on a sample of 1000 EPC records
+
 NB: this pipeline takes the outputs from the following scripts as inputs:
 - asf_heat_pump_suitability/pipeline/flows/run_compute_epc_weights_flow.py
 - asf_heat_pump_suitability/pipeline/flows/run_add_features_flow.py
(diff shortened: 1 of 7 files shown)


asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+    quarter = Parameter(
+        name="quarter",
+        help="EPC data quarter",
+        type=int,
+        required=True,
+    )
+
+    @step
+    def start(self):
+        """
```
crispy-wonton, line 64: Nothing new in this step except refactoring to the metaflow syntax.

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+        self.epc_df = self.epc_df.join(weights, how="left", on="UPRN")
+
+        # logging.info(f"Saving augmented EPC data")
+        # save_as = f"s3://asf-heat-pump-suitability/outputs/{self.year}Q{self.quarter}/augmented_epc/{datetime.today().strftime('%Y%m%d')}_{self.year}_Q{self.quarter}_epc_augmented.parquet"
+        # save_utils.save_to_s3(self.epc_df, save_as)
+        self.next(self.process_features_for_suitability)
+
+    @step
+    def process_features_for_suitability(self):
+        """
```
crispy-wonton, line 97: Nothing new in this step except refactoring to the metaflow syntax.

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+        )
+
+        # TODO remove before merge
+        self.epc_df = self.epc_df.sample(n=1000, seed=2)
+
+        self.next(self.calculate_scores_per_epc_record)
+
+    @step
+    def calculate_scores_per_epc_record(self):
+        """
```
crispy-wonton, line 124: Nothing new in lines 120-142 except refactoring to metaflow syntax.

Lines 144 - 151 are new and require more thorough review.

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+            self.epc_df,
+            group_col="lsoa",
+            n=100,
+        )
+        self.next(self.weight_scores, foreach="chunks")
+
+    # @batch(cpu=2, memory=16000)
+    @batch(cpu=2, memory=1000)
+    @step
+    def weight_scores(self):
```
crispy-wonton, line 175: Nothing new in this step except refactor to metaflow syntax although please review syntax carefully here

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+                calculate_suitability.aggregate_dict_lsoa_suitability_and_features(
+                    self.input, lsoa_code
+                )
+            )
+
+        self.next(self.concatenate_weighted_scores)
+
+    @step
+    def concatenate_weighted_scores(self, inputs):
+        """
```
crispy-wonton, line 201: New code in this step except for more thorough review

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+        suitability_df = pl.DataFrame(self.weighted_scores).filter(
+            pl.col("n_properties") >= 15
+        )
+        self.suitability_df = suitability_df.with_columns(pl.col(pl.Float64).round(3))
+
+        self.next(self.join_proportion_of_flats_and_names)
+
+    @step
+    def join_proportion_of_flats_and_names(self):
+        """
```
crispy-wonton, line 222: Nothing new in this step except refactor to metaflow syntax

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+        )
+        self.suitability_df = self.suitability_df.join(
+            proportion_flats_df, how="left", on="lsoa"
+        ).join(lsoa_names_df, left_on="lsoa", right_on="lsoa_code", how="left")
+
+        self.next(self.save_outputs)
+
+    @step
+    def save_outputs(self):
+        """
```
crispy-wonton, line 252: Nothing new in this step except refactor to metaflow syntax

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
+"""
+Calculate suitability scores of different low-carbon heating technologies in Nesta and 'conventional' views for individual
+properties and LSOAs.
+
+To run:
+python asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_suitability.py run --weights [path/to/weighted/EPC] --features [path/to/EPC/with/features] --gardens [path/to/garden/size/estimates] -y [YYYY] -q [Q] --max-workers 8
```
helloaidank: putting an example run command is useful here!

helloaidank:
```suggestion
python asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_suitability.py run --weights [path/to/weighted/EPC] --features [path/to/EPC/with/features] --gardens [path/to/garden/size/estimates] -y [YYYY] -q [Q] --max-workers 8
Example:
export METAFLOW_USER=yournamewithnospacesorpunctuation
python asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py --datastore=s3 run --gardens s3://asf-heat-pump-suitability/outputs/2023Q4/gardens/20250224_2023_Q4_EPC_garden_size_estimates_EWS_deduplicated.parquet --epc s3://asf-heat-pump-suitability/outputs/2023Q4/features/2023_Q4_EPC_features.parquet --weights s3://asf-heat-pump-suitability/outputs/2023Q4/weights/20250305_2023_Q4_EPC_weights.parquet --year 2023 --quarter 4 --max-workers 4
```

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+    year = Parameter(
+        name="year",
+        help="EPC data year. Format YYYY",
+        type=int,
+        required=True,
+    )
+
+    quarter = Parameter(
+        name="quarter",
+        help="EPC data quarter",
```
helloaidank:
```suggestion
        help="EPC data quarter (i.e. range from 1-4)",
```

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+
+    @step
+    def start(self):
+        """
+        Load input datasets and start flow.
+        """
+        import polars as pl
+        from datetime import datetime
+        from asf_heat_pump_suitability import config
+        from asf_heat_pump_suitability.utils import save_utils
```
helloaidank, line 73: my ruff says this is unused

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+        logging.info("Loading garden size estimates")
+        gardens = pl.read_parquet(self.gardens)
+        logging.info("Loading weights")
+        weights = pl.read_parquet(self.weights)
+
+        logging.info("Joining EPC features data with garden size estimates and weights")
+        self.epc_df = self.epc_df.join(gardens, how="left", on="UPRN")
+        self.epc_df = self.epc_df.join(weights, how="left", on="UPRN")
+
+        # logging.info(f"Saving augmented EPC data")
```
helloaidank: Do we want to delete this comments?

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+            .fill_null(pl.col("msoa_avg_outdoor_space_m2"))
+            .alias("garden_area_m2")
+        ).drop("msoa_avg_outdoor_space_m2")
+
+        logging.info("Filtering EPC data to rows with n_features >= minimum threshold")
+        self.epc_df = calculate_suitability.filter_df_minimum_features(
+            self.epc_df, features=self.features
+        )
+
+        # TODO remove before merge
```
helloaidank: Same as with the parallelised reweighting script, wondering if we can introduce a flag when we want to test the metaflow pipeline

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+        self.next(self.calculate_scores_per_epc_record)
+
+    @step
+    def calculate_scores_per_epc_record(self):
+        """
+        Calculate suitability score per tech type per EPC record.
+        """
+        from datetime import datetime
+        from asf_heat_pump_suitability import config
+        from asf_heat_pump_suitability.utils import parallel_utils, save_utils
```
helloaidank, line 129: again my ruff says save_utils isn't used here

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+            epc_scores_df = calculate_suitability.compute_df_avg_score_per_epc(
+                self.epc_df, tech_type
+            )
+            scores.append(epc_scores_df)
+
+        logging.info("Joining all scores to EPC dataset")
+        for score_df in scores:
+            self.epc_df = self.epc_df.join(score_df, on="UPRN", how="left")
+
+        # save_as = f"s3://asf-heat-pump-suitability/outputs/{self.year}Q{self.quarter}/suitability/{datetime.today().strftime('%Y%m%d')}_{self.year}_Q{self.quarter}_heat_pump_suitability_per_property.parquet"
```
helloaidank: do we want to delete these comments from the script?

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+        use_cols = (
+            ["lsoa", "proportional_weight"]
+            + [col for col in self.epc_df.columns if "score" in col]
+            + self.features
+        )
+        self.epc_df = self.epc_df.select(use_cols)
+
+        # Chunk into dfs of 1000 LSOAs
+        self.chunks = parallel_utils.chunk_df_by_group(
+            # self.epc_df, group_col="lsoa", n=1000
```
helloaidank: assuming again this is because we are running over the sample

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+        import itertools
+
+        self.weighted_scores = list(
+            itertools.chain.from_iterable([input.weighted_scores for input in inputs])
+        )
+        logging.info(
+            "Filtering to LSOAs with data for at least 15 properties to be included in final dataset"
+        )
+        suitability_df = pl.DataFrame(self.weighted_scores).filter(
+            pl.col("n_properties") >= 15
```
helloaidank, line 214: Do you think that at any point we might want to make this parameter configurable?

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+        self.next(self.concatenate_weighted_scores)
+
+    @step
+    def concatenate_weighted_scores(self, inputs):
+        """
+        Concatenate all weighted LSOA suitability scores together and filter to LSOAs with at least 15 records.
+        """
+        import polars as pl
+        import itertools
+
```
helloaidank, line 206: Do you think it would be worth have some exceptions in case a branch produces no data (e.g. an empty list)?

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+        from asf_heat_pump_suitability.pipeline.utils import save_utils
+
+        logging.info("Saving LSOA heat pump suitability scores")
+        # save_as = f"s3://asf-heat-pump-suitability/outputs/{self.year}Q{self.quarter}/suitability/{datetime.today().strftime('%Y%m%d')}_{self.year}_Q{self.quarter}_heat_pump_suitability_per_lsoa"
+
+        save_as = f"s3://asf-heat-pump-suitability/outputs/{self.year}Q{self.quarter}/suitability/{datetime.today().strftime('%Y%m%d')}_{self.year}_Q{self.quarter}_heat_pump_suitability_per_lsoa_SAMPLE"
+        save_utils.save_to_s3(self.suitability_df, f"{save_as}.parquet")
+        save_utils.save_to_s3(self.suitability_df, f"{save_as}.csv")
+
+        # logging.info("Saving open dataset to nesta-open-data S3 bucket")
```
helloaidank: Are we keeping these lines?

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+    # @batch(cpu=2, memory=16000)
+    @batch(cpu=2, memory=1000)
+    @step
+    def weight_scores(self):
+        """
+        Apply weights to scores and aggregate for each LSOA.
+        """
+        import os
+
+        os.system(
```
helloaidank, line 181: Just remember to change to dev when we decide to merge!

asf_heat_pump_suitability/utils/parallel_utils.py
```diff
@@ -0,0 +1,23 @@
+from typing import List
+import polars as pl
+
```
helloaidank, line 3: reviewed in the other script

requirements.txt
```diff
@@ -23,3 +23,4 @@ tenacity
 numpy<2.0.0
 fiona
 bs4
+metaflow
```
helloaidank, line 26: ahhhh, I reviewed the other script first, I was wondering where that was the reweighting PR!

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,263 @@
...
+
+        self.next(self.save_outputs)
+
+    @step
+    def save_outputs(self):
+        """
+        Save outputs to S3.
+        """
+        from datetime import datetime
+        from asf_heat_pump_suitability.pipeline.utils import save_utils
```
helloaidank: Think this is a bug, I get an error when I get to this step and is inconsistent with the rest of our steps


```suggestion
        from asf_heat_pump_suitability.utils import save_utils
```

helloaidank (commented):
Hey!

Thanks for this, all the logic seems good to me and thanks again for demarcating between the code that I should and shouldn't closely look at. I've gone through the different files, I think quite a few of these are similar to the reweighting script and I'll let you answer the questions in that PR with regards to them.

I believe I found a small bug when importing a save_utils package in one of the steps (crashed unfortunately when I ran it first time, but ran after small code change), and just some small comments on documentation as well as introducing a test_run flag (same as for the other PR). 

I've checked the outputs, that's when I noticed the interesting difference between Nesta SGL and Standard SGL, just noting it here for prosperity. Otherwise everything checks out I think!

requirements.txt
crispy-wonton: yep, forgot to add in the other one!

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
crispy-wonton: good idea!

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
crispy-wonton: No, this is to uncomment before merging. This is commented out because I set it up to run on a sample.

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
crispy-wonton: This is because the line to save the data is commented out currently :)

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
crispy-wonton: no - they will be uncommented before merging to dev! this is just so we dont save this out when running on the sample :)

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
crispy-wonton: yep!

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
crispy-wonton: it's because of the commented out line saving the data L79

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
crispy-wonton: Perhaps. Can you think of a reason when this could happen? I can't think of why this would happen?

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
crispy-wonton: We certainly could and we also might want to revisit this number when revisiting the scoring.

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
crispy-wonton: yep! commented out now for sampling purposes. will uncomment before merge

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
crispy-wonton: hmmm.. this should not happen. It might be an artefact of the `sym-linking` i did. I will look into it.

crispy-wonton: I have looked into this. I cannot see that git was tracking the symlink so I'm not sure why this happened for you. Here's the commands I ran in terminal:

1. `find . -type l` - this shows me there is a symlink in my local copy of the repo in `asf_heat_pump_suitability/pipeline/utils`. 
2. `find . -type l -exec ls -ld {} +` - this shows me my local symlink is linking the above folder to `asf_heat_pump_suitability/utils`
3. `git rm asf_heat_pump_suitability/pipeline/utils` - this should remove the file from git's tracking. It outputted an error: `fatal: pathspec 'asf_heat_pump_suitability/pipeline/utils' did not match any files`. This means it's not being tracked by git.
4. `git ls-files -s | grep "utils"` - this lists the paths that git is tracking with the word 'utils' in it. I can see that `'asf_heat_pump_suitability/pipeline/utils` is not present in this list, confirming git is not tracking this directory.
5. `rm asf_heat_pump_suitability/pipeline/utils` - this removes the symlink directory from my local repo
6. `git status` - i can see there are no changes/deletions to commit.

From this, it seems like the symlinking was local only but your fix implies that your local repo also has the symlink. I would suggest pulling the branch again and then running the command in step 1 above to see if you have a local symlink and let me know.

crispy-wonton: 🤦‍♀️ ok now i see that I read your change the wrong way round. Ignore all my comments. What you've done is correct and has fixed an actual bug!!! bro whyyyy am i so incapable of reading

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
crispy-wonton: I will leave this for the README so it's all in one place.

crispy-wonton:
I have added a parameter to the flow to allow the user so specify if they want to run the flow on a sample of the EPC dataset. I tried out several ways to change the `memory` for the `batch` step dynamically - i.e. use less memory when running on a sample. I couldn't find a solution that was readable, robust, and user friendly. Here's what I tried:
1. use an `if` / `else` statement in the `start` step to define `self.batch_memory` depending on whether `self.sample` is `True` or not. Call the `self.batch_memory` in the `@batch(cpu=2, memory=lambda flow: flow.batch_memory)`. This did not work as the whole `lambda` function is being read as a string.
2. Factory pattern where you wrap the step like and define it near the top or outside of the class:
```
from functools import partialmethod
def step_factory(sample: bool)
    @partialmethod
    @batch(memory=1000 if sample else 10000)
    @step
    def your_step():…
```
Then
```
your_step = step_factory(sample=self.sample)
self.next(your_step)
your_step
```
I decided against this because it makes the code less readable. I didn't test it but I'm not sure it will work given the same potential issue as in step 1.
3. Use `Config` object and select different config files depending on whether the code is running on a sample or not. This means 2 additional config files are required. I decided against this because I didn't feel it was robust. It requires the user to pass another argument alongside `--sample` along with a file path to a secondary config file.
4. Conditional branching. Not available in `metaflow` and the [suggested workaround](https://community.outerbounds.com/t/28790044/hey-i-m-trying-to-implement-conditional-branching-to-give-th) is not suitable for this use case because I dont want to set up a batch instance that is immediately torn down on a given condition.

There may be a way to feed in args from a `bash` / trigger script so I can revisit this when I get to that. For now, I'm leaving the memory as a constant.

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
helloaidank: mmm it could be for example in the `filter_df_minimum_feature`s function as an example (something potentially wrong with the dataset itself), guess it just makes it easier for debugging if this does occur but happy to keep as is.

asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py
```diff
@@ -0,0 +1,283 @@
+"""
+Calculate suitability scores of different low-carbon heating technologies in Nesta and 'conventional' views for individual
+properties and LSOAs.
+
+To run:
+python asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py --datastore=s3 run --weights [path/to/weighted/EPC] --features [path/to/EPC/with/features] --gardens [path/to/garden/size/estimates] --year [YYYY] --quarter [Q] --max-workers 18
```
helloaidank, line 6: Update doc string to include sample parameter
```suggestion
python asf_heat_pump_suitability/pipeline/flows/run_calculate_suitability_flow.py --datastore=s3 run --weights [path/to/weighted/EPC] --features [path/to/EPC/with/features] --gardens [path/to/garden/size/estimates] --year [YYYY] --quarter [Q] --max-workers 18 --parameter True
```

helloaidank (dismissed):
Ah interesting, I was going to suggest conditional branching but seems like none of these avenues are simple. I think it's fine? Seems like a bit of an oversight not to be able to dynamically allocate resources. Thanks for making the change! Just suggested one change in terms of adding the --sample flag to the docstring at the beginning of this file.

PR #166, opened 2025-06-24
soft pin requirements.txt versions

### LINKED ISSUE #163: Pin major package versions
We should pin package versions to major releases in our `requirements.txt` so that we don't encounter breaking changes.

crispy-wonton:
Fixes #163 

# Description

Pin major package versions in `requirements.txt` to prevent deprecation issues

# Instructions for Reviewer

Please try running `pip install -r requirements.txt` to ensure that the required packages can be installed without dependency conflicts.

# Checklist:

- [ ] I have tested the code


caldwellst (dismissed):
All looks good, I can successfully install everything. Would still recommend shifting to `uv` in the long run. You can store all of these requirements in the `pyproject.toml` and keep the specific versions pinned in `uv.lock`, which is done automatically by `uv`. In the rare case you actually NEED a specific version for true compatibility reasons, like `scikit-learn<=1.2.2`, this would typically be resolved by default by `uv` and put in the `uv.lock`, but you could also manually specify the same in the `pyproject.toml`.

Issue #171, opened 2025-08-07
Explore building a full property dataset


crispy-wonton:
Exploration to build a dataset with all residential properties (not just EPC) and fill any missing key property-level data points.

Issue #167, opened 2025-06-26
Update conservation area dataset for England


crispy-wonton:
New dataset available: https://www.planning.data.gov.uk/dataset/conservation-area#

PR #175, opened 2025-08-21
Add notebook to create full features dataset for plymouth

crispy-wonton:
Progress towards fixing #171 

**NOTE: to run the notebook you may need to `pip install scikit-learn==1.3`**

# Description

Add new notebook to start creating full dataset of features for Plymouth for new Phase 3 suitability and feasibility calculations. The final dataset should be one with one row per UPRN of all residential UPRNs in Plymouth and a complete set of features relevant to feasibility scoring and tech categorisation. The main purpose of this notebook is to start collating the relevant datasets and explore some methodologies that could be used to impute missing data points. The notebook does not generate the full final dataset - it adds (and imputes, where required) the following features:
- in listed building
- in building conservation area
- in HN zone
- property type
- tenure
- off gas status

Additional features will be added and imputed in ongoing work but I wanted to get this out so you can start sense-checking the methodology.

The purposes of this exploratory work are to:
1. Identify methodologies for imputing missing feature data
2. Create an idea of the flow of what the pipeline will eventually look like
3. Create a minimum viable example dataset for demonstration purposes in Plymouth

**Note: I fully expect us to iterate these methodologies to improve them over time as we develop the pipeline. I want to prioritise getting this data good enough for a demo at the moment so please focus on reviewing from that perspective, but also please do leave any comments/suggestions for improvements that will take more time and exploration with the knowledge that we will make deeper improvements after the Plymouth workshop.**

# Instructions for Reviewer

convert the script to a notebook using jupytext:
```
pip install jupytext
jupytext --to notebook asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```
- The methodologies for imputation and the evaluation/validation of those methodologies. The main one to pay close attention to is the method to impute missing TENURE data. As you will see, I haven't done any hyperparameter tuning or detailed evaluation. I wanted to sense check the general method and datasets used before going too far into model tuning and evaluation. If you think a multi-class classifier model is suitable, I'd be keen to hear your ideas for how to best evaluate it beyond average precision and accuracy scores. I'm very open to other ideas for models/methods.
- Please let me know if you have other ideas for improving the imputation methods of the other features as well.

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
+++ b/asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
@@ -19,21 +19,8 @@ import geopandas as gpd
 import polars as pl
 import pandas as pd
 import fiona
-import momepy
-import folium
-import networkx as nx
-import rustworkx as rx
-import osmnx as ox
 import numpy as np
-from scipy.sparse import csr_matrix
-from sklearn.cluster import DBSCAN, HDBSCAN
-import graphblas_algorithms as ga
-import matplotlib.cm as cm
 import matplotlib.pyplot as plt
-from tqdm import tqdm
-from matplotlib import collections
-import libpysal
-import spopt
 from sklearn.preprocessing import OneHotEncoder
 from sklearn import metrics
 from sklearn.neighbors import NearestNeighbors, KNeighborsClassifier
--- a/asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
+++ b/asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
@@ -50,8 +50,10 @@ import folium
 
 # %% [markdown]
 # ## Load data
+# Some datasets are for Great Britain / England and some are for Plymouth only to increase download speed.
 
 # %%
+# View OpenMap Local dataset layers for GB
 fiona.listlayers("s3://asf-heat-pump-suitability/source_data/opmplc_gb.gpkg")
 
 # %%
@@ -121,7 +123,7 @@ epc_df = prepare_sample.add_col_property_type(epc_df)
 # -------------------------------------------------------------------------#
 
 print("\nLOADING DATASETS TO FILL MISSING DATA")
-print("\nLoading OS Code-Point Open...")
+print("\nLoading OS Code-Point Open (all postcode units in GB)...")
 code_point_df = gpd.read_file(
     "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/codepo_gb.gpkg",
     layers="codepoint",
@@ -241,7 +243,7 @@ plymouth_residential_uprns_gdf.shape
 # Identify UPRNs in listed buildings
 listed_building_uprns = (
     plymouth_residential_uprns_gdf.sjoin_nearest(
-        listed_buildings_gdf, how="inner", max_distance=10
+        listed_buildings_gdf, how="inner", max_distance=10  # 10 metres
     )["UPRN"]
     .unique()
     .tolist()
@@ -265,10 +267,9 @@ len(listed_building_uprns) / len(plymouth_residential_uprns_gdf) * 100
 epc_df = output_areas.standardise_col_postcode(epc_df, pcd_col="POSTCODE")
 epc_df = epc_df.with_columns(
     # Standardise UPRN
-    pl.col("UPRN")
-    .str.replace(".0", "", literal=True)
-    .cast(pl.Int64, strict=False)
-    .alias("UPRN"),
+    pl.col("UPRN").str.replace(".0", "", literal=True)
+    # This will cast filled UPRNs made up of Building Ref Number and Address back to null
+    .cast(pl.Int64, strict=False).alias("UPRN"),
     pl.col("MAIN_FUEL").cast(pl.String).alias("MAIN_FUEL"),
     pl.col("PROPERTY_TYPE").cast(pl.String).alias("PROPERTY_TYPE"),
 )
@@ -294,11 +295,11 @@ features_df = (
         pl.when(pl.col("UPRN").is_in(cons_area_uprns))
         .then(True)
         .otherwise(False)
-        .alias("in_cons_area"),
+        .alias("in_conservation_area"),
         pl.when(pl.col("UPRN").is_in(uprns_in_hn))
         .then(True)
         .otherwise(False)
-        .alias("in_hn"),
+        .alias("in_heat_network_zone"),
     )
     .join(epc_df, how="left", on="UPRN")
 )
@@ -407,13 +408,13 @@ print(len(features_df))
 # Remove space from postcode in code point data
 code_point_df["POSTCODE"] = code_point_df["postcode"].str.replace(" ", "")
 
-# Find nearest postcode point for all UPRNs within a 1km radius
+# Find nearest postcode point within a 1km radius for all UPRNs
 nearest_postcode_df = pl.from_pandas(
     plymouth_residential_uprns_gdf.sjoin_nearest(
         code_point_df[["POSTCODE", "geometry"]],
         how="left",
         max_distance=1000,
-        distance_col="distance_to_postcode_m",
+        distance_col="distance_to_postcode_m",  # distance in metres
     ).drop(columns="index_right")[["UPRN", "POSTCODE", "distance_to_postcode_m"]]
 )
 
@@ -435,6 +436,7 @@ off_gas_df = (
         pl.col("POSTCODE").alias("postcodes"),
         pl.col("POSTCODE").len().alias("n_postcodes"),
         pl.col("off_gas").mode().alias("modal_off_gas"),
+        # All nearest postcodes will be the same distance so we just select the first to keep one
         pl.col("distance_to_postcode_m").first().alias("distance_to_postcode_m"),
     )
     .with_columns(
@@ -545,7 +547,7 @@ uprns_buildings_df = pl.from_pandas(
     .drop(columns=["index_right", "geometry"])
 ).join(features_df.select(["UPRN", "property_type"]), how="left", on="UPRN")
 
-# Identify building IDs which have UPRNs missing property type information
+# Identify building IDs which have one or more UPRNs missing property type information
 missing_property_type_buildings_list = (
     uprns_buildings_df.filter(pl.col("property_type").is_null())["building_id"]
     .unique()
@@ -574,7 +576,7 @@ property_type_df = (
 )
 
 print(
-    "UPRNs that are missing a property type and are in a building with multiple property type labels:"
+    "Proportion of UPRNs that are missing a property type and are in a building with multiple property type labels:"
 )
 print(
     property_type_df.filter(pl.col("nunique_property_types") > 2)["null_count"].sum()
@@ -582,7 +584,7 @@ print(
 )
 
 print(
-    "\nUPRNs that are missing a property type and are in a building with multiple MODAL property type labels:"
+    "\nProportion of UPRNs that are missing a property type and are in a building with multiple MODAL property type labels:"
 )
 print(
     property_type_df.filter(pl.col("n_modal_property_type") > 1)["null_count"].sum()
@@ -590,7 +592,7 @@ print(
 )
 
 print(
-    "\nUPRNs that are missing a property type and are in a multi-property building with NO property type labels:"
+    "\nProportion of UPRNs that are missing a property type and are in a multi-property building with NO property type labels:"
 )
 print(
     property_type_df.filter(
@@ -656,6 +658,7 @@ plt.legend()
 
 # %%
 # CALCULATE NEAREST NEIGHBOUR DISTANCE
+# Calculate NN distance for properties where property type is known and unique within a building
 
 # Calculate distance from nearest neighbour for UPRNs with property type label
 # Get coordinates of each UPRN
@@ -722,14 +725,16 @@ table_df = table_df.set_index("n_UPRNs").sort_index()[
         "Caravan or other mobile or temporary structure",
     ]
 ]
+# Percentage of properties with n_UPRNs, per property type
 table_df / table_df.sum() * 100
 
 # %%
+# Percentage of all properties which are X property type per n_UPRNs
 table_df.div(table_df.sum(axis=1), axis=0) * 100
 
 # %%
 # ----------------------------------------------------------------------------------- #
-# CALCULATE NEAREST NEIGHBOUR DISTANCE
+# APPLY MODEL - CALCULATE NEAREST NEIGHBOUR DISTANCE FOR ALL UPRNs
 
 # Calculate distance from nearest neighbour for all UPRNs
 # Get coordinates of each UPRN
@@ -1133,7 +1138,7 @@ model_df = (
     )
     .join(nn_tenure_df.drop("prop_nn_private_rental"), how="left", on="UPRN")
     .join(
-        features_df.select(["UPRN", "in_cons_area", "in_listed_building"]),
+        features_df.select(["UPRN", "in_conservation_area", "in_listed_building"]),
         how="left",
         on="UPRN",
     )
@@ -1194,9 +1199,10 @@ results_df = (
             "2": y_pred_prob[:, 2],
         }
     )
-    .rename({k: v for k, v in zip(["0", "1", "2"], classifier.classes_)})
-    .with_columns(
+    # Rename columns with class names
+    .rename({k: v for k, v in zip(["0", "1", "2"], classifier.classes_)}).with_columns(
         (pl.col("actual") == pl.col("predicted")).alias("correct_prediction"),
+        # Get maximum class prediction probability across all predictions for a given UPRN
         max_prob=pl.concat_list(
             "owner-occupied", "rental (private)", "rental (social)"
         ).list.max(),
@@ -1271,7 +1277,7 @@ y_true = np.array(y_test["TENURE"].map(tenure_mapping))
 
 y_pred_arr = np.copy(y_pred)
 for k, v in tenure_mapping.items():
-    y_pred_arr[y_pred == k] = v
+    y_pred_arr = [tenure_mapping.get(k) for k in y_pred]
 
 matrix = confusion_matrix(list(y_true), list(y_pred_arr), normalize="true")
 print(matrix.diagonal() / matrix.sum(axis=1))
@@ -1451,7 +1457,7 @@ features = [
     "Y_COORDINATE",
     "prop_nn_owner_occupied",
     "prop_nn_social_rental",
-    "in_cons_area",
+    "in_conservation_area",
     "in_listed_building",
 ]
 
@@ -1719,7 +1725,7 @@ features_df = features_df.with_columns(
 model = HDBSCAN(
     min_cluster_size=5,
     cluster_selection_epsilon=20,
-    algorithm="balltree",
+    algorithm="ball_tree",
     metric="euclidean",
 )
 plymouth_residential_uprns_gdf["cluster"] = model.fit_predict(
@@ -1828,8 +1834,8 @@ select_features = [
     "OA21CD",
     "LSOA21CD",
     "in_listed_building",
-    "in_cons_area",
-    "in_hn",
+    "in_conservation_area",
+    "in_heat_network_zone",
     "filled_off_gas",
     "use_property_type",
     "use_tenure",
@@ -1843,8 +1849,6 @@ select_features = [
 
 rename_selected = {
     "OA21CD": "oa21",
-    "in_cons_area": "in_conservation_area",
-    "in_hn": "in_heat_network_zone",
     "use_garden_area_m2": "garden_area_m2",
     "use_property_type": "predicted_property_type",
     "use_tenure": "predicted_tenure",
--- a/asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
+++ b/asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
@@ -21,12 +21,17 @@ import pandas as pd
 import fiona
 import numpy as np
 import matplotlib.pyplot as plt
+from matplotlib import ticker
+
+from sklearn.cluster import HDBSCAN
 from sklearn.preprocessing import OneHotEncoder
 from sklearn import metrics
+
 from sklearn.neighbors import NearestNeighbors, KNeighborsClassifier
 from sklearn.model_selection import train_test_split
 from sklearn.ensemble import RandomForestClassifier
 
+from asf_heat_pump_suitability.utils import geo_utils, save_utils
 from asf_heat_pump_suitability.getters import get_datasets
 from asf_heat_pump_suitability.pipeline.prepare_features import (
     lat_lon,
@@ -35,9 +40,13 @@ from asf_heat_pump_suitability.pipeline.prepare_features import (
     off_gas,
     anchor_properties,
     output_areas,
+    building_footprint,
+    land_extent,
+    garden_size,
 )
 from asf_heat_pump_suitability.analysis.flats_on_fossils.features import fuel_type
 from asf_heat_pump_suitability.pipeline.reweight_epc import prepare_sample
+import folium
 
 # %% [markdown]
 # ## Load data
@@ -75,23 +84,18 @@ os_openmap_railway_gdf = gpd.read_file(
 print("\nLOADING PROPERTY-LEVEL FEATURE DATASETS...")
 print("\nLoading listed buildings...")
 listed_buildings_gdf = gpd.read_file(
-    "../spatial_clustering/data/National_Heritage_List_for_England_NHLE_v02_VIEW_-464524051049198649/Listed_Building_points.shp"
+    "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/National_Heritage_List_for_England_NHLE_v02_VIEW_-464524051049198649/Listed_Building_points.shp"
 )
 
 print("\nLoading building conservation areas...")
 cons_areas_gdf = gpd.read_file(
-    "../spatial_clustering/data/conservation-area (1).geojson"
+    "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/conservation-area (1).geojson"
 )
 
 print("\nLoading existing / planned HN zones...")
 hn_zones_gdf = gpd.read_file(
     "s3://asf-heat-pump-suitability/heat_network_desnz_data/heat-network-zone-map-Plymouth.gpkg",
     layer="heat-network-zone-map-Plymouth",
-)
-
-print("\nLoading outdoor space...")
-outdoor_space_df = pl.read_parquet(
-    "s3://asf-heat-pump-suitability/outputs/2023Q4/gardens/20250224_2023_Q4_EPC_garden_size_estimates_EWS_deduplicated.parquet"
 )
 
 print("\nLoading off gas postcodes...")
@@ -248,11 +252,6 @@ cons_area_uprns = plymouth_residential_uprns_gdf.sjoin(
     cons_areas_gdf.to_crs(epsg=27700), how="inner", predicate="intersects"
 )["UPRN"].tolist()
 
-# Reformat UPRN for join
-outdoor_space_df = outdoor_space_df.with_columns(
-    pl.col("UPRN").str.replace(".0", "", literal=True).cast(pl.Int64).alias("UPRN")
-)
-
 # UPRNs in HNs
 uprns_in_hn = plymouth_residential_uprns_gdf.sjoin(
     hn_zones_gdf.drop(columns=["index_right"]), predicate="within"
@@ -301,7 +300,6 @@ features_df = (
         .otherwise(False)
         .alias("in_hn"),
     )
-    .join(outdoor_space_df, how="left", on="UPRN")
     .join(epc_df, how="left", on="UPRN")
 )
 
@@ -334,6 +332,7 @@ postcode_lookup_df = output_areas.load_transform_df_area_info(
         "oslaua",
         "oa21",
         "imd",
+        # "LSOA21NM"
     ]
 )
 
@@ -840,7 +839,7 @@ features_df = features_df.join(
 # ## Fill missing data - tenure type
 
 # %%
-# Load census data
+# LOAD CENSUS DATA
 print("Loading age band data...")
 age_bands_df = pl.read_csv(
     "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/2021Census_age_bands_OA_plymouth.csv",
@@ -878,7 +877,7 @@ oa_tenure_df = pl.read_csv(
 )
 
 # %%
-# Preprocess census data to generate features for predictive model
+# PREPROCESS CENSUS DATA TO GENERATE FEATURES FOR PREDICTIVE MODEL
 # Process age band data
 age_bands_df = (
     age_bands_df.filter(
@@ -1011,9 +1010,10 @@ oa_tenure_df = (
 )
 
 # Join census datasets together
-census_df = oa_tenure_df
+# census_df = oa_tenure_df
+census_df = age_bands_df
 for df in [
-    age_bands_df,
+    # age_bands_df,
     disability_df,
     economic_status_df,
     hours_worked_df,
@@ -1022,17 +1022,116 @@ for df in [
     census_df = census_df.join(df, how="left", on="oa_code")
 
 # %%
+# Load output area geospatial boundaries
+oa_boundaries_gdf = gpd.read_file(
+    "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/Output_Areas_2021_EW_BFE_V9_6122634609897870819/OA_2021_EW_BFE_V9.shp"
+)
+
+# Assign UPRNs to their output areas
+uprns_with_oa_df = plymouth_residential_uprns_gdf.sjoin(
+    oa_boundaries_gdf[["OA21CD", "LSOA21CD", "geometry"]],
+    how="left",
+    predicate="intersects",
+).drop(columns="index_right")
+
+# Join output areas to features
+features_df = features_df.join(
+    pl.from_pandas(uprns_with_oa_df[["UPRN", "OA21CD", "LSOA21CD"]]),
+    how="left",
+    on="UPRN",
+).with_columns(pl.col("TENURE").replace("unknown", None))
+
+# %%
+# CALCULATE PROPORTIONS OF TENURE TYPES FOR NEIGHBOURS WITHIN 100m RADIUS FOR EACH UPRN
+nn_tenure_df = features_df.select(
+    [
+        "UPRN",
+        "TENURE",
+        "OA21CD",
+    ]
+).join(
+    os_uprn_df.select(["UPRN", "X_COORDINATE", "Y_COORDINATE"]), how="left", on="UPRN"
+)
+
+# Calculate distance from nearest neighbour for all UPRNs
+# Get coordinates of each UPRN
+nn_tenure_df = lat_lon.generate_gdf_uprn_coords(
+    nn_tenure_df, usecols=["UPRN", "TENURE", "X_COORDINATE", "Y_COORDINATE"]
+)
+
+# Convert coordinates to array
+coords = np.array(nn_tenure_df.geometry.map(lambda p: [p.x, p.y]).tolist())
+
+# Find all neighbours within 100m radius of each UPRN
+knn = NearestNeighbors(radius=100, algorithm="kd_tree").fit(coords)
+knn_dist, knn_idx = knn.radius_neighbors(coords)
+
+# Get the tenures of the neighbours within 100m
+nn_tenure_df["nearest_tenures"] = [
+    nn_tenure_df.TENURE.values[knn_idx[i]] for i in range(0, len(knn_idx))
+]
+
+# Get the UPRNs of the neighbours within 100m
+nn_tenure_df["nearest_UPRNs"] = [
+    nn_tenure_df.UPRN.values[knn_idx[i]] for i in range(0, len(knn_idx))
+]
+
+nn_tenure_df = pd.DataFrame(nn_tenure_df).explode(
+    column=["nearest_tenures", "nearest_UPRNs"]
+)
+nn_tenure_df = (
+    nn_tenure_df[
+        # Filter out neighbours which are 'self' and filter out any neighbours with no tenure
+        (nn_tenure_df["UPRN"] != nn_tenure_df["nearest_UPRNs"])
+        & (nn_tenure_df["nearest_tenures"].notnull())
+        # Get the counts of neighbours with each tenure type per UPRN
+    ]
+    .groupby(["UPRN", "nearest_tenures"])
+    .agg(
+        count=pd.NamedAgg(column="UPRN", aggfunc="count"),
+    )
+    .reset_index()
+)
+
+# Calculate the proportion of neighbours with each tenure type
+total_tenure_df = nn_tenure_df.groupby("UPRN").agg(
+    total=pd.NamedAgg(column="count", aggfunc="sum")
+)
+nn_tenure_df = nn_tenure_df.merge(total_tenure_df, how="left", on="UPRN")
+nn_tenure_df["prop_neighbours"] = nn_tenure_df["count"] / nn_tenure_df["total"]
+
+nn_tenure_df = pl.from_pandas(
+    nn_tenure_df.pivot(
+        index="UPRN", columns="nearest_tenures", values="prop_neighbours"
+    )
+    .fillna(0)
+    .reset_index()
+).rename(
+    {
+        "owner-occupied": "prop_nn_owner_occupied",
+        "rental (private)": "prop_nn_private_rental",
+        "rental (social)": "prop_nn_social_rental",
+    }
+)
+
+# ----------------------------------------------------------------------------------- #
 # CREATE DATAFRAME WITH FEATURES TO TRAIN AND TEST MODEL
 model_df = (
-    epc_df.select(
+    features_df.select(
         [
             "UPRN",
             "property_type",
             "TENURE",
-            "oa21",
+            "OA21CD",
         ]
     )
-    .join(census_df, how="left", left_on="oa21", right_on="oa_code")
+    .join(census_df, how="left", left_on="OA21CD", right_on="oa_code")
+    .join(
+        os_uprn_df.select(["UPRN", "X_COORDINATE", "Y_COORDINATE"]),
+        how="left",
+        on="UPRN",
+    )
+    .join(nn_tenure_df.drop("prop_nn_private_rental"), how="left", on="UPRN")
     .join(
         features_df.select(["UPRN", "in_cons_area", "in_listed_building"]),
         how="left",
@@ -1043,7 +1142,7 @@ model_df = (
         pl.col("property_type") != "Caravan or other mobile or temporary structure",
         pl.col("TENURE") != "unknown",
     )
-    .drop(["oa21", "prop_aged_35_to_64"])
+    .drop(["OA21CD", "prop_aged_35_to_64"])
     .to_pandas()
 )
 
@@ -1065,10 +1164,21 @@ X_train, X_test, y_train, y_test = train_test_split(
     X, y, test_size=0.25, random_state=1
 )
 
+print(y_train["TENURE"].value_counts(normalize=True))
+print(y_test["TENURE"].value_counts(normalize=True))
+
 # Fit and predict with K-nearest neighbours classifier model
-classifier = KNeighborsClassifier(n_neighbors=10)
+# classifier = KNeighborsClassifier(n_neighbors=10)
+# classifier.fit(X_train, y_train)
+# print("K Nearest Neighbour classifier accuracy score:")
+# print(classifier.score(X_test, y_test))
+# y_pred = classifier.predict(X_test)
+# y_pred_prob = classifier.predict_proba(X_test)
+
+# Fit and predict with Random Forest classifier model
+classifier = RandomForestClassifier(class_weight="balanced")
 classifier.fit(X_train, y_train)
-print("K Nearest Neighbour classifier accuracy score:")
+print("\nRandom Forest classifier accuracy score:")
 print(classifier.score(X_test, y_test))
 y_pred = classifier.predict(X_test)
 y_pred_prob = classifier.predict_proba(X_test)
@@ -1093,14 +1203,8 @@ results_df = (
     )
 )
 
-# classifier = RandomForestClassifier()
-# classifier.fit(X_train, y_train)
-# print("\nRandom Forest classifier accuracy score:")
-# print(classifier.score(X_test, y_test))
-
 # ----------------------------------------------------------------------------------- #
 # PLOT RESULTS
-
 # Plot probability of predictions based on prediction outcome
 plot_df = results_df.group_by(["correct_prediction", "max_prob"]).agg(
     pl.col("predicted").count().alias("n_predictions")
@@ -1117,7 +1221,21 @@ plot_df = (
     .to_pandas()
 )
 
-plot_df.index = plot_df["max_prob"]
+bins = [0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1]
+labels = [
+    "(0.3-0.4]",
+    "(0.4-0.5]",
+    "(0.5-0.6]",
+    "(0.6-0.7]",
+    "(0.7-0.8]",
+    "(0.8-0.9]",
+    "(0.9-1]",
+]
+plot_df["max_prob_group"] = pd.cut(
+    plot_df["max_prob"], bins=bins, labels=labels, right=True
+)
+
+plot_df = plot_df.groupby("max_prob_group").sum()
 plot_df = plot_df.drop("max_prob", axis=1)
 
 plot_df.plot(kind="bar")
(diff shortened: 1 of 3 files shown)


asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+    output_areas,
+)
+from asf_heat_pump_suitability.analysis.flats_on_fossils.features import fuel_type
+from asf_heat_pump_suitability.pipeline.reweight_epc import prepare_sample
+
+# %% [markdown]
+# ## Load data
+
+# %%
+fiona.listlayers("s3://asf-heat-pump-suitability/source_data/opmplc_gb.gpkg")
```
sofiapinto, line 57: what does opmplc_gb stand for?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+    "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/OS OpenMap Local (ESRI Shape File) SX/data/SX_RailwayStation.shp"
+)
+
+# --------------------------------------------------------------------------#
+
+print("\nLOADING PROPERTY-LEVEL FEATURE DATASETS...")
+print("\nLoading listed buildings...")
+listed_buildings_gdf = gpd.read_file(
+    "../spatial_clustering/data/National_Heritage_List_for_England_NHLE_v02_VIEW_-464524051049198649/Listed_Building_points.shp"
+)
```
sofiapinto, line 96: needs to be updated later to read from S3

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+
+print("\nLoading building conservation areas...")
+cons_areas_gdf = gpd.read_file(
+    "../spatial_clustering/data/conservation-area (1).geojson"
+)
```
sofiapinto, line 102: needs to be updated later to read from S3

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+# Identify building types which are unlikely to contain residential dwellings
+no_residential_overlap_building_types = [
+    "Port Consisting of Docks and Nautical Berthing",
+    "Fire Station",
+    "Hospital",
+    "Non State Secondary Education",
+    "Secondary Education",
+    "Higher or University Education",
+    "Primary Education",
+    "Post Office",
```
sofiapinto, line 187: I think this one will lead to some false negatives. It's very common to have flats above post offices. But if there's no other way to remove the false positives, then it's better to have some false negatives. Just worth taking note that that's the case.

I guess that there are some UPRNs we could keep, in case they are non-null UPRNs in EPC - since those UPRNs will be only domestic properties.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+    layer="heat-network-zone-map-Plymouth",
+)
+
+print("\nLoading outdoor space...")
+outdoor_space_df = pl.read_parquet(
+    "s3://asf-heat-pump-suitability/outputs/2023Q4/gardens/20250224_2023_Q4_EPC_garden_size_estimates_EWS_deduplicated.parquet"
+)
+
+print("\nLoading off gas postcodes...")
+off_gas_list = off_gas.process_off_gas_data()
```
sofiapinto, line 113: Are these used as postcodes with properties off gas or with the assumption that all properties in those postcodes are off gas? If the second one, I think that's probably likely in rural areas but not necessarily otherwise. Am I wrong?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+    ],
+)
+
+print("\nAdding EPC property type...")
+epc_df = prepare_sample.add_col_property_type(epc_df)
+
+# -------------------------------------------------------------------------#
+
+print("\nLOADING DATASETS TO FILL MISSING DATA")
+print("\nLoading OS Code-Point Open...")
```
sofiapinto: when you refactor this might be nice to say that this is an open dataset of all the current postcode units in Great Britain.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+# Convert UPRN coordinates to geopoints
+os_uprn_gdf = lat_lon.generate_gdf_uprn_coords(os_uprn_df)
+
+# Get polygon of Plymouth LA
+plymouth_polygon = la_boundaries_gdf[
+    la_boundaries_gdf["LAD23NM"].str.contains("Plymouth")
+]["geometry"].values[0]
+
+# Filter UPRN and OpenMap datasets to Plymouth only
+os_uprn_plymouth_gdf = os_uprn_gdf[os_uprn_gdf.within(plymouth_polygon)]
```
sofiapinto, line 155: so does this `within` always compare automatically identify the geometry fields to check if the points are contained within the polygon?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+# View types of building
+os_openmap_importantbuildings_plymouth_gdf.groupby("BUILDGTHEM")[
+    "CLASSIFICA"
+].value_counts(dropna=False)
+
+# %%
+# Identify all UPRNs in a building - these contain residential addresses
+uprns_in_buildings = os_openmap_buildings_plymouth_gdf.sjoin(
+    os_uprn_plymouth_gdf, how="inner", predicate="contains"
+)["UPRN"].tolist()
```
sofiapinto, line 176: why aren't these just all UPRNs? don't all UPRNs have a building ID?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+    & (os_uprn_plymouth_gdf["UPRN"].isin(uprns_in_buildings))
+]
+
+# Filter building footprints to only those which contain residential UPRNs
+plymouth_residential_buildings_gdf = os_openmap_buildings_plymouth_gdf.sjoin(
+    plymouth_residential_uprns_gdf, how="inner", predicate="contains"
+).drop_duplicates(subset=["ID"])
+plymouth_residential_buildings_gdf["building_area_m2"] = (
+    plymouth_residential_buildings_gdf.area
+)
```
sofiapinto, line 232: very cool that this can be done directly with gdfs :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+        listed_buildings_gdf, how="inner", max_distance=10
+    )["UPRN"]
+    .unique()
+    .tolist()
+)
+
+# UPRNs in building conservation areas
+cons_area_uprns = plymouth_residential_uprns_gdf.sjoin(
+    cons_areas_gdf.to_crs(epsg=27700), how="inner", predicate="intersects"
+)["UPRN"].tolist()
```
sofiapinto, line 265: Why did you have to `.to_crs()` in this specific example before doing the sjoin?

Worth adding a documentation note that EPSG:27700 is the British National Grid projection, used for mapping in Great Britain. I didn't know - had to google it.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+
+# Reformat UPRN for join
+outdoor_space_df = outdoor_space_df.with_columns(
+    pl.col("UPRN").str.replace(".0", "", literal=True).cast(pl.Int64).alias("UPRN")
+)
+
+# UPRNs in HNs
+uprns_in_hn = plymouth_residential_uprns_gdf.sjoin(
+    hn_zones_gdf.drop(columns=["index_right"]), predicate="within"
```
sofiapinto, line 269: why did you have to drop this column?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+len(listed_building_uprns) / len(plymouth_residential_uprns_gdf) * 100
+
+# %%
+# Standardise postcode column - remove space and set to upper case
+epc_df = output_areas.standardise_col_postcode(epc_df, pcd_col="POSTCODE")
+epc_df = epc_df.with_columns(
+    # Standardise UPRN
+    pl.col("UPRN")
+    .str.replace(".0", "", literal=True)
+    .cast(pl.Int64, strict=False)
```
sofiapinto: just to flag that you have a `strict=False` here but not above for `outdoor_space_df`. Is this because of our processing of UPRNs in EPC? So you do this so that any UPRN that we create using the building ID and address will be made null here?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+    pl.from_pandas(plymouth_residential_uprns_gdf[["UPRN"]])
+    .with_columns(
+        pl.when(pl.col("UPRN").is_in(listed_building_uprns))
+        .then(True)
+        .otherwise(False)
+        .alias("in_listed_building"),
+        pl.when(pl.col("UPRN").is_in(cons_area_uprns))
+        .then(True)
+        .otherwise(False)
+        .alias("in_cons_area"),
```
sofiapinto:
```suggestion
        .alias("in_conservation_area"),
```
just makes it more readable

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+        pl.when(pl.col("UPRN").is_in(uprns_in_hn))
+        .then(True)
+        .otherwise(False)
+        .alias("in_hn"),
```
sofiapinto:
```suggestion
        .alias("in_heat_network_zone"),
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+        n_households_df.select(["mnemonic", "2021"]),
+        how="left",
+        left_on="lsoa",
+        right_on="mnemonic",
+    )
+    .rename({"UPRN": "EPC_records_count", "2021": "census_count"})
+    .with_columns(
+        (pl.col("census_count") - pl.col("EPC_records_count")).alias("count_diff")
+    )
+)
```
sofiapinto, line 392: `polars` is so elegant! ✨

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+# Remove space from postcode in code point data
+code_point_df["POSTCODE"] = code_point_df["postcode"].str.replace(" ", "")
+
+# Find nearest postcode point for all UPRNs within a 1km radius
+nearest_postcode_df = pl.from_pandas(
+    plymouth_residential_uprns_gdf.sjoin_nearest(
+        code_point_df[["POSTCODE", "geometry"]],
+        how="left",
+        max_distance=1000,
+        distance_col="distance_to_postcode_m",
```
sofiapinto: Worth adding a comment here to say that this saves the distances to a column named `distance_to_postcode_m`.

Also, what does the m stand for? meters?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+
+# %%
+# Identify UPRNs in listed buildings
+listed_building_uprns = (
+    plymouth_residential_uprns_gdf.sjoin_nearest(
+        listed_buildings_gdf, how="inner", max_distance=10
+    )["UPRN"]
+    .unique()
+    .tolist()
+)
```
sofiapinto, line 260: Trying to understand exactly how this works. Does this give us the closest listed building within a 10m radius of a residential UPRN? And so, when no match is found, because we're using "inner", then only residential UPRNs with a match will be considered - hence why we can call this listed_buildings_uprns?

What is the 10m based on? I'm assuming that `max_distance=10` corresponds to 10m. Let me know if i'm wrong.

I think when you refactor this (not a priority now I understand), it would be useful to add some documentation.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+
+# %% [markdown]
+# ## Fill missing data - off gas
+
+# %%
+# FIND NEAREST POSTCODE TO EACH UPRN
+# Remove space from postcode in code point data
+code_point_df["POSTCODE"] = code_point_df["postcode"].str.replace(" ", "")
+
+# Find nearest postcode point for all UPRNs within a 1km radius
```
sofiapinto:
```suggestion
# Find nearest postcode point within a 1km radius of each UPRN
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+        .otherwise(False)
+        .alias("off_gas")
+    )
+    .group_by("UPRN")
+    .agg(
+        # Get number of postcodes per UPRN and modal on/off gas value
+        pl.col("POSTCODE").alias("postcodes"),
+        pl.col("POSTCODE").len().alias("n_postcodes"),
+        pl.col("off_gas").mode().alias("modal_off_gas"),
+        pl.col("distance_to_postcode_m").first().alias("distance_to_postcode_m"),
```
sofiapinto, line 450: why first?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+        max_distance=1000,
+        distance_col="distance_to_postcode_m",
+    ).drop(columns="index_right")[["UPRN", "POSTCODE", "distance_to_postcode_m"]]
+)
+
+# ----------------------------------------------------------------------------------- #
+# ASSIGN OFF GAS LABEL
+
+# Label all UPRNs with on/off gas where possible
+off_gas_df = (
```
sofiapinto, line 435: I believe this is probably the best we can do now as an off gas label, but I think the underlying assumption that close by UPRNs are also on or off gas is a bit too strong. But as long as the team agreed to this for now, then all good.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+        (pl.col("POSTCODE").fill_null(pl.col("POSTCODE_nearest"))).alias(
+            "use_postcode"
+        ),
+        pl.col("off_gas").fill_null(pl.col("off_gas_nearest")).alias("filled_off_gas"),
+    )
+)
+
+# Fill missing off-gas label
+features_df = features_df.with_columns(
+    pl.when((pl.col("off_gas").is_null()) & (pl.col("fuel_type") == "gas"))
```
sofiapinto, line 491: where does this fuel type data come from?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+    .then(False)
+    .otherwise(pl.col("off_gas"))
+    .alias("off_gas")
+)
+
```
sofiapinto, line 496: after this, I think it's worth removing the columns you don't need anymore (after evaluating the accuracy of results of course) as they're not needed in the final dataset. I think it's the ones below if I'm not mistaken:
- POSTCODE_nearest
- off_gas_nearest
- distance_to_postcode_nearest
- use_postcode
- filled_of_gas

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+    ],
+    bins=50,
+    density=True,
+    alpha=0.5,
+    label="Postcode matches EPC",
+)
+plt.title("Distribution of distance from UPRN to postcode point (m)")
+plt.xlabel("Distance between UPRN and postcode point (m)")
+plt.legend()
+plt.show()
```
sofiapinto, line 543: I think I got a little lost, what exactly are we trying to assess here?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+    )
+    .with_columns(
+        ((pl.col("n_UPRNs") - pl.col("null_count")) / pl.col("n_UPRNs")).alias(
+            "non_null_proportion"
+        )
+    )
+)
+
+print(
+    "UPRNs that are missing a property type and are in a building with multiple property type labels:"
```
sofiapinto:
```suggestion
    "Proportion of UPRNs that are missing a property type and are in a building with multiple property type labels:"
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+)
+print(
+    property_type_df.filter(pl.col("nunique_property_types") > 2)["null_count"].sum()
+    / len(plymouth_residential_uprns_gdf)
+)
+
+print(
+    "\nUPRNs that are missing a property type and are in a building with multiple MODAL property type labels:"
```
sofiapinto:
```suggestion
    "\nProportion of UPRNs that are missing a property type and are in a building with multiple MODAL property type labels:"
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+)
+print(
+    property_type_df.filter(pl.col("n_modal_property_type") > 1)["null_count"].sum()
+    / len(plymouth_residential_uprns_gdf)
+)
+
+print(
+    "\nUPRNs that are missing a property type and are in a multi-property building with NO property type labels:"
```
sofiapinto:
```suggestion
    "Proportion of UPRNs that are missing a property type and are in a multi-property building with NO property type labels:"
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+    plymouth_residential_uprns_gdf.sjoin(
+        plymouth_residential_buildings_gdf[["ID", "geometry", "building_area_m2"]],
+        how="left",
+        predicate="within",
+    )
+    .rename(columns={"ID": "building_id"})
+    .drop(columns=["index_right", "geometry"])
+).join(features_df.select(["UPRN", "property_type"]), how="left", on="UPRN")
+
+# Identify building IDs which have UPRNs missing property type information
```
sofiapinto:
```suggestion
# Identify building IDs which have one or more UPRNs missing property type information
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+        pl.col("UPRN").count().alias("n_UPRNs"),
+        pl.col("property_type").alias("property_types"),
+        pl.col("property_type").null_count().alias("null_count"),
+        pl.col("property_type").n_unique().alias("nunique_property_types"),
+        pl.col("property_type").mode().alias("modal_property_type"),
+        pl.col("property_type").mode().len().alias("n_modal_property_type"),
+    )
+    .with_columns(
+        ((pl.col("n_UPRNs") - pl.col("null_count")) / pl.col("n_UPRNs")).alias(
+            "non_null_proportion"
```
sofiapinto, line 583:
```suggestion
            "prop_uprns_non_null_property_type"
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+# Convert coordinates to array
+coords = np.array(nn_dist_gdf.geometry.map(lambda p: [p.x, p.y]).tolist())
+
+# Find two nearest neighbours to each UPRN
+knn = NearestNeighbors(n_neighbors=2, algorithm="kd_tree").fit(coords)
+knn_dist, knn_idx = knn.kneighbors(coords)
+
+# Get distance of second nearest neighbour (the nearest neighbour is itself)
+# The second neighbour will only be itself if there is another neighbour that is 0m away
+nn_dist_gdf[list(map("neighbour_{}".format, range(1, 2)))] = (
```
sofiapinto, line 688: The result of `list(map("neighbour_{}".format, range(1, 2)))` is simply "neighbour_1"

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+
+# Convert coordinates to array
+coords = np.array(nn_dist_gdf.geometry.map(lambda p: [p.x, p.y]).tolist())
+
+# Find two nearest neighbours to each UPRN
+knn = NearestNeighbors(n_neighbors=2, algorithm="kd_tree").fit(coords)
+knn_dist, knn_idx = knn.kneighbors(coords)
+
+# Get distance of second nearest neighbour (the nearest neighbour is itself)
+# The second neighbour will only be itself if there is another neighbour that is 0m away
```
sofiapinto, line 687: you mean the second neighbour _might_ be itself if there the other neighbour is also 0m away, right?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+        "Semi-detached",
+        "Terraced (including end-terrace)",
+        "Flat, maisonette or apartment",
+        "Caravan or other mobile or temporary structure",
+    ]
+]
+table_df / table_df.sum() * 100
+
+# %%
+table_df.div(table_df.sum(axis=1), axis=0) * 100
```
sofiapinto, line 743: could you please add a line of doc for each of these before you refactor

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+        alpha=0.5,
+        label=f"{property_type},\nN={len(density_df.filter(pl.col('property_type') == property_type))}",
+    )
+plt.title("Building footprint area in different property types")
+plt.xlabel("Building footprint area (m2)")
+plt.xlim(0, 500)
+plt.legend()
+
+# %%
+# CALCULATE NEAREST NEIGHBOUR DISTANCE
```
sofiapinto, line 670:
```suggestion
# CALCULATE NEAREST NEIGHBOUR DISTANCE - for properties where property type is known and unique (no multiple property types)
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+    ]
+]
+table_df / table_df.sum() * 100
+
+# %%
+table_df.div(table_df.sum(axis=1), axis=0) * 100
+
+# %%
+# ----------------------------------------------------------------------------------- #
+# CALCULATE NEAREST NEIGHBOUR DISTANCE
```
sofiapinto:
```suggestion
# CALCULATE NEAREST NEIGHBOUR DISTANCE - to assign property type labels to all UPRNs
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+    alpha=0.5,
+    label="Postcode matches EPC",
+)
+plt.title("Distribution of distance from UPRN to postcode point (m)")
+plt.xlabel("Distance between UPRN and postcode point (m)")
+plt.legend()
+plt.show()
+
+# %% [markdown]
+# ## Fill missing data - property type
```
sofiapinto, line 546: This whole section is very smart :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+
+# %% [markdown]
+# ## Fill missing data - tenure type
+
+# %%
+# Load census data
+print("Loading age band data...")
+age_bands_df = pl.read_csv(
+    "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/2021Census_age_bands_OA_plymouth.csv",
+    skip_rows=4,
```
sofiapinto, line 862: I didn't open the original datasets, without `skip_rows` so I haven't checked if this is correct. For age bands `skip_rows=4`, while `skip_rows=6` for all other datasets. Hopefully this is correct, but wanted to flag in case it was a typo.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+        ).alias("prop_owner_occupied"),
+        (pl.col("Social rented") / pl.col("Total: All households")).alias(
+            "prop_social_rented"
+        ),
+    )
+    .select(["oa_code", "prop_owner_occupied", "prop_social_rented"])
+)
+
+# Join census datasets together
+census_df = oa_tenure_df
```
sofiapinto:
```suggestion
census_df = oa_tenure_df.clone()
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+        [
+            "UPRN",
+            "property_type",
+            "TENURE",
+            "oa21",
+        ]
+    )
+    .join(census_df, how="left", left_on="oa21", right_on="oa_code")
+    .join(
+        features_df.select(["UPRN", "in_cons_area", "in_listed_building"]),
```
sofiapinto:
```suggestion
        features_df.select(["UPRN", "in_conservation_area", "in_listed_building"]),
```
if you decide to change above

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+            "1": y_pred_prob[:, 1],
+            "2": y_pred_prob[:, 2],
+        }
+    )
+    .rename({k: v for k, v in zip(["0", "1", "2"], classifier.classes_)})
+    .with_columns(
+        (pl.col("actual") == pl.col("predicted")).alias("correct_prediction"),
+        max_prob=pl.concat_list(
+            "owner-occupied", "rental (private)", "rental (social)"
+        ).list.max(),
```
sofiapinto, line 1227: Obviously not for now, but for when this is refactored: I think this part requires a bit of documentation, it's not immediate obvious what is happening

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1182 @@
...
+)
+
+print(
+    "\nUPRNs that are missing a property type and are in a multi-property building with NO property type labels:"
+)
+print(
+    property_type_df.filter(
+        pl.col("n_UPRNs") > 1, pl.col("non_null_proportion") == 0
+    ).height
+    / len(plymouth_residential_uprns_gdf)
```
sofiapinto, line 611: why are we using height here?

sofiapinto (commented):
@crispy-wonton amazing work! There's so much knowledge of handling gdf's, polars and of UK open datasets! I learnt loads from reviewing this PR. And so much work in this PR. Thank you.

I've left questions where I need clarifications, but haven't found any bugs or problems with the logic.

We have since discussed improvements on the modelling side, so I didn't leave any comments on that part of the code. Feel free to add your newest code and I'm happy to review again and think of improvements on the model/data used.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: sorry - should have added notation! it's the Open Map Local dataset for GB and the layers are all the different geospatial datasets contained in the file :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: Thanks!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: From my understanding, it's the whole postcode doesn't have a gas connection. See [Xoserve's documentation of the data](https://www.xoserve.com/help-centre/supply-points-metering/supply-point-administration-spa/). I've lifted the below from it. I think this means we can assume no properties in the postcode have gas supply.

> The Off Gas Postcode Register is a list of postcodes where Xoserve holds no record of an active gas connection by either large or small gas transporters.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: `within` will check if the left geometry (point or polygon, it doesnt matter) is within the right geometry. If you want to do the reverse, i.e. check if the right geometry is in the left one, you can use `left.contains(right)`.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: I don't think so? The OS UPRN dataset only contains UPRNs and their coordinates. There's no building ID as far as I know. The UPRNs represent the properties / addresses. Is that what you mean?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: Yeh I was thinking this too when I wrote this list. I think in a city it's also reasonably common to have residential properties above certain types of education centres, places of worship, and perhaps sports and leisure centres. The methodology to identify the residential properties generally could do with some improvement.

I think your second point is a great point - I think I need to revisit the UPRN to EPC join so that I retain any EPC UPRNs that might have been filtered out here.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: This is the methodology we used in the listed buildings processing in the pipeline. There's some bug in it at the moment so I just copied the code here and changed it so it worked for Plymouth.

You're correct in your interpretation. The 10m is based on a piece of analysis I did to determine the threshold distance for a UPRN to be considered to be in the listed building using a labelled dataset.

This needs updating to use polygons instead of points which I think are available for Plymouth listed buildings.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: The CRS is the coordinate reference system. Geospatial datasets need to be in the same CRS to interact e.g. with spatial joins or overlays etc. The `cons_area_gdf` is in a different CRS (longitude and latitude) instead of British National Grid (BNG) so the `to_crs()` converts it to BNG.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: For some reason, `geopandas` always adds an `index_right` column when you conduct a join. It's just a duplicate unnecessary column :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: Yeh I believe this is because the `outdoor_space_df` has only valid UPRNs. This is because the dataframe is created by joining garden polygons with UPRNs which have geospatial coordinates - these will only be valid UPRNs. 

You are correct in your interpretation :) Because the EPC data has UPRNs that were generated by DAPs pipeline (i.e. the ones combining building ref number and address), there are errors in conversion to `Int64` here. Setting `strict=False` will cast those UPRNs as null.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: I agree!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: Yes m is for metres :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: I agree, it doesnt work well. Open to other ideas :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: I actually don't know why I did this. Will need to check 😆

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: The processing in lines 277-284

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: yep, will clean the whole dataset at the end before saving :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: I was looking at the distribution of distance of UPRNs from the assigned postcode using the methodology here and splitting it into correct / incorrect postcode assignments (based on whether the assigned postcode matched the one in EPC or not). Just wanted to see if there was any useful pattern.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: height is equivalent to the `len` of the df.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: yeh I copied this line from stack overflow and just edited it to keep one neighbour :D

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: I think it's correct, I checked all the dfs manually before creating these. I remember there were different lengths of headers across the diff files.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: this is just for renaming purposes basically as `oa_tenure_df` isnt used again. also I dont think deep copying dfs is that important in `polars` as there's not much (if any?) in place operations.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: Good point!!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
sofiapinto: interesting, my bad then. Glad this is the case!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
sofiapinto: Yep - what I meant was that you didn't have to pass any information to .within() about the columns in `os_uprn_gdf` and `plymouth_polygon` that hold the geometry information (since there are multiple columns in this gdf). My guess was that it was done automatically (since only one of the columns in each gdf hold a geometry type column).

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
sofiapinto: Sorry I didn't explain myself well. What I don't understand is why the number of UPRNs in `uprns_in_buildings` is smaller than the number of UPRNs in `os_uprn_plymouth_gdf`. Shouldn't `os_openmap_buildings_plymouth_gdf` hold all of the information about buildings in Plymouth?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
sofiapinto: Great that this threshold is based in a previous analysis. Worth writing that down when you find time.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
sofiapinto: I think it's okay for the purpose of this prototype/workshop with Plymouth. Just worth noting down as something to ideate on and improve later.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
sofiapinto: sorry - I got confused with the domain and thought we were looking at the height of the buildings. my bad haha

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1863 @@
...
+print("\nLOADING PROPERTY-LEVEL FEATURE DATASETS...")
+print("\nLoading listed buildings...")
+listed_buildings_gdf = gpd.read_file(
+    "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/National_Heritage_List_for_England_NHLE_v02_VIEW_-464524051049198649/Listed_Building_points.shp"
+)
+
+print("\nLoading building conservation areas...")
+cons_areas_gdf = gpd.read_file(
+    "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/conservation-area (1).geojson"
+)
```
lizgzil, line 102: is this the new dataset (https://github.com/nestauk/asf_heat_pump_suitability/issues/167)?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1863 @@
...
+import folium
+
+# %% [markdown]
+# ## Load data
+
+# %%
+fiona.listlayers("s3://asf-heat-pump-suitability/source_data/opmplc_gb.gpkg")
+
+# %%
+print("LOADING DATASETS TO GET RESIDENTIAL UPRNS FOR PLYMOUTH...")
```
lizgzil, line 60: I can see some of the data you read from `exploration/spatial_clustering_plymouth` is particular to that part of the country (I guess the `SX` prefixed ones), but are some of them for the whole UK? (e.g. listed buildings and conservation zones). Nothing to change - but I'm curious about your logic to include data in this `exploration/spatial_clustering_plymouth` folder vs the `source_data` folder? Perhaps just for speed during exploration?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
lizgzil: I think `pyarrow` needs adding to the requirements. I needed to do `pip install pyarrow` to get off gas data to load.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
lizgzil: When I was plotting I think there were some times when a UPRN coordinate is just outside the building polygon. (see example of the left of slide 9 [here](https://docs.google.com/presentation/d/1Vcc_xnj8tYwYXgklo7zB8CztmrqTxDBHOnjBF-gwDeE/edit?slide=id.g3765b9f7180_0_72#slide=id.g3765b9f7180_0_72)). So when improvements are made - I guess we need to see if we can account for these cases.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -1022,17 +1022,116 @@
...
+# Get coordinates of each UPRN
+nn_tenure_df = lat_lon.generate_gdf_uprn_coords(
+    nn_tenure_df, usecols=["UPRN", "TENURE", "X_COORDINATE", "Y_COORDINATE"]
+)
+
+# Convert coordinates to array
+coords = np.array(nn_tenure_df.geometry.map(lambda p: [p.x, p.y]).tolist())
+
+# Find all neighbours within 100m radius of each UPRN
+knn = NearestNeighbors(radius=100, algorithm="kd_tree").fit(coords)
```
lizgzil, line 1088: at some point it'd be good to play around with this threshold to see what difference it makes to the results

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -1128,37 +1246,250 @@
...
+# %%
+from sklearn.metrics import confusion_matrix
+import seaborn as sns
+
+tenure_mapping = {"owner-occupied": 0, "rental (private)": 1, "rental (social)": 2}
+y_true = np.array(y_test["TENURE"].map(tenure_mapping))
+
+y_pred_arr = np.copy(y_pred)
+for k, v in tenure_mapping.items():
+    y_pred_arr[y_pred == k] = v
```
lizgzil: Not sure why, but this wasn't working as expected for me - `y_pred_arr` wasn't changed to numbers.

`y_pred_arr = [tenure_mapping.get(v) for v in y_pred]` worked though.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1863 @@
...
+
+# One hot encode property type data
+onehotenc_df = pd.DataFrame(
+    enc.transform(
+        pd_features_df[["use_property_type"]].rename(
+            columns={"use_property_type": "property_type"}
+        )
+    ).toarray()
+)
+onehotenc_df.columns = list(col.lower().replace(" ", "_") for col in enc.categories_[0])
```
lizgzil, line 1515: would be good to add all you processing steps to a function so you are sure you cleaned your training data in the same way as you cleaned the data for prediction. Looks consistent to me now though, so don't think there are bugs

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -1173,10 +1504,360 @@
...
+    pl.col("TENURE").fill_null(pl.col("predicted_tenure")).alias("use_tenure"),
+)
+
+# %% [markdown]
+# ## Cluster UPRNs
+
+# %%
+# Cluster UPRNs on distance only
+model = HDBSCAN(
+    min_cluster_size=5,
```
lizgzil, line 1748: Not sure on this - should it be 1, but it's fine for now. Will need to think it through - but just noting!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -1173,10 +1504,360 @@
...
+    cluster_selection_epsilon=20,
+    algorithm="balltree",
```
lizgzil: FYI my version of HDBSCAN required this to be `algorithm="ball_tree". I have scikit-learn=1.7.1

lizgzil (dismissed):
Thanks so much for all this @crispy-wonton !!

The code worked well and was great to see such elegant polars mastery. 

My version of scikit-learn (1.7.1) made it so that I needed to change the code, but I see now you wrote in the description that you might need 1.3 anyway!

I looked over it all since I needed to run it all, but I focused on the tenure model and the extra code from adding the IMD deciles in my review. Think it looks great - no big changes at all - mostly notes.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: Oh I see what you mean. Huh, you're right, I didn't notice that I haven't specified the geometry column here but in general you can only have one active geometry column in a geodataframe so presumably that's why it works although I can't find anything in the documentation about this.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: Yes, what @lizgzil said is true and also not all UPRNs are actually located within buildings. This is because they represent addressable locations, not properties. So things like bus stops and post boxes have a UPRN.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: Yep, it's already in a notebook in the repo :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: All the nearest postcodes (if there are multiple for a UPRN) will be the same distance so we just select the first to keep one.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: Yeh it was the latest version of that data at the time.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: Yes it's for download speed mainly. In cases where we didn't already have the dataset in our pipeline, I chose to use Plymouth-specific datasets because they are faster to load. Have added a note :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: Thanks! I've updated it as that's more readable anyway :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: Yeh that's a good idea, will definitely do that when refactoring or refining the model!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: i updated the requirements to have sklearn 1.7 and then updated the couple of lines you've flagged.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: yeh i think the clustering can be iterated in a separate piece of work :)

asf_heat_pump_suitability/pipeline/prepare_features/garden_size.py
```diff
@@ -99,7 +99,12 @@ def generate_gdf_garden_size(
     """
     building_size = (
         intersections_gdf.groupby("NATIONALCADASTRALREFERENCE")[
-            ["building_id", "building_intersection_area_m2", "height"]
+            [
+                "building_id",
+                "building_intersection_area_m2",
+                # "height",
```
sofiapinto: is this supposed to be commented out now that we're soon going to merge?

asf_heat_pump_suitability/pipeline/prepare_features/garden_size.py
```diff
@@ -110,10 +115,10 @@ def generate_gdf_garden_size(
...
                 "sum",
             ),  # get total building area on land parcel
-            max_building_height=(
-                "height",
-                "max",
-            ),  # get max building height on land parcel
+            # max_building_height=(
+            #     "height",
+            #     "max",
+            # ),  # get max building height on land parcel
```
sofiapinto: same as above

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -75,12 +84,12 @@
 print("\nLOADING PROPERTY-LEVEL FEATURE DATASETS...")
 print("\nLoading listed buildings...")
 listed_buildings_gdf = gpd.read_file(
-    "../spatial_clustering/data/National_Heritage_List_for_England_NHLE_v02_VIEW_-464524051049198649/Listed_Building_points.shp"
```
sofiapinto: is there anywhere else we need to add documentation about these new sources? just flagging, in case we need to create a separate issue for this.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1867 @@
...
+
+# %%
+# USE MODEL TO PREDICT TENURE TYPE
+features = [
+    "prop_aged_16_to_35",
+    "prop_aged_65_and_over",
+    "prop_disabled_residents",
+    "prop_economically_inactive_adults",
+    "prop_employed_adults_working_full_time",
+    "prop_students",
```
sofiapinto, line 1474: can we have all 3? aren't they too correlated?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1867 @@
...
+    )
+    .fillna(0)
+    .reset_index()
+).rename(
+    {
+        "owner-occupied": "prop_nn_owner_occupied",
+        "rental (private)": "prop_nn_private_rental",
+        "rental (social)": "prop_nn_social_rental",
+    }
+)
```
sofiapinto, line 1137: Getting the tenure of the nearest neighbours was such a smart idea!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1867 @@
...
+    .drop(["OA21CD", "prop_aged_35_to_64"])
+    .to_pandas()
+)
+
+X = model_df.drop(columns=["TENURE"])
+y = model_df[["TENURE"]]
+
+# One-hot encode property type feature
+enc = OneHotEncoder()
+onehotenc_df = pd.DataFrame(enc.fit_transform(X[["property_type"]]).toarray())
```
sofiapinto, line 1176: have we tried aggregating some property types before this? do you think it might be useful? Not for now, but something we can test in the future when we refactor the code?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1867 @@
...
+# Fit and predict with K-nearest neighbours classifier model
+# classifier = KNeighborsClassifier(n_neighbors=10)
+# classifier.fit(X_train, y_train)
+# print("K Nearest Neighbour classifier accuracy score:")
+# print(classifier.score(X_test, y_test))
+# y_pred = classifier.predict(X_test)
+# y_pred_prob = classifier.predict_proba(X_test)
+
+# Fit and predict with Random Forest classifier model
+classifier = RandomForestClassifier(class_weight="balanced")
```
sofiapinto: maybe add a note here to say that we're using `class_weight="balanced"` because the target is imbalanced

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
sofiapinto: from our conversation with Liz this morning, her suggestion was to separately account for missingness, e.g. the prop of NN properties with unknown tenure. I think this can be done in a separate PR though, after this one is closed.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1867 @@
...
+classifier.fit(X_train, y_train)
+print("\nRandom Forest classifier accuracy score:")
+print(classifier.score(X_test, y_test))
+y_pred = classifier.predict(X_test)
+y_pred_prob = classifier.predict_proba(X_test)
+
```
sofiapinto, line 1209: I think at one point we discussed potentially modelling whether it's owner occupied or not, rather than doing multiclass classification. Same for rented - or just rental (social), depending on what we end up needing for suitability. Not sure if you tested it, but worth adding to the list of possible improvements to test later.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
sofiapinto: just checked, looks okay

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1867 @@
...
+    .with_columns((pl.col("max") - pl.col("min")).alias("range"))
+)
+
+# Deduplicate UPRNs to the smallest garden size - the join allows us to retain the other columns in the df
+# TODO - use pl.over instead as it's more efficient
+uprns_with_garden_gdf = uprns_with_garden_gdf.join(
+    uprns_with_garden_gdf.group_by("UPRN").agg(pl.min("garden_area_m2")),
+    how="left",
+    on="UPRN",
+).filter(pl.col("garden_area_m2") == pl.col("garden_area_m2_right"))
```
sofiapinto, line 1641: is something like this possible? feels more intuitive when we read it

```suggestion
uprns_with_garden_gdf = uprns_with_garden_gdf.join(
    uprns_with_garden_gdf.group_by("UPRN").agg(pl.min("garden_area_m2").alias("min_garden_area_m2")),
    how="left",
    on="UPRN",
).filter(pl.col("garden_area_m2") == pl.col("garden_area_m2_right"))
```

sofiapinto: but also, if for some reason there are multiple lines for the same UPRN with the same garden size, won't we have the same UPRN multiple times? so, basically not truly deduplicating?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1867 @@
...
+features_df = features_df.join(
+    uprns_with_garden_gdf.select(
+        ["UPRN", "NATIONALCADASTRALREFERENCE", "garden_area_m2"]
+    ),
+    how="left",
+    on="UPRN",
+)
+
+# %%
+# Plot the distribution of range in garden size for UPRNs joined to multiple gardens
```
sofiapinto:
```suggestion
# Plot the distribution of range in garden size for UPRNs joined to multiple garden sizes
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1867 @@
...
+plt.boxplot(duplicated_gdf.filter(pl.col("range") <= 250000)["range"])
+plt.title("Range in garden size (under 250,000m2) for duplicate UPRNs")
+plt.ylabel("Range in garden area (m2)")
+plt.show()
+
```
sofiapinto, line 1657: i don't think we checked, but what proportion of UPRNs with garden have a range of garden sizes higher than 0?
Or higher than a small number.

Even the median seems very high - the median range is about 3 football pitches! Knowing this proportion would help with understanding how trustworthy this data is.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1867 @@
...
+print(duplicated_gdf["range"].describe())
+
+# %%
+# Fill any remaining missing garden size with the median size per property type and OA
+fill_garden_df = features_df.group_by(["predicted_property_type", "OA21CD"]).agg(
```
sofiapinto, line 1662: In addition to the above, i think in future we might want to get rid of some outliers or winsorizing them

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1867 @@
...
+    .then(True)
+    .when(pl.col("building_id").is_in(individual_heating_buildings))
+    .then(False)
+    .otherwise(None)
+    .alias("fill_community_heating")
+).with_columns(
+    pl.col("community_heating")
+    .fill_null(pl.col("fill_community_heating"))
+    .alias("use_community_heating")
+)
```
sofiapinto, line 1728: when refactoring, I think we need to do some work on the names of the variables + adding some additional docs. not just here, but in general to make it easier for future us to follow things easily.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1867 @@
...
+# ## Cluster UPRNs
+
+# %%
+# Cluster UPRNs on distance only
+model = HDBSCAN(
+    min_cluster_size=5,
+    cluster_selection_epsilon=20,
+    algorithm="ball_tree",
+    metric="euclidean",
+)
```
sofiapinto, line 1752: We will obviously need to play with these hyperparameters later. I don't think we need to set a smaller min cluster size because we already have the noise cluster for our outliers. (thinking about Liz's comment above)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1867 @@
...
+# %% [markdown]
+# ## Cluster UPRNs
+
+# %%
+# Cluster UPRNs on distance only
+model = HDBSCAN(
+    min_cluster_size=5,
+    cluster_selection_epsilon=20,
+    algorithm="ball_tree",
+    metric="euclidean",
```
sofiapinto, line 1751: X and Y coord are different from lat/long right? otherwise we might want a different distance metric

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
```diff
@@ -0,0 +1,1867 @@
...
+        ["UPRN", "cluster", "LATITUDE", "LONGITUDE", "geometry"]
+    ].within(stoke_ward)
+].to_crs(epsg=4326)
+wgs84_buildings_df = plymouth_residential_buildings_gdf[
+    plymouth_residential_buildings_gdf.within(stoke_ward)
+].to_crs(epsg=4326)
+x, y = (
+    wgs84_uprns_df.dissolve().centroid.values[0].x,
+    wgs84_uprns_df.dissolve().centroid.values[0].y,
+)
```
sofiapinto, line 1817: some docs here in future would be really helpful.

sofiapinto (commented):
hey @crispy-wonton - so MUCH amazing work in only one PR!

I've mostly left clarifying questions and comments that we should take forward in future PRs. After merging this, I'd suggest doing the refactoring in a couple of separate PRs, so that we're sure there are no bugs.

asf_heat_pump_suitability/pipeline/prepare_features/garden_size.py
crispy-wonton: good point! no!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: you mean like grouping together semi-detached and terraces for example? If that's what you mean i havent tested it but it's a good idea!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: I didn't test it because my impression was that all three tenure categories are worth knowing. However, once we've finalised the tree we can revisit this and potentially create a binary classifier which could improve the accuracy.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: Generally there's some high correlation of the features. However, my understanding is that although high correlation can affect the interpretation of the model, especially feature importance, it has less impact on the actual model performance when using random forests. Let me know if you think something different! :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: yes i think i did check and this doesnt happen (i would need to check again to be sure) but great point, this needs to be fixed when refactoring.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: agree! they are horrible!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
crispy-wonton: yes, X and Y are in British National Grid and lat / lon are in 4326 CRS. what distance metric do you think is better?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
sofiapinto: yes, that was it!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
sofiapinto: In that case euclidean is fine. If we were using latitude/longitude then we would require a spherical metric because they are angular coordinates. But British National Grid are in the projected coordinate system so euclidean works well.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/create_full_dataset_plymouth.py
sofiapinto: That is correct, because we're not using linear models and we don't have very high correlations of >0.8 or similar, it's fine!

sofiapinto (dismissed):

PR #176, opened 2025-08-21
Cluster feasibility scoring and suitability categorisation

### LINKED ISSUE #173: Categorisation logic for most suitable technology and feasibility scoring

sofiapinto:
---

# Description

This PR adds functions to allow for computing cluster feasibility scoring and assigning the most suitable technology scheme for a set of clusters:
- `asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py` contains the feasibility and suitability functions
- `asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/testing_suitability_and_feasibility_functions.py` is a notebook where I tested the feasibility function

Fixes #173 

# Instructions for Reviewer

Hey @crispy-wonton and @lizgzil , here's a first draft of this work! Thanks for taking a look ✨ 

- Could you please review the logic of the functions in `assign_cluster_suitability_and_feasibility.py` and make sure I coded the information in the slides correctly? Also let me know where you think more documentation is needed.
- @crispy-wonton please let me know if any of the repo conventions isn't followed or if you dislike the function/ script names.
- You can test the `create_df_feasibility_scoring()` function by opening `testing_suitability_and_feasibility_functions.py` as a notebook. All you need to do is run `jupytext --to notebook testing_suitability_and_feasibility_functions.py` in the terminal.
- I'm not able to properly test `create_df_suitability_categorisation()` as I don't have a dataset to test it with. @lizgzil I also had a chat with @crispy-wonton about this underlying data. How did you imagine we should identify if a cluster is in a city centre or whether a cluster has shared outdoor space for digging? Is this something you expected me to work on as part of this PR?
- @crispy-wonton I haven't worked with polars extensively, so please flag any better ways of doing anything!

Thanks in advance 💖 

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+This file contains functions for:
+- categorising/assigning each cluster with the most suitable technology scheme.
+- computing the feasibility scores for different heating schemes and clusters;
+"""
+
+# package imports
+import polars as pl
+
+
+def assign_str_suitable_techology_scheme(
```
sofiapinto: hey @lizgzil - is this the correct interpretation of the [categorisation logic](https://docs.google.com/presentation/d/1Yezm5uUh_riMbQkgPcbwy0yIqyIhIosiBTUf-WtHSKw/edit?slide=id.g37497e39a53_0_1276#slide=id.g37497e39a53_0_1276)?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+                row["has_outdoor_space"],
+            )
+        )
+        .alias("suitable_scheme")
+    )
+
+    return cluster_df
+
+
+def create_df_feasibility_scoring(
```
sofiapinto, line 259: I created this function so that features and their weights don't need to be hardcoded to increase flexibility of this function, but let me know if you think it makes it a bit blackbox - hopefully the description of "weights" counteracts that.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+            "social_housing",
+            "flats",
+            "on_communal_heating",
+            "has_outdoor_space",
+            "in_listed_buildings",
+            "in_cons_area",
+            "in_hn",
+            "close_to_anchor_loads",
+            "close_to_city_centre",
+        ]
```
sofiapinto: Once @crispy-wonton finishes her work on the plymouth data, I can improve this part. In the notebook you can see that I did a bit of data processing and prior to calling the function.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+        .alias("suitable_scheme")
+    )
+
+    return cluster_df
+
+
+def create_df_feasibility_scoring(
+    df: pl.DataFrame,
+    weights: dict,
+    cols_to_aggregate: list = None,
```
sofiapinto: We should probably call this `features` or similar

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/testing_suitability_and_feasibility_functions.py
```diff
@@ -0,0 +1,99 @@
...
+    (pl.col("garden_area_m2") > 0).alias("has_outdoor_space")
+)
+
+
+# %%
+from assign_cluster_suitability_and_feasibility import create_df_feasibility_scoring
+
+# %%
+# Dictionary of weights for computing feasibility scores for each scheme
+weights = {
```
sofiapinto: I've set all weights to 1 in this example, but you can play with these numbers @crispy-wonton and @lizgzil

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/testing_suitability_and_feasibility_functions.py
```diff
@@ -0,0 +1,99 @@
...
+    "individual_ashp_feasibility": {
+        "owner_occupied": 1,
+        # "in_high_income_decile": 1,
```
sofiapinto: Some features are commented out because I couldn't find them in the plymouth data. I believe this is because @crispy-wonton is still working on them :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/testing_suitability_and_feasibility_functions.py
```diff
@@ -0,0 +1,99 @@
...
+        "not_listed": 1,
+        "not_in_conservation_area": 1,
+    },
+    "collective_ashp_feasibility": {
+        "owner_occupied": 1,
+        # "in_high_income_decile": 1,
+        "on_gas": 1,
+        "not_listed": 1,
+        "not_in_conservation_area": 1,
+        # "cluster_size": 1
```
sofiapinto: The code works if you uncomment `cluster_size`. However, the feasibility outputs no longer are values between 0 and 100.

How do you want me to deal with the `cluster_size` @lizgzil ? Do you want me to do min-max scaling to the range [0, 100]?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+                    if key != "cluster_size"
+                    else pl.col(f"cluster_size") * weight
+                )
+                for key, weight in hn_feasibility_weights.items()
+            )
+            / sum(hn_feasibility_weights.values())
+        ).alias("hn_feasibility"),
+    )
+
+    return cluster_stats
```
sofiapinto: After a first review I can add a `if __name__ == "__main__":` to call these functions and save the outputs to S3

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
crispy-wonton: Based on an earlier comment from Dan, I think we wanted to move away from calling these 'schemes'. I think the function could be called `assign_str_suitable_tech_type` or similar instead :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+
+def assign_str_suitable_techology_scheme(
+    cluster_size: int,
+    hn_area: bool,
+    city_centre: bool,
+    has_outdoor_space: bool,
+) -> str:
+    """
+    Categorises a cluster (based on its size and other characteristics) into its
+    most suitable technology scheme:
```
crispy-wonton:
```suggestion
    most suitable low carbon heating technology:
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+            if cluster_size > 20:
+                if has_outdoor_space:
+                    return "shared_ground_loop"
+                else:
+                    return "collective_ashp"
+            else:
+                return "collective_ashp"
+
+
+def create_df_suitability_categorisation(cluster_df: pl.DataFrame) -> pl.DataFrame:
```
crispy-wonton: just leaving a note here thinking about future scalability, but i dont think it needs to be done in this PR. This function and the above could be refactored into a `polars` when / then / otherwise chain. It would have to be carefully implemented to make sure the logic is working as expected, but this could work quicker on the full GB dataset.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+        pl.struct(["cluster_size", "hn_area", "city_centre", "has_outdoor_space"])
+        .map_elements(
+            lambda row: assign_str_suitable_techology_scheme(
+                row["cluster_size"],
+                row["hn_area"],
+                row["city_center"],
+                row["has_outdoor_space"],
+            )
+        )
+        .alias("suitable_scheme")
```
crispy-wonton:
```suggestion
        .alias("best_tech")
```
could we rename something along the lines of this? my suggestion isnt the best so open to other things but want something that indicates it's the top / best / selected tech and that it's a tech not a scheme

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
crispy-wonton: agree! `features` is good

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+
+def create_df_feasibility_scoring(
+    df: pl.DataFrame,
+    weights: dict,
+    cols_to_aggregate: list = None,
+    anchor_loads_threshold: float = 500,
+    city_centre_threshold: float = 500,
+) -> pl.DataFrame:
+    """
+    Computes the feasibility score for each scheme and cluster in the dataset.
```
crispy-wonton:
```suggestion
    Computes the feasibility score for each tech type and cluster in the dataset.
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+
+    Schemes include:
+    - Individual Air Source Heat Pump (ASHP)
+    - Collective purchasing of ASHP
+    - Shared Ground Loop (SGL)
+    - Heat Network (HN)
+
+    Args:
+        df (pl.DataFrame): DataFrame containing the dataset with feature values for each UPRN.
+        weights (dict): Dictionary containing the weights for each scheme and features used in feasibility scoring.
```
crispy-wonton:
```suggestion
        weights (dict): Dictionary containing the weights for each tech type and features used in feasibility scoring.
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+            'weights' should contain the following keys (schemes):
+                "individual_ashp_feasibility", "collective_ashp_feasibility", "sgl_feasibility" and "hn_feasibility"
+            The value corresponding to each scheme should be a dictionary whith where keys subsets of features in cols_to_aggregate
```
crispy-wonton:
```suggestion
            'weights' should contain the following keys (tech types):
                "individual_ashp_feasibility", "collective_ashp_feasibility", "sgl_feasibility" and "hn_feasibility"
            The value corresponding to each tech type should be a dictionary where keys are subsets of features in cols_to_aggregate
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+        cols_to_aggregate (list, optional): List of columns to aggregate. If None, defaults to a predefined list.
+        anchor_loads_threshold (float, optional): default threhold for distance to anchor loads in meteres. Defaults to 500 m.
+        city_centre_threshold (float, optional): default threhold for distance to city center in meteres. Defaults to 500 m.
```
crispy-wonton:
```suggestion
        anchor_loads_threshold (float, optional): threshold for distance to anchor loads in metres. Defaults to 500 m.
        city_centre_threshold (float, optional): threshold for distance to city centre in metres. Defaults to 500 m.
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+    Raises:
+        ValueError: if schemes or features provided in weights do not exist.
+
+    Returns:
+        pl.DataFrame: DataFrame with aggregated feasibility scores for each cluster.
+    """
+
+    if cols_to_aggregate is None:
+        # Creating a column for not in conservation area
+        df = df.with_columns((~pl.col("in_cons_area")).alias("not_in_cons_area"))
```
crispy-wonton: What's the reason for adding this col?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+            (pl.col("distance_to_anchor_loads") <= anchor_loads_threshold).alias(
+                "close_to_anchor_loads"
+            )
+        )
+        # Creating close_to_city_centre from distance_to_city_centre and city_centre_threshold
+        df = df.with_columns(
+            (pl.col("distance_to_city_centre") <= city_centre_threshold).alias(
+                "close_to_city_centre"
+            )
+        )
```
crispy-wonton: shouldn't these lines also run if the `cols_to_aggregate` contains `close_to_anchor_loads` and `close_to_city_centre`?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+
+        # Columns to compute percentages for feasibility scoring
+        cols_to_aggregate = [
+            "owner_occupied",
+            "high_income_decile",
+            "on_gas",
+            "not_listed",
+            "not_in_cons_area",
+            "social_housing",
```
crispy-wonton: currently `owner-occupied` and `rental (social)` are in one `tenure` column. We will need to add processing somewhere in the pipeline to get this binary.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+    expected_schemes = {
+        "individual_ashp_feasibility",
+        "collective_ashp_feasibility",
+        "sgl_feasibility",
+        "hn_feasibility",
+    }
+    if set(weights.keys()) != expected_schemes:
+        raise ValueError(
+            "There are incorrect or missing keys in the 'weights' dictionary."
+        )
```
crispy-wonton: could we update the error to print which keys are missing. think that would be really useful for debugging

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
crispy-wonton: can we have a default `weights` that we call from either a config or another `.py` file in the same subdir?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+
+    for scheme in weights.keys():
+        scheme_features = set(weights.get(scheme).keys())
+        if not scheme_features.issubset(set(cols_to_aggregate + ["cluster_size"])):
+            raise ValueError(
+                f"{scheme} scheme: The features you're providing weights for do not exist."
```
crispy-wonton: Here it would be really useful if the value error can print out the features that don't exist :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+
+    # Aggregating data by cluster
+    cluster_stats = df.group_by("cluster").agg(
+        pl.col(cols_to_aggregate).mean().name.prefix("perc_"),
+        pl.col("cluster").count().alias("cluster_size"),
+    )
+
+    # Convert proportions to percentages for columns with prefix "perc_"
+    cluster_stats = cluster_stats.with_columns(
+        (pl.col("^perc_.*$").cast(pl.Float64) * 100).name.keep()
```
crispy-wonton: can this be combined with the line above? Does this work:
`((pl.col(cols_to_aggregate).mean()).cast(pl.Float64) * 100).name.prefix("perc_")`

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+    collective_ashp_feasibility_weights = weights.get("collective_ashp_feasibility")
+    sgl_feasibility_weights = weights.get("sgl_feasibility")
+    hn_feasibility_weights = weights.get("hn_feasibility")
+
+    # Compute feasibility scores using the percentages and the weights
+    cluster_stats = cluster_stats.with_columns(
+        # Cluster identifier
+        pl.col("cluster").alias("cluster"),
+        # Number of UPRNs in the cluster
+        pl.col("cluster_size").alias("cluster_size"),
```
crispy-wonton: what are these lines doing?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,237 @@
...
+        (pl.col("^perc_.*$").cast(pl.Float64) * 100).name.keep()
+    )
+
+    # Extracting weights for each feasibility scoring category
+    individual_ashp_feasibility_weights = weights.get("individual_ashp_feasibility")
+    collective_ashp_feasibility_weights = weights.get("collective_ashp_feasibility")
+    sgl_feasibility_weights = weights.get("sgl_feasibility")
+    hn_feasibility_weights = weights.get("hn_feasibility")
+
+    # Compute feasibility scores using the percentages and the weights
```
crispy-wonton: I suppose if you wanted to reduce the amount of code here, you could use a `for` loop with e.g. a dict of the weights dicts and the aliases you want to assign. Then you could just have the logic of creating the scores once instead of 4 times.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
crispy-wonton: cool, `utils/save_utils.py` may be useful to you :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/testing_suitability_and_feasibility_functions.py
crispy-wonton: thank you!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/testing_suitability_and_feasibility_functions.py
crispy-wonton: yep , can confirm the ones you've commented out are not there yet :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/testing_suitability_and_feasibility_functions.py
```diff
@@ -0,0 +1,99 @@
...
+    },
+    "sgl_feasibility": {
+        "social_housing": 1,
+        "flats": 1,
+        # "on_communal_heating": 1,
+        "has_outdoor_space": 1,
```
crispy-wonton:
```suggestion
        "has_outdoor_space": 1,
        "not_listed": 1,
        "not_in_conservation_area": 1,
```

crispy-wonton (commented):
Hi @sofiapinto !

Thank you so much for putting this together - it looks great!

I think the logic looks mostly correct, there's a couple of questions I've left but nothing major. Also your `polars` syntax looks great to me, as do the function names. The main change on that side is referring to technologies / tech types rather than 'schemes'.

This is really great work and it's great to see the scoring in action. Really well documented too.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: I've called it "most_suitable_tech". Is that okay?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: everything inside this if is up to change. I need to know whether the data processing I did in the notebook is done prior to this or in this script.

As soon as we agree on feature names and where the processing happens then I can change this.

The reason for adding this call is that the plymouth dataset as is does not provide a "not_in_cons_area", only an "in_cons_area".

In the notebook you can see that I actually provide the data ready to go, so all the processing is done prior to the function call and `features` (`cols_to_aggregate`) are provided.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: As above, everything inside this if is up to change. I need to know whether the data processing I did in the notebook is done prior to this or in this script.

We either provide "close_to_city_centre" or we compute it using "distance_to_city_centre" and "city_centre_threshold".

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: Yep - again, as above, in the notebook I do this processing before calling the feasibility function but we need to decide where the processing happens.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: Of course - we just need to agree on some column names. Let's discuss tomorrow.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: I thought the needed to be added to appear in the final dataset. My poor knowledge of polars. Have now removed them.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: implemented!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/testing_suitability_and_feasibility_functions.py
sofiapinto: I implemented this.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: added.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: Added.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: Added

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: Code has now be refactored!

sofiapinto:
Hello @crispy-wonton and @lizgzil ,

I have now done the following:

- all mentions of "scheme" have now been replaced by "tech type"
- created a config with our defaults for: features, weights, expected tech types and oa21 codes in the city centre (these are all OAs in MSOA "City Centre, Barbican and Sutton Harbour")
- refactored `create_df_feasibility_scoring()` including adding more info about ValueError
- refactored `create_df_suitability_categorisation()` to use `polars` when/then/otherwise syntax which is faster than the previous implementation and will enable the script to run faster if ever applied to UK data. Please check if this is aligned with Liz's suitability categorisation diagram. 
- similarly created a function called `prepare_df_for_suitability_categorisation()` which processes the data before suitability categorisation.  A few assumptions based on the current implementation are:
    - if at least one UPRN is in a heat network zone, the cluster is considered to be in the HN zone. Is this okay, or should it be based on the majority of the cluster?
    - similarly, the whole cluster is considered to be close to the city centre if at least one of the properties in the cluster is in the city centre. Currently, I'm considering the city centre to be all OAs in the MSOA "City Centre, Barbican and Sutton Harbour"
    - the total garden area is a sum of the garden areas for properties that aren't flats together with the garden area of the building for flats in the cluster. From the total outdoor space, we create the has_outdoor_space flag based on a `outdoor_space_threshold`. Please take a close look at how I implemented this to make sure it is correct.
- created a function called `prepare_df_for_feasibility_scoring()` which takes the feature df (with features for each UPRN) and processes the data as per our last stand up. Please let me know if you want me to rename any of the features. A few assumptions based on the current implementation are:
    - Same implementation for the `close_to_city_centre` flag
    - `close_to_anchor_loads` flag requires `distance_to_anchor_loads` and `anchor_loads_threshold`. I have no sense of whether this is the right way to do this.
- updated the testing notebook which now enables testing both the feasibility scoring and suitability categorisation with Plymouth and dummy data. To test with Plymouth I had to set a few columns with random values, while we wait for a final version of the plymouth data. To check this yourselves, all you need to do is run `jupytext --to notebook testing_suitability_and_feasibility_functions.py` in the terminal to create a notebook version of this .py file.
- added `if __name__ == __main__` section to enabling running this script to apply the scoring and categorisation to Plymouth data. This won't work at the moment, as it's waiting for a finalised version of the plymouth data.


Thanks in advance for reviewing!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/config.py
```diff
@@ -0,0 +1,106 @@
+"""
+Config file with defaults used for feasibility scoring and suitability categorisation.
+"""
+
+# Features needed for feasibility scoring
+features = [
```
sofiapinto, line 17: let me know if any of these names should be changed.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,313 @@
...
+
+# local imports
+import config
+from asf_heat_pump_suitability.utils.save_utils import save_to_s3
+
+
+def prepare_df_for_suitability_categorisation(
+    df: pl.DataFrame,
+    city_centre_oas: set = config.city_centre_oas,
+    outdoor_space_threshold: float = 0,
```
sofiapinto: this is a random default!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,313 @@
...
+    )
+
+    return df.with_columns(most_suitable_tech=most_suitable_tech)
+
+
+def prepare_df_for_feasibility_scoring(
+    df: pl.DataFrame,
+    features: list = config.features,
+    anchor_loads_threshold: float = 500,
+    outdoor_space_threshold: float = 0,
```
sofiapinto: these are random defaults!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,313 @@
...
+if __name__ == "__main__":
+    plymouth_data = pl.read_parquet(
+        "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/plymouth_uprn_with_features.parquet"
+    )
+
+    feasibility_scoring_data = prepare_df_for_feasibility_scoring(
+        df=plymouth_data,
+        features=config.features,
+        anchor_loads_threshold=500,
+        outdoor_space_threshold=0,
```
sofiapinto: we should change these defaults to values that make sense before running the script

lizgzil:
## Review

In running this code and adding the updates I've thoroughly looked at the code and am happy with it (and if not I changed bits in my commits). Thanks @sofiapinto for making it such a dream to run and understand.

## Update

Writing some notes before I forget!

I realised we need to not calculate feasibility on all the -1 "cluster" at once, they needed to be separate, so I created a unique ID for these (the negative of the UPRN).

I also put the thresholds in the config, fixed a weird thing that was happening with the flats outdoor space (the code isn't very elegant, but I think it works!), and I added the per cluster distance to anchor properties.

We get:
```
>>> suitability_categorisation_data['most_suitable_tech'].value_counts()
shape: (4, 2)
┌────────────────────┬───────┐
│ most_suitable_tech ┆ count │
│ ---                ┆ ---   │
│ str                ┆ u32   │
╞════════════════════╪═══════╡
│ individual_ashp    ┆ 13793 │
│ shared_ground_loop ┆ 1373  │
│ heat_network       ┆ 886   │
│ collective_ashp    ┆ 2898  │
└────────────────────┴───────┘
```
Next week I'll explore the results properly and make some plots 😄 

## Bug/issue?

I'm not sure if this is a bug, but I noticed something a bit weird that seems to be happening with the total garden size - this example should illustrate it:

Cluster 614 contains 6 properties, 5 of which are from the same building ID. However, 3 of the properties are flats and 3 aren't. The 3 flats have the same garden size. So the total garden size is calculated by the unique garden size from the flats with the same building id + the individual 3 building ID garden sizes.

```
┌──────────────┬─────────┬────────────────┬─────────────────────────────────┬─────────────────────────────────┐
│ UPRN         ┆ cluster ┆ garden_area_m2 ┆ building_id                     ┆ predicted_property_type         │
│ ---          ┆ ---     ┆ ---            ┆ ---                             ┆ ---                             │
│ i64          ┆ i64     ┆ f64            ┆ str                             ┆ str                             │
╞══════════════╪═════════╪════════════════╪═════════════════════════════════╪═════════════════════════════════╡
│ 10000103002  ┆ 614     ┆ 311.568512     ┆ 7071C095-41DE-4303-90C3-FF2E54… ┆ Flat, maisonette or apartment   │
│ 10000795860  ┆ 614     ┆ 311.568512     ┆ 7071C095-41DE-4303-90C3-FF2E54… ┆ Flat, maisonette or apartment   │
│ 10090566138  ┆ 614     ┆ 230.171584     ┆ 7071C095-41DE-4303-90C3-FF2E54… ┆ Terraced (including end-terrac… │
│ 10090566139  ┆ 614     ┆ 311.568512     ┆ 7071C095-41DE-4303-90C3-FF2E54… ┆ Flat, maisonette or apartment   │
│ 100040465852 ┆ 614     ┆ 163.218126     ┆ F60005D2-7282-4945-8518-E5318A… ┆ Detached                        │
│ 100041061542 ┆ 614     ┆ 230.171584     ┆ 7071C095-41DE-4303-90C3-FF2E54… ┆ Terraced (including end-terrac… │
└──────────────┴─────────┴────────────────┴─────────────────────────────────┴─────────────────────────────────┘
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+    outdoor_space_threshold: float = config.outdoor_space_threshold,
+) -> pl.DataFrame:
+    """
+    Prepares the dataframe for suitability categorisation. This includes:
+    - Identifying clusters close to the city center or in heat network zones
+    - Identifying clusters with outdoor space
+
+    Args:
+        df (pl.DataFrame): DataFrame containing the dataset with feature values for each UPRN.
+        city_centre_oas (set, optional): OAs that are considered part of the city center. Should be an oa21 format.
```
crispy-wonton:
```suggestion
        city_centre_oas (set, optional): OAs that are considered part of the city center. Should be in OA21 codes format.
```
you're using the codes here right?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+            Defaults to config.city_centre_oas.
+        outdoor_space_threshold (float, optional): Threshold for outdoor space in meters squared.
+        i.e. if garden_area_m2 > 0, then has_outdoor_space = True
```
crispy-wonton: This example is slightly confusing me 😅 should the `0` here be replaced with `outdoor_space_threshold`?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+
+    Returns:
+        pl.DataFrame: DataFrame ready for suitability categorisation.
+    """
+    # Create city_centre column based on whether the UPRN is in the set of city centre OAs
+    df = df.with_columns(pl.col("oa21").is_in(city_centre_oas).alias("in_city_centre"))
+
+    # Transform predicted_tenure and predicted_property_type into dummies
+    df = df.to_dummies("predicted_property_type")
+    df = df.rename({"predicted_property_type_Flat, maisonette or apartment": "flats"})
```
crispy-wonton: you can easily chain these lines together with polars :)
```suggestion
    # Create city_centre column based on whether the UPRN is in the set of city centre OAs
    df = df.with_columns(
        pl.col("oa21").is_in(city_centre_oas).alias("in_city_centre")
        # Transform predicted_tenure and predicted_property_type into dummies
        ).to_dummies("predicted_property_type")
        .rename({"predicted_property_type_Flat, maisonette or apartment": "flats"})
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+                        pl.col("building_id").filter(pl.col("flats") == 1).arg_unique()
+                    )
+                    .struct.field("garden_area_m2")
+                )
+            )
+        )
+        .with_columns(
+            pl.col("building_id").list.sum().alias("total_outdoor_space_flats")
+        )
+    )
```
crispy-wonton, line 91:
```suggestion
flats_garden_sizes = df.filter(pl.col("flats") == 1).group_by(["cluster", "building_id"]).agg(pl.col("garden_area_m2").unique().sum().alias("total_outdoor_space_flats"))
```
I think this does the same thing (?) in a simpler way? The code I've written is getting the total garden size for every unique building ID in a cluster.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+
+    # Join the aggregated data together, summing the flats and the not flats for each cluster
+
+    cluster_df = (
+        cluster_df.join(flats_garden_sizes, on="cluster", how="left")
+        .with_columns(
+            (
+                pl.col("total_outdoor_space")
+                + pl.col("total_outdoor_space_flats").fill_null(0)
```
crispy-wonton: worth testing, but I don't think you need to fill nulls with polars before doing addition operations? I think it ignores nulls by default. let me know if im wrong!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+            ).alias("total_outdoor_space")
+        )
+        .drop(["building_id", "total_outdoor_space_flats"])
+    )
+
+    cluster_df = cluster_df.with_columns(
```
crispy-wonton:
```suggestion
    ).with_columns(
```
chaining again! :D

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+    return cluster_df
+
+
+def create_df_suitability_categorisation(
+    df: pl.DataFrame, sgl_min_properties: int = config.sgl_min_properties
+) -> pl.DataFrame:
     """
-    Categorises a cluster (based on its size and other characteristics) into its
-    most suitable technology scheme:
+    Adds a new column to cluster_df with most suitable low carbon heating technology for each cluster:
```
crispy-wonton:
```suggestion
    Adds a new column to the dataframe with most suitable low carbon heating technology for each cluster:
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
-        cluster_size (int): The number of UPRNs in the cluster.
-        hn_area (bool): Indicates if the cluster is in a heat network area.
-        city_centre (bool): Indicates if the cluster is close to the city centre.
-        has_outdoor_space (bool): Indicates if the cluster has outdoor space.
+        df (pl.DataFrame): DataFrame containing information about each cluster of properties.
+        sgl_min_properties (int): The minimum number of properties in a cluster for a SGL to be considered.
 
     Returns:
-        str: A string representing the suitability category of the cluster.
+        pl.DataFrame: cluster_df with an additional column for the suitability categorisation.
```
crispy-wonton:
```suggestion
        pl.DataFrame: dataframe with an additional column containing the assigned low carbon home heating tech type for each cluster.
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+        .when(pl.col("in_city_centre"))
+        .then(pl.lit("heat_network"))
+        .when(
+            (pl.col("cluster_size") > sgl_min_properties) & pl.col("has_outdoor_space")
+        )
+        .then(pl.lit("shared_ground_loop"))
+        .otherwise(pl.lit("collective_ashp"))
+    )
+
+    return df.with_columns(most_suitable_tech=most_suitable_tech)
```
crispy-wonton:
```suggestion
    df = df.with_columns(
        pl.when(pl.col("cluster_size") == 1)
        .then(pl.lit("individual_ashp"))
        .when(pl.col("in_heat_network_zone"))
        .then(pl.lit("heat_network"))
        .when(pl.col("in_city_centre"))
        .then(pl.lit("heat_network"))
        .when(
            (pl.col("cluster_size") > sgl_min_properties) & pl.col("has_outdoor_space")
        )
        .then(pl.lit("shared_ground_loop"))
        .otherwise(pl.lit("collective_ashp"))
        .alias("most_suitabile_tech")
    )
    return df
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
-            else:
-                return "collective_ashp"
-
-
-def create_df_suitability_categorisation(cluster_df: pl.DataFrame) -> pl.DataFrame:
+    most_suitable_tech = (
+        pl.when(pl.col("cluster_size") == 1)
+        .then(pl.lit("individual_ashp"))
+        .when(pl.col("in_heat_network_zone"))
+        .then(pl.lit("heat_network"))
```
crispy-wonton, line 133: I wonder if this should come first? I'm not sure - something to discuss with others who know more than me but I imagine if it's in a HN zone then it should be assigned HN even if it's just one property. But I can also see the argument for keeping it as ASHP due to the cost of trenching to single properties in a HN.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+        .when(pl.col("in_city_centre"))
+        .then(pl.lit("heat_network"))
```
crispy-wonton, line 135: same question as above here - should this come before ASHP. I don't know the answer but worth discussing with others

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+
+    cluster_df = cluster_df.with_columns(
+        (pl.col("total_outdoor_space") > outdoor_space_threshold).alias(
+            "has_outdoor_space"
+        )
+    )
+    return cluster_df
+
+
+def create_df_suitability_categorisation(
```
crispy-wonton, line 111: I think you've applied the logic from the tree in the slides correctly here but leaving a note to say we need to update and change that logic based on feedback from the workshop!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+    anchor_loads_threshold: float = config.anchor_loads_threshold,
+    outdoor_space_threshold: float = config.outdoor_space_threshold,
+    city_centre_oas: set = config.city_centre_oas,
+) -> pl.DataFrame:
     """
-    Adds a new column to cluster_df with the suitability categorisation for each cluster.
+    Prepares the dataframe for feasibility scoring. This includes:
+    - Transforming categorical variables into dummy/indicator variables.
+    - Creating new binary features based on existing columns and specified thresholds.
+    - Renaming columns for clarity.
```
crispy-wonton: leaving a note to say hopefully this function won't need to rename columns in the future as the pipeline to feed the dataframe going into this script should have consistent names :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
 
     Args:
-        cluster_df (pl.DataFrame): DataFrame containing information about each cluster of properties.
+        df (pl.DataFrame): DataFrame containing the dataset with feature values for each UPRN.
+        features (list, optional): List of features used in the feasibility scoring.
+        anchor_dist_df (pl.DataFrame): DataFrame containing the distance (m) from the centre of each cluster to the nearest anchor load.
+        anchor_loads_threshold (float, optional): threshold for distance to anchor loads in metres.
+        outdoor_space_threshold (float, optional): Threshold for outdoor space in meters squared.
+            i.e. if garden_area_m2 > 0, then has_outdoor_space = True
```
crispy-wonton:
```suggestion
            i.e. if garden_area_m2 > `outdoor_space_threshold`, then has_outdoor_space = True
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+        city_centre_oas (set, optional): OAs that are considered part of the city center. Should be an oa21 format.
```
crispy-wonton:
```suggestion
        city_centre_oas (set, optional): OAs that are considered part of the city center. Should be in OA21 codes format.
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+        (pl.col("garden_area_m2") > outdoor_space_threshold).alias("has_outdoor_space")
+    )
+
+    # Creating close_to_city_centre from distance_to_city_centre and city_centre_threshold
+    df = df.with_columns(
+        pl.col("oa21").is_in(city_centre_oas).alias("close_to_city_centre")
+    )
+
+    # Creating imd_decile_above_avg from imd_decile
+    df = df.with_columns((pl.col("imd_decile") > 5).alias("imd_decile_above_avg"))
```
crispy-wonton:
```suggestion
    df = df.to_dummies("predicted_tenure").to_dummies("predicted_property_type")
        .rename(
            {
                "predicted_tenure_owner-occupied": "owner_occupied",
                "predicted_tenure_rental (social)": "social_housing",
                "predicted_property_type_Flat, maisonette or apartment": "flats"
            }
        ).with_columns(
                (~pl.col("use_off_gas")).alias("on_gas"),
                (~pl.col("in_listed_building")).alias("not_in_listed_building"),
                (~pl.col("in_conservation_area")).alias("not_in_conservation_area"),
                (pl.col("garden_area_m2") > outdoor_space_threshold).alias("has_outdoor_space"),
                pl.col("oa21").is_in(city_centre_oas).alias("close_to_city_centre"),
                (pl.col("imd_decile") > 5).alias("imd_decile_above_avg")
        )
```
`with_columns` allows you to add as many manipulations as you want into one chain :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+        (
+            (pl.col([f for f in features if f != "close_to_anchor_loads"]).mean()).cast(
+                pl.Float64
             )
+            * 100
+        ).name.prefix("perc_"),
+        pl.col("UPRN").count().alias("cluster_size"),
+    )
+
+    # Creating close_to_anchor_loads from distance_from_anchor_property_m and anchor_loads_threshold
```
crispy-wonton, line 208: obviously not for you to sort out in this PR but leaving a note to say in future ideally the distance_from_anchor_property_m feature would be in the same features df as all the others if possible :) this can come when we refactor mine and Liz's code into a proper pipeline

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+
+    anchor_dist_df = anchor_dist_df.with_columns(
+        (pl.col("distance_from_anchor_property_m") <= anchor_loads_threshold).alias(
+            "close_to_anchor_loads"
         )
-        .alias("suitable_scheme")
+    ).select(
+        pl.col("cluster").cast(pl.Int64),
+        (pl.col("close_to_anchor_loads").cast(pl.Float64) * 100).name.prefix("perc_"),
```
crispy-wonton, line 216: am I right in thinking this will either give 0 or 100%?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+    df = df.join(anchor_dist_df, on="cluster")
+
+    # scale cluster size to be between 0 and 100
+    df = df.with_columns(
+        (
+            (pl.col("cluster_size") - pl.col("cluster_size").min())
+            / (pl.col("cluster_size").max() - pl.col("cluster_size").min())
+            * 100
+        ).alias("cluster_size")
+    )
```
crispy-wonton:
```suggestion
    df = df.join(anchor_dist_df, on="cluster").with_columns(
        # scale cluster size to be between 0 and 100
        (
            (pl.col("cluster_size") - pl.col("cluster_size").min())
            / (pl.col("cluster_size").max() - pl.col("cluster_size").min())
            * 100
        ).alias("cluster_size")
    )
```
I wonder if this is worth renaming as `scaled_cluster_size` so we don't get confused

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+
+    return df
+
+
+def calculate_feasibility_expression(tech_specific_weights: dict) -> pl.Expr:
```
crispy-wonton: the name of the arg here doesnt match the one in the docstring. I think `weights_dict` is a good name

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
+    # Compute weighted sum i.e. sum(feature_value * weight)
+    weighted_cols_sum = pl.sum_horizontal(
+        (
+            pl.col(f"perc_{feature}") * weight
+            if feature != "cluster_size"  # most features will have a "perc_" prefix
+            else pl.col("cluster_size")
+            * weight  # cluster_size does not have a "perc_" prefix
+        )
+        for feature, weight in tech_specific_weights.items()
+    )
```
crispy-wonton: I think you can do this but haven't tested (and then just return `weighted_cols_sum`)
```suggestion
    ).truediv(total_weight)
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -2,236 +2,402 @@
...
-                "individual_ashp_feasibility", "collective_ashp_feasibility", "sgl_feasibility" and "hn_feasibility"
-            The value corresponding to each scheme should be a dictionary whith where keys subsets of features in cols_to_aggregate
-        cols_to_aggregate (list, optional): List of columns to aggregate. If None, defaults to a predefined list.
-        anchor_loads_threshold (float, optional): default threhold for distance to anchor loads in meteres. Defaults to 500 m.
-        city_centre_threshold (float, optional): default threhold for distance to city center in meteres. Defaults to 500 m.
+        features (list, optional): List of features used in the feasibility scoring.
+            Defaults to a predefined list.
+        expected_tech_types (set, optional): Set of expected technology types for feasibility scoring.
+        weights (dict, optional): Dictionary containing the weights for each feature used to compute feasibility for specific tech types.
+            in the format {'tech_type_1': {'feature_1': weight_1, 'feature_2': weight_2, ...}, ..., 'tech_type_n': {...}}
```
crispy-wonton, line 280: great documentation

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,403 @@
...
+    """
+    if set(weights.keys()) != expected_tech_types:
+        incorrect_techs = [w for w in weights.keys() if w not in expected_tech_types]
+        missing_techs = [t for t in expected_tech_types if t not in weights.keys()]
+        raise ValueError(
+            f"There are incorrect or missing keys in the 'weights' dictionary.\n Incorrect techs: {incorrect_techs}\n Missing techs: {missing_techs}"
+        )
+
+    for tech in weights.keys():
+        tech_features = set(weights.get(tech).keys())
```
crispy-wonton:
```suggestion
        input_features = set(weights.get(tech).keys())
```
suggest renaming this variable so it's more readable :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,403 @@
...
+        for tech in expected_tech_types
+    ]
+
+    # Add feasibility scores as new columns to cluster_stats
+    df = df.with_columns(tech_feasibility_scores)
+
+    return df
+
+
+def assign_no_cluster_unique_code(
```
crispy-wonton:
```suggestion
def assign_df_no_cluster_unique_code(
```

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -0,0 +1,403 @@
...
+
+    plymouth_data = pl.read_parquet(
+        "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/results/plymouth_features_selected_with_clusters.parquet"
+    )
+    plymouth_data = assign_no_cluster_unique_code(plymouth_data)
+
+    # The building polygon data per cluster, and the distance to anchor loads from the centre of the cluster
+    cluster_polygons = gpd.read_file(
+        "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/merged_uprns/per_cluster_merged_polygons.geojson"
+    ).to_crs(epsg=4326)
```
crispy-wonton: How come you are converting the crs here?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/config.py
crispy-wonton: All good for now. We can change in a later PR if need to but i think the names are good :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/testing_suitability_and_feasibility_functions.py
```diff
@@ -0,0 +1,123 @@
...
+# Importing Plymouth data
+plymouth_data = pl.read_parquet(
+    "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/results/plymouth_features_selected_with_clusters.parquet"
+)
+
+# %%
+# The building polygon data per cluster, and the distance to anchor loads from the centre of the cluster
+cluster_polygons = gpd.read_file(
+    "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/merged_uprns/per_cluster_merged_polygons.geojson"
+).to_crs(epsg=4326)
```
crispy-wonton: as before - i dont think you need to conver the crs here?

crispy-wonton (commented):
Hey @sofiapinto :)

Thanks for making these changes, the code looks great - I couldn't find any faults in the logic. I think it's all working as it should be based on what we've discussed so far. There will be refinements to the logic based on the workshop with Plymouth but that will come in a later PR.

I've left some suggestions for reducing the lines of code and some minor suggestions for some of the docstrings.

Some really clean code and documentation in here! Thanks!

crispy-wonton:
> ## Bug/issue?
> I'm not sure if this is a bug, but I noticed something a bit weird that seems to be happening with the total garden size - this example should illustrate it:
> 
> Cluster 614 contains 6 properties, 5 of which are from the same building ID. However, 3 of the properties are flats and 3 aren't. The 3 flats have the same garden size. So the total garden size is calculated by the unique garden size from the flats with the same building id + the individual 3 building ID garden sizes.

As discussed offline, I don't think this is an issue. There can be cases where a building is split into flats and terraces - e.g. a house converted into 2 or more flats within a block of terraces. The flats may share a garden but not all the properties within the same building ID would be expected to share the same garden.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: I won't change this for now, so that we have a merged record of the categorisation as per the Plymouth workshop. I think we can open a separate PR later to deal with the changes. But let me know if you disagree.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: It doesn't actually rename anything now, this was left here from the previous version and I forgot to remove it.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: Good shout. I was also thinking that something else that would make sense to compute would be the percentage of homes in each cluster, so cluster_size divided by the total number of homes in analysis *100. do you have a preference?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: @lizgzil can you please clarify if this is needed? This is part of what you added. thank you!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/testing_suitability_and_feasibility_functions.py
sofiapinto: @lizgzil can you please clarify if this is needed? This is part of what you added. thank you!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: Yeah, this was added by @lizgzil, but yes I believe so, since `close_to_anchor_loads` will be a boolean that we're then transforming to float to enable multiplying by 100. (We could use proportions all over rather than percentages, if it makes it easier to understand what's happening & then we could keep close_to_anchor_loads rather than a percentage.)

But do we want this - or do we want the percentage of homes in the cluster close to anchor loads? @crispy-wonton @lizgzil

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: yep, I originally had 0 as default and then Liz created a threshold var, `outdoor_space_threshold`, but forgot to change the docs. I'll update it now.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
crispy-wonton: All good! We will def be revisiting the logic after the tree is finalised!

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
crispy-wonton: I think in general we want the whole cluster as a binary so I'm happy with 0 or 100. Something we might want to refine is how we determine if the cluster is close to an anchor property. I.e. if one property is in range, as you have it now, or does the whole cluster have to be in range. I think one to take to a sprint review.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
crispy-wonton: I think the scaled cluster size is better. They both kind of contain the same information but the scaled size is more intuitive to me and I think will scale better as well if we include other areas.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: I think this is similar to what I had originally, but it actually outputs something different. Liz spent some time looking at this methodology, to make sure it works correctly, so I'll leave her version for now and maybe later we can spend time improving the code (but making sure the computations are correct).

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: yep, this is correct. have removed the fill_null :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
lizgzil: I think this might have solved an issue at some point (potentially it helped in having the data in a format for Flourish) and from then on it's how I've read every geojson without much thought of whether it's needed!

Is there a reason to not do this? I'm not being aware that it causes anything detrimental?

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
crispy-wonton: Ok no worries, in that case it would be great to add some documentation here to the lines within the `.agg` method.

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
crispy-wonton: The main reason against converting CRS generally (if avoidable) is because it can introduce conversion errors to the coordinates. Generally, if we have datasets in British National Grid (CRS 27700) then that's the ideal format because it's a projected coordinate system (meaning coordinates projected to a flat surface) and it's specific and accurate for the British Isles. Projected CRS is less computationally demanding (i think) and more accurate to use when calculating distances. Latitude and longitude (EPSG 4326, what you have here) are geographic CRS, which is not projected.

I think probably you were converting to 4326 for plotting? A lot of the geospatial plotting/ mapping libraries require the data to be in 4326 CRS.

I think the `to_crs()` doesn't really have effect here because the later lines aren't using the geometry column anyway. I would suggest removing it to avoid confusion here.

In general, if you're having problems with loading geodataframes and CRS, you might not have `pyogrio` installed in your environment. This is the engine which `geopandas` uses by default IF it's installed. It can automatically detect the CRS of a geodata file. If you have `pyogrio` and still get errors, I think it's most likely that it's then an error in the file itself? See the `engine` kwarg in `.read_file()` here for more details: https://geopandas.org/en/v1.1.1/docs/reference/api/geopandas.read_file.html

sofiapinto: I want to keep this answer in a picture frame! ✨

sofiapinto:
hey @crispy-wonton I have now implemented your suggestions and will merge this :)

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
```diff
@@ -133,9 +138,9 @@ def create_df_suitability_categorisation(
         )
         .then(pl.lit("shared_ground_loop"))
         .otherwise(pl.lit("collective_ashp"))
+        .alias("most_suitabile_tech")
```
crispy-wonton:
```suggestion
        .alias("most_suitable_tech")
```

crispy-wonton (changes requested):
One typo to fix in a column name and then good to merge

asf_heat_pump_suitability/analysis/exploratory/create_full_dataset/assign_cluster_suitability_and_feasibility.py
sofiapinto: it wouldn't be me without a typo

crispy-wonton (approved):

PR #184, opened 2025-10-20
183 identify residential UPRNs

### LINKED ISSUE #183: Add script to filter residential UPRNs and buildings
Add script to identify residential UPRNs and building footprints as a first pass. This can be improved and scaled later.

crispy-wonton:
Fixes #183 

# Description

Data processing to identify residential UPRNs from the main OS Open UPRN dataset. 

New files:
- `getters/load_boundaries.py` - getters to load any boundary data
- `getters/load_tree_input.py` - getters to load input datasets used to create the dataset that goes into the decision tree
- `pipeline/transform/__init__.py` - required
- `pipeline/transform/non_residential_entities.py` - functions to process datasets related to identifying non-residential buildings / entities
- `pipeline/transform/uprns.py` - functions to process UPRN data and script to filter the full UPRN dataset to residential UPRNs only. NB: In writing this script (and the code in general), I have focussed only on making it work for Plymouth and similar cities (i.e. the small scale). I've added placeholder code for running it for all of GB but I haven't tested it and I imagine it will not be able to handle the large datasets. As such, this will need to be refined and developed later when we scale up.

Changed files
- `config/base.yaml` - add new data sources and constants and start restructuring to be more organised
- `getters/base_getters.py` - new function to load dataframes
- `getters/get_datasets.py` - minor change to add print statement
- `pipeline/prepare_features/lat_lon.py` - minor change to add print statement

# Instructions for Reviewer

To run the script to filter UPRNs in Plymouth, Liverpool, Portsmouth, Southampton, and Swansea to residential only, run:
`python asf_heat_pump_suitability/pipeline/transform/uprns.py --scale plymouth_similar`

- The general logic leading to the filtering to ensure we are filtering the UPRNs correctly
- Any ideas you have for making the UPRN filtering more accurate - e.g. datasets we could use or things we could do differently
- Any places in the code where you think we could make improvements to make scaling easier later. This can be a general view of the structure of the repository and how we ingest and process data as well as specific to individual functions. I've tried to keep scaling in mind as much as possible so that it's easy to change the code but please let me know if there's places where you can see it being very complicated or difficult.

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/config/base.yaml
+++ b/asf_heat_pump_suitability/config/base.yaml
@@ -1,3 +1,35 @@
+data:
+  geodata:
+    # OpenMap Local Geopackage
+    gb_os_openmap_local: "s3://asf-heat-pump-suitability/source_data/opmplc_gb.gpkg"
+    # OpenMap Local layers for certain grid squares
+    grid_square_os_openmap_local: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/v042025_OSOpenMapLocal_geometries_selected/{square}/{square}_{layer}.shp"
+
+    boundaries:
+      # UK LA boundaries
+      UK_ons_lad_bounds: "s3://asf-heat-pump-suitability/source_data/Local_Authority_Districts_December_2023_Boundaries_UK_BFE_-2600600853110041429/LAD_DEC_2023_UK_BFE.shp"
+  epc:
+    domestic: "s3://asf-daps/lakehouse/2025_Q1/processed/epc/deduplicated/processed_dedupl-0.parquet"
+    commercial: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/v092025_EPC_NonDomestic_EW.csv"
+
+constant:
+  grid_squares:
+    plymouth: "SX"
+    plymouth_similar_cities: # Liverpool, Plymouth, Portsmouth, Southampton, Swansea
+      - "SD"
+      - "SJ"
+      - "SN"
+      - "SS"
+      - "SU"
+      - "SX"
+      - "SZ"
+  plymouth_similar_cities:
+    - "Liverpool"
+    - "Plymouth"
+    - "Portsmouth"
+    - "Southampton"
+    - "Swansea"
+
 data_source:
   gb_ons_postcode_dir_url: "https://www.arcgis.com/sharing/rest/content/items/487a5ba62c8b4da08f01eb3c08e304f6/data" # Aug 2023 data
   gb_ons_postcode_dir_file_path: "Data/ONSPD_AUG_2023_UK.csv" # Aug 2023 data
--- a/asf_heat_pump_suitability/getters/base_getters.py
+++ b/asf_heat_pump_suitability/getters/base_getters.py
@@ -1,3 +1,7 @@
+"""
+Generic loaders of specific file types. These functions shouldn’t load specific datasets and can be used in multiple specific getters.
+"""
+
 import requests
 from zipfile import ZipFile
 from io import BytesIO
--- a/asf_heat_pump_suitability/getters/base_getters.py
+++ b/asf_heat_pump_suitability/getters/base_getters.py
@@ -1,11 +1,29 @@
 import requests
-import polars as pl
 from zipfile import ZipFile
 from io import BytesIO
 import logging
 import s3fs
 import geojson
 import geopandas as gpd
+from fnmatch import fnmatch
+import polars as pl
+
+
+def load_df_from_s3(uri: str, **kwargs) -> pl.DataFrame:
+    """
+    Load polars dataframe from S3.
+
+    Args:
+        uri (str): S3 URI
+        **kwargs for polars file reader.
+
+    Returns:
+        pl.DataFrame
+    """
+    if fnmatch(uri, "*.parquet"):
+        return pl.read_parquet(uri, **kwargs)
+    elif fnmatch(uri, "*.csv"):
+        return pl.read_csv(uri, **kwargs)
 
 
 def get_df_from_excel_url(url: str, **kwargs) -> pl.DataFrame:
--- a/asf_heat_pump_suitability/getters/load_boundaries.py
+++ b/asf_heat_pump_suitability/getters/load_boundaries.py
@@ -27,9 +27,16 @@ def load_gdf_local_authority_boundaries(
         return la_boundaries_gdf
     elif isinstance(select_las, str):
         print(f"Loading Local Authority boundaries for {select_las}...")
-        return la_boundaries_gdf[
+        la_boundaries_gdf = la_boundaries_gdf[
             la_boundaries_gdf["LAD23NM"].str.contains(select_las.title())
         ]
+        # Raise exception if boundaries are not found for LA specified by select_las
+        if len(la_boundaries_gdf) == 0:
+            raise Exception(
+                f"Could not find boundaries for the following Local Authority: {select_las}"
+            )
+        else:
+            return la_boundaries_gdf
     else:
         print(f"Loading Local Authority boundaries for {select_las}...")
         select_las = [la.title() for la in select_las]
--- a/asf_heat_pump_suitability/getters/load_boundaries.py
+++ b/asf_heat_pump_suitability/getters/load_boundaries.py
@@ -1,0 +1,51 @@
+import geopandas as gpd
+import shapely
+from typing import List
+
+from asf_heat_pump_suitability import config
+
+
+def load_gdf_local_authority_boundaries(
+    select_las: str | List[str] = None,
+) -> gpd.GeoDataFrame:
+    """
+    Load boundaries for specified Local Authority Districts. CRS British National Grid (27700).
+
+    Args:
+        select_las (str | List[str]): selected Local Authorities to load boundaries for. Optional. Default None to load all UK
+        Local Authority boundaries.
+
+    Returns:
+        gpd.GeoDataFrame: boundaries for specified Local Authority Districts
+    """
+    la_boundaries_gdf = gpd.read_file(
+        config["data"]["geodata"]["boundaries"]["UK_ons_lad_bounds"]
+    )
+
+    if not select_las:
+        print("Loading Local Authority boundaries for UK...")
+        return la_boundaries_gdf
+    elif isinstance(select_las, str):
+        print(f"Loading Local Authority boundaries for {select_las}...")
+        return la_boundaries_gdf[
+            la_boundaries_gdf["LAD23NM"].str.contains(select_las.title())
+        ]
+    else:
+        print(f"Loading Local Authority boundaries for {select_las}...")
+        select_las = [la.title() for la in select_las]
+        la_boundaries_gdf = la_boundaries_gdf[
+            la_boundaries_gdf["LAD23NM"].str.contains("|".join(select_las))
+        ]
+        matches = [
+            la
+            for la in select_las
+            if any(la_boundaries_gdf["LAD23NM"].str.contains(la))
+        ]
+
+        # Raise exception if any boundaries are not found for LAs in select_las
+        if len(set(select_las).difference(matches)) > 0:
+            raise Exception(
+                f"Could not find boundaries for the following Local Authorities: {set(select_las).difference(matches)}"
+            )
+        else:
+            return la_boundaries_gdf
--- a/asf_heat_pump_suitability/getters/load_boundaries.py
+++ b/asf_heat_pump_suitability/getters/load_boundaries.py
@@ -1,3 +1,7 @@
+"""
+Functions to load raw census geography boundaries, like Local Authority; ward; output areas, etc. No/minimal preprocessing occurs in these functions.
+"""
+
 import geopandas as gpd
 import shapely
 from typing import List
@@ -50,9 +54,10 @@ def load_gdf_local_authority_boundaries(
         ]
 
         # Raise exception if any boundaries are not found for LAs in select_las
-        if len(set(select_las).difference(matches)) > 0:
+        missing_las = set(select_las).difference(matches)
+        if len(missing_las) > 0:
             raise Exception(
-                f"Could not find boundaries for the following Local Authorities: {set(select_las).difference(matches)}"
+                f"Could not find boundaries for the following Local Authorities: {missing_las}"
             )
         else:
             return la_boundaries_gdf
--- a/asf_heat_pump_suitability/getters/load_tree_input.py
+++ b/asf_heat_pump_suitability/getters/load_tree_input.py
@@ -1,3 +1,7 @@
+"""
+Functions to load specific raw datasets used in decision tree pipeline using base getters and sources in config. No/minimal preprocessing occurs in these functions.
+"""
+
 import geopandas as gpd
 import pandas as pd
 from typing import Optional, List
@@ -8,12 +12,13 @@ def load_gdf_os_openmap_local_layer(
     layer: str, grid_squares: Optional[List[str]] = None, **kwargs
 ) -> gpd.GeoDataFrame:
     """
-    Load specified OS OpenMap Local layer. CRS British National Grid (27700).
+    Load specified OS OpenMap Local layer for Great Britain or optionally for a specific grid square. CRS British National Grid (27700).
+
+    Find grid square information at: https://www.ordnancesurvey.co.uk/documents/resources/guide-to-nationalgrid.pdf
 
     Args:
-        layer (str): name of layer to load.
-        grid_squares (Optional[List[str]]): names of grid squares in OS mapping for regions of Great Britain to be loaded.
-        Default None to load whole GB.
+        layer (str): name of layer to load. See layer options below.
+        grid_squares (Optional[List[str]]): names of grid squares in OS mapping for regions of Great Britain to be loaded. Default None to load whole GB.
         **kwargs for geopandas.read_file()
 
     Layer options:
@@ -50,6 +55,7 @@ def load_gdf_os_openmap_local_layer(
     else:
         if not isinstance(grid_squares, List):
             grid_squares = [grid_squares]
+        # Reformat layer name to how it appears in file name
         layer = layer.replace("_", " ").title().replace(" ", "")
         file_path = config["data"]["geodata"]["grid_square_os_openmap_local"]
         files = [file_path.format(square=code, layer=layer) for code in grid_squares]
--- a/asf_heat_pump_suitability/getters/load_tree_input.py
+++ b/asf_heat_pump_suitability/getters/load_tree_input.py
@@ -14,6 +14,7 @@ def load_gdf_os_openmap_local_layer(
         layer (str): name of layer to load.
         grid_squares (Optional[List[str]]): names of grid squares in OS mapping for regions of Great Britain to be loaded.
         Default None to load whole GB.
+        **kwargs for geopandas.read_file()
 
     Layer options:
         'building',
--- a/asf_heat_pump_suitability/getters/load_tree_input.py
+++ b/asf_heat_pump_suitability/getters/load_tree_input.py
@@ -1,0 +1,60 @@
+import geopandas as gpd
+import pandas as pd
+from typing import Optional, List
+from asf_heat_pump_suitability import config
+
+
+def load_gdf_os_openmap_local_layer(
+    layer: str, grid_squares: Optional[List[str]] = None, **kwargs
+) -> gpd.GeoDataFrame:
+    """
+    Load specified OS OpenMap Local layer. CRS British National Grid (27700).
+
+    Args:
+        layer (str): name of layer to load.
+        grid_squares (Optional[List[str]]): names of grid squares in OS mapping for regions of Great Britain to be loaded.
+        Default None to load whole GB.
+
+    Layer options:
+        'building',
+        'car_charging_point',
+        'electricity_transmission_line',
+        'foreshore',
+        'functional_site',
+        'glasshouse',
+        'important_building',
+        'motorway_junction',
+        'named_place',
+        'railway_station',
+        'railway_track',
+        'railway_tunnel',
+        'road',
+        'road_tunnel',
+        'roundabout',
+        'surface_water_area',
+        'surface_water_line',
+        'tidal_boundary',
+        'tidal_water',
+        'woodland'
+
+    Returns:
+        gpd.GeoDataFrame: OS OpenMap Local geometries for specified layer
+    """
+    if not grid_squares:
+        print(f"\nLoading OS OpenMap Local - {layer.title()}...")
+        return gpd.read_file(
+            config["data"]["geodata"]["gb_os_openmap_local"], layer=layer, **kwargs
+        )
+
+    else:
+        layer = layer.replace("_", " ").title().replace(" ", "")
+        file_path = config["data"]["geodata"]["grid_square_os_openmap_local"]
+        files = [file_path.format(square=code, layer=layer) for code in grid_squares]
+
+        gdfs = []
+
+        for file in files:
+            print(f"\nLoading OS OpenMap Local - {layer.title()} file: {file}")
+            gdfs.append(gpd.read_file(file, **kwargs))
+
+        return pd.concat(gdfs)
--- a/asf_heat_pump_suitability/getters/load_tree_input.py
+++ b/asf_heat_pump_suitability/getters/load_tree_input.py
@@ -48,6 +48,8 @@ def load_gdf_os_openmap_local_layer(
         )
 
     else:
+        if not isinstance(grid_squares, List):
+            grid_squares = [grid_squares]
         layer = layer.replace("_", " ").title().replace(" ", "")
         file_path = config["data"]["geodata"]["grid_square_os_openmap_local"]
         files = [file_path.format(square=code, layer=layer) for code in grid_squares]
--- a/asf_heat_pump_suitability/pipeline/prepare_features/lat_lon.py
+++ b/asf_heat_pump_suitability/pipeline/prepare_features/lat_lon.py
@@ -29,7 +29,7 @@ def generate_gdf_uprn_coords(
 
     Args:
         df (pl.DataFrame): dataframe with x, y coordinates in BNG (CRS: EPSG:27700) and UPRNs
-        usecols (list): columns of dataframe to use. Default None.
+        usecols (list): columns of dataframe to use. Default None to use all columns in dataframe.
         x_col (str): name of BNG x coordinate column
         y_col (str): name of BNG y coordinate column
 
--- a/asf_heat_pump_suitability/pipeline/prepare_features/lat_lon.py
+++ b/asf_heat_pump_suitability/pipeline/prepare_features/lat_lon.py
@@ -36,6 +36,7 @@ def generate_gdf_uprn_coords(
     Returns:
         gpd.GeoDataFrame: UPRNs with BNG coordinate point geometries
     """
+    print("Generating point geometries from coordinates...")
     if not usecols:
         usecols = ["*"]
     else:
--- a/asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
+++ b/asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
@@ -10,7 +10,6 @@ NO_RESIDENTIAL_OVERLAP_BUILDING_TYPES = [
     "Secondary Education",
     "Higher or University Education",
     "Primary Education",
-    "Post Office",
     "Place Of Worship",
     "Medical Care Accommodation",
     "Museum",
@@ -23,6 +22,8 @@ NO_RESIDENTIAL_OVERLAP_BUILDING_TYPES = [
     "Vehicular Ferry Terminal",
     "Hospice",
     "Bus Station",
+    "Road User Services",
+    "Passenger Ferry Terminal",
 ]
 
 
--- a/asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
+++ b/asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
@@ -1,3 +1,7 @@
+"""
+Functions to transform data related to identifying non-domestic buildings.
+"""
+
 import geopandas as gpd
 import pandas as pd
 
@@ -33,7 +37,7 @@ def transform_gdf_non_residential_buildings(
     building_gdf: gpd.GeoDataFrame,
 ) -> gpd.GeoDataFrame:
     """
-    Transform important buildings and railway data to create a dataframe of polygons representing buildings
+    Combine important buildings and railway data to create a dataframe of polygons representing buildings
     which are unlikely to contain residential properties, e.g. hospitals, train stations, museums etc.
 
     Args:
@@ -68,8 +72,8 @@ def transform_gdf_non_residential_buildings(
     ]
 
     # Get buildings which are railway stations (railway stations are only given as point geometries)
-    railway_station_gdf = building_gdf.sjoin(
-        railway_station_gdf, how="inner", predicate="contains"
+    railway_station_gdf = railway_station_gdf.sjoin(
+        building_gdf, how="inner", predicate="within"
     )
 
     return pd.concat(
--- a/asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
+++ b/asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
@@ -1,0 +1,65 @@
+import geopandas as gpd
+import pandas as pd
+
+# TODO these may need refinement for large cities where overlap with residential is possible
+NO_RESIDENTIAL_OVERLAP_BUILDING_TYPES = [
+    "Port Consisting of Docks and Nautical Berthing",
+    "Fire Station",
+    "Hospital",
+    "Non State Secondary Education",
+    "Secondary Education",
+    "Higher or University Education",
+    "Primary Education",
+    "Post Office",
+    "Place Of Worship",
+    "Medical Care Accommodation",
+    "Museum",
+    "Special Needs Education",
+    "Further Education",
+    "Non State Primary Education",
+    "Coach Station",
+    "Police Station",
+    "Sports And Leisure Centre",
+    "Vehicular Ferry Terminal",
+    "Hospice",
+    "Bus Station",
+]
+
+
+def transform_gdf_non_residential_buildings(
+    important_building_gdf: gpd.GeoDataFrame,
+    railway_station_gdf: gpd.GeoDataFrame,
+    building_gdf: gpd.GeoDataFrame,
+) -> gpd.GeoDataFrame:
+    """
+    Transform important buildings and railway data to create a dataframe of polygons representing buildings
+    which are unlikely to contain residential properties, e.g. hospitals, train stations, museums etc.
+
+    Args:
+        important_building_gdf (gpd.GeoDataFrame): OS OpenMap Local important building footprints in area of interest
+        railway_station_gdf (gpd.GeoDataFrame): OS OpenMap Local railway station point geometries in area of interest
+        building_gdf (gpd.GeoDataFrame): all building footprints in area of interest
+
+    Returns:
+        gpd.GeoDataFrame: geometries of buildings which are unlikely to contain residential properties
+    """
+    print("Creating non-residential buildings dataset...")
+    # Assert all gdfs have the same CRS
+    assert (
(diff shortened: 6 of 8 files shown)


asf_heat_pump_suitability/getters/load_boundaries.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+    if not select_las:
+        print("Loading Local Authority boundaries for UK...")
+        return la_boundaries_gdf
+    elif isinstance(select_las, str):
+        print(f"Loading Local Authority boundaries for {select_las}...")
+        return la_boundaries_gdf[
+            la_boundaries_gdf["LAD23NM"].str.contains(select_las.title())
+        ]
```
eglucas:
```suggestion
        la_boundaries_gdf = la_boundaries_gdf[
            la_boundaries_gdf["LAD23NM"].str.contains(select_las.title())
        ]

        # Raise exception if boundaries are not found for LA specified by select_las
        if len(la_boundaries_gdf) == 0:
            raise Exception(
                f"Could not find boundaries for the following Local Authorities: {select_las}"
            )
        else:
            return la_boundaries_gdf
```

In the same way it's done for no matches in a list of LAs, I suggest also raising an exception if no match is found when just a single LA string is specified

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -0,0 +1,194 @@
...
+    from asf_heat_pump_suitability.utils import save_utils
+
+    args = parse_arguments()
+
+    uprns_df = get_datasets.get_df_osopen_uprn_latlon()
+    uprns_gdf = lat_lon.generate_gdf_uprn_coords(uprns_df)
+
+    if args.scale.lower() == "plymouth":
+        print("Creating residential UPRN dataset for Plymouth Local Authority...")
+        grid_squares = config["constant"]["grid_squares"]["plymouth"]
```
eglucas, line 140:
```suggestion
        grid_squares = [config["constant"]["grid_squares"]["plymouth"]]
```

For the case when running the pipeline only for plymouth:

I get a `DataSourceError: HTTP response code: 404` when running:
`layers = {
    f"{layer}_gdf": load_tree_input.load_gdf_os_openmap_local_layer(
        layer=layer, grid_squares=grid_squares
    )
    for layer in ["important_building", "railway_station", "building"]
}`

As there's only one grid_square and it's a string, the line in `load_gdf_os_openmap_local_layer`:
`files = [file_path.format(square=code, layer=layer) for code in grid_squares]`
doesn't produce the right file path for each layer.

It seems that when `grid_squares = "SX"`, file paths are created where each character in the string is iterated over -> i.e. 
`files = ['s3://asf-heat-pump-suitability/local_heat_planning/inputs/v042025_OSOpenMapLocal_geometries_selected/S/S_ImportantBuilding.shp',
 's3://asf-heat-pump-suitability/local_heat_planning/inputs/v042025_OSOpenMapLocal_geometries_selected/X/X_ImportantBuilding.shp']`

Ensuring `grid_squares` is a list fixes the bug - I've suggested just putting square brackets here but it could also be changed in config :)

asf_heat_pump_suitability/getters/load_tree_input.py
```diff
@@ -0,0 +1,60 @@
...
+    layer: str, grid_squares: Optional[List[str]] = None, **kwargs
+) -> gpd.GeoDataFrame:
+    """
+    Load specified OS OpenMap Local layer. CRS British National Grid (27700).
+
+    Args:
+        layer (str): name of layer to load.
+        grid_squares (Optional[List[str]]): names of grid squares in OS mapping for regions of Great Britain to be loaded.
+        Default None to load whole GB.
+
```
eglucas, line 23:
```suggestion
   **kwargs for geopandas file reader
```
Suggestion for additional documentation

asf_heat_pump_suitability/getters/base_getters.py
```diff
@@ -1,11 +1,29 @@
...
+        uri (str): S3 URI
+        **kwargs for polars file reader.
+
+    Returns:
+        pl.DataFrame
+    """
+    if fnmatch(uri, "*.parquet"):
+        return pl.read_parquet(uri, **kwargs)
+    elif fnmatch(uri, "*.csv"):
+        return pl.read_csv(uri, **kwargs)
```
eglucas, line 30: Just a question for my own curiosity - I was wondering about the reasoning for using `fnmatch` rather than `.endswith(".csv")`? Is it for future flexibility?

asf_heat_pump_suitability/pipeline/prepare_features/lat_lon.py
```diff
@@ -36,6 +36,7 @@ def generate_gdf_uprn_coords(
     Returns:
         gpd.GeoDataFrame: UPRNs with BNG coordinate point geometries
     """
+    print("Generating point geometries from coordinates...")
     if not usecols:
```
eglucas, line 40: I can't comment on line 32 (description of the usecols argument) as it's not a change in this commit but I think for clarity it should be clearer that when usecols defaults to None it means that all columns are used, not that none are used

asf_heat_pump_suitability/pipeline/transform/uprns.py
eglucas: Although, I can see that you've got an output for Plymouth only in S3 which suggests this wasn't an issue when you generated that output?

eglucas (changes requested):
Really cool work, @crispy-wonton! The logic all made sense to me and found everything well documented.

I've left a few suggestions for minor improvements, but the main fix is ensuring that `grid_squares` is a list in the code block if `if args.scale.lower() == "plymouth"` in `pipeline/transform/uprns.py`.  

As we've discussed, I wasn't able to run the functions specific to loading the entire OS Open UPRNs dataset, as well as the domestic EPC dataset - so I haven't been able to test run the entire script from terminal. As far as I can tell from inspecting the code though, the way these are loaded and used looks good.

Besides needing to verify whether the Plymouth only grid square bug is indeed something to fix (since you seem to have generated an output for it without encountering this error), everything looks good to me for merging.

asf_heat_pump_suitability/getters/base_getters.py
crispy-wonton: No particular reason here, could be either, but yeh generally `fnmatch` is more flexible.

asf_heat_pump_suitability/getters/load_boundaries.py
crispy-wonton: Good idea :)

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: This is weird! Looking at the code briefly, I can't understand why it has worked for me. I will need to look into this in more detail and test it out again!

crispy-wonton: I have fixed the bug and pushed changes. Because I had two `if` statements earlier in this script to use the scaling arg to identify the grid_squares and lad_boundaries, when I was running with `--scale plymouth`, my script was going into the first `if` and then into the `else`. The `else` statement was setting the `grid_squares` to `None`, which means that the data for the whole country is loaded rather than for the individual grid square. This worked because it was never attempting to load the specific grid square for SX.

I have fixed the `else` statement and the `load_tree_input.load_gdf_os_openmap_local_layer()` function. This should now work as intended.

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -1,3 +1,35 @@
+data:
+  geodata:
+    # OpenMap Local Geopackage
+    gb_os_openmap_local: "s3://asf-heat-pump-suitability/source_data/opmplc_gb.gpkg"
+    # OpenMap Local layers for certain grid squares
+    grid_square_os_openmap_local: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/v042025_OSOpenMapLocal_geometries_selected/{square}/{square}_{layer}.shp"
+
+    boundaries:
+      # UK LA boundaries
+      UK_ons_lad_bounds: "s3://asf-heat-pump-suitability/source_data/Local_Authority_Districts_December_2023_Boundaries_UK_BFE_-2600600853110041429/LAD_DEC_2023_UK_BFE.shp"
```
sofiapinto, line 10: How are you differentiating `s3://asf-heat-pump-suitability/source_data/` from `s3://asf-heat-pump-suitability/local_heat_planning/inputs/`? i.e. what goes to which one?

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -1,3 +1,35 @@
...
+  grid_squares:
+    plymouth: "SX"
+    plymouth_similar_cities: # Liverpool, Plymouth, Portsmouth, Southampton, Swansea
+      - "SD"
+      - "SJ"
+      - "SN"
+      - "SS"
+      - "SU"
+      - "SX"
+      - "SZ"
```
sofiapinto, line 25: Where do these two letters come from? and could we have the city names together with these in a dict?

sofiapinto: Also, how would you see this changing with the scaling to other areas?

asf_heat_pump_suitability/getters/base_getters.py
sofiapinto, line 1: I know this is very minor in the grand scheme of things but `base_getters.py` does hint at what this file is for. Not a priority of course, but maybe in future we can rename this to `getter_utils.py` or perhaps just adding a line of doc at the top of the file explaining the type of functions that can found here.

Also, there's lots of files in this in this getters folders. We need to add a README.md to guide people in future :)

asf_heat_pump_suitability/getters/load_boundaries.py
sofiapinto, line 1: Could we have a line of doc at the top of the file? :)

asf_heat_pump_suitability/getters/load_boundaries.py
```diff
@@ -0,0 +1,58 @@
...
+from typing import List
+
+from asf_heat_pump_suitability import config
+
+
+def load_gdf_local_authority_boundaries(
+    select_las: str | List[str] = None,
+) -> gpd.GeoDataFrame:
+    """
+    Load boundaries for specified Local Authority Districts. CRS British National Grid (27700).
```
sofiapinto, line 16: Is this correct?
```suggestion
    Load boundaries for specified Local Authority Districts, in the CRS British National Grid (27700) coordination system.
```

asf_heat_pump_suitability/getters/load_boundaries.py
```diff
@@ -0,0 +1,58 @@
...
+        # Raise exception if boundaries are not found for LA specified by select_las
+        if len(la_boundaries_gdf) == 0:
+            raise Exception(
+                f"Could not find boundaries for the following Local Authority: {select_las}"
+            )
+        else:
+            return la_boundaries_gdf
+    else:
+        print(f"Loading Local Authority boundaries for {select_las}...")
+        select_las = [la.title() for la in select_las]
```
sofiapinto, line 46: what is la.title() doing here?

asf_heat_pump_suitability/getters/load_boundaries.py
```diff
@@ -0,0 +1,58 @@
...
+        la_boundaries_gdf = la_boundaries_gdf[
+            la_boundaries_gdf["LAD23NM"].str.contains("|".join(select_las))
```
sofiapinto, line 48: why are we joining again?

asf_heat_pump_suitability/getters/load_boundaries.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Returns:
+        gpd.GeoDataFrame: boundaries for specified Local Authority Districts
+    """
+    la_boundaries_gdf = gpd.read_file(
+        config["data"]["geodata"]["boundaries"]["UK_ons_lad_bounds"]
+    )
+
+    if not select_las:
+        print("Loading Local Authority boundaries for UK...")
```
sofiapinto, line 30: was there any reason why you've decided to use printing and not logging across these files?

asf_heat_pump_suitability/getters/load_boundaries.py
```diff
@@ -0,0 +1,58 @@
...
+            la
+            for la in select_las
+            if any(la_boundaries_gdf["LAD23NM"].str.contains(la))
+        ]
+
+        # Raise exception if any boundaries are not found for LAs in select_las
+        if len(set(select_las).difference(matches)) > 0:
+            raise Exception(
+                f"Could not find boundaries for the following Local Authorities: {set(select_las).difference(matches)}"
+            )
```
sofiapinto:
```suggestion
        not_found = set(select_las).difference(matches)
        if len(not_found) > 0:
            raise Exception(
                f"Could not find boundaries for the following Local Authorities: {not_found}"
            )
```

asf_heat_pump_suitability/getters/load_tree_input.py
```diff
@@ -0,0 +1,63 @@
...
+import pandas as pd
+from typing import Optional, List
+from asf_heat_pump_suitability import config
+
+
+def load_gdf_os_openmap_local_layer(
+    layer: str, grid_squares: Optional[List[str]] = None, **kwargs
+) -> gpd.GeoDataFrame:
+    """
+    Load specified OS OpenMap Local layer. CRS British National Grid (27700).
```
sofiapinto: is this correct?
```suggestion
    Load specified OS OpenMap Local layer, in the CRS British National Grid (27700) coordination system, for specific areas (through grid squares).
```

asf_heat_pump_suitability/getters/load_tree_input.py
```diff
@@ -0,0 +1,63 @@
+import geopandas as gpd
```
sofiapinto, line 5: Can we have a line of doc at the top of the file?

asf_heat_pump_suitability/getters/load_tree_input.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def load_gdf_os_openmap_local_layer(
+    layer: str, grid_squares: Optional[List[str]] = None, **kwargs
+) -> gpd.GeoDataFrame:
+    """
+    Load specified OS OpenMap Local layer. CRS British National Grid (27700).
+
+    Args:
+        layer (str): name of layer to load.
```
sofiapinto:
```suggestion
        layer (str): name of layer to load. See list of options below.
```

asf_heat_pump_suitability/getters/load_tree_input.py
```diff
@@ -0,0 +1,63 @@
...
+        grid_squares (Optional[List[str]]): names of grid squares in OS mapping for regions of Great Britain to be loaded.
```
sofiapinto: What's an example of a grid_square? was it those SX, etc?

asf_heat_pump_suitability/getters/load_tree_input.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    if not grid_squares:
+        print(f"\nLoading OS OpenMap Local - {layer.title()}...")
+        return gpd.read_file(
+            config["data"]["geodata"]["gb_os_openmap_local"], layer=layer, **kwargs
+        )
+
+    else:
+        if not isinstance(grid_squares, List):
+            grid_squares = [grid_squares]
```
sofiapinto, line 57: is this in the case of only one grid square being provided?

asf_heat_pump_suitability/getters/load_tree_input.py
```diff
@@ -0,0 +1,63 @@
...
+        layer = layer.replace("_", " ").title().replace(" ", "")
```
sofiapinto, line 59: perhaps line of doc explaining what's happening here?

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -0,0 +1,77 @@
+import geopandas as gpd
```
sofiapinto, line 5: Can we have a line of doc at the top of the file for this one too?

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -0,0 +1,77 @@
...
+import pandas as pd
+
+# TODO these may need refinement for large cities where overlap with residential is possible
+NO_RESIDENTIAL_OVERLAP_BUILDING_TYPES = [
```
sofiapinto, line 9: should this be in a config?

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Returns:
+        gpd.GeoDataFrame: geometries of buildings which are unlikely to contain residential properties
+    """
+    print("Creating non-residential buildings dataset...")
+    # Assert all gdfs have the same CRS
+    assert (
+        len({important_building_gdf.crs, railway_station_gdf.crs, building_gdf.crs})
+        == 1
+    )
```
sofiapinto, line 56: should we raise an exception here?

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -0,0 +1,77 @@
...
+    "Sports And Leisure Centre",
+    "Vehicular Ferry Terminal",
+    "Hospice",
+    "Bus Station",
+    "Road User Services",
+    "Passenger Ferry Terminal",
+]
+
+
+def transform_gdf_non_residential_buildings(
```
sofiapinto, line 34: the name of this function kind of hinted that there was going to be a transformation e.g. changing columns, but we're just creating a new gfd consisting of non residential buildings right?

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -0,0 +1,77 @@
...
+    ]
+
+    # Get buildings which are railway stations (railway stations are only given as point geometries)
+    railway_station_gdf = building_gdf.sjoin(
+        railway_station_gdf, how="inner", predicate="contains"
+    )
+
+    return pd.concat(
+        [exclude_buildings_gdf[["geometry"]], railway_station_gdf[["geometry"]]]
+    )
```
sofiapinto, line 81: should we keep some kind of ID beyond the geometry?

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def transform_gdf_non_residential_buildings(
+    important_building_gdf: gpd.GeoDataFrame,
+    railway_station_gdf: gpd.GeoDataFrame,
+    building_gdf: gpd.GeoDataFrame,
+) -> gpd.GeoDataFrame:
+    """
+    Transform important buildings and railway data to create a dataframe of polygons representing buildings
+    which are unlikely to contain residential properties, e.g. hospitals, train stations, museums etc.
```
sofiapinto:
```suggestion
    Creates a dataframe of polygons representing buildings which are unlikely to contain residential properties, e.g. hospitals, train stations, museums etc. based important buildings and railway.
```

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Args:
+        important_building_gdf (gpd.GeoDataFrame): OS OpenMap Local important building footprints in area of interest
+        railway_station_gdf (gpd.GeoDataFrame): OS OpenMap Local railway station point geometries in area of interest
+        building_gdf (gpd.GeoDataFrame): all building footprints in area of interest
+
+    Returns:
+        gpd.GeoDataFrame: geometries of buildings which are unlikely to contain residential properties
```
sofiapinto, line 49:
```suggestion
        gpd.GeoDataFrame: geodataframe containing only one column with geometries of buildings which are unlikely to contain residential properties
```

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -0,0 +1,194 @@
+"""
+Functions to transform UPRN data.
+
+Contains script to filter OS UPRNs to residential UPRNs only. UPRNs that meet any of the following criteria are assumed to be
+residential:
+- UPRNs geolocated inside a building footprint but not in a 'non-residential' building type (see non_residential_entities.py) or in the non-domestic EPC register
```
sofiapinto:
```suggestion
- UPRNs geolocated inside a building footprint but not in a 'non-residential' building type (see non_residential_entities.py) nor in the non-domestic EPC register
```

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -0,0 +1,194 @@
...
+- UPRNs found in the domestic EPC register
+
+To run the script:
+python asf_heat_pump_suitability/pipeline/transform/uprns.py
+
+Set the optional `scale` parameter to `plymouth` to run for Plymouth Local Authority; `plymouth_similar` to run for
+Plymouth plus four other similar Local Authorities (Liverpool, Portsmouth, Southampton, Swansea); or leave blank to run
+for all of Great Britain
```
sofiapinto: could we call them `similar_las` rather than `plymouth_similar`?

sofiapinto: so what are the options for `scale`?

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -0,0 +1,194 @@
...
+
+    Returns:
+        set: valid UPRNs from specified EPC dataset
+    """
+    print(f"Loading UPRNs from {epc_type} EPC register...")
+    df = base_getters.load_df_from_s3(config["data"]["epc"][epc_type], columns="UPRN")
+    before = len(df)
+    df = df.with_columns(
+        pl.col("UPRN").cast(pl.Float64, strict=False).cast(pl.Int64).alias("UPRN")
+    ).drop_nulls()
```
sofiapinto, line 44: can we have a line of documentation here?

sofiapinto: are we basically just removing things which aren't true UPRNs?

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -0,0 +1,194 @@
...
+        ]
+    )
+
+    # Get valid non-residential UPRNs
+    non_residential_uprns.update(load_set_valid_epc_uprns(epc_type="commercial"))
+
+    # Find UPRNs which are in any building (i.e. remove UPRNs which represent outdoor addressable locations)
+    uprns_in_buildings = set(
+        uprn_gdf.sjoin(buildings_gdf, how="inner", predicate="intersects")["UPRN"]
+    )
```
sofiapinto, line 85: I don't understand what's happening here

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -0,0 +1,194 @@
...
+
+    # Get valid residential UPRNs
+    epc_residential_uprns = load_set_valid_epc_uprns(epc_type="domestic")
+
+    return uprn_gdf[
+        (
+            # Filter to UPRNs which are in buildings AND not in non-residential UPRNs list
+            (~uprn_gdf["UPRN"].isin(non_residential_uprns))
+            & (uprn_gdf["UPRN"].isin(uprns_in_buildings))
+        )
```
sofiapinto, line 95: I'm a little confused by this

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -0,0 +1,194 @@
...
+    Create ArgumentParser and parse.
+
+    Returns:
+        argparse.Namespace: populated `Namespace`
+    """
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument(
+        "--scale",
+        help="Run script for either all of Great Britain; Plymouth only {plymouth}; or Plymouth and 4 similar local authorities {plymouth_similar}. Default to all of GB",
```
sofiapinto: i thought we had two different arguments here, not just scale. scale + plymouth_similar (which i'd propose renaming to similar_las which could be None or a list of LAs when scale != GB)

asf_heat_pump_suitability/pipeline/transform/uprns.py
sofiapinto: I was confused by this documentation originally and i though plymouth_similar and scale were two diff arguments!

asf_heat_pump_suitability/pipeline/transform/uprns.py
sofiapinto: it makes sense to have just one argument though, but the original documentation confused me.

can we call it `areas` or `local_authorities`?

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -0,0 +1,194 @@
...
+    import polars as pl
+
+    from asf_heat_pump_suitability.getters import (
+        get_datasets,
+        load_tree_input,
+        load_boundaries,
+    )
+    from asf_heat_pump_suitability.pipeline.prepare_features import lat_lon
+    from asf_heat_pump_suitability.pipeline.transform import non_residential_entities
+    from asf_heat_pump_suitability.utils import save_utils
```
sofiapinto, line 131: was there any reason for importing these here?

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -0,0 +1,194 @@
...
+    layers = {
+        f"{layer}_gdf": load_tree_input.load_gdf_os_openmap_local_layer(
+            layer=layer, grid_squares=grid_squares
+        )
+        for layer in ["important_building", "railway_station", "building"]
+    }
+
+    # Identify assumed non-residential buildings
+    non_residential_buildings_gdf = (
+        non_residential_entities.transform_gdf_non_residential_buildings(**layers)
```
sofiapinto, line 180: what happens if the order of the layers is wrong?

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -0,0 +1,194 @@
...
+        uprn_gdf=uprns_gdf,
+        buildings_gdf=layers["building_gdf"],
+        non_residential_buildings_gdf=non_residential_buildings_gdf,
+    )
+
+    # Save residential UPRNs to S3
+    df = pl.from_pandas(
+        residential_uprns_gdf[
+            ["UPRN", "X_COORDINATE", "Y_COORDINATE", "LATITUDE", "LONGITUDE"]
+        ]
```
sofiapinto, line 194: maybe add a ref of diff between (x,y) and (lat,lon) and where to use which one?

asf_heat_pump_suitability/pipeline/transform/__init__.py
sofiapinto, line 1: can we add a README to /transform in a future PR?

asf_heat_pump_suitability/getters/base_getters.py
sofiapinto: For the documentation to add at the top of the file, you can even use the descriptions that you created for the repo style guide.

sofiapinto (commented):
Hey @crispy-wonton , thanks for this!

I left some additional comments, mostly questions for my own understanding and requests of additional documentation.

I've ran the script for `plymouth` and `plymouth_similar` and both worked after your changes!

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: any new data from phase 4 should go into `s3://asf-heat-pump-suitability/local_heat_planning/inputs/`. The other folder is data from earlier phases.

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: These will only be used for small scale work. They load individual data files for a given area (a grid square) in GB.

asf_heat_pump_suitability/getters/base_getters.py
crispy-wonton: I will add docstrings in a separate PR when the repo structure is more settled.

asf_heat_pump_suitability/getters/load_boundaries.py
crispy-wonton: CRS stands for coordinate reference system :)

asf_heat_pump_suitability/getters/load_boundaries.py
crispy-wonton: Yes, I used printing because metaflow required printing and it was more reliable. I figured we can change it to logging at a later point if we think it's more useful and we want to retain the logs in a file somewhere.

asf_heat_pump_suitability/getters/load_boundaries.py
crispy-wonton: putting the strings into [title case](https://www.w3schools.com/python/ref_string_title.asp) :)

asf_heat_pump_suitability/getters/load_boundaries.py
crispy-wonton: Here we are joining the list of strings which are Local Authority names into one regex line.

asf_heat_pump_suitability/getters/load_tree_input.py
crispy-wonton: Yes! :)

crispy-wonton: https://www.ordnancesurvey.co.uk/documents/resources/guide-to-nationalgrid.pdf

asf_heat_pump_suitability/getters/load_tree_input.py
crispy-wonton: yep

asf_heat_pump_suitability/pipeline/transform/__init__.py
crispy-wonton: yes deffo!

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: I think it's easier to configure here. Maybe if we confirm the types then we can move it.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: I chose not to because if the assertion fails it will raise an error anyway and we can see where it failed.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: fair enough. there's a spatial join to identify railway station buildings so we are doing some level of transformation. I don't like using 'create' because I think it implies that we are somehow creating data rather than modifying existing data. I can think about renaming to something else.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: As above, will reword :)

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: we could (i need to check if they actually have unique IDs though).

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: I don't expect us to keep `plymouth_similar` for long. It's more of an intermediate testing parameter in case we need it. I haven't actually ended up using it yet.

I will keep it as `plymouth_similar` because I wanted it to be clear that the LAs are similar to Plymouth specifically :)

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: Yes exactly :) we're removing those rows from the EPC data which don't have a true UPRN but have the concatenated address and building reference number. Will add some doc

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: it's a dictionary with the argument names as the key so it shouldn't matter.

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: You mean as opposed to at the top of the file? If so, it was to avoid unnecessary importing when calling the functions in this file from other files.

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: We are joining UPRNs to buildings by identifying which UPRNs are within a building or on the building footprint boundary. This is an inner join so we're left with UPRNs which are in or on the boundary of buildings only and we create a `set` of them so there are no duplicates.

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: Here we are filtering to UPRNs which we want to keep. I.e. we want to filter out UPRNs which are not located within a building footprint, AND those which are also not in a commercial (non-domestic) building.

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: These are all original columns from the OS Open UPRN dataset so that's documented in their documentation :)

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: Actually, I looked at this again and I don't think so because we are filtering these buildings out anyway so the IDs and the geometries are not actually retained or saved anywhere.

asf_heat_pump_suitability/pipeline/transform/uprns.py
eglucas: Ahh that makes sense! Thanks for checking. All looks good and it's clear that it's working as intended since Sofia was able to run it :)

eglucas:
Thanks @crispy-wonton - the changes you made based on my comments all look good! Happy for you to merge.

asf_heat_pump_suitability/getters/load_boundaries.py
sofiapinto: of courseeeeee :)

asf_heat_pump_suitability/getters/load_tree_input.py
sofiapinto: can we add this link to the base.yaml or the docs somewhere for future reference?

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
sofiapinto: you did rename this in the follow up PR btw!

asf_heat_pump_suitability/pipeline/transform/uprns.py
sofiapinto: ah awesome!

asf_heat_pump_suitability/pipeline/transform/uprns.py
sofiapinto: are we linking to that documentation anywhere? if not, I think that would be a good idea. We obviously have the datasets saved on S3, but i'm just thinking of someone starting the project now, having the links available somewhere would be great.

crispy-wonton: Yes, this is currently linked in the `config/README` and will be retained when the READMEs are updated :)

asf_heat_pump_suitability/getters/load_tree_input.py
crispy-wonton: I've linked it in this function itself, you might need to refresh the page to see it :)

sofiapinto (dismissed):

eglucas (dismissed):

PR #189, opened 2025-10-27
185 improve uprn filtering

### LINKED ISSUE #185: Use POI data to improve UPRN filtering

crispy-wonton:
Fixes #185 

# Description

Further improvements to UPRN filtering to residential UPRNs only using Points of Interest (POI) dataset. The Points of Interest dataset contains other POI beyond what is found in the OS OpenMap Local Important Building layer dataset. This dataset should help us identify additional buildings which contain only non-domestic POI.

New files:
- `pipeline/transform/poi.py` - functions to transform Points of Interest data

Modified files:
- `config/base.yaml` - add new data sources and constants
- `getters/load_tree_input.py` - add new getter for POI dataset (modified from the original in `pipeline/prepare_features/anchor_properties.py`
- `pipline/transform/non_residential_entities.py` - add POI data to the processing to identify buildings which are unlikely to contain residential properties
- `pipeline/transform/uprns.py` - add POI data to the script to filter to residential UPRNs only

# Instructions for Reviewer

To run the script to filter UPRNs to those assumed to be residential, run the following line on terminal:
`python -i asf_heat_pump_suitability/pipeline/transform/uprns.py --local_authorities plymouth`

- The decision to retain any buildings which contain domestic UPRNs as identified using the domestic EPC register. I can't think of a reason not to do this but let me know if you think otherwise.
- The logic of filtering the UPRNs to make sure I'm not accidentally filtering too many UPRNs in or out

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/config/base.yaml
+++ b/asf_heat_pump_suitability/config/base.yaml
@@ -3,7 +3,7 @@ data:
     # OpenMap Local Geopackage
     gb_os_openmap_local: "s3://asf-heat-pump-suitability/source_data/opmplc_gb.gpkg"
     # OpenMap Local layers for certain grid squares
-    grid_square_os_openmap_local: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/v202510_OSOpenMapLocal_geometries_selected/{square}/{square}_{layer}.shp"
+    grid_square_os_openmap_local: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/v202510_OSOpenMapLocal_geometries_selected/{square}/{square}_{layer}.shp"
 
     boundaries:
       # UK LA boundaries
--- a/asf_heat_pump_suitability/config/base.yaml
+++ b/asf_heat_pump_suitability/config/base.yaml
@@ -3,7 +3,7 @@ data:
     # OpenMap Local Geopackage
     gb_os_openmap_local: "s3://asf-heat-pump-suitability/source_data/opmplc_gb.gpkg"
     # OpenMap Local layers for certain grid squares
-    grid_square_os_openmap_local: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/v042025_OSOpenMapLocal_geometries_selected/{square}/{square}_{layer}.shp"
+    grid_square_os_openmap_local: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/v202510_OSOpenMapLocal_geometries_selected/{square}/{square}_{layer}.shp"
 
     boundaries:
       # UK LA boundaries
@@ -27,12 +27,27 @@ constant:
       - "SU"
       - "SX"
       - "SZ"
+    sampling_areas: # Plymouth, Nottingham, Bradford, Glasgow, Manchester, Bath
+      - "NS"
+      - "SD"
+      - "SE"
+      - "SJ"
+      - "SK"
+      - "ST"
+      - "SX"
   plymouth_similar_cities:
     - "Liverpool"
     - "Plymouth"
     - "Portsmouth"
     - "Southampton"
     - "Swansea"
+  sampling_areas:
+    - "Bath and North East Somerset"
+    - "Bradford"
+    - "Glasgow City"
+    - "Manchester"
+    - "Nottingham"
+    - "Plymouth"
   target_crs: 27700 # British National Grid
 
 data_source:
--- a/asf_heat_pump_suitability/config/base.yaml
+++ b/asf_heat_pump_suitability/config/base.yaml
@@ -12,6 +12,10 @@ data:
     domestic: "s3://asf-daps/lakehouse/2025_Q1/processed/epc/deduplicated/processed_dedupl-0.parquet"
     commercial: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/v092025_EPC_NonDomestic_EW.csv"
 
+  # Preprocessed data
+  processed:
+    non_domestic_poi_categories: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/processed/non_residential_poi_categories.csv"
+
 constant:
   grid_squares:
     plymouth: "SX"
@@ -29,6 +33,7 @@ constant:
     - "Portsmouth"
     - "Southampton"
     - "Swansea"
+  target_crs: 27700 # British National Grid
 
 data_source:
   gb_ons_postcode_dir_url: "https://www.arcgis.com/sharing/rest/content/items/487a5ba62c8b4da08f01eb3c08e304f6/data" # Aug 2023 data
--- a/asf_heat_pump_suitability/getters/load_boundaries.py
+++ b/asf_heat_pump_suitability/getters/load_boundaries.py
@@ -28,19 +28,15 @@ def load_gdf_local_authority_boundaries(
     elif isinstance(select_las, str):
         print(f"Loading Local Authority boundaries for {select_las}...")
         return la_boundaries_gdf[
-            la_boundaries_gdf["LAD23NM"].str.contains(select_las.title())
+            la_boundaries_gdf["LAD23NM"].str.lower() == select_las.lower()
         ]
     else:
         print(f"Loading Local Authority boundaries for {select_las}...")
-        select_las = [la.title() for la in select_las]
         la_boundaries_gdf = la_boundaries_gdf[
-            la_boundaries_gdf["LAD23NM"].str.contains("|".join(select_las))
+            # Filter to exact LA name matches, case insensitive
+            la_boundaries_gdf["LAD23NM"].str.fullmatch("|".join(select_las), case=False)
         ]
-        matches = [
-            la
-            for la in select_las
-            if any(la_boundaries_gdf["LAD23NM"].str.contains(la))
-        ]
+        matches = set(la_boundaries_gdf["LAD23NM"].unique())
 
         # Raise exception if any boundaries are not found for LAs in select_las
         if len(set(select_las).difference(matches)) > 0:
--- a/asf_heat_pump_suitability/getters/load_tree_input.py
+++ b/asf_heat_pump_suitability/getters/load_tree_input.py
@@ -46,7 +46,7 @@ def load_gdf_os_openmap_local_layer(
             filename=config["data"]["geodata"]["gb_os_openmap_local"],
             layer=layer,
             **kwargs,
-        )
+        ).drop_duplicates(subset="ID")
 
     else:
         layer = layer.replace("_", " ").title().replace(" ", "")
@@ -59,7 +59,7 @@ def load_gdf_os_openmap_local_layer(
             print(f"\nLoading OS OpenMap Local - {layer.title()} file: {file}")
             gdfs.append(gpd.read_file(file, **kwargs))
 
-        return pd.concat(gdfs)
+        return pd.concat(gdfs).drop_duplicates(subset="ID")
 
 
 def load_gdf_poi() -> gpd.GeoDataFrame:
--- a/asf_heat_pump_suitability/getters/load_tree_input.py
+++ b/asf_heat_pump_suitability/getters/load_tree_input.py
@@ -43,7 +43,9 @@ def load_gdf_os_openmap_local_layer(
     if not grid_squares:
         print(f"Loading OS OpenMap Local - {layer.title()}...")
         return gpd.read_file(
-            config["data"]["geodata"]["gb_os_openmap_local"], layer=layer, **kwargs
+            filename=config["data"]["geodata"]["gb_os_openmap_local"],
+            layer=layer,
+            **kwargs,
         )
 
     else:
@@ -80,7 +82,7 @@ def load_gdf_poi() -> gpd.GeoDataFrame:
         "geometry",
     ]
     poi = gpd.read_file(
-        config["data_source"]["UK_poi_locations"],
+        filename=config["data_source"]["UK_poi_locations"],
         columns=required_columns,
         layer="poi_uk",
     ).to_crs("EPSG:4326")
--- a/asf_heat_pump_suitability/getters/load_tree_input.py
+++ b/asf_heat_pump_suitability/getters/load_tree_input.py
@@ -41,7 +41,7 @@ def load_gdf_os_openmap_local_layer(
         gpd.GeoDataFrame: OS OpenMap Local geometries for specified layer
     """
     if not grid_squares:
-        print(f"\nLoading OS OpenMap Local - {layer.title()}...")
+        print(f"Loading OS OpenMap Local - {layer.title()}...")
         return gpd.read_file(
             config["data"]["geodata"]["gb_os_openmap_local"], layer=layer, **kwargs
         )
@@ -58,3 +58,31 @@ def load_gdf_os_openmap_local_layer(
             gdfs.append(gpd.read_file(file, **kwargs))
 
         return pd.concat(gdfs)
+
+
+def load_gdf_poi() -> gpd.GeoDataFrame:
+    """
+    Load and process Points of Interest data. CRS EPSG 4326.
+
+    Returns:
+        gpd.GeoDataFrame: Processed POI data containing types of POI specified
+
+    Raises:
+        ValueError: If required columns are missing
+    """
+    print("Loading POI data...")
+
+    required_columns = [
+        "id",
+        "country",
+        "main_category",
+        "alternate_category",
+        "geometry",
+    ]
+    poi = gpd.read_file(
+        config["data_source"]["UK_poi_locations"],
+        columns=required_columns,
+        layer="poi_uk",
+    ).to_crs("EPSG:4326")
+    print(f"POI CRS: {poi.crs}")
+    return poi
--- a/asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
+++ b/asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
@@ -64,7 +64,7 @@ def generate_gdf_non_residential_buildings(
             }
         )
         == 1
-    )
+    ), "All GeoDataFrame inputs must have the same CRS"
 
     # Find important building classification column name
     col = None
--- a/asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
+++ b/asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
@@ -1,6 +1,8 @@
 import geopandas as gpd
 import pandas as pd
 
+from asf_heat_pump_suitability.pipeline.transform import uprns
+
 # TODO these may need refinement for large cities where overlap with residential is possible
 NO_RESIDENTIAL_OVERLAP_BUILDING_TYPES = [
     "Port Consisting of Docks and Nautical Berthing",
@@ -27,14 +29,15 @@ NO_RESIDENTIAL_OVERLAP_BUILDING_TYPES = [
 ]
 
 
-def transform_gdf_non_residential_buildings(
+def generate_gdf_non_residential_buildings(
     important_building_gdf: gpd.GeoDataFrame,
     railway_station_gdf: gpd.GeoDataFrame,
     poi_gdf: gpd.GeoDataFrame,
     building_gdf: gpd.GeoDataFrame,
+    uprns_gdf: gpd.GeoDataFrame,
 ) -> gpd.GeoDataFrame:
     """
-    Transform important buildings, railway station, and points of interest data to create a dataframe of polygons representing buildings
+    Use important buildings, railway station, and points of interest data to create a dataframe of polygons representing buildings
     which are unlikely to contain residential properties, e.g. hospitals, train stations, museums etc. Important buildings
     and POI geopoints should include only types which are unlikely to be in mixed-use (residential and commercial) buildings.
 
@@ -43,6 +46,7 @@ def transform_gdf_non_residential_buildings(
         railway_station_gdf (gpd.GeoDataFrame): OS OpenMap Local railway station point geometries in area of interest
         poi_gdf (gpd.GeoDataFrame): non-domestic Points of Interest geopoints in area of interest
         building_gdf (gpd.GeoDataFrame): all building footprints in area of interest
+        uprns_gdf (gpd.GeoDataFrame): UPRNs with point geometries in area of interest
 
     Returns:
         gpd.GeoDataFrame: geometries of buildings which are unlikely to contain residential properties
@@ -56,6 +60,7 @@ def transform_gdf_non_residential_buildings(
                 railway_station_gdf.crs,
                 poi_gdf.crs,
                 building_gdf.crs,
+                uprns_gdf.crs,
             }
         )
         == 1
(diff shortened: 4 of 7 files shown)


asf_heat_pump_suitability/config/base.yaml
```diff
@@ -12,6 +12,10 @@ data:
     domestic: "s3://asf-daps/lakehouse/2025_Q1/processed/epc/deduplicated/processed_dedupl-0.parquet"
     commercial: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/v092025_EPC_NonDomestic_EW.csv"
 
+  # Preprocessed data
+  processed:
+    non_domestic_poi_categories: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/processed/non_residential_poi_categories.csv"
```
sofiapinto, line 17: Let me know if i'm wrong but isn't there another file in the repo that needs to be updated that has info on all datasets we're using?

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -23,12 +27,28 @@ constant:
...
       - "SX"
       - "SZ"
+    sampling_areas: # Plymouth, Nottingham, Bradford, Glasgow, Manchester, Bath
+      - "NS"
+      - "SD"
+      - "SE"
+      - "SJ"
+      - "SK"
+      - "ST"
+      - "SX"
```
sofiapinto, line 37: similar comment to the one i've made in the previous PR. can we have a dict mapping these letters to the name of the areas. and also what exactly are these? how do we find these?

asf_heat_pump_suitability/getters/load_boundaries.py
```diff
@@ -28,19 +28,15 @@ def load_gdf_local_authority_boundaries(
...
-        select_las = [la.title() for la in select_las]
         la_boundaries_gdf = la_boundaries_gdf[
-            la_boundaries_gdf["LAD23NM"].str.contains("|".join(select_las))
-        ]
-        matches = [
-            la
-            for la in select_las
-            if any(la_boundaries_gdf["LAD23NM"].str.contains(la))
+            # Filter to exact LA name matches, case insensitive
+            la_boundaries_gdf["LAD23NM"].str.fullmatch("|".join(select_las), case=False)
```
sofiapinto, line 48: ah these changes make sense! I made some comments in the previous PR about this.

asf_heat_pump_suitability/getters/load_tree_input.py
```diff
@@ -58,3 +58,31 @@ def load_gdf_os_openmap_local_layer(
             gdfs.append(gpd.read_file(file, **kwargs))
 
         return pd.concat(gdfs)
+
+
+def load_gdf_poi() -> gpd.GeoDataFrame:
+    """
+    Load and process Points of Interest data. CRS EPSG 4326.
```
sofiapinto, line 76:
```suggestion
    Load and process Points of Interest data, in CRS EPSG 4326 coordinate system.
```

asf_heat_pump_suitability/getters/load_tree_input.py
```diff
@@ -58,3 +58,31 @@ def load_gdf_os_openmap_local_layer(
...
+
+    required_columns = [
+        "id",
+        "country",
+        "main_category",
+        "alternate_category",
+        "geometry",
+    ]
+    poi = gpd.read_file(
+        config["data_source"]["UK_poi_locations"],
```
sofiapinto: maybe you could add the parameter name, since you do for the remaining two :)

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -27,29 +29,42 @@
 ]
 
 
-def transform_gdf_non_residential_buildings(
+def generate_gdf_non_residential_buildings(
```
sofiapinto, line 36: I made a comment about this in the previous PR. :) I think it makes more sense to call it generate too!

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -27,29 +29,42 @@
...
+    poi_gdf: gpd.GeoDataFrame,
     building_gdf: gpd.GeoDataFrame,
+    uprns_gdf: gpd.GeoDataFrame,
 ) -> gpd.GeoDataFrame:
     """
-    Transform important buildings and railway data to create a dataframe of polygons representing buildings
-    which are unlikely to contain residential properties, e.g. hospitals, train stations, museums etc.
+    Use important buildings, railway station, and points of interest data to create a dataframe of polygons representing buildings
+    which are unlikely to contain residential properties, e.g. hospitals, train stations, museums etc. Important buildings
+    and POI geopoints should include only types which are unlikely to be in mixed-use (residential and commercial) buildings.
```
sofiapinto, line 46: Is there an overlap between POI and important buildings? i guess even if there is, that is not problematic as we're just using them to exclude buildings, not to add (So no problem with duplicates)

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -27,29 +29,42 @@
...
+                important_building_gdf.crs,
+                railway_station_gdf.crs,
+                poi_gdf.crs,
+                building_gdf.crs,
+                uprns_gdf.crs,
+            }
+        )
         == 1
-    )
+    ), "All GeoDataFrame inputs must have the same CRS"
```
sofiapinto, line 71: as per the previous PR, we might want to raise an exception?

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -66,12 +81,35 @@ def transform_gdf_non_residential_buildings(
...
+        [
+            exclude_buildings_gdf[["geometry"]],
+            railway_station_gdf[["geometry"]],
+            poi_buildings_gdf[["geometry"]],
+        ]
+    )
+
+    # Get locations of UPRNs in domestic EPC
+    include_uprns = uprns.load_set_valid_epc_uprns(epc_type="domestic")
+    uprns_gdf = uprns_gdf[uprns_gdf["UPRN"].isin(include_uprns)].copy()
```
sofiapinto, line 105: why are we using copy here? just for my own understanding

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -66,12 +81,35 @@ def transform_gdf_non_residential_buildings(
...
+
+    # Find buildings that contain a domestic UPRN
+    exclude_buildings_gdf = exclude_buildings_gdf.sjoin(
```
sofiapinto, line 108: shall we call it non_residential_buildings_gdf?

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -66,12 +81,35 @@ def transform_gdf_non_residential_buildings(
     exclude_buildings_gdf = important_building_gdf[
```
sofiapinto, line 85: shall we call it non_residential_important_buildings or similar?

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -66,12 +81,35 @@ def transform_gdf_non_residential_buildings(
...
+    uprns_gdf = uprns_gdf[uprns_gdf["UPRN"].isin(include_uprns)].copy()
+
+    # Find buildings that contain a domestic UPRN
+    exclude_buildings_gdf = exclude_buildings_gdf.sjoin(
+        uprns_gdf[["UPRN", "geometry"]], how="left", predicate="contains"
     )
+
+    # Filter to buildings that don't contain a domestic UPRN
+    exclude_buildings_gdf = exclude_buildings_gdf[
+        exclude_buildings_gdf["UPRN"].isnull()
```
sofiapinto, line 114: I guess this will:
- delete buildings where all UPRNs exist
- however if for some reason there's a building with at least one missing UPRN then that will count as a building to exclude - do we perhaps want the opposite? i.e. to only consider buildings if all the UPRNs are missing?

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -66,12 +81,35 @@ def transform_gdf_non_residential_buildings(
...
+    ].drop(columns=["UPRN", "index_right"])
+
+    # Normalize to drop duplicate building footprints
+    exclude_buildings_gdf["geometry"] = exclude_buildings_gdf.normalize()
+    return exclude_buildings_gdf.drop_duplicates()
```
sofiapinto, line 119: why are geometries different for the same building?
and wouldn't it be enough to drop by building ID, if there is one in this df?

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -1,6 +1,8 @@
 import geopandas as gpd
```
sofiapinto, line 5: can we have a little docstring at the top?

asf_heat_pump_suitability/pipeline/transform/poi.py
```diff
@@ -0,0 +1,49 @@
+import geopandas as gpd
```
sofiapinto, line 5: can we have a little docstring at the top?

asf_heat_pump_suitability/pipeline/transform/poi.py
```diff
@@ -0,0 +1,49 @@
...
+import csv
+import smart_open
+
+from asf_heat_pump_suitability import config
+
+
+def transform_gdf_poi(
+    poi: gpd.GeoDataFrame,
+    filter_categories: Iterable[str] = None,
+    target_crs: str | int = config["constant"]["target_crs"],
```
sofiapinto, line 16: nice, I was reading about this and I see that using Union is kind of out date now :)

asf_heat_pump_suitability/pipeline/transform/poi.py
```diff
@@ -0,0 +1,49 @@
...
+    Args:
+        poi (gpd.GeoDataFrame): Points Of Interest dataset
+        filter_categories (List[str]): types of POI to filter the dataframe for from the `main_category` column. Default None to load all categories.
+        target_crs (str | int): coordinate reference system to reproject data to
+
+    Returns:
+        gpd.GeoDataFrame: Processed POI data containing types of POI specified
+    """
+    # Filter to Great Britain
+    poi = poi[poi.country == "GB"].copy()
```
sofiapinto, line 30: again, any reason for creating a copy here?

asf_heat_pump_suitability/pipeline/transform/poi.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def load_set_non_domestic_poi_categories() -> set:
+    """
+    Load set of Points of Interest categories which are unlikely to be located in buildings with domestic properties.
+
+    Returns:
+        set: non-domestic POI categories
+    """
+    with smart_open.open(
```
sofiapinto, line 49: uhh what is the reason for using smart_open? never used it before

asf_heat_pump_suitability/pipeline/transform/poi.py
sofiapinto, line 1: i thought we would save some data here and hence why we have "s3://asf-heat-pump-suitability/local_heat_planning/inputs/processed/non_residential_poi_categories.csv". What am i missing?

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -139,10 +147,10 @@ def parse_arguments() -> argparse.Namespace:
         uprns_gdf = uprns_gdf.sjoin(
             la_boundaries_gdf[["LAD23CD", "LAD23NM", "geometry"]],
             how="inner",
-            predicate="within",
+            predicate="intersects",
         ).drop(columns="index_right")
 
-    if args.scale.lower() == "plymouth_similar":
+    elif args.scale.lower() == "plymouth_similar":
```
sofiapinto: uff i missed this if that should've been an elif in the previous PR! good you found it.

sofiapinto: in general we could just have a list of areas as input and deal with it accordingly.

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -139,10 +147,10 @@ def parse_arguments() -> argparse.Namespace:
         uprns_gdf = uprns_gdf.sjoin(
             la_boundaries_gdf[["LAD23CD", "LAD23NM", "geometry"]],
             how="inner",
-            predicate="within",
+            predicate="intersects",
```
sofiapinto, line 154: can you guide me through the change here to make sure i understand it correctly? for some reason these sjoins and predicates always confuse me

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -153,7 +161,21 @@ def parse_arguments() -> argparse.Namespace:
...
+        ).drop(columns="index_right")
+
+    elif args.scale.lower() == "sampling_areas":
```
sofiapinto: I think we can do this in a nicer way without so many elif's, but you're also aware of it from your TODO comment above.

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -153,7 +161,21 @@ def parse_arguments() -> argparse.Namespace:
...
+            "Creating residential UPRN dataset for Bath, Bradford, Glasgow, Manchester, Nottingham, and Plymouth Local Authorities..."
+        )
+        grid_squares = config["constant"]["grid_squares"]["sampling_areas"]
+        la_boundaries_gdf = load_boundaries.load_gdf_local_authority_boundaries(
+            select_las=config["constant"]["sampling_areas"]
+        )
+        uprns_gdf = uprns_gdf.sjoin(
+            la_boundaries_gdf[["LAD23CD", "LAD23NM", "geometry"]],
+            how="inner",
+            predicate="intersects",
```
sofiapinto, line 182: btw i think this part is the same across all if's and elif's so it could just appear once outside of the if-elif-else statements.

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -172,7 +200,9 @@ def parse_arguments() -> argparse.Namespace:
 
     # Identify assumed non-residential buildings
     non_residential_buildings_gdf = (
-        non_residential_entities.transform_gdf_non_residential_buildings(**layers)
+        non_residential_entities.generate_gdf_non_residential_buildings(
+            **layers, poi_gdf=poi_gdf, uprns_gdf=uprns_gdf
```
sofiapinto, line 208: again, wouldn't it be better to pass arguments separately rather than using **layers? i'm afraid if inputs are in the wrong order, things might be incorrect. but let me know if this would never be the case.

sofiapinto (commented):
Hello @crispy-wonton ,

Thanks for this PR! The changes make lots of sense and reflect some of the comments i left in the other PR.

I left some questions & documentation requests, but haven't found any errors in the logic.

sofiapinto:
Forgot to say that I haven't re-run the code - can do that after you merge the other branch and resolve the conflicts :)

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: Yes there is. I will add READMEs in a separate PR. This is the same dataset we were using as before.

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: Answered in #184

asf_heat_pump_suitability/getters/load_tree_input.py
crispy-wonton: As before, CRS stands for coordinate reference system. I will make sure to state this in a README :)

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: Added in #184

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: There is most likely some overlap, but yes as you said, we are using them to filter out buildings so I don't think it's an issue :)

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: I think this is enough. It will error out and then a user would see this error message.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: As `important_building` is the name of the layer from the OS geospatial file, I'll call it the same here for consistency and clarity :)

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: I was getting a `SettingWithCopyWarning` here so I added the `.copy()` to avoid unintentionally modifying the original DataFrame object.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: I find `exclude_buildings` more intuitive :)

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: I don't think this can happen but your comment did make me doubt it so please double check my thinking which I list here:
- we start with a gdf of building polygons that we want to exclude because we think they are not domestic. Each row is a unique building.
- we conduct a spatial join with UPRNs. UPRNs which are inside a building will be joined to the relevant building. We are doing a left join of the UPRNs onto the buildings.
- This means we will end up with a gdf which contains a combination of: buildings joined to one or more UPRNs (i.e. we want to keep the UPRNs in these buildings); AND buildings that did not get joined to anything and have a null value (we want to exclude these).
- if we filter this gdf to only the buildings with null values, we will keep only buildings which are not joined to any UPRNs.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: The dataframe should already be deduplicated by ID. However, this is an extra check in case there are two copies of the same building which have different IDs. you have to `normalize` any geometry first before you can deduplicate on it because sometimes the same geometry can have the coordinates in a slightly different order. Normalising them corrects for this.

asf_heat_pump_suitability/pipeline/transform/poi.py
crispy-wonton: As before, it's probably avoiding the SettingWithCopyWarning but I didn't write this function originally, I just updated it and moved it here.

asf_heat_pump_suitability/pipeline/transform/poi.py
crispy-wonton: I can't remember the exact reason now. It was related to me wanting to stream a csv file from S3 as a list and I came across it in that context.

asf_heat_pump_suitability/pipeline/transform/poi.py
crispy-wonton: That file just contains a list of strings which represent categories of points of interest which are unlikely to appear in buildings which contain domestic properties. It is probably worth reviewing. It wasn't created with code, it was created using a combination of manual selection and ChatGPT with manual review. I think there were about 8000 categories to choose from.

I will document this in the repository - it's already documented in the sprint review slides as well.

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: No worries :)
- `predicate=within` here means that the UPRN point must be fully contained within the boundaries of the local authority to be joined to it
- `predicate=intersects` here means that the UPRN point must either be fully contained or on the boundary of the local authority to be joined to it.

I don't think it really matters much here because we are just looking at one LA of Plymouth. When we scale, we need to be careful with these joins that we are a) not excluding any UPRNs which fall directly on boundaries; or b) duplicating UPRNs and joining them to multiple Local Authorities if they are on a boundary.

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: If we want a list of areas, we need to download all the separate grid square files (there are quite a few), which is why I haven't done it this way. But yeh we could write a script to stream the files to S3 and automate that if it's necessary later.

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: Yes I don't think this will be like this at the end. It's just to move forward with what we need for now.

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: It doesn't appear in the `else` statement otherwise I would do this :) as I said before, I'm not too worried about this part at the moment because I intend for it to be tidied later

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: The `**layers` unpacks the dictionary into a list of named arguments, so it doesn't matter the order because it will enter them with the key names from the dict. See more info here: https://docs.python.org/3/tutorial/controlflow.html#unpacking-argument-lists

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
sofiapinto: I meant `non_residential_important_buildings` rather than `exclude_buildings_gdf`. Since it's the `important_building_gdf` after exclusion. Feel free to keep it if you wish, I guess my problem was with `exclude_buildings_gdf` which is used for different purposes and it gets a little confusing. Nothing special, it just takes longer to read.

asf_heat_pump_suitability/pipeline/transform/poi.py
sofiapinto: was this a phase 3 thing?

asf_heat_pump_suitability/pipeline/transform/uprns.py
sofiapinto: sounds great!

asf_heat_pump_suitability/pipeline/transform/uprns.py
sofiapinto: yep yep, I didn't see this was a dict, my bad

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
sofiapinto: yes, correct!

asf_heat_pump_suitability/pipeline/transform/uprns.py
sofiapinto: SO IMPORTANT, thanks for flagging! What is the best place to keep track beyond to make sure we don't forget to check it?

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: I should also say, I can't really make this into a dictionary because it isn't a 1-1 mapping. E.g. some LAs are in the same grid square, and other LAs are spread across multiple grid squares.

sofiapinto (dismissed):

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: Maybe Asana, I'll add a task to the backlog.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: Ah my bad I see what you mean. I think because we are just appending to `exclude_buildings_gdf` i'm happy to keep it as the same name throughout the function. I find it more confusing to have lots of variable names within a function.

asf_heat_pump_suitability/pipeline/transform/poi.py
crispy-wonton: No I did this in the last sprint or the one before.

crispy-wonton: My bad it was 1800 categories, not 8000.

sofiapinto (approved):

PR #190, opened 2025-11-07
188 impute property type (flats)

### LINKED ISSUE #188: Refactor property type imputation
Refactor from `create_full_dataset_plymouth.py/ipynb`

crispy-wonton:
Fixes #188 

# Description

Add functions and script to impute boolean flag for whether an individual UPRN is a flat or not.

New files:
- `asf_heat_pump_suitability/getters/load_geodata.py` - new file to add getters for geodata specifically
- `asf_heat_pump_suitability/pipeline/impute/property_type.py` - functions to impute property type
- `asf_heat_pump_suitability/pipeline/run/flat_blocks.py` - script to impute flat property type
- `asf_heat_pump_suitability/pipeline/transform/uprns.py` - functions to transform OS Open UPRN data

Modified files:
- `asf_heat_pump_suitability/config/base.yaml` - added latest OS Open UPRN file to source data

# Instructions for Reviewer

Run the following line in the terminal:
`python asf_heat_pump_suitability/pipeline/run/flat_blocks.py --uprns s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_residential_uprns.parquet`

This will run the script to take the domestic UPRNs for Plymouth (which we have preprocessed in a separate PR #184) and add a boolean flag to say whether or not the UPRN is a flat.

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/config/base.yaml
+++ b/asf_heat_pump_suitability/config/base.yaml
@@ -1,3 +1,6 @@
+data:
+  geodata:
+    uk_osopen_uprn: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/osopenuprn_202510_csv.zip"
 data_source:
   gb_ons_postcode_dir_url: "https://www.arcgis.com/sharing/rest/content/items/487a5ba62c8b4da08f01eb3c08e304f6/data" # Aug 2023 data
   gb_ons_postcode_dir_file_path: "Data/ONSPD_AUG_2023_UK.csv" # Aug 2023 data
--- a/asf_heat_pump_suitability/getters/load_geodata.py
+++ b/asf_heat_pump_suitability/getters/load_geodata.py
@@ -12,7 +12,7 @@ def load_df_osopen_uprn(**kwargs) -> pl.DataFrame:
     coordinates for all UPRNs in Great Britain.
 
     Args:
-        **kwargs fo pl.read_csv
+        **kwargs for pl.read_csv
 
     Returns:
         pl.DataFrame: raw OS Open UPRN dataset with lat/lon and x/y coordinates for every UPRN
@@ -20,10 +20,10 @@ def load_df_osopen_uprn(**kwargs) -> pl.DataFrame:
     print("Loading OSOpen UPRNs...")
     path = config["data"]["geodata"]["uk_osopen_uprn"]
     # extract_file = re.search(r"osopenuprn_\d{6}", path).group()
-    extract_file = os.path.basename(path).split("_csv")[0]
+    filename = os.path.basename(path).split("_csv")[0]
     df = base_getters.get_df_from_zip_csv_s3(
         path,
-        extract_file=f"{extract_file}.csv",
+        extract_file=f"{filename}.csv",
         **kwargs,
     )
 
--- a/asf_heat_pump_suitability/getters/load_geodata.py
+++ b/asf_heat_pump_suitability/getters/load_geodata.py
@@ -1,0 +1,30 @@
+import polars as pl
+import regex as re
+import os
+
+from asf_heat_pump_suitability import config
+from asf_heat_pump_suitability.getters import base_getters
+
+
+def load_df_osopen_uprn(**kwargs) -> pl.DataFrame:
+    """
+    Get raw OS (Ordnance Survey) Open UPRN dataset containing latitude and longitude and British National Grid X and Y
+    coordinates for all UPRNs in Great Britain.
+
+    Args:
+        **kwargs fo pl.read_csv
+
+    Returns:
+        pl.DataFrame: raw OS Open UPRN dataset with lat/lon and x/y coordinates for every UPRN
+    """
+    print("Loading OSOpen UPRNs...")
+    path = config["data"]["geodata"]["uk_osopen_uprn"]
+    # extract_file = re.search(r"osopenuprn_\d{6}", path).group()
+    extract_file = os.path.basename(path).split("_csv")[0]
+    df = base_getters.get_df_from_zip_csv_s3(
+        path,
+        extract_file=f"{extract_file}.csv",
+        **kwargs,
+    )
+
+    return df
--- a/asf_heat_pump_suitability/getters/load_geodata.py
+++ b/asf_heat_pump_suitability/getters/load_geodata.py
@@ -19,7 +19,6 @@ def load_df_osopen_uprn(**kwargs) -> pl.DataFrame:
     """
     print("Loading OSOpen UPRNs...")
     path = config["data"]["geodata"]["uk_osopen_uprn"]
-    # extract_file = re.search(r"osopenuprn_\d{6}", path).group()
     filename = os.path.basename(path).split("_csv")[0]
     df = base_getters.get_df_from_zip_csv_s3(
         path,
--- a/asf_heat_pump_suitability/pipeline/impute/property_type.py
+++ b/asf_heat_pump_suitability/pipeline/impute/property_type.py
@@ -1,0 +1,25 @@
+import geopandas as gpd
+
+
+def impute_set_flat_properties(uprns_gdf: gpd.GeoDataFrame) -> set:
+    """
+    Identify domestic UPRNs which are flats from their geometries.
+
+    Args:
+        uprns_gdf (gpd.GeoDataFrame): all domestic UPRNs and their point geometries in area of interest
+
+    Returns:
+        set: UPRNs which are flats / apartments
+    """
+    # Count how many times each geometry occurs
+    geom_counts = uprns_gdf["geometry"].value_counts()
+
+    # Get the geometries that appear more than once (the index is the geometry)
+    duplicate_geoms = geom_counts[geom_counts > 1].index
+
+    # Filter the GeoDataFrame to only those geometries
+    flats = set(uprns_gdf[uprns_gdf["geometry"].isin(duplicate_geoms)]["UPRN"])
+    print(
+        f"{len(flats)} flats found in UPRN dataset, N={len(uprns_gdf)}, {round(len(flats)/len(uprns_gdf)*100, 2)}%"
+    )
+    return flats
--- a/asf_heat_pump_suitability/pipeline/run/flat_blocks.py
+++ b/asf_heat_pump_suitability/pipeline/run/flat_blocks.py
@@ -2,7 +2,7 @@
 Script to label UPRNs with flat / apartment property type boolean flag.
 
 Run:
-python asf_heat_pump_suitability/pipeline/run/impute.py
+python asf_heat_pump_suitability/pipeline/run/flat_blocks.py
 """
 
 import argparse
@@ -29,12 +29,12 @@ def parse_arguments() -> argparse.Namespace:
 
 
 if __name__ == "__main__":
-    import polars as pl
     import os
+    import polars as pl
 
-    from asf_heat_pump_suitability.pipeline.transform import uprns
-    from asf_heat_pump_suitability.pipeline.impute import property_type
     from asf_heat_pump_suitability.utils import save_utils
+    from asf_heat_pump_suitability.pipeline.impute import property_type
+    from asf_heat_pump_suitability.pipeline.transform import uprns
 
     args = parse_arguments()
 
@@ -55,5 +55,5 @@ if __name__ == "__main__":
 
     save_utils.save_to_s3(
         uprns_df,
-        path=f"s3://asf-heat-pump-suitability/local_heat_planning/outputs/{os.path.basename(args.uprns)}_with_flats.parquet",
+        path=f"s3://asf-heat-pump-suitability/local_heat_planning/outputs/{os.path.basename(args.uprns).split('.')[0]}_with_flats.parquet",
     )
--- a/asf_heat_pump_suitability/pipeline/run/flat_blocks.py
+++ b/asf_heat_pump_suitability/pipeline/run/flat_blocks.py
@@ -2,7 +2,7 @@
 Script to label UPRNs with flat / apartment property type boolean flag.
 
 Run:
-python asf_heat_pump_suitability/pipeline/run/flat_blocks.py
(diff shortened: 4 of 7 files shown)


asf_heat_pump_suitability/config/base.yaml
sofiapinto, line 1: we might need to go through this base.yaml and reorganise it. What do you think? Just think about these first 6 lines of code I find it confusing that we have `data` and then `data_source`.

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -1,3 +1,6 @@
+data:
+  geodata:
+    uk_osopen_uprn: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/osopenuprn_202510_csv.zip"
```
sofiapinto, line 3: is this the one which is updated every couple of months? hence why we have a date attached to the filename?

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -0,0 +1,30 @@
...
+from asf_heat_pump_suitability.getters import base_getters
+
+
+def load_df_osopen_uprn(**kwargs) -> pl.DataFrame:
+    """
+    Get raw OS (Ordnance Survey) Open UPRN dataset containing latitude and longitude and British National Grid X and Y
+    coordinates for all UPRNs in Great Britain.
+
+    Args:
+        **kwargs fo pl.read_csv
```
sofiapinto:
```suggestion
        **kwargs for pl.read_csv
```

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -0,0 +1,30 @@
...
+
+    Returns:
+        pl.DataFrame: raw OS Open UPRN dataset with lat/lon and x/y coordinates for every UPRN
+    """
+    print("Loading OSOpen UPRNs...")
+    path = config["data"]["geodata"]["uk_osopen_uprn"]
+    # extract_file = re.search(r"osopenuprn_\d{6}", path).group()
```
sofiapinto: can this be removed?

asf_heat_pump_suitability/config/base.yaml
sofiapinto: have we renamed this ourselves, or is this how these files are always named?

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -0,0 +1,30 @@
...
+    extract_file = os.path.basename(path).split("_csv")[0]
```
sofiapinto: if we expect multiple versions of this, we might want to be checking if the filename always ends like this.

sofiapinto:
```suggestion
    extract_file_name = os.path.basename(path).split("_csv")[0]
```

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -0,0 +1,30 @@
...
+    df = base_getters.get_df_from_zip_csv_s3(
+        path,
+        extract_file=f"{extract_file}.csv",
```
sofiapinto:
```suggestion
        extract_file=f"{extract_file_name}.csv",
```

asf_heat_pump_suitability/pipeline/impute/property_type.py
```diff
@@ -0,0 +1,25 @@
+import geopandas as gpd
+
+
+def impute_set_flat_properties(uprns_gdf: gpd.GeoDataFrame) -> set:
```
sofiapinto, line 4: I'm guessing there can't be cases where the same building has different geometries (unless it's from two different versions as we've been talking about)

asf_heat_pump_suitability/pipeline/impute/property_type.py
```diff
@@ -0,0 +1,25 @@
...
+        uprns_gdf (gpd.GeoDataFrame): all domestic UPRNs and their point geometries in area of interest
+
+    Returns:
+        set: UPRNs which are flats / apartments
+    """
+    # Count how many times each geometry occurs
+    geom_counts = uprns_gdf["geometry"].value_counts()
+
+    # Get the geometries that appear more than once (the index is the geometry)
+    duplicate_geoms = geom_counts[geom_counts > 1].index
```
sofiapinto, line 18: So basically any property with a geometry that is common to properties is being assigned as a flat?

asf_heat_pump_suitability/pipeline/run/flat_blocks.py
```diff
@@ -0,0 +1,59 @@
...
+    # Load UPRN data
+    print(f"Loading domestic UPRNs from: {args.uprns}")
+    uprns_df = pl.read_parquet(
+        args.uprns, columns=["UPRN", "X_COORDINATE", "Y_COORDINATE"]
+    )
+
+    # Get geopoints of UPRNs
+    uprns_gdf = uprns.generate_gdf_uprn_coords(df=uprns_df)
+
+    # Label flats
```
sofiapinto:
```suggestion
    # Create boolean column called `property_type_flat` to identify flats
```

asf_heat_pump_suitability/pipeline/impute/property_type.py
sofiapinto: But I guess this won't capture flats in the same building with different footprints/geometries?

sofiapinto: and what about multiple terraces if they share a single footprint?

sofiapinto: maybe sharing a single footprint doesn't mean sharing the same geometry? sorry I think I am a little confused :)

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -0,0 +1,39 @@
...
+        df (pl.DataFrame): dataframe with x, y coordinates in BNG (CRS: EPSG:27700) and UPRNs
+        usecols (list): columns of dataframe to use. Default None.
+        x_col (str): name of BNG x coordinate column
+        y_col (str): name of BNG y coordinate column
+
+    Returns:
+        gpd.GeoDataFrame: UPRNs with BNG coordinate point geometries
+    """
+    if not usecols:
+        usecols = ["*"]
```
sofiapinto, line 46: can you add some documentation?

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -0,0 +1,39 @@
...
+            if col not in usecols:
+                usecols.append(col)
+    df = df.select(usecols)
+    df = df.to_pandas()
+
+    gdf = gpd.GeoDataFrame(
+        df,
+        geometry=gpd.points_from_xy(df[x_col], df[y_col]),
+        crs="EPSG:27700",
+    )
```
sofiapinto, line 59: excuse my ignorance on this, but are we basically going from two separate columns into one column that holds the point geometry?

sofiapinto (commented):
Hello @crispy-wonton ,

Thanks for another great PR - your code is a beauty.

I know the file names are self explanatory, but i do also see the value in a little docstring at the top. Sorry for always being a pain about this!

I think I got confused about building footprints and geometries & how we're identifying flats. Can you clarify?

Thank you!

asf_heat_pump_suitability/pipeline/run/flat_blocks.py
```diff
@@ -0,0 +1,59 @@
+"""
+Script to label UPRNs with flat / apartment property type boolean flag.
+
+Run:
+python asf_heat_pump_suitability/pipeline/run/flat_blocks.py --uprns path/to/domestic/UPRNs
```
sofiapinto, line 5: Any reason for not adding the path here?
python asf_heat_pump_suitability/pipeline/run/flat_blocks.py --uprns s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_residential_uprns.parquet

asf_heat_pump_suitability/pipeline/run/flat_blocks.py
sofiapinto, line 1: this is in run, not run_scripts. any reason why we have a separate folder?

asf_heat_pump_suitability/pipeline/run/flat_blocks.py
sofiapinto: I've ran this btw (commented the line where the data is saved) and it all ran smoothly

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: Yeh absolutely, that's the plan! I imagine we will do this much later in a separate PR. You can ignore the `data_source` part for now.

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: The OS datasets in general are updated every 6 months I believe but this is the UPRN geopoints rather than building footprints. This is how the file is named by OS.

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: yes this or the line below can be removed. I couldnt decide which was more robust to keep

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: so far I have used a few versions and they have all been names in the same format

asf_heat_pump_suitability/pipeline/impute/property_type.py
crispy-wonton: I don't think this is a problem (at least a common one) but I don't know for sure. I couldnt find it mentioned in any of their documentation that there are duplicates and I couldn't see cases of it from visually inspecting the data but I suppose it is possible that it can happen.

asf_heat_pump_suitability/pipeline/impute/property_type.py
crispy-wonton: This is the geometry of the UPRN rather than the building footprint. The geometry of the UPRN is a point (i.e. an X and Y coordinate). So the intention in this part of the code, is to identify any UPRNs which have coordinates which are not unique. I.e. another UPRN is in the same coordinates, indicating there are two homes located on top of each other.

asf_heat_pump_suitability/pipeline/run/flat_blocks.py
crispy-wonton: `run_scripts` is from the old repo structure that I want to move away from. I just think `run` is cleaner.

asf_heat_pump_suitability/pipeline/run/flat_blocks.py
crispy-wonton: I just didn't include the path because it changes depending on the input parameters :)

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: Yes, it will still retain the X and Y coordinates as separate columns but there will be a new geometry column with a point created from combining X and Y. A GeoDataFrame requires a geometry column and then we can use the GeoDataFrame to do spatial joins and such. Basically, we are converting the polars dataframe to a geopandas geodataframe for use in spatial transformations.

asf_heat_pump_suitability/getters/load_geodata.py
sofiapinto: the one you have commented is less intuitive, but do check for robustness as that's more important :)

asf_heat_pump_suitability/getters/load_geodata.py
sofiapinto: maybe we could just have a little warning, in case it changes and we don't notice?

asf_heat_pump_suitability/pipeline/impute/property_type.py
sofiapinto: I guess when things slow down we can check for polygons with >X% of overlap, for peace of mind

asf_heat_pump_suitability/pipeline/impute/property_type.py
sofiapinto: can there ever be a case where the coordinates are not an exact match but the properties still being on top of each other?

asf_heat_pump_suitability/pipeline/run/flat_blocks.py
sofiapinto: niiiice! so i guess we will eventually remove run_scripts

sofiapinto (dismissed):

asf_heat_pump_suitability/pipeline/impute/property_type.py
crispy-wonton: I think there probably can, that's why this method isn't perfect. I suppose there could also be cases where the coordinates are on top of each other but they are actually neighbouring properties (e.g. beside each other rather than on top of each other). 

I have a confusion matrix using the EPC data as a ground truth for this which I will send you on slack.

asf_heat_pump_suitability/pipeline/impute/property_type.py
crispy-wonton: Good idea!

asf_heat_pump_suitability/pipeline/run/flat_blocks.py
crispy-wonton: yes! I'm hoping to delete most of the old files!

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: I've added a print statement into the base getter to print out the name of the file being loaded. I think that combined with the fact that this will fail if there isn't a file matching the extracted filename is enough. Or were you imagining a different warning?

PR #193, opened 2025-11-12
192 sample buildings

### LINKED ISSUE #192: Create sample of buildings for labelling
Take a sample of buildings containing flats from Plymouth and other areas (with a mixture of homogeneous and heterogeneous housing types). These will be manually labelled to create a ground truth dataset for the block of flats model.

crispy-wonton:
Fixes #192 

# Description

Add a notebook to take a sample of buildings containing flats from Plymouth and some additional sampling areas for manual labelling. The data will be used to train a classification model. The final samples are saved in separate `.kml` files for each labeller to load into MyMaps and label.

New files:
- `asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py` - notebook to conduct the sampling and create the `.kml` files for plotting in MyMaps

Files from separate PRs (unchanged from other PRs and in review - these don't need to be reviewed in detail here because they are being thoroughly reviewed separately. It would still be useful to give a cursory review but **please focus on the notebook**)
- `asf_heat_pump_suitability/getters/load_tree_input.py` - load geodata required for sampling
- `asf_heat_pump_suitability/getters/load_boundaries.py` - load LA boundaries required for sampling
- `asf_heat_pump_suitability/config/base.yaml` - add params and data sources required for sampling

# Instructions for Reviewer

use jupytext to convert the py file to a notebook and then run it:
`jupytext --to notebook asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py`

**The main part of the code to review is the notebook. The functions are being reviewed across separate PRs. I had to copy them to here so that I could use them while the PRs are unmerged.**

- We are sampling in an appropriately random way
- We are creating a reproducible sample. This is especially important to watch out for when using polars. Dataframes need to be sorted before sampling as there is no index.
- Are we sampling from the right subset of buildings? Is anything being excluded / included that shouldn't?

Some questions to you:
- Do we want to set a threshold for proportion of UPRNs in the building which are flats to sample from? I didn’t do this in my original sample and the proportions are anywhere between low and 1.
- Do we want to change the groupings of flat counts? Or are we happy with 2-6, 7-15, 16+?
- Do we want to change the counts of samples per Local Authority? Relatedly, do we want to remove any Local Authorities and have a higher number of samples from fewer authorities?
- Because I have already labelled 300 buildings and will continue labelling more, it means that the majority of buildings which will receive 2 labels will receive one from another labeller, and one from me. We can change this so that we are only taking the ‘second sample’ (I.e. creating a subsample of buildings to receive 2 labels) evenly from all labellers. It would be a simple code change. What do you think?

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/config/base.yaml
+++ b/asf_heat_pump_suitability/config/base.yaml
@@ -1,6 +1,57 @@
 data:
   geodata:
+    # OpenMap UPRN with coordinates
     uk_osopen_uprn: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/osopenuprn_202510_csv.zip"
+    # OpenMap Local Geopackage
+    gb_os_openmap_local: "s3://asf-heat-pump-suitability/source_data/opmplc_gb.gpkg"
+    # OpenMap Local layers for certain grid squares
+    grid_square_os_openmap_local: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/v202510_OSOpenMapLocal_geometries_selected/{square}/{square}_{layer}.shp"
+
+    boundaries:
+      # UK LA boundaries
+      UK_ons_lad_bounds: "s3://asf-heat-pump-suitability/source_data/Local_Authority_Districts_December_2023_Boundaries_UK_BFE_-2600600853110041429/LAD_DEC_2023_UK_BFE.shp"
+  epc:
+    domestic: "s3://asf-daps/lakehouse/2025_Q1/processed/epc/deduplicated/processed_dedupl-0.parquet"
+    commercial: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/v092025_EPC_NonDomestic_EW.csv"
+
+  # Preprocessed data
+  processed:
+    non_domestic_poi_categories: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/processed/non_residential_poi_categories.csv"
+
+constant:
+  grid_squares:
+    plymouth: "SX"
+    plymouth_similar_cities: # Liverpool, Plymouth, Portsmouth, Southampton, Swansea
+      - "SD"
+      - "SJ"
+      - "SN"
+      - "SS"
+      - "SU"
+      - "SX"
+      - "SZ"
+    sampling_areas: # Plymouth, Nottingham, Bradford, Glasgow, Manchester, Bath
+      - "NS"
+      - "SD"
+      - "SE"
+      - "SJ"
+      - "SK"
+      - "ST"
+      - "SX"
+  plymouth_similar_cities:
+    - "Liverpool"
+    - "Plymouth"
+    - "Portsmouth"
+    - "Southampton"
+    - "Swansea"
+  sampling_areas:
+    - "Bath and North East Somerset"
+    - "Bradford"
+    - "Glasgow City"
+    - "Manchester"
+    - "Nottingham"
+    - "Plymouth"
+  target_crs: 27700 # British National Grid
+
 data_source:
   gb_ons_postcode_dir_url: "https://www.arcgis.com/sharing/rest/content/items/487a5ba62c8b4da08f01eb3c08e304f6/data" # Aug 2023 data
   gb_ons_postcode_dir_file_path: "Data/ONSPD_AUG_2023_UK.csv" # Aug 2023 data
--- a/asf_heat_pump_suitability/getters/load_boundaries.py
+++ b/asf_heat_pump_suitability/getters/load_boundaries.py
@@ -1,3 +1,7 @@
+"""
+Functions to load raw census geography boundaries, like Local Authority; ward; output areas, etc. No/minimal preprocessing occurs in these functions.
+"""
+
 import geopandas as gpd
 import shapely
 from typing import List
@@ -27,9 +31,16 @@ def load_gdf_local_authority_boundaries(
         return la_boundaries_gdf
     elif isinstance(select_las, str):
         print(f"Loading Local Authority boundaries for {select_las}...")
-        return la_boundaries_gdf[
-            la_boundaries_gdf["LAD23NM"].str.lower() == select_las.lower()
+        la_boundaries_gdf = la_boundaries_gdf[
+            la_boundaries_gdf["LAD23NM"].str.contains(select_las.title())
         ]
+        # Raise exception if boundaries are not found for LA specified by select_las
+        if len(la_boundaries_gdf) == 0:
+            raise Exception(
+                f"Could not find boundaries for the following Local Authority: {select_las}"
+            )
+        else:
+            return la_boundaries_gdf
     else:
         print(f"Loading Local Authority boundaries for {select_las}...")
         la_boundaries_gdf = la_boundaries_gdf[
@@ -39,9 +50,10 @@ def load_gdf_local_authority_boundaries(
         matches = set(la_boundaries_gdf["LAD23NM"].unique())
 
         # Raise exception if any boundaries are not found for LAs in select_las
-        if len(set(select_las).difference(matches)) > 0:
+        missing_las = set(select_las).difference(matches)
+        if len(missing_las) > 0:
             raise Exception(
-                f"Could not find boundaries for the following Local Authorities: {set(select_las).difference(matches)}"
+                f"Could not find boundaries for the following Local Authorities: {missing_las}"
             )
         else:
             return la_boundaries_gdf
--- a/asf_heat_pump_suitability/getters/load_boundaries.py
+++ b/asf_heat_pump_suitability/getters/load_boundaries.py
@@ -1,0 +1,47 @@
+import geopandas as gpd
+import shapely
+from typing import List
+
+from asf_heat_pump_suitability import config
+
+
+def load_gdf_local_authority_boundaries(
+    select_las: str | List[str] = None,
+) -> gpd.GeoDataFrame:
+    """
+    Load boundaries for specified Local Authority Districts. CRS British National Grid (27700).
+
+    Args:
+        select_las (str | List[str]): selected Local Authorities to load boundaries for. Optional. Default None to load all UK
+        Local Authority boundaries.
+
+    Returns:
+        gpd.GeoDataFrame: boundaries for specified Local Authority Districts
+    """
+    la_boundaries_gdf = gpd.read_file(
+        config["data"]["geodata"]["boundaries"]["UK_ons_lad_bounds"]
+    )
+
+    if not select_las:
+        print("Loading Local Authority boundaries for UK...")
+        return la_boundaries_gdf
+    elif isinstance(select_las, str):
+        print(f"Loading Local Authority boundaries for {select_las}...")
+        return la_boundaries_gdf[
+            la_boundaries_gdf["LAD23NM"].str.lower() == select_las.lower()
+        ]
+    else:
+        print(f"Loading Local Authority boundaries for {select_las}...")
+        la_boundaries_gdf = la_boundaries_gdf[
+            # Filter to exact LA name matches, case insensitive
+            la_boundaries_gdf["LAD23NM"].str.fullmatch("|".join(select_las), case=False)
+        ]
+        matches = set(la_boundaries_gdf["LAD23NM"].unique())
+
+        # Raise exception if any boundaries are not found for LAs in select_las
+        if len(set(select_las).difference(matches)) > 0:
+            raise Exception(
+                f"Could not find boundaries for the following Local Authorities: {set(select_las).difference(matches)}"
+            )
+        else:
+            return la_boundaries_gdf
--- a/asf_heat_pump_suitability/getters/load_tree_input.py
+++ b/asf_heat_pump_suitability/getters/load_tree_input.py
@@ -1,0 +1,55 @@
+import geopandas as gpd
+import pandas as pd
+from typing import Optional, List
+from asf_heat_pump_suitability import config
+
+
+def load_gdf_os_openmap_local_layer(
+    layer: str, grid_squares: Optional[List[str]] = None, **kwargs
+) -> gpd.GeoDataFrame:
+    """
+    Load specified OS OpenMap Local layer. CRS British National Grid (27700).
+    Args:
+        layer (str): name of layer to load.
+        grid_squares (Optional[List[str]]): names of grid squares in OS mapping for regions of Great Britain to be loaded.
+        Default None to load whole GB.
+    Layer options:
+        'building',
+        'car_charging_point',
+        'electricity_transmission_line',
+        'foreshore',
+        'functional_site',
+        'glasshouse',
+        'important_building',
+        'motorway_junction',
+        'named_place',
+        'railway_station',
+        'railway_track',
+        'railway_tunnel',
+        'road',
+        'road_tunnel',
+        'roundabout',
+        'surface_water_area',
+        'surface_water_line',
+        'tidal_boundary',
+        'tidal_water',
+        'woodland'
+    Returns:
+        gpd.GeoDataFrame: OS OpenMap Local geometries for specified layer
+    """
+    if not grid_squares:
+        print(f"\nLoading OS OpenMap Local - {layer.title()}...")
+        print(f"Loading OS OpenMap Local - {layer.title()}...")
+        return gpd.read_file(
+            config["data"]["geodata"]["gb_os_openmap_local"], layer=layer, **kwargs
+        )
+    else:
+        layer = layer.replace("_", " ").title().replace(" ", "")
+        file_path = config["data"]["geodata"]["grid_square_os_openmap_local"]
+        files = [file_path.format(square=code, layer=layer) for code in grid_squares]
+        gdfs = []
+        for file in files:
+            print(f"\nLoading OS OpenMap Local - {layer.title()} file: {file}")
+            gdfs.append(gpd.read_file(file, **kwargs))
+
+        return pd.concat(gdfs)
--- a/asf_heat_pump_suitability/getters/load_tree_input.py
+++ b/asf_heat_pump_suitability/getters/load_tree_input.py
@@ -1,3 +1,7 @@
+"""
+Functions to load specific raw datasets used in decision tree pipeline using base getters and sources in config. No/minimal preprocessing occurs in these functions.
+"""
+
 import geopandas as gpd
 import pandas as pd
 from typing import Optional, List
@@ -8,11 +12,15 @@ def load_gdf_os_openmap_local_layer(
     layer: str, grid_squares: Optional[List[str]] = None, **kwargs
 ) -> gpd.GeoDataFrame:
     """
-    Load specified OS OpenMap Local layer. CRS British National Grid (27700).
+    Load specified OS OpenMap Local layer for Great Britain or optionally for a specific grid square. CRS British National Grid (27700).
+
+    Find grid square information at: https://www.ordnancesurvey.co.uk/documents/resources/guide-to-nationalgrid.pdf
+
     Args:
-        layer (str): name of layer to load.
-        grid_squares (Optional[List[str]]): names of grid squares in OS mapping for regions of Great Britain to be loaded.
-        Default None to load whole GB.
+        layer (str): name of layer to load. See layer options below.
+        grid_squares (Optional[List[str]]): names of grid squares in OS mapping for regions of Great Britain to be loaded. Default None to load whole GB.
+        **kwargs for geopandas.read_file()
+
     Layer options:
         'building',
         'car_charging_point',
@@ -34,22 +42,58 @@ def load_gdf_os_openmap_local_layer(
         'tidal_boundary',
         'tidal_water',
         'woodland'
+
     Returns:
         gpd.GeoDataFrame: OS OpenMap Local geometries for specified layer
     """
     if not grid_squares:
-        print(f"\nLoading OS OpenMap Local - {layer.title()}...")
         print(f"Loading OS OpenMap Local - {layer.title()}...")
         return gpd.read_file(
-            config["data"]["geodata"]["gb_os_openmap_local"], layer=layer, **kwargs
-        )
+            filename=config["data"]["geodata"]["gb_os_openmap_local"],
+            layer=layer,
+            **kwargs,
+        ).drop_duplicates(subset="ID")
+
     else:
+        if not isinstance(grid_squares, List):
+            grid_squares = [grid_squares]
+        # Reformat layer name to how it appears in file name
         layer = layer.replace("_", " ").title().replace(" ", "")
         file_path = config["data"]["geodata"]["grid_square_os_openmap_local"]
         files = [file_path.format(square=code, layer=layer) for code in grid_squares]
+
         gdfs = []
+
         for file in files:
             print(f"\nLoading OS OpenMap Local - {layer.title()} file: {file}")
             gdfs.append(gpd.read_file(file, **kwargs))
 
-        return pd.concat(gdfs)
+        return pd.concat(gdfs).drop_duplicates(subset="ID")
+
+
+def load_gdf_poi() -> gpd.GeoDataFrame:
+    """
+    Load and process Points of Interest data. CRS EPSG 4326.
+
+    Returns:
+        gpd.GeoDataFrame: Processed POI data containing types of POI specified
+
+    Raises:
+        ValueError: If required columns are missing
+    """
+    print("Loading POI data...")
+
+    required_columns = [
+        "id",
+        "country",
+        "main_category",
+        "alternate_category",
+        "geometry",
+    ]
+    poi = gpd.read_file(
+        filename=config["data_source"]["UK_poi_locations"],
+        columns=required_columns,
+        layer="poi_uk",
+    ).to_crs("EPSG:4326")
+    print(f"POI CRS: {poi.crs}")
+    return poi
--- a/asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
+++ b/asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
@@ -35,7 +35,56 @@ s3 = boto3.resource("s3")
 BUCKET = "asf-heat-pump-suitability"
 
 # %%
-labellers = ["reindeer", "springbok", "anteater", "emu"]
+labellers = ["reindeer", "springbok", "anteater", "raccoon", "salamander"]
+
+# %% [markdown]
+# ## Compare April and October buildings files
+
+# %%
+# Load April 2025 building footprints
+apr_buildings_gdf = gpd.read_file(
+    "s3://asf-heat-pump-suitability/local_heat_planning/inputs/v042025_OSOpenMapLocal_geometries_selected/SX/SX_Building.shp"
+)
+
+# Load October 2025 building footprints
+oct_buildings_gdf = gpd.read_file(
+    "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/v202510_OSOpenMapLocal_geometries_selected/SX/SX_Building.shp"
+)
+
+# Print diffs
+print("N building footprints in April 2025 dataset:")
+print(len(apr_buildings_gdf))
+
+print("\nN building footprints in October 2025 dataset:")
+print(len(oct_buildings_gdf))
+
+print("\nN extra building footprints in October 2025 dataset:")
+print(len(oct_buildings_gdf) - len(apr_buildings_gdf))
+
+# Normalize geometry columns to make sure the coordinates are in the same order
+apr_buildings_gdf["geometry"] = apr_buildings_gdf.normalize()
+oct_buildings_gdf["geometry"] = oct_buildings_gdf.normalize()
+
+# Merge April and October on geometry
+joined_buildings_gdf = apr_buildings_gdf.merge(
+    oct_buildings_gdf, how="outer", on="geometry", suffixes=("_apr", "_oct")
+)
+
+# Get proportion of building footprints which persist across April and October 2025
+persistent_buildings_gdf = joined_buildings_gdf[
+    (~joined_buildings_gdf["ID_apr"].isna()) & (~joined_buildings_gdf["ID_oct"].isna())
+]
+
+print("\nCount of building footprints which persist:")
+print(len(persistent_buildings_gdf))
+
+print("\nProportion of building footprints which persist:")
+print(len(persistent_buildings_gdf) / len(apr_buildings_gdf))
+
+# %% [markdown]
+# ## Original random samples of buildings containing flats from Plymouth
+#
+# See instructions to save polygons to kml with simplekml here: https://simplekml.readthedocs.io/en/latest/gettingstarted.html#creating-a-polygon
 
 # %%
 # Load Plymouth buildings and residential UPRNs with property type label
@@ -56,8 +105,9 @@ uprns_df = pl.from_pandas(uprns_gdf.drop(columns="geometry")).rename(
     {"ID": "building_id"}
 )
 
-# %% [markdown]
-# ## Original random samples of buildings containing flats from Plymouth
+# %%
+# We need to assert there are only Polygons in the building footprints, otherwise taking the outerboundary is not sufficient
+assert (buildings_gdf["geometry"].geom_type == "Polygon").all()
 
 # %%
 # Sample 1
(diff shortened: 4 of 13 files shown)


asf_heat_pump_suitability/config/base.yaml
crispy-wonton, line 1: This file is being reviewed in detail separately.

asf_heat_pump_suitability/getters/load_boundaries.py
crispy-wonton, line 1: This file is being reviewed in detail separately.

asf_heat_pump_suitability/getters/load_tree_input.py
crispy-wonton, line 1: This file is being reviewed in detail separately.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+
+from datetime import date
+
+today = date.today().strftime("%Y%m%d")
+
+# %%
+seed = 10
+
+# %%
+labellers = ["reindeer", "springbok", "anteater", "emu"]
```
sofiapinto: funny 😆!

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+
+# %%
+# Load Plymouth buildings and residential UPRNs with property type label
+buildings_gdf = gpd.read_file(
+    "s3://asf-heat-pump-suitability/local_heat_planning/inputs/v042025_OSOpenMapLocal_geometries_selected/SX/SX_Building.shp"
+)
+uprns_df = pl.read_parquet(
+    f"s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_residential_uprns_with_flats.parquet"
+)
```
sofiapinto, line 97: What's currently happening to buildings with no information about the property types? They could still be all flats or have some flats in the building.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+today = date.today().strftime("%Y%m%d")
+
+# %%
+seed = 10
+
+# %%
+labellers = ["reindeer", "springbok", "anteater", "emu"]
+
+# %%
+# Load Plymouth buildings and residential UPRNs with property type label
```
sofiapinto, line 91: Is this correct? uprns_df doesn't have all residential UPRNs does it? Only those in buildings with at least one flat. If not, then I guess my question would be why did we name the file "plymouth_residential_uprns_with_flats.parquet"
```suggestion
# Load Plymouth buildings and residential UPRNs in a building with at least one flat
```

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+    "s3://asf-heat-pump-suitability/local_heat_planning/inputs/v042025_OSOpenMapLocal_geometries_selected/SX/SX_Building.shp"
+)
+uprns_df = pl.read_parquet(
+    f"s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_residential_uprns_with_flats.parquet"
+)
+
+# Join buildings to UPRNs
+uprns_gdf = (
+    uprns.generate_gdf_uprn_coords(df=uprns_df)
+    .sjoin(buildings_gdf, how="inner", predicate="within")
```
sofiapinto, line 102: what happens during this sjoin?

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+    {"ID": "building_id"}
+)
+
+# %% [markdown]
+# ## Original random samples of buildings containing flats from Plymouth
+
+# %%
+# Sample 1
+n = 200
+pl.set_random_seed(seed)
```
sofiapinto, line 116: i've never used this but I'm guessing this actually set this seed for everything after this? I'd still explicitly set it in functions (but I think you're doing that anyway!).

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+# Filter to buildings with flats, then get total count of flats and UPRNs per building
+sample_df = (
+    uprns_df.filter(
+        pl.col("building_id").is_in(buildings_containing_flats),
+    )
+    .group_by("building_id")
+    .agg(
+        pl.col("property_type_flat").sum().alias("n_flats"),
+        pl.col("UPRN").count().alias("n_total"),
+        # Take a random sample from any building with >4 flats
```
sofiapinto, line 132: maybe document the reason behind the >4. Something like: If <=4, we do not consider it a block of flats.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+    )
+    .filter(pl.col("n_flats") > 4)
+    .sort(by="building_id")
+    .sample(n=n, seed=seed)
+)
+
+# Merge sample buildings with their geometries
+sample_gdf = buildings_gdf[["ID", "geometry"]].merge(
+    sample_df.to_pandas(), how="inner", left_on="ID", right_on="building_id"
+)
```
sofiapinto, line 143: I guess `buildings_gdf` doesn't have any missing geometries? If it does the sample could then be smaller than `n`. But I guess that's the reason why you then check the length after.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+    axis=1,
+)
+
+# Save as kml file
+kml = simplekml.Kml()
+for idx, r in sample_4326_gdf.iterrows():
+    pol = kml.newpolygon(
+        name="unlabelled",
+        description=f"Location: https://www.google.com/maps/search/?api=1&query={r['y']},{r['x']}\nN flats: {r['n_flats']}\nN total: {r['n_total']}",
+        outerboundaryis=list(r["geometry"].exterior.coords),
```
sofiapinto, line 168: what is this?

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+    )
+    pol.style.polystyle.color = "9939FF14"
```
sofiapinto, line 170: this is not an hex, but I guess it's something similar to set the color of your pologyon?

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+    lambda row: f"https://www.google.com/maps/search/?api=1&query={row['y']},{row['x']}",
+    axis=1,
+)
+
+# Save as kml file
+kml = simplekml.Kml()
+for idx, r in sample_4326_gdf.iterrows():
+    pol = kml.newpolygon(
+        name="unlabelled",
+        description=f"Location: https://www.google.com/maps/search/?api=1&query={r['y']},{r['x']}\nN flats: {r['n_flats']}\nN total: {r['n_total']}",
```
sofiapinto, line 166: This is very smart, have never used it. Then I guess you need to create a new map in your own account and upload the kml file?

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+        pl.col("building_id").is_in(buildings_containing_flats),
+        ~pl.col("building_id").is_in(already_labelled),
+    )
+    .group_by("building_id")
+    .agg(
+        pl.col("property_type_flat").sum().alias("n_flats"),
+        pl.col("UPRN").count().alias("n_total"),
+        # Sample from any buildings with >1 flat and <=15 UPRNs
+    )
+    .filter(pl.col("n_flats") > 1, pl.col("n_total") <= 15)
```
sofiapinto, line 214: why did we change the approach for the second sample?

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+# Save as kml file
+kml = simplekml.Kml()
+for idx, r in sample_4326_gdf.iterrows():
+    pol = kml.newpolygon(
+        name="unlabelled",
+        description=f"Location: https://www.google.com/maps/search/?api=1&query={r['y']},{r['x']}\nN flats: {r['n_flats']}\nN total: {r['n_total']}",
+        outerboundaryis=list(r["geometry"].exterior.coords),
+    )
+    pol.style.polystyle.color = "9939FF14"
+    pol.style.polystyle.outline = 1
```
sofiapinto, line 172: is this the size of the outline? or a specific style of line?

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+
+# %%
+# LOAD REQUIRED DATA TO SAMPLE BUILDINGS FROM
+# Load building footprints for sampling areas and load UPRNs
+buildings_gdf = load_tree_input.load_gdf_os_openmap_local_layer(
+    layer="building", grid_squares=config["constant"]["grid_squares"]["sampling_areas"]
+)
+sampling_areas_df = pl.read_parquet(
+    "s3://asf-heat-pump-suitability/local_heat_planning/outputs/sampling_areas_residential_uprns_with_flats.parquet"
+)
```
sofiapinto, line 293: In which PR is this file created?

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+
+# Identify already labelled buildings
+already_labelled = set(sample_df["building_id"].unique()) | set(
+    small_building_sample_df["building_id"].unique()
+)
+already_labelled_df = pl.concat([sample_df, small_building_sample_df])
+
+# Prepare already labelled dataframe with same columns for later sampling
+already_labelled_df = already_labelled_df.with_columns(
+    pl.Series("la_name", ["Plymouth"] * already_labelled_df.height),
```
sofiapinto, line 355:
```suggestion
    # Setting LA name to Plymouth for all lines of already labelled data
    pl.Series("la_name", ["Plymouth"] * already_labelled_df.height), 
```

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+    # Group number of flats
+    pl.when(pl.col("n_flats").is_between(2, 6, closed="both"))
+    .then(pl.lit("2-6_flats"))
+    .when(pl.col("n_flats").is_between(7, 15, closed="both"))
+    .then(pl.lit("7-15_flats"))
+    .when(pl.col("n_flats") > 15)
+    .then(pl.lit("16+_flats"))
+    .otherwise(None)
+    .alias("n_flats_grouped"),
+    # Add labeller
```
sofiapinto:
```suggestion
    # Set labeller name to "reindeer" for all lines of labelled data
```

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+    .filter(pl.col("n_flats") > 1)
+    .with_columns(
+        # Group number of flats
+        pl.when(pl.col("n_flats").is_between(2, 6, closed="both"))
+        .then(pl.lit("2-6_flats"))
+        .when(pl.col("n_flats").is_between(7, 15, closed="both"))
+        .then(pl.lit("7-15_flats"))
+        .when(pl.col("n_flats") > 15)
+        .then(pl.lit("16+_flats"))
+        .otherwise(None)
```
sofiapinto, line 398: Do we need this otherwise? Or did you just add it as good practice?

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+    # Add labeller
+    pl.Series("labeller", ["reindeer"] * already_labelled_df.height),
+)
+
+# Get buildings containing flats
+buildings_containing_flats = sampling_areas_df.filter(pl.col("property_type_flat"))[
+    "building_id"
+].unique(maintain_order=True)
+
+# Sample buildings containing flats but only if they haven't been labelled
```
sofiapinto:
```suggestion
# Create a dataframe of potential buildings data for labelling by
# 1) Removing buildings that have already been sampled & labelled
# 2) Enhancing dataframe with info such as: local authority where building is located, number of flats in the building and additional dummy variables
```

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+samples = []
+
+n_flats_groups = ["2-6_flats", "7-15_flats", "16+_flats"]
+
+# Take samples for each Local Authority for each group of flat counts to create sample of 700 buildings
+for la, df in sample_from_df.sort(by="la_name").group_by(
+    "la_name", maintain_order=True
+):
+    # Get the number of samples per LA
+    n = n_samples[la[0]]
```
sofiapinto, line 412: Does this work?
```suggestion
    n = n_samples[la]
```

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+    # Take one third of samples from each group of flat counts
+    for n_flats in n_flats_groups:
+        _n = np.ceil(n / len(n_flats_groups))
```
sofiapinto, line 416:
```suggestion
        # proportion of samples from each group of flat counts, in this case it's one third because there are 3 groups
        prop = np.ceil(n / len(n_flats_groups))
```

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+n_samples_per_labeller = int(np.ceil(sample_3_df.height / len(labellers)))
+# Create list of labellers corresponding to total sample length
+labeller_assignments = labellers * n_samples_per_labeller
+
+# Shuffle the list for random assignment
+random.seed(seed)
+random.shuffle(labeller_assignments)
+
+# Assign a labeller to each row
+labeller_assignments = labeller_assignments[: sample_3_df.height]
```
sofiapinto, line 438: Why do we need to do this? Is it because of rounding we end up with len(labeller_assignments)>len(sample_3_df)?

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+final_samples = {}
+
+# Create the final subset for each labeller
+for labeller in labellers:
+    # Get the number of primary samples selected for the labeller
+    n_samples = sample_3_df.filter(pl.col("labeller") == labeller).height
+
+    # Calculate number of additional samples so that some buildings get 2 labels from different people
+    total_sample = n_samples_per_labeller + 30
+    additional_sample = total_sample - n_samples
```
sofiapinto, line 452: Apologies if this is obvious, but I'm getting a little bit confused with the + 30 here. Are we adding 30 extra samples for each labeller? If so, why are we computing the additional_sample?

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+
+    # Get the second sample
+    if labeller != "reindeer":
+        print(
+            f"Labeller: {labeller}, adding already labelled sample for second sampling..."
+        )
```
sofiapinto, line 458: I don't understand what we're trying to say here.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+        second_sample = (
+            sample_3_df.filter(
+                pl.col("labeller") != labeller,
+                ~pl.col("building_id").is_in(second_samples),
+            )
+            .sort("building_id")
+            .sample(additional_sample, seed=seed)
+        )
+
+    # Add the IDs of the second sample to the list so that they cannot be sampled again
```
sofiapinto, line 478: So only two people can label the same line, but no more than 2?

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
+# %%
+import geopandas as gpd
+import polars as pl
+import numpy as np
+import random
+
+import simplekml
```
sofiapinto, line 21: This one is missing from requirements.txt

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
+# %%
```
sofiapinto, line 15: It would be great if you could add a few lines of documentation explaining what we're trying to do here - so that when we come back to this notebook in future we can just read it and quickly understand.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+        [sample_3_df.filter(pl.col("labeller") == labeller), second_sample]
+    )
+    final_samples[labeller] = final_sample
+
+    print(f"Labeller: {labeller}, N samples: {final_sample.height}")
+
+# %%
+for l, sample in final_samples.items():
+    print(l)
+    print(sample["labeller"].value_counts())
```
sofiapinto, line 493: what do these tables represent?

sofiapinto (commented):
Hey @crispy-wonton ,

Thanks for putting this together so quickly. I'm currently waiting for the notebook to finish running but have taken a look at the code and left some questions. Also answered your questions below.

> Please pay special attention to ...
> 
> We are sampling in an appropriately random way

Looks good!

> We are creating a reproducible sample. This is especially important to watch out for when using polars. Dataframes need to be sorted before sampling as there is no index.

I think this looks good because you're using a seed, you're sorting the datasets by buildin ID and you're saving the building IDs themselves.

> Are we sampling from the right subset of buildings? Is anything being excluded / included that shouldn't?

Looks okay, but I did leave some questions about the underlying data which will help me understand if this really is the case.

> Some questions to you:
> 
> Do we want to set a threshold for proportion of UPRNs in the building which are flats to sample from? I didn’t do this in my original sample and the proportions are anywhere between low and 1.

Hummm... I don't see why we should? Unless we end up with too little examples of the "difficult" cases to actually annotate.

> Do we want to change the groupings of flat counts? Or are we happy with 2-6, 7-15, 16+?

These groupings make sense to me based on the conversations we've been having, but we can also share this decision with the team and check if they see any caveat.


> Do we want to change the counts of samples per Local Authority? Relatedly, do we want to remove any Local Authorities and have a higher number of samples from fewer authorities?

Looks like a good amount. If anything, we might want to increase the size of the sample for Plymouth specifically later.

> Because I have already labelled 300 buildings and will continue labelling more, it means that the majority of buildings which will receive 2 labels will receive one from another labeller, and one from me. We can change this so that we are only taking the ‘second sample’ (I.e. creating a subsample of buildings to receive 2 labels) evenly from all labellers. It would be a simple code change. What do you think?

I was initially thinking that you've spent the most time thinking about this, so that wouldn't be a problem from my perspective. The only thing I'd worry is if all the common samples we're labelling are all the same type of building/same number of flats - so maybe because of that it would be better to focus only on this latest sample, or you can just check the common samples to see if they vary in the number of flats in each building.

---
Separately, I think you could create some functions. Some of the code is repeated and can lead to errors (using the wrong df for example).

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -0,0 +1,401 @@
...
+            name="unlabelled",
+            description=f"Location: https://www.google.com/maps/search/?api=1&query={r['y']},{r['x']}\nN flats: {r['n_flats']}\nN total: {r['n_total']}",
+            outerboundaryis=list(r["geometry"].exterior.coords),
+        )
+        pol.style.polystyle.color = "9939FF14"
+        pol.style.polystyle.outline = 1
+
+    kml.save(
+        f"{today}_{labeller}_UNLABELLED_small_plymouth_buildings_containing_flats_sample_n{l}_seed{seed}.kml"
+    )
```
sofiapinto, line 531: maybe worth saving to s3?

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: This dataset is based on a boolean label assigned to each UPRN which indicates if it's a flat / not. The boolean is based on imputation in PR #189 and is calculated for ALL UPRNs. Therefore, there should be no missing data for the flats label.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: No this is the full set of residential UPRNs with a label which is either `True` or `False` depending on whether it's a flat or not. See PR #190 `flat_blocks.py` for the script which produces this file.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: UPRN point geometries are getting joined to building footprints. A point will be joined to a building if it is fully within the boundaries of the footprint.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: Yes great point!

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: Thank you!

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: Yep have set explicit seeds elsewhere when possible.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: As `buildings_gdf` is primarily a geo-dataset, it doesn't have missing geometries :)

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: Yes, either you can create your own map, or I can create the maps (probably the better way to make sure there's no upload errors) and then make them editable by others.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: The `.exterior` gets the [outer boundary of the polygon](https://geopandas.org/en/latest/docs/reference/api/geopandas.GeoSeries.exterior.html) for each building footprint. This converts it from a Polygon to a [LinearRing,](https://shapely.readthedocs.io/en/stable/reference/shapely.LinearRing.html#shapely.LinearRing) which you can then get the CoordinateSequence for using `.coords`. 

I followed the instructions from the `simplekml` [docs here](https://simplekml.readthedocs.io/en/latest/gettingstarted.html#creating-a-polygon).

But now you've made me realise I should double check if there's any MultiPolygons (a polygon that has a hole in it) - can't remember if I did!

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: It is a hex code with an [opacity value (60%)](https://gist.github.com/lopspower/03fb1cc0ac9f32ef38f4).

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: This is a binary that can be 1 or 0 for if you want an outline or not.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: This was the method I used to try to collect more samples from the ambiguous blocks. I will add a note :)

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: PR #189 contains the script to filter UPRNs to residential only and PR #190 adds the boolean label of 'flat' or not.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: Not necessary, it defaults to adding this anyway but I added to be explicit and I checked there were none that received a `None` label.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: No because the `la` variable is a tuple that looks like this: `("Plymouth", )`

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: Yep, will add a note!

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: We are adding around 30 extra samples each. This is for the double labelling.

The reason I computed `additional_sample` is because, due to this earlier line:
`labeller_assignments = labeller_assignments[: sample_3_df.height]`
the labellers have uneven numbers of samples. This computation makes it even. It is slightly overkill but I thought it's just an opportunity to get an extra couple of labels.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: Updated:
`f"Labeller: {labeller}, adding reindeer's original sample for additional sampling/ double labelling..."`
This is just a print statement to make sure that all labellers take their extra 30 samples from the full pool of 1000 buildings minus their own sample, except `reindeer`, who will take 30 extra samples from the pool of 700 newer samples minus their own sample.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: Yes, let me know if this should change.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: Will add a note.
For each labeller, it shows the number of samples in your set for labelling that is being labelled by each other labeller.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
sofiapinto: cool, maybe add a comment in front of the line of code, for future reference. thank u

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
sofiapinto: There's just too many similar words in the sentence labelling/labeller/sample/sampling that even after your explanation, it's hard for me to read it and understanding the meaning. but it's fine, don't worry about it

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
sofiapinto: sounds good

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
sofiapinto: nice!

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
sofiapinto: nice & maybe add a little docstring & the link or some explanation for future reference. thanks!

asf_heat_pump_suitability/getters/load_tree_input.py
```diff
@@ -34,22 +42,58 @@ def load_gdf_os_openmap_local_layer(
...
+        "country",
+        "main_category",
+        "alternate_category",
+        "geometry",
+    ]
+    poi = gpd.read_file(
+        filename=config["data_source"]["UK_poi_locations"],
+        columns=required_columns,
+        layer="poi_uk",
+    ).to_crs("EPSG:4326")
```
sofiapinto, line 97: Apologies for my ignorance on this, but is it problematic that we're using CRS 27700 and EPSG 4326 here?

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -1,24 +1,90 @@
...
+#
+# We started with a sample of 200 buildings in Plymouth, then added an additional 100. Finally, we are taking a sample of 700 from a mixture of Plymouth and the other towns/cities listed below.
+# - (South) Plymouth (He) - x700. 300 samples already. 400 left to sample.
+# - (North) Nottingham (He) - x60
+# - (North) Bradford (Ho - terraces) - x60
+# - (Scotland) Glasgow (Ho - tenements) - x60
+# - (North) Manchester (Ho - blocks) x60
+# - (South) Bath (Ho - terraced) x60
+#
+
```
sofiapinto, line 14: maybe just add a mention of what the He and Ho are, for future knowledge.

asf_heat_pump_suitability/getters/load_tree_input.py
crispy-wonton: Thank you for flagging this - I double checked the code and I think it's all good but good to check.

The CRS here is WGS84 (which is the same as EPSG 4326) - i.e. the coordinates are latitude and longitude. We have loaded the dataset in the WGS84 CRS here because that's the CRS of the original file.

In `asf_heat_pump_suitability/pipeline/transform/poi.py` in the `transform_gdf_poi` function, we convert it to the target CRS for this project which is 27700. [See line here](https://github.com/nestauk/asf_heat_pump_suitability/blob/9576f7ed9193bff92d82ef2ed5127a5a854a7bb5/asf_heat_pump_suitability/pipeline/transform/poi.py#L38). So whenever we conduct any joins or use the POI data, it will be transformed to 27700 first and then used.

But otherwise, you are correct, if we were mixing two different CRS then we would get silent failure in our spatial joins.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
crispy-wonton: good idea :)

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
sofiapinto: can you add a little bit more text about this, for future reference?

sofiapinto (dismissed):

PR #199, opened 2025-12-02
194 filter uprns in heat network zones

### LINKED ISSUE #194: Add script to filter UPRNs in heat network zones
Develop code and add script to identify UPRNs in heat network zones.

eglucas:
Fixes https://github.com/nestauk/asf_heat_pump_suitability/issues/194

# Description

Code to process UPRNs and identify whether they are in an existing heat network zone.

Changed files:
- `config/base.yaml` - add path to dataset containing existing heat network zone polygons in Plymouth, and (placeholder) path for processed OSOpen UPRNs filtered to residential UPRNs in Plymouth only
- `getters/get_datasets.py` - add getter for existing heat network zone data in Plymouth

New files:
- `pipeline/transform/heat_network_zones.py` - add function to filter UPRNs to those located within any heat network zone
- `pipeline/run/heat_network_city_centres.py` - add script to add boolean flags labelling residential UPRNs indicating whether they're in a heat network zone (this will be further developed to add code that will add another boolean column identifying UPRNs are in a city centre). This script currently only contains code processing residential UPRN and heat network zone for Plymouth.

# Instructions for Reviewer

To run the script: `python asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py --local_authorities plymouth`
This should save an output file to S3 that you can read and inspect.

- General logic in the filtering function (`heat_network_zones.filter_gdf_heat_network_zone_uprns`) and whether there are any other checks in addition to CRS are needed.
- General logic in the processing run script
- Schema of the saved output dataframe - i.e. are the columns included in the saved output dataframe appropriate for how it will be further processed.

Note: The Plymouth heat network data has more columns (including qualitative descriptions under `RefType` and `Just`), but I think that the most critical is `ZoneID` (which also has a PLYM- prefix identifier) - which is the only one I've included at the moment. There are two other ID columns called `OBJECTID` and `RefID` but it's not clear why/how they differ from `ZoneID`. Not sure if we have any other documentation explaining the columns but please shout if we do and if you think any other columns are important! 

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/getters/get_datasets.py
+++ b/asf_heat_pump_suitability/getters/get_datasets.py
@@ -404,20 +404,3 @@ def load_df_dz_lookup(**kwargs) -> pl.DataFrame:
         **kwargs,
     )
     return df
-
-
-def load_gdf_plymouth_heat_network_zone_areas(**kwargs) -> gpd.GeoDataFrame:
-    """
-    Load GeoDataFrame with Plymouth heat network zone polygons from Plymouth City Council (CRS: EPSG:27700).
-
-    Args:
-        **kwargs for `gpd.read_file()`
-
-    Returns:
-        gpd.GeoDataFrame: polygons of heat network zones in Plymouth.
-    """
-    gdf = base_getters.get_gdf_from_gpkg_s3_path(
-        path=config["data"]["geodata"]["heat_network_zones"]["plymouth"],
-        **kwargs,
-    )
-    return gdf
--- a/asf_heat_pump_suitability/getters/get_datasets.py
+++ b/asf_heat_pump_suitability/getters/get_datasets.py
@@ -404,3 +404,20 @@ def load_df_dz_lookup(**kwargs) -> pl.DataFrame:
         **kwargs,
     )
     return df
+
+
+def load_gdf_plymouth_heat_network_zone_areas(**kwargs) -> gpd.GeoDataFrame:
+    """
+    Load GeoDataFrame with Plymouth heat network zone polygons from Plymouth City Council (CRS: EPSG:27700).
+
+    Args:
+        **kwargs for `gpd.read_file()`
+
+    Returns:
+        gpd.GeoDataFrame: polygons of heat network zones in Plymouth.
+    """
+    gdf = base_getters.get_gdf_from_gpkg_s3_path(
+        path=config["data"]["geodata"]["heat_network_zones"]["plymouth"],
+        **kwargs,
+    )
+    return gdf
--- a/asf_heat_pump_suitability/getters/load_geodata.py
+++ b/asf_heat_pump_suitability/getters/load_geodata.py
@@ -30,18 +30,26 @@ def load_df_osopen_uprn(**kwargs) -> pl.DataFrame:
     return df
 
 
-def load_gdf_plymouth_heat_network_zones(**kwargs) -> gpd.GeoDataFrame:
+def load_gdf_heat_network_zones(local_authority: str, **kwargs) -> gpd.GeoDataFrame:
     """
-    Load GeoDataFrame with Plymouth heat network zone polygons from Plymouth City Council (CRS: EPSG:27700).
+    Load GeoDataFrame with heat network zone polygons in given Local Authority.
 
     Args:
         **kwargs for `gpd.read_file()`
 
     Returns:
-        gpd.GeoDataFrame: polygons of heat network zones in Plymouth.
+        gpd.GeoDataFrame: polygons of heat network zones in given Local Authority.
     """
+
+    local_authority = local_authority.lower()
+
+    if local_authority not in config["data"]["geodata"]["heat_network_zones"].keys():
+        raise ValueError(
+            f"No path found for heat network zone geodata in Local Authority: {local_authority}"
+        )
+
     gdf = base_getters.get_gdf_from_gpkg_s3_path(
-        path=config["data"]["geodata"]["heat_network_zones"]["plymouth"],
+        path=config["data"]["geodata"]["heat_network_zones"][local_authority],
         **kwargs,
     )
     return gdf
--- a/asf_heat_pump_suitability/getters/load_geodata.py
+++ b/asf_heat_pump_suitability/getters/load_geodata.py
@@ -30,7 +30,7 @@ def load_df_osopen_uprn(**kwargs) -> pl.DataFrame:
     return df
 
 
-def load_gdf_plymouth_heat_network_zone_areas(**kwargs) -> gpd.GeoDataFrame:
+def load_gdf_plymouth_heat_network_zones(**kwargs) -> gpd.GeoDataFrame:
     """
     Load GeoDataFrame with Plymouth heat network zone polygons from Plymouth City Council (CRS: EPSG:27700).
 
--- a/asf_heat_pump_suitability/getters/load_geodata.py
+++ b/asf_heat_pump_suitability/getters/load_geodata.py
@@ -52,4 +52,8 @@ def load_gdf_heat_network_zones(local_authority: str, **kwargs) -> gpd.GeoDataFr
         path=config["data"]["geodata"]["heat_network_zones"][local_authority],
         **kwargs,
     )
+
+    print(
+        f"Heat network zone geodataframe successfully loaded for {local_authority} with CRS {gdf.crs}."
+    )
     return gdf
--- a/asf_heat_pump_suitability/getters/load_geodata.py
+++ b/asf_heat_pump_suitability/getters/load_geodata.py
@@ -1,4 +1,5 @@
 import polars as pl
+import geopandas as gpd
 import regex as re
 import os
 
@@ -27,3 +28,20 @@ def load_df_osopen_uprn(**kwargs) -> pl.DataFrame:
     )
 
     return df
+
+
+def load_gdf_plymouth_heat_network_zone_areas(**kwargs) -> gpd.GeoDataFrame:
+    """
+    Load GeoDataFrame with Plymouth heat network zone polygons from Plymouth City Council (CRS: EPSG:27700).
+
+    Args:
+        **kwargs for `gpd.read_file()`
+
+    Returns:
+        gpd.GeoDataFrame: polygons of heat network zones in Plymouth.
+    """
+    gdf = base_getters.get_gdf_from_gpkg_s3_path(
+        path=config["data"]["geodata"]["heat_network_zones"]["plymouth"],
+        **kwargs,
+    )
+    return gdf
--- a/asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
+++ b/asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
@@ -65,7 +65,7 @@ if __name__ == "__main__":
         print(
             "Identifying residential UPRNs in heat network zones for Plymouth Local Authority..."
         )
-        hn_zone_uprn_df = heat_network_zones.filter_gdf_heat_network_zone_uprns(
+        hn_zone_uprn_df = heat_network_zones.label_gdf_heat_network_zone_uprns(
             uprn_gdf=uprn_gdf,
             hn_zone_gdf=plymouth_hn_zones_gdf,
             usecols=[
--- a/asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
+++ b/asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
@@ -74,10 +74,9 @@ if __name__ == "__main__":
         )
 
         # Clean up columns
-        hn_zone_uprn_df = hn_zone_uprn_df[
+        hn_zone_uprn_df = hn_zone_uprn_df.select(
             ["UPRN", "LAD23NM", "X_COORDINATE", "Y_COORDINATE", "in_hn_zone", "ZoneID"]
-        ]
-        hn_zone_uprn_df = hn_zone_uprn_df.rename({"ZoneID": "HNZoneID"})
+        ).rename({"ZoneID": "HNZoneID"})
 
         # TODO filter and add label for UPRNs in city centres
 
--- a/asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
+++ b/asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
@@ -38,8 +38,6 @@ def parse_arguments() -> argparse.Namespace:
 
 
 if __name__ == "__main__":
-    import polars as pl
-
     from asf_heat_pump_suitability import config
     from asf_heat_pump_suitability.getters import base_getters, load_geodata
     from asf_heat_pump_suitability.pipeline.transform import uprns
@@ -65,22 +63,19 @@ if __name__ == "__main__":
         print(
             "Identifying residential UPRNs in heat network zones for Plymouth Local Authority..."
         )
-        hn_zone_uprn_gdf = heat_network_zones.filter_gdf_heat_network_zone_uprns(
+        hn_zone_uprn_df = heat_network_zones.filter_gdf_heat_network_zone_uprns(
             uprn_gdf=uprn_gdf,
             hn_zone_gdf=plymouth_hn_zones_gdf,
-            hn_zone_usecols=[  # unique to Plymouth heat network zone dataset
+            usecols=[
                 "ZoneID",  # zone unique identifier
-                "geometry",  # polygons
             ],
         )
 
-        # Add boolean flag for whether UPRN is in hn zone in original uprn_gdf
-        uprn_gdf["in_hn_zone"] = uprn_gdf["UPRN"].isin(hn_zone_uprn_gdf["UPRN"])
-
-        # Merge ZoneID into original uprn_gdf
-        uprn_gdf = uprn_gdf.merge(
-            hn_zone_uprn_gdf[["UPRN", "ZoneID"]], on="UPRN", how="left"
-        )
+        # Clean up columns
+        hn_zone_uprn_df = hn_zone_uprn_df[
+            ["UPRN", "LAD23NM", "X_COORDINATE", "Y_COORDINATE", "in_hn_zone", "ZoneID"]
+        ]
+        hn_zone_uprn_df = hn_zone_uprn_df.rename({"ZoneID": "HNZoneID"})
 
         # TODO filter and add label for UPRNs in city centres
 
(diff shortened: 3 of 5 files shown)


asf_heat_pump_suitability/getters/get_datasets.py
crispy-wonton, line 1: This new function can be moved to the `load_geodata.py` file. I would like to delete `get_datasets.py` eventually.

asf_heat_pump_suitability/getters/get_datasets.py
```diff
@@ -404,3 +404,20 @@ def load_df_dz_lookup(**kwargs) -> pl.DataFrame:
...
+    """
+    Load GeoDataFrame with Plymouth heat network zone polygons from Plymouth City Council (CRS: EPSG:27700).
+
+    Args:
+        **kwargs for `gpd.read_file()`
+
+    Returns:
+        gpd.GeoDataFrame: polygons of heat network zones in Plymouth.
+    """
+    gdf = base_getters.get_gdf_from_gpkg_s3_path(
```
crispy-wonton: Nice use of existing getters! :D

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -0,0 +1,42 @@
+"""
+Functions to filter OS UPRNs to UPRNs located within heat network zones.
```
crispy-wonton:
```suggestion
Functions to filter UPRNs to those located within heat network zones.
```
To clarify that this function can be used for UPRNs of any origin.

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+
+    # CRS checks
+    assert (
+        uprn_gdf.crs == config["constant"]["target_crs"]
+    ), f"Target CRS is ESPG:{config['constant']['target_crs']}, uprn_gdf is {uprn_gdf.crs}"
+
+    assert (
+        uprn_gdf.crs == hn_zone_gdf.crs
+    ), f"CRS mismatch: uprn_gdf is {uprn_gdf.crs} and hn_zone_gdf is {hn_zone_gdf.crs}"
```
crispy-wonton: Instead of erroring out with these statements, maybe you could convert the CRS to the target CRS and have a print statement logging that the code has done that.

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -0,0 +1,42 @@
...
+
+import geopandas as gpd
+
+from asf_heat_pump_suitability import config
+
+
+def filter_gdf_heat_network_zone_uprns(
+    uprn_gdf: gpd.GeoDataFrame,
+    hn_zone_gdf: gpd.GeoDataFrame,
+    hn_zone_usecols: list = None,
```
crispy-wonton:
```suggestion
    usecols: list = None,
```
I'd rename this for readability. Also, it needs to be added to docstring :)

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -0,0 +1,42 @@
...
+
+    assert (
+        uprn_gdf.crs == hn_zone_gdf.crs
+    ), f"CRS mismatch: uprn_gdf is {uprn_gdf.crs} and hn_zone_gdf is {hn_zone_gdf.crs}"
+
+    # If hn_zone_usecols not specified, use all columns in hn_zone_gdf
+    hn_zone_gdf = hn_zone_gdf if not hn_zone_usecols else hn_zone_gdf[hn_zone_usecols]
+
+    filtered_uprn_gdf = uprn_gdf.sjoin(
+        hn_zone_gdf, how="inner", predicate="intersects"
```
crispy-wonton: This intersection is something we can flag to the rest of the team in a stand up / sprint review and double check how they want us to deal with HNZ borders.

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
```diff
@@ -0,0 +1,106 @@
...
+    args = parse_arguments()
+
+    if args.local_authorities.lower() == "plymouth":
+
+        # Load residential OS UPRNs in Plymouth
+        print("Loading residential UPRN dataset for Plymouth Local Authority...")
+        uprn_df = base_getters.load_df_from_s3(
+            config["data"]["processed"]["plymouth_residential_uprns"]
+        )
+        uprn_gdf = lat_lon.generate_gdf_uprn_coords(uprn_df)
```
```suggestion
        uprn_gdf = uprns.generate_gdf_uprn_coords(uprn_df)
```
What you have isn't incorrect, but will be removed at some point. I've replaced with the same function from a different file. The `lat_lon.py` file will be removed during the transition from v1.0.0 to v2.0.0.

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
```diff
@@ -0,0 +1,106 @@
...
+
+    return parser.parse_args()
+
+
+if __name__ == "__main__":
+    import polars as pl
+
+    from asf_heat_pump_suitability import config
+    from asf_heat_pump_suitability.getters import base_getters, get_datasets
+    from asf_heat_pump_suitability.pipeline.prepare_features import lat_lon
```
```suggestion
    from asf_heat_pump_suitability.pipeline.transform import uprns
```

asf_heat_pump_suitability/getters/get_datasets.py
```diff
@@ -404,3 +404,20 @@ def load_df_dz_lookup(**kwargs) -> pl.DataFrame:
         **kwargs,
     )
     return df
+
+
+def load_gdf_plymouth_heat_network_zone_areas(**kwargs) -> gpd.GeoDataFrame:
```
crispy-wonton:
```suggestion
def load_gdf_plymouth_heat_network_zones(**kwargs) -> gpd.GeoDataFrame:
```
shortened for readability

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -0,0 +1,42 @@
...
+    ).drop(columns="index_right")
+
+    return filtered_uprn_gdf
```
crispy-wonton: I think this could return a `polars` df rather than `gdf` here :) let me know if there's a reason you retained the geometry that I'm missing tho!

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
```diff
@@ -0,0 +1,106 @@
...
+            ],
+        )
+
+        # Add boolean flag for whether UPRN is in hn zone in original uprn_gdf
+        uprn_gdf["in_hn_zone"] = uprn_gdf["UPRN"].isin(hn_zone_uprn_gdf["UPRN"])
+
+        # Merge ZoneID into original uprn_gdf
+        uprn_gdf = uprn_gdf.merge(
+            hn_zone_uprn_gdf[["UPRN", "ZoneID"]], on="UPRN", how="left"
+        )
```
crispy-wonton: If you decide to create a `polars` dataframe to replace `hn_zone_uprn_gdf` here as suggested in `heat_network_zones.py` then this section could become a bit cleaner with `polars` chaining.

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -0,0 +1,42 @@
...
+) -> gpd.GeoDataFrame:
+    """
+    Filter UPRNs to UPRNs that are located within a heat network zone.
+
+    Args:
+        uprn_gdf (gpd.GeoDataFrame): UPRNs with point geometries to be filtered
+        hn_zone_gdf (gpd.GeoDataFrame): polygons of heat network zones
+
+    Returns:
+        gpd.GeoDataFrame: UPRNs located within any of the heat network zones
```
crispy-wonton: Since you are passing in the full `uprn_gdf` here, it could clean up the main script to return a df of ALL UPRNs and label them with ZoneID (either ID or null) and the `in_hnz` boolean. Then you can just join it back onto the main UPRN df in the script.

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
```diff
@@ -0,0 +1,106 @@
...
+
+    else:  # all GB
+        # Placeholder for future implementation
+        raise NotImplementedError("Processing for all of GB is not yet supported.")
+
+    # Save residential UPRNs with heat network zone flags to S3
+    df_with_flags = pl.from_pandas(
+        uprn_gdf[
+            ["UPRN", "LAD23NM", "X_COORDINATE", "Y_COORDINATE", "in_hn_zone", "ZoneID"]
```
crispy-wonton: Might rename `ZoneID` to `HNZoneID` so it's clear what it refers to :)

crispy-wonton (changes requested):
Thanks so much for your hard work on this @eglucas :D it looks great and I couldn't find any issues in the logic.

I've left some comments and suggestions to streamline the code a bit more. I've also flagged a couple of places where we have new `.py` files to use instead of V1.0.0 files (hence requesting changes) - this will be something that we need to keep checking for at the moment as we transition away from the old repo structure to the new one.

I had a look at the output dataframe, I think it looks sensible and the proportions look reasonable. One thing that you could do to check if you haven't already is plot the UPRNs in HNZ by their coordinates and double check that it matches the HNZ shapes as you expect. Shouldn't take too long if you've built the code in a notebook. You can use `geopandas.GeoDataFrame.plot()`.

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
eglucas: Good point, thanks! Yes, there was no reason to retain the geometry, just hadn't thought through at that point in development that the convention was to save output dataframes in polars :)

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -27,3 +28,20 @@ def load_df_osopen_uprn(**kwargs) -> pl.DataFrame:
...
+        **kwargs for `gpd.read_file()`
+
+    Returns:
+        gpd.GeoDataFrame: polygons of heat network zones in Plymouth.
+    """
+    gdf = base_getters.get_gdf_from_gpkg_s3_path(
+        path=config["data"]["geodata"]["heat_network_zones"]["plymouth"],
+        **kwargs,
+    )
+    return gdf
```
sofiapinto: With scaling in mind, would it make sense to have the following?
```suggestion
def load_gdf_heat_network_zones(council_or_area="plymouth":str, **kwargs) -> gpd.GeoDataFrame:
    Load GeoDataFrame with Plymouth heat network zone polygons from Plymouth City Council (CRS: EPSG:27700).

    Args:
        council_or_area (str): council or local area considered. Defaults to "plymouth"
        **kwargs for `gpd.read_file()`

    Returns:
        gpd.GeoDataFrame: polygons of heat network zones in Plymouth.
    """
    gdf = base_getters.get_gdf_from_gpkg_s3_path(
        path=config["data"]["geodata"]["heat_network_zones"][council_or_area],
        **kwargs,
    )
    return gdf
```

sofiapinto: cc: @crispy-wonton

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
sofiapinto: and something we should add to the documentation of the app! :)

sofiapinto (commented):
great job @eglucas :) amazing how quickly you got up to speed on the project and codebase!

asf_heat_pump_suitability/getters/load_geodata.py
eglucas: I think it makes sense! If you agree @crispy-wonton, I'll make that change

eglucas:
Thank you for your super helpful suggestions @crispy-wonton !

I've addressed all your comments and implemented all suggestions:
- moving the getter function to `load_geodata.py` (https://github.com/nestauk/asf_heat_pump_suitability/pull/199/commits/2bc3133ac22da062be70d145c3a9d83595fc3c6d)
- renaming getter function for readability (https://github.com/nestauk/asf_heat_pump_suitability/pull/199/commits/e2ecc48799d031d6c2bbbfdb1dc94682c23ed32c)
- modifying imports in main script (https://github.com/nestauk/asf_heat_pump_suitability/pull/199/commits/b67a0fd44676c5afc474d9171d0d861412ea93ca)
- cleaning up the main script by modifying the filter function to add HNZ labels to all UPRNs and return as a polars df (https://github.com/nestauk/asf_heat_pump_suitability/pull/199/commits/81a855ab6618dafda38d88261c77a1717f380260) (NB the HNZ labels that are included in the returned polars df include the `in_hn_zone` boolean as well as any other descriptive label specified in the `usecols` argument - in the case for the Plymouth HNZ dataset, it's just ZoneID).
-  replaced the CRS assert statements with reprojections (https://github.com/nestauk/asf_heat_pump_suitability/pull/199/commits/81a855ab6618dafda38d88261c77a1717f380260)
- renamed the `usecols` argument and added it to docstring (https://github.com/nestauk/asf_heat_pump_suitability/pull/199/commits/81a855ab6618dafda38d88261c77a1717f380260)

I also did the visual check you suggested as we've discussed on slack! 

I'll also implement the change that Sofia suggested for scalability of the heat network data getter once you've had a look at that.

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: Yep looks good. 2 minor things, please could we use `local_authority` as the arg and config names for consistency, and secondly please no default, but instead call `local_authority="Plymouth"` explicitly where the function is used. This is to avoid unintentional silent fails. For example, if you didn't input the LA name and called the function with the Plymouth default but were using it for another LA, you would conduct the spatial join and it would erroneously look like none of the UPRNs fall in a HN zone but the code wouldn't error out.

eglucas:
Hi @crispy-wonton @sofiapinto - thanks so much for spotting that opportunity to make the HNZ data getter general for any local authority. I've made those changes (https://github.com/nestauk/asf_heat_pump_suitability/pull/199/commits/0f950fd077ef87c3b1c3fdbf79d7b62df96e5d57):
- `local_authority` is now an argument, with no default
- an error is raised if a HNZ geodata path for the given `local_authority` doesn't exist in the config file

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -0,0 +1,62 @@
...
+    hn_zone_gdf: gpd.GeoDataFrame,
+    usecols: list = None,
+) -> pl.DataFrame:
+    """
+    Filter UPRNs to UPRNs that are located within a heat network zone.
+
+    Args:
+        uprn_gdf (gpd.GeoDataFrame): UPRNs with point geometries to be filtered
+        hn_zone_gdf (gpd.GeoDataFrame): polygons of heat network zones
+        usecols: names of descriptive columns in hn_zone_gdf (excluding "geometry") to be joined with uprn_gdf. Default is None to use all columns.
```
eglucas:
```suggestion
        usecols (list, optional): names of descriptive columns in hn_zone_gdf (excluding "geometry") to be joined with uprn_gdf. Default is None to use all columns.
```

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -30,18 +30,26 @@ def load_df_osopen_uprn(**kwargs) -> pl.DataFrame:
...
+    local_authority = local_authority.lower()
+
+    if local_authority not in config["data"]["geodata"]["heat_network_zones"].keys():
+        raise ValueError(
+            f"No path found for heat network zone geodata in Local Authority: {local_authority}"
+        )
+
     gdf = base_getters.get_gdf_from_gpkg_s3_path(
-        path=config["data"]["geodata"]["heat_network_zones"]["plymouth"],
+        path=config["data"]["geodata"]["heat_network_zones"][local_authority],
```
crispy-wonton, line 52: Maybe add a print statement with the CRS here :)

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -30,18 +30,26 @@ def load_df_osopen_uprn(**kwargs) -> pl.DataFrame:
...
-        gpd.GeoDataFrame: polygons of heat network zones in Plymouth.
+        gpd.GeoDataFrame: polygons of heat network zones in given Local Authority.
     """
+
+    local_authority = local_authority.lower()
+
+    if local_authority not in config["data"]["geodata"]["heat_network_zones"].keys():
+        raise ValueError(
+            f"No path found for heat network zone geodata in Local Authority: {local_authority}"
+        )
```
crispy-wonton, line 49: Nice!

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -1,42 +1,56 @@
...
+def label_gdf_heat_network_zone_uprns(
     uprn_gdf: gpd.GeoDataFrame,
     hn_zone_gdf: gpd.GeoDataFrame,
-    hn_zone_usecols: list = None,
-) -> gpd.GeoDataFrame:
+    usecols: list = None,
+) -> pl.DataFrame:
     """
-    Filter UPRNs to UPRNs that are located within a heat network zone.
+    Labels UPRNs that are located within a heat network zone.
```
crispy-wonton:
```suggestion
    Label UPRNs that are located within a heat network zone.
```

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -1,42 +1,56 @@
...
-        gpd.GeoDataFrame: UPRNs located within any of the heat network zones
+        pl.DataFrame: input UPRNs labelled with heat network zone identifiers
     """
 
-    # CRS checks
-    assert (
-        uprn_gdf.crs == config["constant"]["target_crs"]
-    ), f"Target CRS is ESPG:{config['constant']['target_crs']}, uprn_gdf is {uprn_gdf.crs}"
+    # CRS checks and reprojection if needed
+    target_crs = config["constant"]["target_crs"]
```
crispy-wonton, line 29: It's not necessary, but this is a good example of where you could use a global variable. e.g. set `TARGET_CRS` at the top of the script outside the function.

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
```diff
@@ -59,28 +57,27 @@ def parse_arguments() -> argparse.Namespace:
...
-
-        # Merge ZoneID into original uprn_gdf
-        uprn_gdf = uprn_gdf.merge(
-            hn_zone_uprn_gdf[["UPRN", "ZoneID"]], on="UPRN", how="left"
-        )
+        # Clean up columns
+        hn_zone_uprn_df = hn_zone_uprn_df[
+            ["UPRN", "LAD23NM", "X_COORDINATE", "Y_COORDINATE", "in_hn_zone", "ZoneID"]
+        ]
+        hn_zone_uprn_df = hn_zone_uprn_df.rename({"ZoneID": "HNZoneID"})
```
```suggestion
        hn_zone_uprn_df = hn_zone_uprn_df.select(
            ["UPRN", "LAD23NM", "X_COORDINATE", "Y_COORDINATE", "in_hn_zone", "ZoneID"]
        ).rename({"ZoneID": "HNZoneID"})
```

crispy-wonton (commented):
Thanks for making those changes - looks great! I left some minor comments -  then good to merge.

sofiapinto (approved):

crispy-wonton (approved):

PR #203, opened 2025-12-05
200 label uprns in city centre

### LINKED ISSUE #200: Add function to filter UPRNs in city centre areas
Further development of the script created addressing https://github.com/nestauk/asf_heat_pump_suitability/issues/194. Function(s) are needed to identify whether a UPRN is located within a "city centre" area and add a boolean flag.

More precisely, the following need to be developed:
* Approach that provides polygons indicating "city centre" areas
* Function to filter UPRNs located within these polygons
* Add new code to the existing run script that flags whether UPRNs are in a heat network zone or in a "city centre"

eglucas:
Fixes https://github.com/nestauk/asf_heat_pump_suitability/issues/200

# Description

Code to process UPRNs and identify whether they are in a potential heat network zone, based on which spatial signature type area they are in.

Changed files:
- `config/base.yaml` - add path to datasets containing the spatial signatures dataset for GB. There is one version with the suffix `_simplified` and the other without a suffix. **The code and run script I've developed is based only on the `_simplified` version. (I've not been able to inspect the full version as loading it keeps crashing my kernel 🥲)**

- `getters/load_geodata.py` - add getter for spatial signatures dataset for GB with arguments to select `simplified` or `full` version.

- `pipeline/transform/heat_network_zones.py` - add function to label UPRNs that are in one of the given spatial signature types. This function does the same processing as the HNZ labeller, so there might be opportunity to streamline but I think keeping them distinctly separate is best for clarity at this stage.
Note: I've changed the logic for `usecols` in both the HNZ and spatial signature labeller so that default `usecols=None` no longer means that all columns are used but rather only the columns essential for processing.

- `pipeline/run/heat_network_city_centres.py` - add code block to existing script that adds boolean flags labelling residential UPRNs indicating whether they're in one of spatial signature types we're currently considering to be representative of a city centre. This script now saves two separate dataframes to S3 - one with existing HNZ labels, and one with spatial signature types and city centre boolean labels. For simplicity, I've kept them separate but we could in future, combine both HNZ and city centre labels one dataframe by converting `hp_zone_uprn_df` to a `gdf` and passing that to `heat_network_zones.label_gdf_heat_network_spatial_signature_uprns()`.

- `analysis/spatial_signatures_hn_zones.py` - add analytical notebook that compares existing HNZ and city centre spatial signature types. This is specific to Plymouth LA. Open as Jupyter Notebook.

# Instructions for Reviewer

To run the script: python asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py --local_authorities plymouth
- Please inspect the new `city_centre_uprn_df` that is saved out to S3

* Any bugs or opportunities for improvement in the loading and labelling functions, as well as the new code block in the run script
* Whether you're happy with the saved location of the spatial signatures dataset, as well as how their paths are inserted in config
* Schema of the saved output dataframe

With regards to the analysis notebook, please run all the cells and have a look at the outputs. Let me know if you have any other thoughts on how we're evaluating whether the city centre signature types do a good enough job at flagging potential heat network zones. Please also flag if there are any errors in the analysis or interpretation.

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/analysis/spatial_signatures_hn_zones/20251205_plymouth_investigation.py
+++ b/asf_heat_pump_suitability/analysis/spatial_signatures_hn_zones/20251205_plymouth_investigation.py
@@ -1,0 +1,530 @@
+# %% [markdown]
+# ### **Investigating areas in Plymouth LA with existing heat network zones or potential for a heat network**
+
+# %%
+import pandas as pd
+import geopandas as gpd
+import polars as pl
+import textwrap
+import matplotlib.pyplot as plt
+import matplotlib.patches as mpatches
+from matplotlib.lines import Line2D
+
+from asf_heat_pump_suitability import config
+from asf_heat_pump_suitability.getters import (
+    base_getters,
+    load_geodata,
+    load_boundaries,
+)
+from asf_heat_pump_suitability.pipeline.transform import uprns, heat_network_zones
+
+# %% [markdown]
+# Loading data
+
+# %%
+# OSOpen UPRNS filtered for Plymouth only
+plymouth_uprns_df = base_getters.load_df_from_s3(
+    "s3://asf-heat-pump-suitability/dump/plymouth_OSOpen_UPRNs_filtered.parquet"
+)
+plymouth_uprns_gdf = uprns.generate_gdf_uprn_coords(plymouth_uprns_df)
+
+# %%
+# Plymouth residential UPRNs
+plymouth_residential_uprns_df = base_getters.load_df_from_s3(
+    config["data"]["processed"]["plymouth_residential_uprns"]
+)
+plymouth_residential_uprns_gdf = uprns.generate_gdf_uprn_coords(
+    plymouth_residential_uprns_df
+)
+
+# %%
+# Plymouth LA boundary
+plymouth_la_boundaries_gdf = load_boundaries.load_gdf_local_authority_boundaries(
+    select_las="plymouth"
+)
+
+# %%
+# Existing heat network zones in plymouth
+plymouth_hn_zones_gdf = load_geodata.load_gdf_heat_network_zones(
+    local_authority="plymouth"
+)
+
+# %%
+# Simplified form of the spatial signatures gb dataset
+spatial_signatures_gb_simplified_gdf = load_geodata.load_gdf_spatial_signatures_gb(
+    detail_level="simplified"
+)
+
+# %%
+# Confirming spatial signature types
+print("Spatial signatures:")
+counter = 1
+for type in spatial_signatures_gb_simplified_gdf["type"].unique():
+    print(f"{counter}. {type}")
+    counter += 1
+
+# %%
+# CRS checks
+gdfs = [
+    plymouth_uprns_gdf,
+    plymouth_residential_uprns_gdf,
+    plymouth_la_boundaries_gdf,
+    plymouth_hn_zones_gdf,
+    spatial_signatures_gb_simplified_gdf,
+]
+len({gdf.crs for gdf in gdfs}) == 1
+
+# %% [markdown]
+# #### **Inspecting existing heat network zones**
+
+# %%
+# Inspecting each HNZ
+plymouth_boundary = plymouth_la_boundaries_gdf.plot(
+    color="white", edgecolor="black"
+)  # Plymouth LA boundary
+
+plymouth_hn_zones_gdf.plot(
+    ax=plymouth_boundary,
+    edgecolor="blue",
+    legend=True,
+    column="ZoneID",
+)  # Existing heat network zones
+
+# %%
+# Label residential UPRNs in HNZ
+plymouth_uprns_hnz_labelled_df = heat_network_zones.label_gdf_heat_network_zone_uprns(
+    uprn_gdf=plymouth_residential_uprns_gdf,
+    hn_zone_gdf=plymouth_hn_zones_gdf,
+    usecols=["ZoneID"],
+)
+
+# Filter only in HNZ
+plymouth_hn_zone_uprns_df = plymouth_uprns_hnz_labelled_df.filter(
+    pl.col("in_hn_zone") == True
+)
+plymouth_hn_zone_uprns_gdf = uprns.generate_gdf_uprn_coords(plymouth_hn_zone_uprns_df)
+
+# %%
+# Inspecting layers of UPRNs, boundaries and HNZ
+
+# Layers
+plymouth_boundary = plymouth_la_boundaries_gdf.plot(color="white", edgecolor="black")
+
+plymouth_uprns_gdf.plot(
+    ax=plymouth_boundary, color="grey", markersize=2, label="All Plymouth UPRNs"
+)
+
+plymouth_residential_uprns_gdf.plot(
+    ax=plymouth_boundary, color="red", markersize=2, label="Residential UPRNs only"
+)
+
+plymouth_hn_zones_gdf.plot(
+    ax=plymouth_boundary, facecolor="none", edgecolor="blue", linewidth=2
+)
+
+plymouth_hn_zone_uprns_gdf.plot(
+    ax=plymouth_boundary, color="yellow", markersize=2, label="Residential UPRNs in HNZ"
+)
+
+# Manual legend patches for boundaries
+boundary_patch = mpatches.Patch(
+    facecolor="white", edgecolor="black", label="Plymouth LA boundary"
+)
+
+hn_zone_patch = mpatches.Patch(
+    facecolor="none", edgecolor="blue", label="Existing heat network zones"
+)
+
+# Combine auto and manual labels
+handles, labels = plymouth_boundary.get_legend_handles_labels()
+
+new_handles = handles + [boundary_patch, hn_zone_patch]
+new_labels = labels + ["Plymouth LA boundary", "Existing heat network zones"]
+
+plt.legend(
+    handles=new_handles,
+    labels=new_labels,
+    loc="upper left",
+    bbox_to_anchor=(1.05, 1),
+    title="Layers",
+)
+
+plt.show()
+
+# %%
+# What % of residential UPRNs are in existing HNZ?
+prop_in_hn_zone = plymouth_uprns_hnz_labelled_df.select(
+    pl.col("in_hn_zone").mean()
+).item()
+
+print(f"Proportion in HN zone: {prop_in_hn_zone:.3f}")
+
+# %% [markdown]
+# #### **Investigating which spatial signature types are in existing heat network zones**
+
+# %%
+# Spatial signatures in Plymouth LA boundaries only
+spatial_signatures_plymouth_gdf = spatial_signatures_gb_simplified_gdf.sjoin(
+    plymouth_la_boundaries_gdf, how="inner", predicate="intersects"
+).drop(columns="index_right")
+
+# %%
+# colour for spatial signature types
+unique_types = spatial_signatures_plymouth_gdf["type"].unique()
+colors = plt.cm.tab20.colors
+color_dict = {t: colors[i % len(colors)] for i, t in enumerate(unique_types)}
+
+
+fig, ax = plt.subplots(figsize=(15, 7.5))
+
+# plot spatial signatures polygons with mapped colors
+spatial_signatures_plymouth_gdf.plot(
+    ax=ax, color=spatial_signatures_plymouth_gdf["type"].map(color_dict)
+)
+
+# overlay Plymouth LA boundaries
+plymouth_la_boundaries_gdf.plot(ax=ax, facecolor="none", edgecolor="black", linewidth=1)
+
+# overlay Plymouth heat network zones
+plymouth_hn_zones_gdf.plot(ax=ax, facecolor="none", edgecolor="blue", linewidth=1)
+
+# manual legend for spatial signature types
+legend_handles = [mpatches.Patch(color=color_dict[t], label=t) for t in unique_types]
+
+# lines for boundaries and heat network zones
+legend_handles += [
+    Line2D([0], [0], color="black", lw=2, label="Plymouth LA boundary"),
+    Line2D([0], [0], color="blue", lw=2, label="Existing heat network zones"),
+]
+
+ax.legend(
+    handles=legend_handles,
+    title="Spatial Signature Type",
+    loc="upper left",
+    bbox_to_anchor=(1.05, 1),
+)
+
+plt.show()
+
+# %% [markdown]
+# 05/12/25: There is no clear consistency in the spatial signature types in existing heat network zones.
+
+# %% [markdown]
+# Inspecting existing HNZ one at a time:
+
+# %%
+plymouth_hn_zones_with_labels_gdf = gpd.sjoin(
+    plymouth_hn_zones_gdf,
+    spatial_signatures_plymouth_gdf[["type", "geometry"]],
+    how="left",
+    predicate="within",
+)
+
+plymouth_hn_zones_with_labels_agg_gdf = (
+    plymouth_hn_zones_with_labels_gdf.groupby("ZoneID")["type"]
+    .apply(lambda x: list(x.dropna().unique()))
+    .reset_index()
+)
+
+
+# %%
+# helper function to visually inspect each HNZ
+def inspect_spatial_signatures_in_hn_zone(zone_id: str):
+    zone_gdf = plymouth_hn_zones_gdf[plymouth_hn_zones_gdf["ZoneID"] == zone_id]
+
+    just_text_raw = str(zone_gdf.Just.values[0])
+    just_text = "\n".join(textwrap.wrap(just_text_raw, width=90))
+
+    spatial_signatures_gdf = spatial_signatures_plymouth_gdf.sjoin(
+        zone_gdf, how="inner", predicate="intersects"
+    )
+
+    fig, ax = plt.subplots(figsize=(12, 8))
+
+    base = spatial_signatures_gdf.plot(column="type", legend=True, ax=ax)
+    zone_gdf.plot(ax=base, facecolor="none", edgecolor="red")
+    ax.set_title(f"Zone: {zone_id}", fontsize=12)
+
+    fig.text(
+        0.5,
+        0.01,
+        just_text,
+        ha="center",
+        va="bottom",
+        fontsize=10,
+    )
+
+    return ax
+
+
+# %%
+for zone_id in plymouth_hn_zones_with_labels_agg_gdf["ZoneID"].unique():
+    inspect_spatial_signatures_in_hn_zone(zone_id)
+
+# %% [markdown]
+# #### **Testing urban spatial signatures to indicate "city centre" areas - i.e. areas potentially suitable for a heat network**
+
+# %% [markdown]
+# Labelling original residential UPRNs
+
+# %%
+# Test set of signature types
+city_centre_types = [
+    "Hyper concentrated urbanity",
+    "Concentrated urbanity",
+    "Metropolitan urbanity",
+    "Regional urbanity",
+    "Local urbanity",
+    "Dense urban neighbourhoods",
+]
+
+plymouth_uprns_spatial_signature_labelled_df = (
+    heat_network_zones.label_gdf_heat_network_spatial_signatures_uprns(
+        uprn_gdf=plymouth_residential_uprns_gdf,
+        spatial_signatures_gdf=spatial_signatures_gb_simplified_gdf,
+        types=city_centre_types,
+    )
+)
+
+# %%
+# Process filtered UPRNs for plotting
+plymouth_uprns_spatial_signature_df = (
+    plymouth_uprns_spatial_signature_labelled_df.filter(
+        pl.col("in_potential_hn_zone") == True
+    )
+)
+plymouth_uprns_spatial_signature_gdf = uprns.generate_gdf_uprn_coords(
+    plymouth_uprns_spatial_signature_df
+)
+
+# %%
+# Filter spatial signatures in Plymouth to selected types
+selected_spatial_signatures_plymouth_gdf = spatial_signatures_plymouth_gdf[
+    spatial_signatures_plymouth_gdf["type"].isin(city_centre_types)
+]
+
+# %%
+# Inspect selected spatial signatures in relation to existing HNZ
+
+# colour for spatial signature types
+unique_types = selected_spatial_signatures_plymouth_gdf["type"].unique()
+colors = plt.cm.tab20.colors
+color_dict = {t: colors[i % len(colors)] for i, t in enumerate(unique_types)}
+
+
+fig, ax = plt.subplots(figsize=(15, 7.5))
+
+# plot spatial signatures polygons with mapped colors
+selected_spatial_signatures_plymouth_gdf.plot(
+    ax=ax, color=selected_spatial_signatures_plymouth_gdf["type"].map(color_dict)
+)
+
+# overlay Plymouth LA boundaries
+plymouth_la_boundaries_gdf.plot(ax=ax, facecolor="none", edgecolor="black", linewidth=1)
+
+# overlay Plymouth heat network zones
+plymouth_hn_zones_gdf.plot(ax=ax, facecolor="none", edgecolor="blue", linewidth=1)
+
+# manual legend for spatial signature types
+legend_handles = [mpatches.Patch(color=color_dict[t], label=t) for t in unique_types]
+
+# lines for boundaries and heat network zones
+legend_handles += [
+    Line2D([0], [0], color="black", lw=2, label="Plymouth LA boundary"),
+    Line2D([0], [0], color="blue", lw=2, label="Existing heat network zones"),
+]
+
+ax.legend(
+    handles=legend_handles,
+    title="Spatial Signature Type",
+    loc="upper left",
+    bbox_to_anchor=(1.05, 1),
+)
+
+plt.show()
+
+# %% [markdown]
+# 05/12/25: Only 3 of the 6 city centre spatial signature types are present within Plymouth LA boundaries (Dense urban neighbourhoods, Local urbanity and Regional urbanity). Only the existing HNZ PLYM-011 contain any of these city centre types. The remaining 7 existing HN zones are in a wide range of other spatial signature types.
+
+# %% [markdown]
+# **Comparison of zones by spatial coverage**
+
+# %%
+# Comparing total spatial overlap
+overlap = gpd.overlay(
+    plymouth_hn_zones_gdf, selected_spatial_signatures_plymouth_gdf, how="intersection"
+)
+overlap_area = overlap.geometry.area.sum()
+
+# % overlap of each layer
+area_hnz = plymouth_hn_zones_gdf.geometry.area.sum()
+area_selected = selected_spatial_signatures_plymouth_gdf.geometry.area.sum()
+
+percent_hnz_covered_by_selected = (overlap_area / area_hnz) * 100
+percent_selected_covered_by_hnz = (overlap_area / area_selected) * 100
+
+# summary table
+summary_df = pd.DataFrame(
+    {
+        "Area (m²)": [area_hnz, area_selected, overlap_area],
+        "% of HNZ covered by selected signatures": [
+            "100%",
+            f"{percent_selected_covered_by_hnz:.1f}%",
+            f"{(overlap_area / area_hnz) * 100:.1f}%",
+        ],
+        "% of selected signatures covered by HNZ": [
+            f"{(area_hnz / area_selected) * 100:.1f}%",
+            "100%",
+            f"{(overlap_area / area_selected) * 100:.1f}%",
+        ],
+    },
+    index=["Existing HNZ", "Selected signatures", "Actual overlap"],
+)
+
+summary_df
+
+# %% [markdown]
+# 05/12/25: Computing the spatial overlap, the **city centre types cover 35% of the area covered by existing heat network zones**. Conversely, the **existing heat network zones cover 83% of the area covered by city centre types**.
+# - 35% metric -> The selected signature types only partially align with the existing heat network coverage; most of the current HNZ lie outside "city centre" areas. This indicates that using "city centre" areas is not able to capture all areas that have been deemed suitable for a heat network in reality.
+# - 83% metric -> Most of the areas identified as city centre/potentially suitable for a heat network are already part of the existing heat network.
+# - Using city centre areas as proxies for heat network suitability: most areas selected are part of an existing HNZ (high precision) - but many existing HNZ areas outside these urban centres are missed, since HNZ are not chosen based on urban location alone.
+
+# %%
+# Inspecting layers of UPRNs, boundaries and spatial signatures
+
+# Layers
+plymouth_boundary = plymouth_la_boundaries_gdf.plot(color="white", edgecolor="grey")
+
+
+plymouth_residential_uprns_gdf.plot(
+    ax=plymouth_boundary, color="#8B0000", markersize=2, label="Residential UPRNs only"
+)
+
+plymouth_hn_zones_gdf.plot(
+    ax=plymouth_boundary,
+    facecolor="none",
+    edgecolor="#1f77b4",
+    linewidth=2,
+    alpha=0.8,
+    label="HNZ Zones",
+)
+
+plymouth_hn_zone_uprns_gdf.plot(
+    ax=plymouth_boundary,
+    color="#FFD700",
+    markersize=3,
+    label="Residential UPRNs in HNZ",
+)
+
+selected_spatial_signatures_plymouth_gdf.plot(
+    ax=plymouth_boundary,
+    facecolor="none",
+    edgecolor="#800080",
+    linewidth=2,
+    label="Selected Spatial Signatures",
+)
+
+plymouth_uprns_spatial_signature_gdf.plot(
+    ax=plymouth_boundary,
+    color="#228B22",
+    markersize=3,
+    label="Residential UPRNs in selected spatial signatures",
+)
+
+# Manual legend patches for boundaries
+boundary_patch = mpatches.Patch(
+    facecolor="white", edgecolor="grey", label="Plymouth LA boundary"
+)
+
+hn_zone_patch = mpatches.Patch(
+    facecolor="none", edgecolor="#1f77b4", label="Existing heat network zones"
+)
+
+spatial_signature_patch = mpatches.Patch(
+    facecolor="none", edgecolor="#800080", label="Selected spatial signatures"
+)
+
+# Combine auto and manual labels
+handles, labels = plymouth_boundary.get_legend_handles_labels()
+
+new_handles = handles + [boundary_patch, hn_zone_patch, spatial_signature_patch]
+new_labels = labels + [
+    "Plymouth LA boundary",
+    "Existing heat network zones",
+    "Selected spatial signatures",
+]
+
+plt.legend(
+    handles=new_handles,
+    labels=new_labels,
+    loc="upper left",
+    bbox_to_anchor=(1.05, 1),
+    title="Layers",
+)
+
+plt.show()
+
+# %% [markdown]
+# Labelling UPRNs with existing HNZ labels **and** spatial signature types
+
+# %%
+plymouth_uprns_hnz_labelled_gdf = uprns.generate_gdf_uprn_coords(
+    plymouth_uprns_hnz_labelled_df
+)
+
+
+# Test set of signature types
+city_centre_types = [
+    "Hyper concentrated urbanity",
+    "Concentrated urbanity",
+    "Metropolitan urbanity",
+    "Regional urbanity",
+    "Local urbanity",
+    "Dense urban neighbourhoods",
+]
+
+plymouth_uprns_hnz_spatial_signature_labelled_df = (
+    heat_network_zones.label_gdf_heat_network_spatial_signatures_uprns(
+        uprn_gdf=plymouth_uprns_hnz_labelled_gdf,
+        spatial_signatures_gdf=spatial_signatures_gb_simplified_gdf,
+        types=city_centre_types,
+    )
+)
+
+# %% [markdown]
+# **Comparison of zones by UPRNs coverage**
+
+# %%
+plymouth_uprns_hnz_spatial_signature_labelled_df.group_by(
+    ["in_hn_zone", "in_potential_hn_zone"]
+).len().with_columns((pl.col("len") / pl.col("len").sum()).alias("percent"))
+
+# %%
+# UPRNs in existing HNZ
+hnz_true_df = plymouth_uprns_hnz_spatial_signature_labelled_df.filter(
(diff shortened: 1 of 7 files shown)


asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -57,3 +57,39 @@ def load_gdf_heat_network_zones(local_authority: str, **kwargs) -> gpd.GeoDataFr
...
+    Load GeoDataFrame with polygons in GB from the Spatial Signatures Framework  classified by their
+    Spatial Signature type. (Source: https://doi.org/10.6084/m9.figshare.16691575).
+
+    The dataset can be loaded at two levels of detail:
+    - "simplified": geometries provided with only fields "id" and "type"
+    - "full": geometries provided with all fields
+
+    Args:
+        detail_level (str, optional): Which level of descriptive detail to load.
+            Must be either "simplified" or "full". Defaults to "simplified".
```
crispy-wonton: Does `simplified` just mean fewer columns loaded? If so, I would change the word to something else here. 'simplified' already has a meaning when talking about GeoDataFrames and refers to a simplification of the geometries themselves, e.g. reducing the number of points in a polygon.

You could change it to a `boolean` argument and call it something like `required_fields` or whatever you think is intuitive and then have it load either the full dataset or only the required fields.

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -57,3 +57,39 @@ def load_gdf_heat_network_zones(local_authority: str, **kwargs) -> gpd.GeoDataFr
...
     )
     return gdf
+
+
+def load_gdf_spatial_signatures_gb(
+    detail_level: str = "simplified", **kwargs
+) -> gpd.GeoDataFrame:
+    """
+    Load GeoDataFrame with polygons in GB from the Spatial Signatures Framework  classified by their
+    Spatial Signature type. (Source: https://doi.org/10.6084/m9.figshare.16691575).
```
crispy-wonton: Please can you add the crs to the description here :)

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -54,3 +57,64 @@ def label_gdf_heat_network_zone_uprns(
...
+    """
+    Labels UPRNs that are located within a given Spatial Signature type deemed to be potentially
+    suitable for a heat network.
+
+    Args:
+        uprn_gdf (gpd.GeoDataFrame): UPRNs with point geometries to be labelled
+        spatial_signatures_gdf (gpd.GeoDataFrame): polygons of spatial signatures
+        usecols (list, optional): names of descriptive columns from spatial_signatures_gdf to be joined with uprn_gdf (in addition to "geometry" and "type").
+            Default is None to use only "geometry" and "type".
+        types (list, optional): spatial signature types potentially suitable for a heat network. Default is None to use all spatial signature types.
```
crispy-wonton: I would probably set the default spatial signature types to those you've selected. You could put them as a global variable at the top of this file. Otherwise, here I think we could get another case of silent fail where everything gets labelled as potentially suitable for a HNZ and we can't immediately see why.

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -54,3 +57,64 @@ def label_gdf_heat_network_zone_uprns(
...
+    types: list = None,
+) -> pl.DataFrame:
+    """
+    Labels UPRNs that are located within a given Spatial Signature type deemed to be potentially
+    suitable for a heat network.
+
+    Args:
+        uprn_gdf (gpd.GeoDataFrame): UPRNs with point geometries to be labelled
+        spatial_signatures_gdf (gpd.GeoDataFrame): polygons of spatial signatures
+        usecols (list, optional): names of descriptive columns from spatial_signatures_gdf to be joined with uprn_gdf (in addition to "geometry" and "type").
```
crispy-wonton: Is there a reason you allowed this functionality instead of just going with the default options you've selected? If not, we can cut this out to simplify the function.

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -54,3 +57,64 @@ def label_gdf_heat_network_zone_uprns(
...
+        spatial_signatures_gdf = spatial_signatures_gdf.to_crs(target_crs)
+        print(f"spatial_signatures_gdf reprojected to target CRS: {target_crs}")
+
+    # If usecols specified, filter columns in spatial_signatures_gdf
+    if usecols:
+        spatial_signatures_gdf = spatial_signatures_gdf[["geometry", "type"] + usecols]
+    else:
+        spatial_signatures_gdf = spatial_signatures_gdf[["geometry", "type"]]
+
+    # Spatial join for for labelling UPRN with signature type
```
crispy-wonton:
```suggestion
    # Spatial join for labelling UPRN with signature type
```

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -54,3 +57,64 @@ def label_gdf_heat_network_zone_uprns(
...
+    labelled_uprn_gdf = uprn_gdf.sjoin(
+        spatial_signatures_gdf,
+        how="left",
+        predicate="intersects",  # include properties intersecting spatial signature cell boundary
```
crispy-wonton: We have to be careful that we don't end up with UPRNs (if not in Plymouth, then it could happen with scaling) with multiple spatial signatures here in cases when a UPRN is on a border because of using the `intersects` predicate. This could mean you end up with UPRNs which are labelled as both being within and not within a potential HN zone.

I think the solution would be to keep the predicate at `intersects`, but set a rule so that any UPRNs which end up in this situation get assigned `in_potential_hn_zone == True`. 

I've checked the `s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_residential_uprns_with_city_centres.parquet` output and there are no duplicated UPRNs so it doesn't seem to be a problem at the moment but let's put some code in place to address it in case these edge cases exist when scaling.

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -54,3 +57,64 @@ def label_gdf_heat_network_zone_uprns(
...
+    ).drop(columns="index_right")
+
+    # Add potential heat network zone boolean label to original uprn_gdf
+    labelled_uprn_gdf["in_potential_hn_zone"] = (
```
crispy-wonton: I would relabel this as `in_city_centre`. I think this label has the potential to cause some confusion given that DESNZ's zones could be thought of as 'potential' HN zones.

I know that `in_city_centre` also isn't perfectly accurate but since it aligns with our phrasing in the decision tree I think that's clearer for now. Open to other ideas.

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -54,3 +57,64 @@ def label_gdf_heat_network_zone_uprns(
...
+        labelled_uprn_gdf["type"].isin(types) if types else True
+    )
+    labelled_uprn_gdf["in_potential_hn_zone"] = labelled_uprn_gdf[
+        "in_potential_hn_zone"
+    ].fillna(False)
```
crispy-wonton: Why would there be NA values in this column? Is it because some UPRNs do not get assigned a spatial signature?

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -54,3 +57,64 @@ def label_gdf_heat_network_zone_uprns(
...
+
+    # Return as polars df without geometry
+    labelled_uprn_df = pl.from_pandas(labelled_uprn_gdf.drop(columns="geometry"))
+
+    labelled_uprn_df = labelled_uprn_df.rename({"type": "spatial_signature_type"})
```
crispy-wonton:
```suggestion
    labelled_uprn_df = pl.from_pandas(labelled_uprn_gdf.drop(columns="geometry")).rename({"type": "spatial_signature_type"})
```

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
```diff
@@ -90,8 +130,14 @@ def parse_arguments() -> argparse.Namespace:
         # Placeholder for future implementation
         raise NotImplementedError("Processing for all of GB is not yet supported.")
 
-    # Save residential UPRNs with labels to S3
+    # Save residential UPRNs with existing hn zone labels to S3
```
crispy-wonton: I think we can save HNZ and city centre labels into the same output file - lmk if there's a reason you kept separate :)

asf_heat_pump_suitability/config/base.yaml
sofiapinto, line 1: We need to also update this https://github.com/nestauk/asf_heat_pump_suitability/blob/dev/asf_heat_pump_suitability/config/README.md either in this or in a separate PR with the two new datasets, the HNZ doc and the spatial signatures framework.

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -57,3 +57,39 @@ def load_gdf_heat_network_zones(local_authority: str, **kwargs) -> gpd.GeoDataFr
...
+            f"detail_level must be 'full' or 'simplified', not {detail_level}"
+        )
+
+    gdf = base_getters.get_gdf_from_gpkg_s3_path(
+        path=config["data"]["geodata"]["gb_spatial_signatures"][detail_level],
+        **kwargs,
+    )
+
+    print(
+        f"Spatial signatures {detail_level} geodataframe successfully loaded with CRS {gdf.crs}."
```
sofiapinto, line 103: Nice! I love that we print the CRS too, so useful.

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
```diff
@@ -55,13 +55,15 @@ def parse_arguments() -> argparse.Namespace:
...
+        """ Existing heat network zones """
+
         # Load Plymouth existing heat network zone polygons
         print("Loading heat network zone data for Plymouth Local Authority...")
         plymouth_hn_zones_gdf = load_geodata.load_gdf_heat_network_zones(
             local_authority="plymouth"
         )
 
-        # Filter for UPRNs in existing heat network zones
+        # Label UPRNs in existing heat network zones
```
sofiapinto: I wonder if it might be useful to clarify
```suggestion
        # Label UPRNs in existing, potential and planned heat network zones
```

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
```diff
@@ -78,7 +80,45 @@ def parse_arguments() -> argparse.Namespace:
...
+
+        # Label UPRNs in city centres
+        city_centre_types = [  # TODO confirm types
+            "Hyper concentrated urbanity",
+            "Concentrated urbanity",
+            "Metropolitan urbanity",
+            "Regional urbanity",
+            "Local urbanity",
+            "Dense urban neighbourhoods",
+        ]
```
sofiapinto: i'd check with @crispy-wonton about how she's been doing this across the repo, but i'd suggest we maybe want to move this to `base.yaml`?

sofiapinto: I think the areas selected sound good for our MVP!

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
```diff
@@ -78,7 +80,45 @@ def parse_arguments() -> argparse.Namespace:
...
+
+        # Clean up columns
+        city_centre_uprn_df = city_centre_uprn_df.select(
+            [
+                "UPRN",
+                "LAD23NM",
+                "X_COORDINATE",
+                "Y_COORDINATE",
+                "spatial_signature_type",
+                "in_potential_hn_zone",
```
sofiapinto: i''d call this one `in_city_centre` in the original function

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
```diff
@@ -90,8 +130,14 @@ def parse_arguments() -> argparse.Namespace:
...
     save_utils.save_to_s3(
         hn_zone_uprn_df,
         f"s3://asf-heat-pump-suitability/local_heat_planning/outputs/{args.local_authorities}_residential_uprns_with_hn_zones.parquet",
     )
+
+    # Save residential UPRNs with city centre labels to S3
+    save_utils.save_to_s3(
+        city_centre_uprn_df,
+        f"s3://asf-heat-pump-suitability/local_heat_planning/outputs/{args.local_authorities}_residential_uprns_with_city_centres.parquet",
+    )
```
sofiapinto, line 132: I'll leave it to @crispy-wonton on how data should be saved but perhaps these two flags could be saved in the same file instead of separate files? Two separate columns, but same file.

We also have a `plymouth_residential_uprns_with_features.parquet` where eventually all relevant columns will be saved, I imagine. I'm guessing that is a product of Roisin's work in a separate PR.

asf_heat_pump_suitability/research/exploratory/20251205_plymouth_investigation.py
```diff
@@ -0,0 +1,530 @@
...
+    detail_level="simplified"
+)
+
+# %%
+# Confirming spatial signature types
+print("Spatial signatures:")
+counter = 1
+for type in spatial_signatures_gb_simplified_gdf["type"].unique():
+    print(f"{counter}. {type}")
+    counter += 1
```
crispy-wonton, line 74: I love this! 😄 the counter delighted me

asf_heat_pump_suitability/research/exploratory/20251205_plymouth_investigation.py
```diff
@@ -0,0 +1,530 @@
...
+# %%
+# colour for spatial signature types
+unique_types = spatial_signatures_plymouth_gdf["type"].unique()
+colors = plt.cm.tab20.colors
+color_dict = {t: colors[i % len(colors)] for i, t in enumerate(unique_types)}
+
+
+fig, ax = plt.subplots(figsize=(15, 7.5))
+
+# plot spatial signatures polygons with mapped colors
```
crispy-wonton, line 188: nice plot!

asf_heat_pump_suitability/research/exploratory/20251205_plymouth_investigation.py
```diff
@@ -0,0 +1,530 @@
...
+)
+
+
+# %%
+# helper function to visually inspect each HNZ
+def inspect_spatial_signatures_in_hn_zone(zone_id: str):
+    zone_gdf = plymouth_hn_zones_gdf[plymouth_hn_zones_gdf["ZoneID"] == zone_id]
+
+    just_text_raw = str(zone_gdf.Just.values[0])
+    just_text = "\n".join(textwrap.wrap(just_text_raw, width=90))
```
crispy-wonton, line 245:
```suggestion
    # Add HN zone notes 
    just_text_raw = str(zone_gdf.Just.values[0])
    just_text = "\n".join(textwrap.wrap(just_text_raw, width=90))
```
Something like this, because it's not clear from looking at it what `Just` is.

crispy-wonton, line 245: Nice! haven't seen `textwrap` before - looks very useful

asf_heat_pump_suitability/analysis/spatial_signatures_hn_zones/20251205_plymouth_investigation.py
```diff
@@ -0,0 +1,530 @@
...
+    ax=plymouth_boundary, color="#8B0000", markersize=2, label="Residential UPRNs only"
+)
+
+plymouth_hn_zones_gdf.plot(
+    ax=plymouth_boundary,
+    facecolor="none",
+    edgecolor="#1f77b4",
+    linewidth=2,
+    alpha=0.8,
+    label="HNZ Zones",
```
crispy-wonton:
```suggestion
    label="HN Zones",
```

asf_heat_pump_suitability/research/exploratory/20251205_plymouth_investigation.py
```diff
@@ -0,0 +1,530 @@
...
+plymouth_uprns_hnz_spatial_signature_labelled_df = (
+    heat_network_zones.label_gdf_heat_network_spatial_signatures_uprns(
+        uprn_gdf=plymouth_uprns_hnz_labelled_gdf,
+        spatial_signatures_gdf=spatial_signatures_gb_simplified_gdf,
+        types=city_centre_types,
+    )
+)
+
+# %% [markdown]
+# **Comparison of zones by UPRNs coverage**
```
crispy-wonton, line 505: You could also plot a confusion matrix here: https://scikit-learn.org/stable/modules/generated/sklearn.metrics.ConfusionMatrixDisplay.html

asf_heat_pump_suitability/analysis/spatial_signatures_hn_zones/20251205_plymouth_investigation.py
```diff
@@ -0,0 +1,530 @@
...
+    title="Layers",
+)
+
+plt.show()
+
+# %%
+# What % of residential UPRNs are in existing HNZ?
+prop_in_hn_zone = plymouth_uprns_hnz_labelled_df.select(
+    pl.col("in_hn_zone").mean()
+).item()
```
crispy-wonton:
```suggestion
prop_in_hn_zone = plymouth_uprns_hnz_labelled_df["in_hn_zone"].mean()
```

asf_heat_pump_suitability/research/exploratory/20251205_plymouth_investigation.py
```diff
@@ -0,0 +1,530 @@
...
+            "100%",
+            f"{(overlap_area / area_selected) * 100:.1f}%",
+        ],
+    },
+    index=["Existing HNZ", "Selected signatures", "Actual overlap"],
+)
+
+summary_df
+
+# %% [markdown]
```
crispy-wonton, line 396: This is a great summary and much easier to understand than the table above. Especially with the 236% value which I'm not sure how to interpret..

asf_heat_pump_suitability/research/exploratory/20251205_plymouth_investigation.py
```diff
@@ -0,0 +1,530 @@
...
+    spatial_signatures_plymouth_gdf[["type", "geometry"]],
+    how="left",
+    predicate="within",
+)
+
+plymouth_hn_zones_with_labels_agg_gdf = (
+    plymouth_hn_zones_with_labels_gdf.groupby("ZoneID")["type"]
+    .apply(lambda x: list(x.dropna().unique()))
+    .reset_index()
+)
```
crispy-wonton, line 235: Is this meant to be deleted?

asf_heat_pump_suitability/analysis/spatial_signatures_hn_zones/20251205_plymouth_investigation.py
```diff
@@ -0,0 +1,530 @@
...
+plymouth_uprns_hnz_spatial_signature_labelled_df.group_by(
+    ["in_hn_zone", "in_potential_hn_zone"]
+).len().with_columns((pl.col("len") / pl.col("len").sum()).alias("percent"))
+
+# %%
+# UPRNs in existing HNZ
+hnz_true_df = plymouth_uprns_hnz_spatial_signature_labelled_df.filter(
+    (pl.col("in_hn_zone") == True)
+)
+prop_true = hnz_true_df.select(pl.col("in_potential_hn_zone").mean()).item()
```
crispy-wonton:
```suggestion
prop_true = hnz_true_df["in_potential_hn_zone"].mean()
```

asf_heat_pump_suitability/analysis/spatial_signatures_hn_zones/20251205_plymouth_investigation.py
```diff
@@ -0,0 +1,530 @@
...
+summary_df = pd.DataFrame(
+    {
+        "Area (m²)": [area_hnz, area_selected, overlap_area],
+        "% of HNZ covered by selected signatures": [
+            "100%",
+            f"{percent_selected_covered_by_hnz:.1f}%",
+            f"{(overlap_area / area_hnz) * 100:.1f}%",
+        ],
+        "% of selected signatures covered by HNZ": [
+            f"{(area_hnz / area_selected) * 100:.1f}%",
```
crispy-wonton: This is just a size comparison right because there's no overlay involved and doesn't actually calculate percentage coverage?

asf_heat_pump_suitability/analysis/spatial_signatures_hn_zones/20251205_plymouth_investigation.py
```diff
@@ -0,0 +1,530 @@
...
+)
+
+prop_true = potential_hnz_true_df.select(pl.col("in_hn_zone").mean()).item()
+
+print(
+    f"Proportion of UPRNs in a potential heat network zone (as classified by the set of spatial signature types) which are already in an existing heat network zone [precision]: {prop_true:.3f}"
+)
+
+# %% [markdown]
+# 05/12/25: City centres as a proxy performs better at capturing properties in existing HNZ, compared to land area. Mainly because of the higher density of properties in the selected urban area types.
```
crispy-wonton: What do you mean 'compared to land area'?

asf_heat_pump_suitability/research/exploratory/20251205_plymouth_investigation.py
crispy-wonton, line 1: Please could you move this file to `asf_heat_pump_suitability/research/exploratory/` - I dont think you'll need a subdir there :)

Please could you add a summary description of the notebook at the top. Can be just a few bullet points about what it contains.

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
```diff
@@ -78,7 +80,45 @@ def parse_arguments() -> argparse.Namespace:
...
+        """ City centre areas """
+
+        # Load spatial signature polygons
+        print("Loading spatial signatures dataset...")
+        spatial_signatures_gb_simplified_gdf = (
+            load_geodata.load_gdf_spatial_signatures_gb(detail_level="simplified")
+        )
+
+        # Label UPRNs in city centres
+        city_centre_types = [  # TODO confirm types
```
```suggestion
        city_centre_types = [  # TODO confirm types with scaling
```
I think this is especially important!

crispy-wonton (changes requested):
Thanks for this @eglucas ! A great piece of analysis, great plotting, and thoroughly thought through - great work :D 

I've left some comments and suggestions. The main requested change is to make sure that we're addressing any edge cases of UPRNs that may arise when scaling which end up on the boundary between two spatially signatures meaning a single UPRN could receive 2 conflicting labels. You will find my comment about this in the `heat_network_zones.py` file.

To answer your qs in the PR description:

> Please inspect the new city_centre_uprn_df that is saved out to S3

I checked a few things - number of UPRNs is correct; proportions in city centre looks reasonable; no nulls in city centre label; spatial signatures are appropriately categorised into city centre and not. 

The only weird thing is there are some UPRNs which get `null` values for spatial signatures. I saw in your visualisations that there seems to be some gaps in the signatures, and can see you ultimately categorise these as not in city centres. I think it would be useful to note that somewhere as we might want to revisit it - especially when scaling, if there are larger gaps.

E.g. if there's an area with no spatial signature but is entirely enclosed by city centre, then maybe it should be classed as a city centre. Or maybe we should use a density metric in these areas instead? Not for this PR but good to think about if you have ideas.

> Whether you're happy with the saved location of the spatial signatures dataset, as well as how their paths are inserted in config

Paths of outputs and signatures datasets look good - I would just suggest combining city centre and HNZ output dfs into one file :)

> Schema of the saved output dataframe

Looks good !

> With regards to the analysis notebook, please run all the cells and have a look at the outputs. Let me know if you have any other thoughts on how we're evaluating whether the city centre signature types do a good enough job at flagging potential heat network zones. Please also flag if there are any errors in the analysis or interpretation.

Looks good to me and I like your visualisations, really helped to understand the data.

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -54,3 +57,64 @@ def label_gdf_heat_network_zone_uprns(
...
+
+def label_gdf_heat_network_spatial_signatures_uprns(
+    uprn_gdf: gpd.GeoDataFrame,
+    spatial_signatures_gdf: gpd.GeoDataFrame,
+    usecols: list = None,
+    types: list = None,
+) -> pl.DataFrame:
+    """
+    Labels UPRNs that are located within a given Spatial Signature type deemed to be potentially
+    suitable for a heat network.
```
sofiapinto: Happy to be challenged on this, but rather than talking about suitability for a heat network at this point I'd say something like:
"Labels UPRNs that are located within a given Spatial Signature type identified as a city centre" or similar.

Same about the function name.

In general it's a little confusing if we talk about DESNZ's heat network zones and then separately areas which are potentially suitable for HN. DESNZ's HN zones are also "potential"/planned. So i'd use different wording for spatial signatures - we're using them to identify high density areas/city centres, etc.

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -54,3 +57,64 @@ def label_gdf_heat_network_zone_uprns(
...
+
+    Args:
+        uprn_gdf (gpd.GeoDataFrame): UPRNs with point geometries to be labelled
+        spatial_signatures_gdf (gpd.GeoDataFrame): polygons of spatial signatures
+        usecols (list, optional): names of descriptive columns from spatial_signatures_gdf to be joined with uprn_gdf (in addition to "geometry" and "type").
+            Default is None to use only "geometry" and "type".
+        types (list, optional): spatial signature types potentially suitable for a heat network. Default is None to use all spatial signature types.
+
+    Returns:
+        pl.DataFrame: input UPRNs labelled with spatial signature identifiers and booleans indicating potential for a heat network
```
sofiapinto: same as above

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
sofiapinto: i'd use `in_city_centre`, like in the decision tree. otherwise it's confusing.

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
sofiapinto: same as above

asf_heat_pump_suitability/research/exploratory/20251205_plymouth_investigation.py
```diff
@@ -0,0 +1,530 @@
...
+plymouth_boundary = plymouth_la_boundaries_gdf.plot(
+    color="white", edgecolor="black"
+)  # Plymouth LA boundary
+
+plymouth_hn_zones_gdf.plot(
+    ax=plymouth_boundary,
+    edgecolor="blue",
+    legend=True,
+    column="ZoneID",
+)  # Existing heat network zones
```
sofiapinto, line 101:
```suggestion
)  # Existing/planned heat network zones
```

asf_heat_pump_suitability/analysis/spatial_signatures_hn_zones/20251205_plymouth_investigation.py
```diff
@@ -0,0 +1,530 @@
+# %% [markdown]
+# ### **Investigating areas in Plymouth LA with existing heat network zones or potential for a heat network**
```
sofiapinto: Or similar wording...
```suggestion
# ###**Investigating areas in Plymouth LA with existing/planned heat network zones or areas which are city centres**
```

asf_heat_pump_suitability/research/exploratory/20251205_plymouth_investigation.py
```diff
@@ -0,0 +1,530 @@
...
+# %%
+# CRS checks
+gdfs = [
+    plymouth_uprns_gdf,
+    plymouth_residential_uprns_gdf,
+    plymouth_la_boundaries_gdf,
+    plymouth_hn_zones_gdf,
+    spatial_signatures_gb_simplified_gdf,
+]
+len({gdf.crs for gdf in gdfs}) == 1
```
sofiapinto, line 85: smart :)

asf_heat_pump_suitability/research/exploratory/20251205_plymouth_investigation.py
sofiapinto, line 1: lots of great work here! :)

asf_heat_pump_suitability/research/exploratory/20251205_plymouth_investigation.py
```diff
@@ -0,0 +1,530 @@
...
+        ],
+    },
+    index=["Existing HNZ", "Selected signatures", "Actual overlap"],
+)
+
+summary_df
+
+# %% [markdown]
+# 05/12/25: Computing the spatial overlap, the **city centre types cover 35% of the area covered by existing heat network zones**. Conversely, the **existing heat network zones cover 83% of the area covered by city centre types**.
+# - 35% metric -> The selected signature types only partially align with the existing heat network coverage; most of the current HNZ lie outside "city centre" areas. This indicates that using "city centre" areas is not able to capture all areas that have been deemed suitable for a heat network in reality.
```
sofiapinto, line 398: And that's fine! I think we never expected this to capture all areas in the planned HNZ. The idea of having both is exactly to capture all areas which either have plans for a HNZ (which we know is also based on cost and other feasibility factors) with areas which we know from this new source that are of high density/which also represent city centres. We're not trying to see if the spatial signatures is a good proxy for DESNZ's HNZ.

asf_heat_pump_suitability/research/exploratory/20251205_plymouth_investigation.py
sofiapinto: i'd rephrase things using "city centre" rather than "potential HNZ" when talking about signatures data, as per comments left in other files.

sofiapinto (commented):
So much great work @eglucas! Also love how quickly you learnt about all these datasets & using/visualising geo data.

I did a quick review and left comments mostly about how we phrase things on planned HNZ vs identifying city centres.

I quickly ran the notebook, but didn't review in detail. I also haven't run the main script, but hopefully @crispy-wonton has :)

asf_heat_pump_suitability/getters/load_geodata.py
eglucas: Thank you so much for raising this - I didn't know about this "simplified" convention with geodataframes. As we've discussed on slack, I compared the number of polygon coordinate points for a chosen geometry in the simplified vs. full-res and there was 264 in simple and 720 in full. This does mean that the simplified suffix in the file name is referring to the simplified polygon geometry.

I have rewritten the getter docstring to reflect this and changed the default to be `detail_level = "full"`

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
eglucas: It was initially because I didn't know what other columns were in the full-res version of the dataset. Now that I've been able to load just a chunk of it and inspected the column names - I agree that it's best to remove this functionality.

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
eglucas: Thanks Sofia, that makes complete sense. 

FYI @crispy-wonton as well - I think it's best to make a clearer distinction between the existing HNZ labeller and this one for city centres. For complete clarity, I've moved this transformation function to a new separate module `transform/city_centres.py`. I'll rename the function and rewrite the docstring so that it's only related to city centre labelling (without making the leap that it could be a potential HNZ).

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
eglucas: I've ended up implementing @crispy-wonton 's suggestion to set this list of types as a global variable `CITY_CENTRE_TYPES` at the top of new `transform/city_centres.py` (and is now the default argument for `types`)

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
eglucas: Agree! Will combine into one output file

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
eglucas: Thanks Sofia, yes - Roisin made the same suggestion. Will combine into one output file!

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
eglucas: Thanks so much for raising this important point that I hadn't considered!

I've ended up implementing (in https://github.com/nestauk/asf_heat_pump_suitability/pull/203/commits/04316c7d3a77dc46fc2615b0c1d2a04a728b3095) what I think is pretty clean:
- all matched spatial signature types for each UPRN will be combined into a list
- `in_city_centre=True` for that UPRN if any type in the list is in `CITY_CENTRE_TYPES`

This way also means that all matched spatial signatures are preserved if they need to be inspected later.

asf_heat_pump_suitability/config/base.yaml
eglucas: Thanks Sofia, I've created a specific issue for this: https://github.com/nestauk/asf_heat_pump_suitability/issues/205

asf_heat_pump_suitability/research/exploratory/20251205_plymouth_investigation.py
eglucas: No, I've just not cleaned up the way that the helper function `inspect_spatial_signatures_in_hn_zone` is used. I need to create `plymouth_hn_zones_with_labels_agg_gdf` which is then looped over in the following code in subsequent lines

`for zone_id in plymouth_hn_zones_with_labels_agg_gdf["ZoneID"].unique():
    inspect_spatial_signatures_in_hn_zone(zone_id)`

This section can definitely be cleaned up but given that it's just for the purpose of this notebook and visualising specific outputs for exploration, I've left it as is.

asf_heat_pump_suitability/analysis/spatial_signatures_hn_zones/20251205_plymouth_investigation.py
eglucas: Ah good spot! This is a mistake. I've now corrected it to:
```suggestion
            f"{percent_hnz_covered_by_selected:.1f}%",
```

I've also cleaned up the summary table so it's clearer.

asf_heat_pump_suitability/analysis/spatial_signatures_hn_zones/20251205_plymouth_investigation.py
eglucas: I meant spatial coverage - have corrected this text.

```suggestion
# 05/12/25: City centres as a potential proxy performs better at capturing properties in existing HNZ (i.e. UPRN coverage), compared to spatial coverage. Mainly because of the higher density of properties in the selected urban area types.
```

eglucas:
Hi @crispy-wonton and @sofiapinto 

Thank you so much for your comments, super helpful!

I've addressed all your comments and made changes. To summarise the main changes:

`load_gdf_spatial_signatures_gb()`
- docstrings have been rewritten to more accurately reflect the difference between the "simplified" and the "full" geopackage file being loaded
- `detail_level = "full"` is now the default

Labeller transformation function has been renamed to `label_gdf_city_centre_spatial_signatures_uprns()` and is now in a new file `pipeline/transform/city_centres.py`
- City centre spatial signature types are defined as a global variable at the top of the module, and is the default for the `types` argument
- the boolean column is now named `"in_city_centre"
- To address the cases where multiple spatial signature types are matched to one UPRN: assigned spatial signature types are now compiled in a list for each UPRN under the column "spatial_signature_types"
- To address how UPRNs that do not get a spatial signature match and therefore do not get classified to be in a city centre: No code fix for now but I have put a TODO comment on line 78 as a reminder to revisit in future

All language used when using the city centre labeller function in the run script and in the exploratory notebook has been modified from referring to these areas as 'potential hn zones' to simply 'city centre areas'.

`run/heat_network_city_centres.py`
- `in_hn_zone` and `in_city_centre` boolean columns are added to the same dataframe and saved as a single output
- **NOTE: I still specify `detail_level="simplified" here because I can't test run the script otherwise. Feel free to change this in the final run script for all features.**

Exploratory notebook
- this has been moved to `research/exploratory`
- explainer markdown has been added at the top
- language modifications
- small code fixes and improvements

Please have a look and merge when/if you're happy!

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
crispy-wonton: Great idea to retain them in a list :)

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
crispy-wonton: Great, I think that's a good idea. Thanks @eglucas

asf_heat_pump_suitability/pipeline/transform/city_centres.py
```diff
@@ -0,0 +1,88 @@
...
+            **{col: "first" for col in uprn_columns},  # keep original UPRN columns
+            "geometry": "first",  # keep the point geometry
+            "type": list,  # combine types into a list
+        }
+    )
+
+    # Add city centre boolean label
+    labelled_uprn_gdf["in_city_centre"] = labelled_uprn_gdf["type"].apply(
+        lambda x: any(t in types for t in x)
+    )
```
crispy-wonton:
```suggestion
    # Add city centre boolean label
    labelled_uprn_gdf["in_city_centre"] = labelled_uprn_gdf["type"].isin(types)
    labelled_uprn_gdf = labelled_uprn_gdf.groupby("UPRN", group_keys=False).agg(
        {
            **{col: "first" for col in uprn_columns},  # keep original UPRN columns
            "geometry": "first",  # keep the point geometry
            "type": list,  # combine types into a list
            "in_city_centre": sum,  # sums >0 indicate UPRN in a city centre signature
        }
    # Convert city centre to boolean label
    ).astype({"in_city_centre": bool})
```
Just thinking about scaling, we might not want to use `apply` because of speed constraints. I think this works but haven't tested.

eglucas: Just tested it, and yes it works! Will make that change and push.

crispy-wonton (approved):

PR #210, opened 2026-01-15
209 enable running pipelines for any LA + prep data for manchester workshop

sofiapinto:
---

# Description

The main aim of this PR was to get the most suitable low carbon heating tech for Greater Manchester. To do that I have changed the existing code so that it can now be ran for any local authority (LA) or list of LAs. Before this, it was only straightforward to run the code for Plymouth.

Changes to the code in this PR include:
- `base.yaml`: changed how we import grid squares/LA data to enable running pipelines for other LAs/CAs beyond Plymouth
- `add_features.py`, `heat_network_city_centres.py`, `heat_network_zones.py`, `uprns.py`: changes to enable running pipelines for other LAs beyond Plymouth

Also added a new notebook:
- `asf_heat_pump_suitability/research/exploratory/decision_tree_categorisation/manchester_applying_decision_tree.py`: with code from PRs #207 and #206. This still needs to be refacotored.

Fixes #209 #212 

Read this if you don't know what a grid square is: https://www.ordnancesurvey.co.uk/documents/resources/guide-to-nationalgrid.pdf

# Instructions for Reviewer

Hey @simran-dave,

Because this is your first time reviewing a PR in this repository please follow the instructions below:
- Clone the repo with `git clone git@github.com:nestauk/asf_heat_pump_suitability.git`
- Navigate to your local repo folder in the terminal/through the Visual studio terminal
- Run `direnv allow`

To get the changes from this specific PR do:
- Fetch latest from origin`git fetch origin`
- Select the specific `git checkout 209_manchester_workshop`
- Run `make install` to configure the development environment
- Run `conda activate asf_heat_pump_suitability`

Let us know if you run into any problem with the above & we can help.

**Instructions for reviewing the code:**
- @simran-dave and @crispy-wonton: Please review the changes in `base.yaml`, `add_features.py`, `heat_network_city_centres.py`, `heat_network_zones.py`, `uprns.py` to check the logic is correct.

- @simran-dave please check the scripts run in test mode for Greater Manchester. To do so run the following on your command line:
     - `python asf_heat_pump_suitability/pipeline/transform/uprns.py --local_authorities greater_manchester_las --test_mode`
     - `python asf_heat_pump_suitability/pipeline/run/add_features.py --uprns s3://asf-heat-pump-suitability/local_heat_planning/outputs/greater_manchester_las_residential_uprns.parquet --test_mode`
[...]

PR #207, opened 2025-12-12
181 Build MVP block of flats classification model

crispy-wonton:
Fixes #181 

# Description

Add notebook to build MVP of block of flats building classification model.

 This notebook trains a model for the binary classification of buildings into blocks of flats or not. The notebook does the following:
 - Load building and UPRN data for Plymouth and sample areas
 - Create features for each building
 - Load and preprocess a manually labelled sample of data
 - Plot features for labelled data
 - Train and evaluate random forest binary classifier model, with hyperparameter selection and cross validation

New files:
-  `asf_heat_pump_suitability/research/exploratory/build_classification_model_block_of_flats.py` - notebook containing code for above.

# Instructions for Reviewer

@sofiapinto please could you do a full review of all the code
@caldwellst if you have time, it would be great if you could just look at the section called 'Train and evaluate model and tune hyperparameters'
@eglucas if you are interested, please feel free to take a look through and leave any comments or questions 

In order to test the code in this PR you need to ...
`pip install jupytext` if not already. Then run the following line in your terminal:
`jupytext --to notebook asf_heat_pump_suitability/research/exploratory/build_classification_model_block_of_flats.py`

Please then run the full notebook and review the outputs. It should take about 10-15 mins to run all the way through.

Please pay special attention to ...
- **Feature engineering** Is there any additional plots you would create to review the features or the labelled data and explore them? Also, are there any easy features to add that you think are important and missing?
- **Model hyperparameters**: Any changes you would make to the parameter distributions for hyperparameter tuning, or to the parameters in cross validation or classifier objects?
- **Robustness of model and trust in evaluation:** This is my main concern. I'm interested to hear if you have any concerns about any low counts of particular archetypes or how the nine archetypes could be affecting model performance. Let me know if you think the training, cross validation, and evaluation is appropriately robust or if there's something we could do better to train for the variation within the binary classes. Is there anything you thing is 

# Checklist:

- [ ] I have refactored my code out from `notebooks/`
- [x] I have checked the code runs
- [ ] I have tested the code
- [x] I have run `pre-commit` and addressed any issues not automatically fixed
- [x] I have merged any new changes from `dev`
- [x] I have documented the code
  - [x] Major functions have docstrings
  - [ ] Appropriate information has been added to `README`s
- [x] I have explained this PR above
- [x] I have requested a code review

PR #219, opened 2026-02-12
214 apply block classification model to dataset

### LINKED ISSUE #214: Refactor block of flats classification prototype

crispy-wonton:
Fixes #214 

# Description

Apply the trained random forest classifier model to unlabelled data to predict class of building (block of flats / not). Note, when PR #210 has been merged, we can merge the scaling of the `add_features.py` script to other selected LAs here.

Modified files:
- `asf_heat_pump_suitability/config/base.yaml` - update config path to a specific variable
- `asf_heat_pump_suitability/getters/base_getters.py` - add function to load pickle files 
- `asf_heat_pump_suitability/pipeline/impute/property_type.py` - update a print statement
- `asf_heat_pump_suitability/pipeline/model/block_of_flats/feature_engineering.py` - add a print statement
- `asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py` - add functions to apply trained model to unlabelled data
- `asf_heat_pump_suitability/pipeline/run/add_features.py` - call trained model in script to add features to UPRNs and use it to add the `in_block_of_flats` feature
- `asf_heat_pump_suitability/pipeline/transform/uprns.p`y - create new function to map UPRNs to their buildings 
- `asf_heat_pump_suitability/utils/geo_utils.py` - add function to verify the CRS of a geodataframe and convert to target CRS if not in it already

# Instructions for Reviewer

Run the following piece of code from the terminal to run the updated `add_features.py` script:
`python asf_heat_pump_suitability/pipeline/run/add_features.py --uprns s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_residential_uprns.parquet`

- The output data of the pipeline in relation to the two new features added: `in_block_of_flats` and `block_of_flats_label_proba`. I had a look at it myself and compared it to the outputs of my original notebook in #207 . The results look the same for UPRNs which are common across both datasets (the original notebook covers the full 6 sampling areas, whereas the script covers Plymouth only). Do you see anything weird or unexpected about the results?
- Do the updated predicates on the sjoins make sense? Is there a reason we shouldn't use `intersects`? I thought that it could create cases where a UPRN gets joined to multiple buildings, but in testing it on Plymouth, I saw that didn't seem to happen.

Fyi there are a small number of UPRNs (1.5% of Plymouth data) which don't receive a label. I investigated this to see if it was caused by a bug. It's caused by our domestic filtering methodology - not a bug but seems to be a data quality issue. Basically, the domestic filtering retains all UPRNs which are in the domestic EPC register. A small number of these seem to not get sjoined to a building - this could be because (a) the UPRN is very old and so there no longer is a valid building, (b) the UPRN is wrong, or (c) the UPRN is located just outside the building due to small mapping errors. These UPRNs will not receive any label for block of flats, and therefore will not receive a tech type from decision tree. I suggest we exclude these for now, and then see if it's a big percentage / count of the data when scaling up. We could then try to remediate issue (c).

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/pipeline/model/block_of_flats/feature_engineering.py
+++ b/asf_heat_pump_suitability/pipeline/model/block_of_flats/feature_engineering.py
@@ -21,6 +21,7 @@ def generate_df_features(
     Return:
         pl.DataFrame: all features for each building ID
     """
+    print("Generating features required for block of flats classifier...")
     buildings_w_uprns_gdf = buildings_gdf.sjoin(
         uprns_gdf, how="inner", predicate="contains"
     )
--- a/asf_heat_pump_suitability/pipeline/model/block_of_flats/feature_engineering.py
+++ b/asf_heat_pump_suitability/pipeline/model/block_of_flats/feature_engineering.py
@@ -23,11 +23,11 @@ def generate_df_features(
     """
     print("Generating features required for block of flats classifier...")
     buildings_w_uprns_gdf = buildings_gdf.sjoin(
-        uprns_gdf, how="inner", predicate="contains"
-    )
+        uprns_gdf, how="inner", predicate="intersects"
+    ).dropna(subset=id_col)
     uprns_w_buildings_gdf = uprns_gdf.sjoin(
-        buildings_gdf, how="inner", predicate="within"
-    )
+        buildings_gdf, how="inner", predicate="intersects"
+    ).dropna(subset="UPRN")
 
     features_dfs = [
         _generate_df_building_features(buildings_w_uprns_gdf, id_col),
--- a/asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
+++ b/asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
@@ -165,7 +165,7 @@ def predict_class_block_of_flats(
     concat_dfs = []
     labelled_ids = labelled_df[id_col].unique()
     X_df = (
-        features_df.filter(pl.col("n_flats") > 0, pl.col(id_col).is_in(labelled_ids))
+        features_df.filter(pl.col("n_flats") > 0, ~pl.col(id_col).is_in(labelled_ids))
         .to_pandas()
         .set_index(id_col)[FEATURES]
     )
--- a/asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
+++ b/asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
@@ -27,6 +27,7 @@ from sklearn.model_selection import (
 from sklearn.model_selection._search import BaseSearchCV
 from sklearn.metrics import f1_score
 import argparse
+from asf_heat_pump_suitability.pipeline.transform import uprns
 
 # Set random state int and RandomState instance
 RANDOM_STATE = 8
@@ -140,6 +141,85 @@ def train_block_of_flats_classifier(
     return final_model
 
 
+def predict_class_block_of_flats(
+    model: RandomForestClassifier,
+    features_df: pl.DataFrame,
+    labelled_df: pl.DataFrame,
+    id_col: str,
+) -> pl.DataFrame:
+    """
+    Args:
+        model (RandomForestClassifier): trained model for binary classification of buildings into blocks of flats or not
+        features_df (pl.DataFrame): buildings to predict classes on with engineered features for model
+        labelled_df (pl.DataFrame): labelled training data with features and target variable
+        id_col (str): name of ID column in both `features_df` and `labelled_df` (must be the same).
+    """
+    print(
+        "Predicting classes (block of flats / not) and class probability of buildings..."
+    )
+    concat_dfs = []
+    labelled_ids = labelled_df[id_col].unique()
+    X_df = (
+        features_df.filter(pl.col("n_flats") > 0, pl.col(id_col).is_in(labelled_ids))
+        .to_pandas()
+        .set_index(id_col)[FEATURES]
+    )
+    predictions_df = X_df.copy()
+    predictions_df["block_of_flats"] = model.predict(X_df)
+
+    # Add probability of predictions
+    predictions_df[f"block_of_flats_proba_{model.classes_[0]}"] = model.predict_proba(
+        X_df
+    )[:, 0]
+    predictions_df[f"block_of_flats_proba_{model.classes_[1]}"] = model.predict_proba(
+        X_df
+    )[:, 1]
+
+    # Combine into one probability label
+    concat_dfs.append(
+        pl.from_pandas(predictions_df, include_index=True).with_columns(
+            pl.when(pl.col("block_of_flats"))
+            .then(pl.col("block_of_flats_proba_True"))
+            .when(~pl.col("block_of_flats"))
+            .then(pl.col("block_of_flats_proba_False"))
+            .alias("block_of_flats_label_proba")
+        )
+    )
+
+    # Set manually labelled block probability to 1
+    concat_dfs.append(
+        labelled_df.filter(pl.col(id_col).is_in(labelled_ids)).with_columns(
+            pl.lit(1.0).alias("block_of_flats_label_proba")
+        )
+    )
+
+    # Add required columns to buildings which do not contain flats for concatenation
+    concat_dfs.append(
+        features_df.filter(pl.col("n_flats") == 0).with_columns(
+            pl.lit(False).alias("block_of_flats"),
+            pl.lit(None).alias("block_of_flats_label_proba"),
+        )
+    )
+
+    cols = [id_col, "block_of_flats", "block_of_flats_label_proba"]
+
+    return pl.concat([df.select(cols) for df in concat_dfs])
+
+
+def extend_df_in_block_of_flats_label(
+    uprns_df: pl.DataFrame, mapping: dict, predictions_df: pl.DataFrame, id_col: str
+) -> pl.DataFrame:
+    """ """
+    print("Adding `in_block_of_flats` label to UPRNs...")
+    return (
+        uprns_df.with_columns(
+            pl.col("UPRN").replace(mapping, default=None).alias(id_col)
+        )
+        .join(predictions_df, how="left", on=id_col)
+        .rename({"block_of_flats": "in_block_of_flats"})
+    )
+
+
 def parse_arguments() -> argparse.Namespace:
     """
     Create ArgumentParser and parse.
@@ -222,5 +302,5 @@ if __name__ == "__main__":
         param_search="default",
     )
 
-    save_as = config["output"]["save_as"]["block_of_flats_model"]
+    save_as = config["output"]["save_as"]["model"]["block_of_flats_model"]
     save_utils.save_model_to_pkl_s3(model, save_as)
--- a/asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
+++ b/asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
@@ -12,6 +12,8 @@ Set the required parameters:
 `labelled_data` takes a path to a parquet file containing labelled data. Requires a boolean 'block_of_flats' column and a building
 ID column with one row per building.
 `uprns` must contain all domestic UPRNs within the area(s) that `labelled_data` samples from.
+
+`labelled_data` takes a path to labelled data to train binary classification model on in parquet file format. Building ID required.
 """
 
 import numpy as np
@@ -78,7 +80,7 @@ def train_block_of_flats_classifier(
         df (pl.DataFrame): engineered features with labelled target variable
         id_col (str): name of ID column
         features (Iterable[str]): features used to train the model
-        target (str): name of target variable,
+        target (str): name of target variable
         param_search (Type[BaseSearchCV]): a class (not an instance) of `BaseSearchCV`, e.g. `HalvingRandomSearchCV` or `HalvingGridSearchCV` etc.
         Defaults to using `HalvingRandomSearchCV` which will create an instance of this class with selected custom arguments for `param_distributions`, `factor`, `cv`,
         and `n_jobs`, using F1 `scoring` metric. If using something other than the default option, kwargs for the selected `BaseSearchCV` class must be given, including param_distributions.
@@ -148,7 +150,18 @@ def predict_class_block_of_flats(
     id_col: str,
 ) -> pl.DataFrame:
     """
-    Predict binary class (block of flats / not) on buildings using trained Random Forest Classifier model.
+    Predict binary class (block of flats / not) on buildings using trained Random Forest Classifier model. Features required:
+        - n_UPRNs
+        - n_flats
+        - building_area_m2
+        - building_perimeter_m
+        - proportion_flats
+        - UPRNs_per_building_m2
+        - concave_hull_area_m2
+        - uprns_per_hull_area_m2
+        - flats_per_hull_area_m2
+        - avg_n_stacked_uprns
+        - std_n_stacked_uprns
 
     Args:
         model (RandomForestClassifier): trained model for binary classification of buildings into blocks of flats or not
@@ -180,7 +193,7 @@ def predict_class_block_of_flats(
         X_df
     )[:, 1]
 
-    # Combine into one probability label
+    # Combine into one probability label - final probability label indicates probability of the class assigned
     concat_dfs.append(
         pl.from_pandas(predictions_df, include_index=True).with_columns(
             pl.when(pl.col("block_of_flats"))
@@ -229,9 +242,14 @@ def extend_df_in_block_of_flats_label(
     print("Adding `in_block_of_flats` label to UPRNs...")
     return (
         uprns_df.with_columns(
-            pl.col("UPRN").replace(mapping, default=None).alias(id_col)
+            # Map building IDs to the UPRNs they contain
+            pl.col("UPRN")
+            .replace(mapping, default=None)
+            .alias(id_col)
         )
+        # Join the predicted block of flats label to the UPRNs via the building ID
         .join(predictions_df, how="left", on=id_col)
+        # Rename class for UPRN-level data
         .rename({"block_of_flats": "in_block_of_flats"})
     )
 
@@ -254,7 +272,7 @@ def parse_arguments() -> argparse.Namespace:
 
     parser.add_argument(
         "--labelled_data",
-        help="Labelled data to train binary classification model on in parquet file format. Building ID required.",
+        help="Path to labelled data to train binary classification model on in parquet file format. Building ID required.",
         type=str,
         required=True,
     )
--- a/asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
+++ b/asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
@@ -1,5 +1,5 @@
 """
-Functions to train a random forest binary classifier model.
+Functions to train and apply a random forest binary classifier model.
 
 Contains script to train a random forest classifier to identify buildings as blocks of flats or not, given features derived
 from building footprint and UPRN geospatial information.
@@ -148,11 +148,16 @@ def predict_class_block_of_flats(
     id_col: str,
 ) -> pl.DataFrame:
     """
+    Predict binary class (block of flats / not) on buildings using trained Random Forest Classifier model.
+
     Args:
         model (RandomForestClassifier): trained model for binary classification of buildings into blocks of flats or not
         features_df (pl.DataFrame): buildings to predict classes on with engineered features for model
         labelled_df (pl.DataFrame): labelled training data with features and target variable
-        id_col (str): name of ID column in both `features_df` and `labelled_df` (must be the same).
+        id_col (str): name of building ID column in both `features_df` and `labelled_df` (must be the same)
+
+    Returns:
+        pl.DataFrame: one row per building with predicted class and probability of predicted class
     """
     print(
         "Predicting classes (block of flats / not) and class probability of buildings..."
@@ -209,7 +214,18 @@ def predict_class_block_of_flats(
 def extend_df_in_block_of_flats_label(
     uprns_df: pl.DataFrame, mapping: dict, predictions_df: pl.DataFrame, id_col: str
 ) -> pl.DataFrame:
-    """ """
+    """
+    Join predicted building class (block of flats / not) to the corresponding UPRNs located within.
+
+    Args:
+        uprns_df (pl.DataFrame): dataset with UPRN column
+        mapping (dict): mapping of UPRNs (keys) to building IDs (values), where the building ID represents the building the UPRN is located within
+        predictions_df (pl.DataFrame): one row per building with predicted class and probability of predicted class
+        id_col (str): name of building ID column in `predictions_df`
+
+    Returns:
+        pl.DataFrame: one row per UPRN with `in_block_of_flats` label indicating the class of building the UPRN is located within
+    """
     print("Adding `in_block_of_flats` label to UPRNs...")
     return (
         uprns_df.with_columns(
--- a/asf_heat_pump_suitability/pipeline/run/add_features.py
+++ b/asf_heat_pump_suitability/pipeline/run/add_features.py
@@ -1,6 +1,7 @@
 """
 Script to add features to UPRNs:
 - flat / apartment property type boolean flag
+-
 - estimated max contiguous and total outdoor space (m2)
 
 Run:
@@ -35,9 +36,14 @@ if __name__ == "__main__":
     import polars as pl
     import geopandas as gpd
 
+    from asf_heat_pump_suitability import config
     from asf_heat_pump_suitability.utils import save_utils
-    from asf_heat_pump_suitability.getters import load_tree_input
+    from asf_heat_pump_suitability.getters import load_tree_input, base_getters
     from asf_heat_pump_suitability.pipeline.impute import property_type
+    from asf_heat_pump_suitability.pipeline.model.block_of_flats import (
+        feature_engineering,
+        train_model,
+    )
     from asf_heat_pump_suitability.pipeline.transform import uprns, outdoor_space
 
     args = parse_arguments()
@@ -60,14 +66,49 @@ if __name__ == "__main__":
     )
 
     # ------------------------ #
+    # PREDICT BLOCK OF FLATS CLASSIFICATION
+    # Load building footprint data
+    # TODO scale beyond sampling areas
+    building_footprints_gdf = load_tree_input.load_gdf_os_openmap_local_layer(
+        layer="building", grid_squares="SX"
+    )
+
+    uprn_building_id_dict = uprns.map_dict_uprns_to_building_id(
+        uprns_gdf=uprns_gdf, buildings_gdf=building_footprints_gdf, id_col="ID"
+    )
+
+    uprns_gdf["property_type_flat"] = uprns_gdf["UPRN"].isin(flat_uprns)
+    building_features_df = feature_engineering.generate_df_features(
+        buildings_gdf=building_footprints_gdf,
+        uprns_gdf=uprns_gdf,
+        id_col="ID",
+    )
+
+    # TODO make this robust by loading same labelled data used to train model
+    labelled_df = pl.read_parquet(
+        "s3://asf-heat-pump-suitability/local_heat_planning/inputs/processed/manually_labelled_block_of_flats.parquet"
+    )
+    rfc = base_getters.load_pickle(
+        config["output"]["save_as"]["model"]["block_of_flats_model"]
+    )
+    features_df = train_model.extend_df_in_block_of_flats_label(
+        uprns_df=features_df,
+        mapping=uprn_building_id_dict,
+        predictions_df=train_model.predict_class_block_of_flats(
+            model=rfc,
+            features_df=building_features_df,
+            labelled_df=labelled_df,
+            id_col="ID",
+        ),
+        id_col="ID",
+    )
+
+    # ------------------------ #
     # ESTIMATE OUTDOOR SPACE
     # TODO scale beyond Plymouth
     print("Loading land registry data...")
     land_parcels_gdf = gpd.read_file(
         "s3://asf-heat-pump-suitability/local_heat_planning/plymouth_inputs/Plymouth_Land_Registry_Cadastral_Parcels.gml"
-    )
-    building_footprints_gdf = load_tree_input.load_gdf_os_openmap_local_layer(
-        layer="building", grid_squares="SX"
     )
 
     # Get intersection of building footprint polygons and land polygons
--- a/asf_heat_pump_suitability/pipeline/run/add_features.py
+++ b/asf_heat_pump_suitability/pipeline/run/add_features.py
@@ -1,7 +1,7 @@
 """
 Script to add features to UPRNs:
 - flat / apartment property type boolean flag
--
+- boolean flag to indicate whether UPRN is in a block of flats
 - estimated max contiguous and total outdoor space (m2)
 
 Run:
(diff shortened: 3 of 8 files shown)


asf_heat_pump_suitability/utils/geo_utils.py
```diff
@@ -71,3 +72,23 @@ def parse_binary_geometry(
...
+    Verify GeoDataFrame uses the defined target CRS. If not, convert to target CRS.
+
+    Args:
+        gdf (gpd.GeoDataFrame): geodataset to check CRS of
+        target_crs (str|int): target CRS. Defaults to target CRS defined in config base.yaml
+
+    Returns:
+        gpd.GeoDataFrame: geodataset in target CRS
+    """
+    if gdf.crs != target_crs:
```
sofiapinto, line 90: do we need to convert from str to int?

asf_heat_pump_suitability/utils/geo_utils.py
```diff
@@ -71,3 +72,23 @@ def parse_binary_geometry(
         return wkb.loads(binary_data, hex=True)
     else:
         return None
+
+
+def verify_gdf_crs(
```
sofiapinto, line 77: this is very useful, we keep doing it manually!

asf_heat_pump_suitability/pipeline/model/block_of_flats/feature_engineering.py
```diff
@@ -21,12 +21,13 @@ def generate_df_features(
...
     buildings_w_uprns_gdf = buildings_gdf.sjoin(
-        uprns_gdf, how="inner", predicate="contains"
-    )
+        uprns_gdf, how="inner", predicate="intersects"
+    ).dropna(subset=id_col)
     uprns_w_buildings_gdf = uprns_gdf.sjoin(
-        buildings_gdf, how="inner", predicate="within"
-    )
+        buildings_gdf, how="inner", predicate="intersects"
+    ).dropna(subset="UPRN")
```
sofiapinto, line 33: I felt like this made sense when I first looked at it in the other PR, but now i'm confused. Does this result in the same output? or were the previous predicates wrong?

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
```diff
@@ -140,6 +141,101 @@ def train_block_of_flats_classifier(
...
+    id_col: str,
+) -> pl.DataFrame:
+    """
+    Predict binary class (block of flats / not) on buildings using trained Random Forest Classifier model.
+
+    Args:
+        model (RandomForestClassifier): trained model for binary classification of buildings into blocks of flats or not
+        features_df (pl.DataFrame): buildings to predict classes on with engineered features for model
+        labelled_df (pl.DataFrame): labelled training data with features and target variable
+        id_col (str): name of building ID column in both `features_df` and `labelled_df` (must be the same)
```
sofiapinto, line 183: do we want to add a check for this below and make it error if not the same?

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
```diff
@@ -140,6 +141,101 @@ def train_block_of_flats_classifier(
...
+    labelled_df: pl.DataFrame,
+    id_col: str,
+) -> pl.DataFrame:
+    """
+    Predict binary class (block of flats / not) on buildings using trained Random Forest Classifier model.
+
+    Args:
+        model (RandomForestClassifier): trained model for binary classification of buildings into blocks of flats or not
+        features_df (pl.DataFrame): buildings to predict classes on with engineered features for model
+        labelled_df (pl.DataFrame): labelled training data with features and target variable
```
sofiapinto, line 182: I think this might need to be optional. We will make predictions for areas where we have no labelled data.

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
```diff
@@ -140,6 +141,101 @@ def train_block_of_flats_classifier(
...
+        id_col (str): name of building ID column in both `features_df` and `labelled_df` (must be the same)
+
+    Returns:
+        pl.DataFrame: one row per building with predicted class and probability of predicted class
+    """
+    print(
+        "Predicting classes (block of flats / not) and class probability of buildings..."
+    )
+    concat_dfs = []
+    labelled_ids = labelled_df[id_col].unique()
```
sofiapinto, line 192:
```suggestion
    # IDs of labelled buildings
    labelled_ids = labelled_df[id_col].unique()
```

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
```diff
@@ -140,6 +141,101 @@ def train_block_of_flats_classifier(
...
+    X_df = (
```
sofiapinto, line 193:
```suggestion
    # Features df for buildings that have not been manually labelled with at least 1 UPRN which is a flat
    X_df = (
```

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
```diff
@@ -140,6 +141,101 @@ def train_block_of_flats_classifier(
...
+        features_df.filter(pl.col("n_flats") > 0, ~pl.col(id_col).is_in(labelled_ids))
+        .to_pandas()
+        .set_index(id_col)[FEATURES]
+    )
+    predictions_df = X_df.copy()
+    predictions_df["block_of_flats"] = model.predict(X_df)
```
sofiapinto, line 199:
```suggestion
    predictions_df["is_block_of_flats"] = model.predict(X_df)
```

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
```diff
@@ -140,6 +141,101 @@ def train_block_of_flats_classifier(
...
+    cols = [id_col, "block_of_flats", "block_of_flats_label_proba"]
+
+    return pl.concat([df.select(cols) for df in concat_dfs])
+
+
+def extend_df_in_block_of_flats_label(
+    uprns_df: pl.DataFrame, mapping: dict, predictions_df: pl.DataFrame, id_col: str
+) -> pl.DataFrame:
+    """
+    Join predicted building class (block of flats / not) to the corresponding UPRNs located within.
```
sofiapinto, line 244: I wouldn't say this is what's happening based on the name of the function.

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
```diff
@@ -140,6 +141,101 @@ def train_block_of_flats_classifier(
...
+        id_col (str): name of building ID column in `predictions_df`
+
+    Returns:
+        pl.DataFrame: one row per UPRN with `in_block_of_flats` label indicating the class of building the UPRN is located within
+    """
+    print("Adding `in_block_of_flats` label to UPRNs...")
+    return (
+        uprns_df.with_columns(
+            pl.col("UPRN").replace(mapping, default=None).alias(id_col)
+        )
```
sofiapinto, line 262: Let me see if i understand. we're taking the UPRN column, doing the mapping from UPRNs to building IDS so that we have the mapped building ID. and then calling new col the bulding ID column name. But we're keeping the UPRN call right?

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -59,16 +65,51 @@ def parse_arguments() -> argparse.Namespace:
         pl.col("UPRN").is_in(flat_uprns).alias("property_type_flat")
     )
 
+    # ------------------------ #
+    # PREDICT BLOCK OF FLATS CLASSIFICATION
+    # Load building footprint data
+    # TODO scale beyond sampling areas
+    building_footprints_gdf = load_tree_input.load_gdf_os_openmap_local_layer(
+        layer="building", grid_squares="SX"
+    )
```
sofiapinto, line 74: This will need changing after I merge by PRs :) we'll be able to load the correct grid_squares from base.yaml

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -1,6 +1,7 @@
 """
 Script to add features to UPRNs:
 - flat / apartment property type boolean flag
+-
```
sofiapinto: did you forget to add something here?

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -59,16 +65,51 @@ def parse_arguments() -> argparse.Namespace:
...
 
+    # ------------------------ #
+    # PREDICT BLOCK OF FLATS CLASSIFICATION
+    # Load building footprint data
+    # TODO scale beyond sampling areas
+    building_footprints_gdf = load_tree_input.load_gdf_os_openmap_local_layer(
+        layer="building", grid_squares="SX"
+    )
+
+    uprn_building_id_dict = uprns.map_dict_uprns_to_building_id(
```
sofiapinto, line 77:
```suggestion
    # Create a mapping between UPRNs and respective building they're in
    uprn_building_id_dict = uprns.map_dict_uprns_to_building_id(
```

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -59,16 +65,51 @@ def parse_arguments() -> argparse.Namespace:
...
+    building_features_df = feature_engineering.generate_df_features(
+        buildings_gdf=building_footprints_gdf,
+        uprns_gdf=uprns_gdf,
+        id_col="ID",
+    )
+
+    # TODO make this robust by loading same labelled data used to train model
+    labelled_df = pl.read_parquet(
+        "s3://asf-heat-pump-suitability/local_heat_planning/inputs/processed/manually_labelled_block_of_flats.parquet"
+    )
```
sofiapinto, line 92: This path can be in args, without it being required

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -59,16 +65,51 @@ def parse_arguments() -> argparse.Namespace:
...
+    uprn_building_id_dict = uprns.map_dict_uprns_to_building_id(
+        uprns_gdf=uprns_gdf, buildings_gdf=building_footprints_gdf, id_col="ID"
+    )
+
+    uprns_gdf["property_type_flat"] = uprns_gdf["UPRN"].isin(flat_uprns)
+    building_features_df = feature_engineering.generate_df_features(
+        buildings_gdf=building_footprints_gdf,
+        uprns_gdf=uprns_gdf,
+        id_col="ID",
+    )
```
sofiapinto, line 87:
```suggestion
    # Generates features at building level for the blocks of flats estimation
    uprns_gdf["property_type_flat"] = uprns_gdf["UPRN"].isin(flat_uprns)
    building_features_df = feature_engineering.generate_df_features(
        buildings_gdf=building_footprints_gdf,
        uprns_gdf=uprns_gdf,
        id_col="ID",
    )
```

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -59,16 +65,51 @@ def parse_arguments() -> argparse.Namespace:
...
+
+    # TODO make this robust by loading same labelled data used to train model
+    labelled_df = pl.read_parquet(
+        "s3://asf-heat-pump-suitability/local_heat_planning/inputs/processed/manually_labelled_block_of_flats.parquet"
+    )
+    rfc = base_getters.load_pickle(
```
sofiapinto: We could class it `model` or `classifier` so that, in case we change the type of classifier, we don't have to change this.

Also improves readability. Someone going through this code might not know that `rfc` stands for random forest classifier.
```suggestion
    # Load model to identify buildings that are blocks of flats
    model = base_getters.load_pickle(
```

asf_heat_pump_suitability/pipeline/run/add_features.py
sofiapinto, line 1: A comment that's not about this PR specifically about about this `pipeline/run/` folder. I find it very confusing that we have this run folder, but we actually have other pipeline folders where we run pipelines.

sofiapinto: In part this will be improved by having a README. but i wonder if it would be better to get rid of this run folder and then have pipeline organised into the different parts of the pipelines. As long as we have an ordered list of pipelines to run, we're good.

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -140,6 +141,35 @@ def filter_gdf_residential_uprns(
...
+    predicate: str = "intersects",
+) -> dict:
+    """
+    Create a mapping of UPRNs (keys) to the building ID (values) of the building they are located within or intersect with.
+
+    Args:
+        uprns_gdf (gpd.GeoDataFrame): UPRNs with geospatial point data
+        buildings_gdf (gpd.GeoDataFrame): building footprints
+        id_col (str): name of building ID column in `buildings_gdf`
+        predicate (str): how to join buildings and UPRNs, of `intersects` which joins UPRNs with building footprints
```
sofiapinto: the second part of this sentence doesn't make sense to me.

sofiapinto (commented):
Hi @crispy-wonton , I had a first pass at reviewing this! Thank you for putting it together so quickly.


Have left comments throughout and answers to your questions/feedback requests.

> In order to test the code in this PR you need to ...
> Run the following piece of code from the terminal to run the updated add_features.py script:
> python asf_heat_pump_suitability/pipeline/run/add_features.py --uprns s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_residential_uprns.parquet

Was able to run this with my conda env from the Manchester PR. Here are the logs from the command line:

```
Loading domestic UPRNs from: s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_residential_uprns.parquet
56968 flats found in UPRN dataset, N=148418, 38.38% of all UPRNs

Loading OS OpenMap Local - Building file: s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/v202510_OSOpenMapLocal_geometries_selected/SX/SX_Building.shp
Generating features required for block of flats classifier...
2026-02-16 17:40:55,303 - botocore.credentials - INFO - Found credentials in shared credentials file: ~/.aws/credentials
Predicting classes (block of flats / not) and class probability of buildings...
Adding `in_block_of_flats` label to UPRNs...
Loading land registry data...
/Users/anasofiapinto/Documents/repos/asf_heat_pump_suitability/asf_heat_pump_suitability/pipeline/transform/outdoor_space.py:128: FutureWarning: The provided callable <built-in function max> is currently using SeriesGroupBy.max. In a future version of pandas, the provided callable will be used directly. To keep current behavior pass the string "max" instead.
  .agg(
/Users/anasofiapinto/Documents/repos/asf_heat_pump_suitability/asf_heat_pump_suitability/pipeline/transform/outdoor_space.py:128: FutureWarning: The provided callable <built-in function sum> is currently using SeriesGroupBy.sum. In a future version of pandas, the provided callable will be used directly. To keep current behavior pass the string "sum" instead.
  .agg(
2026-02-16 17:43:12,505 - root - INFO - Saving file to s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_residential_uprns_with_features.parquet
/Users/anasofiapinto/Documents/repos/asf_heat_pump_suitability/asf_heat_pump_suitability/utils/save_utils.py:41: UserWarning: Polars found a filename. Ensure you pass a path to the file instead of a python file object when possible for best performance.
  df.write_parquet(f
```

> Please pay special attention to ...
> 
> - The output data of the pipeline in relation to the two new features added: in_block_of_flats and block_of_flats_label_proba. I had a look at it myself and compared it to the outputs of my original notebook in https://github.com/nestauk/asf_heat_pump_suitability/pull/207 . The results look the same for UPRNs which are common across both datasets (the original notebook covers the full 6 sampling areas, whereas the script covers Plymouth only). Do you see anything weird or unexpected about the results?

Can you point me to the files I should be looking at?

> - Do the updated predicates on the sjoins make sense? Is there a reason we shouldn't use intersects? I thought that it could create cases where a UPRN gets joined to multiple buildings, but in testing it on Plymouth, I saw that didn't seem to happen.

I was wondering why you had changed it. I always have a little trouble with the predicates. But here's how I interpret it:
- buildings_w_uprns_gdf: when using contains, the UPRN point bust be strictly inside the polygon, not on the edge. With intersects, the point can be inside or touching the boundary. You've taken a look at the location of UPRNs in comparison to the building boundaries - where there cases where the UPRN will sit on top of the boundary? Either way, I think i'd stick with contains.
- uprns_w_buildings_gdf: when using within, we're asking if the point is strictly inside the polygon. The intersects matches if the UPRN point is inside or sitting exactly on the edge - if a UPRN point sits on the shared boundary of two buildings, intersects will return two rows for that single UPRN.

I think intersects can indeed result in duplication. On the other hand, using contains/within might result in UPRNs having no building that they "belong to".

> Fyi there are a small number of UPRNs (1.5% of Plymouth data) which don't receive a label. I investigated this to see if it was caused by a bug. It's caused by our domestic filtering methodology - not a bug but seems to be a data quality issue. Basically, the domestic filtering retains all UPRNs which are in the domestic EPC register. A small number of these seem to not get sjoined to a building - this could be because (a) the UPRN is very old and so there no longer is a valid building, (b) the UPRN is wrong, or (c) the UPRN is located just outside the building due to small mapping errors. These UPRNs will not receive any label for block of flats, and therefore will not receive a tech type from decision tree. I suggest we exclude these for now, and then see if it's a big percentage / count of the data when scaling up. We could then try to remediate issue (c).

I had seen this happening too! I think we need to investigate this one separately to understand whether it's a mix of a), b) and c) or just one of those. Are you happy to add this to our backlog?

asf_heat_pump_suitability/pipeline/model/block_of_flats/feature_engineering.py
sofiapinto: Have since seen your question about this, and left my opinion in the main PR comment.

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
```diff
@@ -140,6 +141,101 @@ def train_block_of_flats_classifier(
...
+    features_df: pl.DataFrame,
+    labelled_df: pl.DataFrame,
+    id_col: str,
+) -> pl.DataFrame:
+    """
+    Predict binary class (block of flats / not) on buildings using trained Random Forest Classifier model.
+
+    Args:
+        model (RandomForestClassifier): trained model for binary classification of buildings into blocks of flats or not
+        features_df (pl.DataFrame): buildings to predict classes on with engineered features for model
```
simran-dave, line 181: it might be useful here to say what the features for the model should be

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
```diff
@@ -140,6 +141,101 @@ def train_block_of_flats_classifier(
...
+
+    # Add probability of predictions
+    predictions_df[f"block_of_flats_proba_{model.classes_[0]}"] = model.predict_proba(
+        X_df
+    )[:, 0]
+    predictions_df[f"block_of_flats_proba_{model.classes_[1]}"] = model.predict_proba(
+        X_df
+    )[:, 1]
+
+    # Combine into one probability label
```
simran-dave: after a while I understood that (I think) the probability label is the probability of the building being in the category it has been assigned by the model but that could be made clearer here

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
```diff
@@ -140,6 +141,101 @@ def train_block_of_flats_classifier(
...
+    """
+    print("Adding `in_block_of_flats` label to UPRNs...")
+    return (
+        uprns_df.with_columns(
+            pl.col("UPRN").replace(mapping, default=None).alias(id_col)
+        )
+        .join(predictions_df, how="left", on=id_col)
+        .rename({"block_of_flats": "in_block_of_flats"})
+    )
+
```
simran-dave, line 268: I don't fully follow what happened in this function

asf_heat_pump_suitability/utils/geo_utils.py
```diff
@@ -71,3 +72,23 @@ def parse_binary_geometry(
...
+    gdf: gpd.GeoDataFrame, target_crs: str | int = config["constant"]["target_crs"]
+) -> gpd.GeoDataFrame:
+    """
+    Verify GeoDataFrame uses the defined target CRS. If not, convert to target CRS.
```
simran-dave, line 81: maybe just me but I had to look up what CRS is

simran-dave (commented):
Hi @crispy-wonton ! :) 
I ran python `asf_heat_pump_suitability/pipeline/run/add_features.py --uprns s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_residential_uprns.parquet` which worked fine for me. 

On this:

> Fyi there are a small number of UPRNs (1.5% of Plymouth data) which don't receive a label. I investigated this to see if it was caused by a bug.

I am happy to look into this as part of the domestic filtering work. It sounds reasonable to me that some UPRNs might be just outside a building and that some might be old. Do we know how accurate the building footprint/ land parcel data is? If it's say ~1 metre off but the UPRN is more precise you could end up with this mismatch?

I've left a couple of comments where the code wasn't super clear to me.

asf_heat_pump_suitability/utils/geo_utils.py
crispy-wonton: No - either int or str will work accurately in this evaluation

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
crispy-wonton: No, it will error out if it's not the same with a `ColumnNotFoundError`.

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
crispy-wonton: That's fair. I'm inclined to change it to optional when that becomes the case and leave it for now, because we might just run the model on the full GB when scaling.

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
crispy-wonton: This is clear from the code so I won't add for succinctness.

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
crispy-wonton: I'm confused - can you elaborate what you mean?

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
crispy-wonton: Yes, it's adding the ID of the building each UPRN is in , essentially.

asf_heat_pump_suitability/pipeline/run/add_features.py
crispy-wonton: Yes but since only one file exists at the moment, I've hard-coded it. Can be updated if we end up with more labelled datasets.

asf_heat_pump_suitability/pipeline/run/add_features.py
crispy-wonton: Good point - I used the `sklearn` standard reference name for classifier models.

asf_heat_pump_suitability/pipeline/run/add_features.py
crispy-wonton: The purpose of the `/run` subdir is to contain final scripts to run the full pipeline. The other subdirs within the pipeline folder contain the modules required to run the pipeline beyond getters or can be run as independent scripts as pipeline components. This is covered in the style guide.

I suppose it's slightly confusing now because we need to run the pipeline in steps from the other subdirs, but eventually, we should just be able to run scripts in the `/run` subdir and only run individual scripts elsewhere if we want to run a particular part for some reason. 

This could be renamed if too confusing.

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: me neither lol

crispy-wonton: oh I understand what I tried to say. Have clarified.

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
crispy-wonton: Good point

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
crispy-wonton: lmk if clearer or if you think the column header itself should change

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
crispy-wonton: added annotation - lmk if you have further qs.

asf_heat_pump_suitability/utils/geo_utils.py
crispy-wonton: CRS is standard language for geospatial work and we use the term throughout the repository so I won't define it here. Let me know if you have questions about it :)

crispy-wonton:
@sofiapinto responding to your questions:
> Can you point me to the files I should be looking at?

Sorry for not specifying, it would be this file: `s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_residential_uprns_with_features.parquet`. Just to check if anything looks weird to you.

> I think intersects can indeed result in duplication. On the other hand, using contains/within might result in UPRNs having no building that they "belong to".

Your interpretation is correct, `within` / `contains` is stricter and the UPRNs must be fully inside the buildings to be matched to them, `intersects` allows UPRNs to be on the edge of the building footprint. As mentioned, it doesn't match any UPRNs to multiple buildings in the case of Plymouth when using `intersects`. I think I'd lean to `intersects`.
 
RE: the small number of UPRNs (1.5% of Plymouth data) which don't receive a label.

> I think we need to investigate this one separately to understand whether it's a mix of a), b) and c) or just one of those. Are you happy to add this to our backlog?

@simran-dave suggested to look into this when looking at domestic UPRNs, I think that's a great idea. Have added to sprint tasks. To answer your question @simran-dave :

>Do we know how accurate the building footprint/ land parcel data is? If it's say ~1 metre off but the UPRN is more precise you could end up with this mismatch?

There may be some mapping errors in the building footprints and UPRNs (I imagine there must be some) - I couldn't find any information about the accuracy of the OS datasets in the data documentation but you might have more luck. Also, UPRNs don't need to be located within a centre of a building footprint, they could be along the edges of one. So we need to be careful of how we are conducting our spatial joins (e.g. more strict with `within` / `contains` or less strict with `intersects`).

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
crispy-wonton: This is not a trivial change as it requires also changing the labelled data processing which isnt in this PR. Therefore I'll leave as is as I think it's descriptive enough and the final variable for UPRN-level data is renamed anyway.

simran-dave (approved):

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
sofiapinto: i think that's fair!

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
sofiapinto: that makes sense!

asf_heat_pump_suitability/pipeline/model/block_of_flats/train_model.py
sofiapinto: I just meant that from `extend_df_in_block_of_flats_label()` I don't necessarily understand what the function does. I do understand the docstring though.

asf_heat_pump_suitability/pipeline/run/add_features.py
sofiapinto: That all makes sense, and I think having the clear README's will help. thank you for clarifying.

sofiapinto:
Hello @crispy-wonton,

Have reviewed the changes :) thank you for actioning them! It looks good. Happy to review once again after you resolve the conflicts with dev.

> Sorry for not specifying, it would be this file: `s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_residential_uprns_with_features.parquet`. Just to check if anything looks weird to you.

I got this summary table & added the "Meaning" column with my interpretation. Do you agree?

148,418 UPRNs and 32,944 buildings. 

<table border="1" class="dataframe">
  <thead>
    <tr style="text-align: right;">
      <th></th>
      <th></th>
      <th>min</th>
      <th>max</th>
      <th>median</th>
      <th>size</th>
      <th>Meaning</th>
    </tr>
    <tr>
      <th>property_type_flat</th>
      <th>in_block_of_flats</th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
      <th></th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th rowspan="3" valign="top">False</th>
      <th>False</th>
      <td>0.500324</td>
      <td>1.0</td>
      <td>0.906338</td>
      <td>88044</td>
      <td>There's at least one flat in the same building, but property not a flat and building not a block of flats.</td>
    </tr>
    <tr>
      <th>NaN</th>
      <td>NaN</td>
      <td>NaN</td>
      <td>NaN</td>
      <td>1352</td>
      <td>No flats in this building, so this data doesn't go through block of flats model.</td>
    </tr>
    <tr>
      <th>True</th>
      <td>0.500127</td>
      <td>1.0</td>
      <td>0.629260</td>
      <td>2054</td>
      <td>Our model outputs building as being a block of flats, although property not a flat (perhaps because it doesn't stack when it should?).</td>
    </tr>
    <tr>
      <th rowspan="3" valign="top">True</th>
      <th>False</th>
      <td>0.500324</td>
      <td>1.0</td>
      <td>0.697241</td>
      <td>31110</td>
      <td>Property is a flat, but not in a block of flats according to our model. </td>
    </tr>
    <tr>
      <th>NaN</th>
      <td>NaN</td>
      <td>NaN</td>
      <td>NaN</td>
      <td>896</td>
      <td>Is this the case where the property is not within any building footprint?</td>
    </tr>
    <tr>
      <th>True</th>
      <td>0.500127</td>
      <td>1.0</td>
      <td>0.724472</td>
      <td>24962</td>
      <td>Flat in a block of flats.</td>
    </tr>
  </tbody>
</table>
</div>




> Your interpretation is correct, `within` / `contains` is stricter and the UPRNs must be fully inside the buildings to be matched to them, `intersects` allows UPRNs to be on the edge of the building footprint. As mentioned, it doesn't match any UPRNs to multiple buildings in the case of Plymouth when using `intersects`. I think I'd lean to `intersects`.

I think this will become less relevant now that Simran is checking for UPRNs close to buildings. But just to check, for now we will stick with intersects for both? We need to add a check for whether a UPRN ends up belonging to two buildings. I'm guessing if that happens, we can assign it to the closest building?

> @simran-dave suggested to look into this when looking at domestic UPRNs, I think that's a great idea. Have added to sprint tasks. 

That's great! Thank you.

crispy-wonton:
I've added some text in the meaning column in bold:

| property_type_flat | in_block_of_flats | min      | max | median   | size  | Meaning |
|--------------------|------------------|----------|-----|----------|-------|---------|
| False              | False            | 0.500324 | 1.0 | 0.906338 | 88044 | **Property is not a flat and the building it's in is not a block of flats. There isn't necessarily at least one flat in the same building, there could be none.** |
| NaN                | NaN              | NaN      | NaN | NaN      | 1352  | **UPRNs not joined to a building footprint (these are from the EPC domestic register).** |
| False              | True             | 0.500127 | 1.0 | 0.629260 | 2054  | Our model outputs building as being a block of flats, although property not a flat. **UPRNs may not always be stacked when they are a flat but since this is our current methodology for detecting flats, the label for `property_type_flat` is `False`.** |
| True               | False            | 0.500324 | 1.0 | 0.697241 | 31110 | Property is a flat, but not in a block of flats according to our model. |
| NaN                | NaN              | NaN      | NaN | NaN      | 896   | **UPRNs not joined to a building footprint (these are from the EPC domestic register).** |
| True               | True             | 0.500127 | 1.0 | 0.724472 | 24962 | Flat in a block of flats. |

All UPRNs which have been joined to a building footprint will receive a `True` or `False` label for the `in_block_of_flats` variable. Only those which are not joined to buildings receive a `null` at the moment. 

> I think this will become less relevant now that Simran is checking for UPRNs close to buildings. But just to check, for now we will stick with intersects for both? We need to add a check for whether a UPRN ends up belonging to two buildings. I'm guessing if that happens, we can assign it to the closest building?

Yes, let's stick with intersects and if it starts causing major issues, go back to a stricter filter. If a UPRN ends up belonging to multiple buildings, then that means the point is equidistant to all of them, so we will have to choose a replicable rule to assign them to one, but there is no closest building.

sofiapinto (approved):
happy for you to merge ✨

PR #247, opened 2026-02-24
246 opportunity area counts

### LINKED ISSUE #246: Calculate value counts of features for opportunity areas
Use EPC data to get value counts of TENURE, ATTACHMENT TYPE, and EPC RATING for 3 example opportunity areas in Plymouth.

crispy-wonton:
Fixes #246 

# Description

A short one-off script to use EPC data to get value counts of TENURE, ATTACHMENT TYPE, and EPC RATING for 3 example opportunity areas in Plymouth. This is work for the playbook.

New files:
- /research/analysis/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py - script to do above

# Instructions for Reviewer

run the following code:
`python asf_heat_pump_suitability/research/exploratory/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py`

- the spatial joins and preprocessing of the EPC to ensure the final numbers are accurate
- if you have better suggestions for how to save the data without making it more difficult to extract manually (as the output will be used by design team in manual playbook development)

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/research/analysis/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py
```diff
@@ -0,0 +1,103 @@
...
+            .otherwise(pl.col("BUILT_FORM"))
+            .alias("attachment")
+        )
+    )
+
+    print("Joining geospatial coordinates to EPC data...")
+    # Add geospatial coordinates to EPC UPRNs
+    epc_gdf = uprns_gdf.merge(
+        epc_df.to_pandas(), how="inner", left_on="UPRN", right_on="processed_UPRN"
+    )
```
sofiapinto: Let me know if I misinterpreting your code, but I think it's important to do a left join here, so that any UPRNs that are not in EPC don't get deleted. We can (and should) show how many unknown tenures, attachment, etc there are.

```suggestion
    epc_gdf = uprns_gdf.merge(
        epc_df.to_pandas(), how="left", left_on="UPRN", right_on="processed_UPRN"
    )
```

asf_heat_pump_suitability/research/analysis/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py
```diff
@@ -0,0 +1,103 @@
...
+            .then(pl.lit("End-Terrace"))
+            .when(pl.col("BUILT_FORM") == "")
+            .then(pl.lit("unknown"))
+            .otherwise(pl.col("BUILT_FORM"))
+            .alias("attachment")
+        )
+    )
+
+    print("Joining geospatial coordinates to EPC data...")
+    # Add geospatial coordinates to EPC UPRNs
```
sofiapinto: Am I mistaken?

```suggestion
    # Adding EPC information to UPRNs
```

asf_heat_pump_suitability/research/analysis/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py
```diff
@@ -0,0 +1,103 @@
...
+    epc_gdf = uprns_gdf.merge(
+        epc_df.to_pandas(), how="inner", left_on="UPRN", right_on="processed_UPRN"
+    )
+
+    # Filter to UPRNs which are in opportunity areas
+    areas_epc_gdf = epc_gdf.sjoin(
+        areas_gdf[["Name", "geometry"]], how="inner", predicate="within"
```
sofiapinto: To be safe
```suggestion
        areas_gdf[["Name", "geometry"]], how="right", predicate="within"
```

asf_heat_pump_suitability/research/analysis/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py
```diff
@@ -0,0 +1,103 @@
...
+    ).drop(columns="index_right")
+
+    fname = "plymouth_example_opportunity_areas_feature_counts.txt"
+    print(f"Saving outputs to {fname}")
+    # Save the value counts of features of interest to txt file
+    with open(fname, "w") as file:
+        for tech in areas_epc_gdf["Name"].unique():
```
sofiapinto: We should probably use the area ID right? In case multiple areas in the file have the same tech. What we want is the distribution for each opportunity area, not for each tech - correct?

asf_heat_pump_suitability/research/analysis/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py
```diff
@@ -0,0 +1,103 @@
...
+            file.write(f"\n\n{tech} tenure:")
+            file.write(_gdf["TENURE"].value_counts().to_string())
+
+            file.write(f"\n\n{tech} attachment:")
+            file.write(_gdf["attachment"].value_counts().to_string())
+
+            file.write(f"\n\n{tech} EPC rating:")
+            file.write(
+                _gdf["CURRENT_ENERGY_RATING"].value_counts().sort_index().to_string()
+            )
```
sofiapinto: You could save as a csv with columns opportunity_area_ID, tech, number_properties, attach_flat, attach_mid_terrace, ..., attach_unknown, EPC_A, EPC_B, ..., EPC_unknown, etc

sofiapinto (changes requested):
Hey @crispy-wonton,

I ran `python asf_heat_pump_suitability/research/analysis/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py` with no issues. I haven't open the file though to see the contents though!

I left a few comments about the joins & how to save the data, let me know what you think.

More generally, I don't think this needs to be a separate research script, but happy for you to keep it this way if you think it's helpful. My reasoning is: We will eventually create a very similar script that will gather contextual information for each opportunity area/cluster. I think this could be the first version of that script. You can use argsparse to:
- receive the path where the kml with the clusters is stored (flexible so that it works whether it's the final set of clusters or 3 opportunity areas)
- a path where to store the stats (so we can use it for these bespoke asks and to save the final dataset of contextual information)

Let me know what you think!

crispy-wonton:
> More generally, I don't think this needs to be a separate research script, but happy for you to keep it this way if you think it's helpful. 

This was only meant to be a quick script written sub-optimally for the purpose of getting it out quickly. I'd prefer to leave this as it's own script, and then refactor the functions separately in another PR to create a proper pipeline script.

But yes, the changes you've suggested are ultimately the kind of changes we need to make in refactoring.

asf_heat_pump_suitability/research/analysis/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py
crispy-wonton: This was deliberate, as the original request was to recreate some existing EPC count data but it was on my mind to update it to a left merge and retain unknowns. Will make the changes.

asf_heat_pump_suitability/research/analysis/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py
crispy-wonton: Because of the original inner join, this was correct, but now it will be joining the EPC to UPRNs.

asf_heat_pump_suitability/research/analysis/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py
crispy-wonton: Is the purpose of this to highlight areas where there is no data (if that happens)?

asf_heat_pump_suitability/research/analysis/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py
crispy-wonton: This is true. Again, this was a choice I made based on the fact that it worked for the 3 opportunity areas and the understanding that anything in this script would be refactored and scaled separately.

asf_heat_pump_suitability/research/analysis/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py
crispy-wonton: Good idea

asf_heat_pump_suitability/research/analysis/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py
crispy-wonton: Also, upon a second look, there is no area ID field in this dataset because these were specifically created in a separate MyMaps file.

asf_heat_pump_suitability/research/analysis/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py
sofiapinto: yes - if that's the case, there's probably a mistake somewhere

asf_heat_pump_suitability/research/analysis/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py
sofiapinto: Yes, I can imagine. Do all of these opportunity areas have a different tech? If not, it's problematic. otherwise, it's okay for now.

But you could also create an ID within the script, so you can do the counts per cluster rather than tech.

sofiapinto:
> > More generally, I don't think this needs to be a separate research script, but happy for you to keep it this way if you think it's helpful.
> 
> This was only meant to be a quick script written sub-optimally for the purpose of getting it out quickly. I'd prefer to leave this as it's own script, and then refactor the functions separately in another PR to create a proper pipeline script.
> 
> But yes, the changes you've suggested are ultimately the kind of changes we need to make in refactoring.

I could imagine that was the case, very happy for that to be the case of course.

asf_heat_pump_suitability/research/analysis/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py
sofiapinto: Looking at the data, the ID exists. it's `opportunity_area_code`

sofiapinto (approved):
Have looked at the changes to the code and checked the data & counts. All looks good! ✨

asf_heat_pump_suitability/research/analysis/example_opportunity_areas_counts/plymouth_example_opportunity_area_counts.py
crispy-wonton: Which file are you looking at? In this script I'm using this for the opportunity areas: `"s3://asf-heat-pump-suitability/dump/example_opportunity_areas_plymouth.kml"` and I don't see an `opportunity_area_code` col - what am I missing? `Name` appears to be the most unique ID equivalent.

Either way, in this case it works here but when scaling we won't use that column. This data has a different format to what we would pass into this script as it was created manually on MyMaps.

PR #255, opened 2026-02-25
fix(#234): reduce Polars↔pandas↔GeoDataFrame round-trips in add_features

### LINKED ISSUE #223: Use load_settings() in CLI scripts instead of raw config dict
## Problem
`pipeline/uprns.py` and `pipeline/add_features.py` import config via the legacy raw dict (`from asf_heat_pump_suitability import config`). The new Pydantic-settings model (`config/settings.py`) was written precisely to replace this, but the CLI entry scripts were never updated to use it.

This means:
- No type-safety or validation on config values
- Env-var overrides defined in `Settings` (e.g. `DATA_MODE`, `AWS_ENDPOINT_URL`) are silently ignored by the CLI scripts
- Config values are accessed with fragile string-key dict lookups (`config["data"]["geodata"][...]`) instead of typed attributes

## Acceptance criteria
- Both `pipeline/uprns.py` and `pipeline/add_features.py` call `load_settings()` at startup and pass the resulting `Settings` object to all getter and transform calls that need config
- Raw `from asf_heat_pump_suitability import config` import removed from both scripts
- Any getter/transform function that currently accepts a raw dict should be updated to accept a `Settings` instance (or the relevant typed sub-object)

### LINKED ISSUE #234: Reduce Polars ↔ Pandas ↔ GeoDataFrame round-trips in add_features pipeline
## Problem
`pipeline/add_features.py` and the transform modules it calls perform several unnecessary format conversions:
1. Read parquet → Polars DataFrame
2. Convert to pandas → then to GeoDataFrame (for spatial ops)
3. After spatial join → convert back to Polars
4. Several intermediate `pl.from_pandas()` / `.to_pandas()` calls within the same function

Each conversion has memory and CPU overhead, and the back-and-forth makes the data flow hard to follow.

## Acceptance criteria
- The pipeline stays in GeoDataFrame format for the spatial processing section and converts to Polars once at the end before writing
- Intermediate `to_pandas()` / `from_pandas()` calls within a single function are eliminated where possible
- The overall logic (inputs → outputs) is unchanged; only the intermediate representation is simplified
- A brief comment at the top of `add_features.py` documents the intended data types at each stage

caldwellst:
## Summary

Eliminates unnecessary format conversions in the `add_features` pipeline, as specified in #234.

**`outdoor_space.py`**
- `sjoin_df_uprn_to_outdoor_space` now returns `gpd.GeoDataFrame` instead of `pl.DataFrame` (removes the intermediate `pl.from_pandas()` wrapper).
- `deduplicate_df_outdoor_space` accepts `pd.DataFrame` (a GeoDataFrame subclass) and is rewritten with equivalent pandas operations, removing its `polars` dependency.
- `import polars as pl` removed (no longer used).

**`add_features.py`**
- A docstring note in `run()` documents the data type at each pipeline stage (Input → Spatial → Output).
- `property_type_flat` is added directly to `uprns_gdf` as a pandas column instead of forking a parallel Polars DataFrame.
- The entire spatial section (flat detection + outdoor space sjoin + dedup) stays in `gpd.GeoDataFrame`.
- A single `pl.from_pandas(uprns_gdf[keep_cols])` at the end converts to Polars for writing.
- Stale `from asf_heat_pump_suitability import config` import removed (also fixes #223 on this base).

**`tests/unit/test_transform_outdoor_space.py`**
- `test_deduplicate_df_outdoor_space_keeps_smallest_total`: uses `pd.DataFrame` input and pandas-style assertions.
- `test_sjoin_df_uprn_to_outdoor_space`: asserts `isinstance(result, gpd.GeoDataFrame)`.

## Test plan

- [x] `tests/unit/test_transform_outdoor_space.py` — all 4 tests pass
- [ ] Pre-commit (ruff, prettier) — passes

Closes #234.

### DIFF
--- a/asf_heat_pump_suitability/pipeline/add_features.py
+++ b/asf_heat_pump_suitability/pipeline/add_features.py
@@ -18,7 +18,6 @@ import os
 import geopandas as gpd
 import polars as pl
 
-from asf_heat_pump_suitability import config
 from asf_heat_pump_suitability.config.settings import load_settings
 from asf_heat_pump_suitability.getters import load_tree_input
 from asf_heat_pump_suitability.pipeline.impute import property_type
@@ -63,6 +62,11 @@ def parse_arguments() -> argparse.Namespace:
 def run(uprns_path: str, area: str = "plymouth") -> None:
     """Add features to domestic UPRNs and write enriched dataset to S3.
 
+    Data types at each stage:
+        Input   — pl.DataFrame    (pl.read_parquet; efficient columnar reader)
+        Spatial — gpd.GeoDataFrame (one conversion at the top; all spatial ops stay here)
+        Output  — pl.DataFrame    (one pl.from_pandas() conversion before writing)
+
     Args:
         uprns_path: Path (S3 URI or local) to the domestic UPRNs parquet file.
         area: Geographic area identifier used to look up INSPIRE land-registry parcels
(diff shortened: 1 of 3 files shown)


caldwellst:
Reverting: the Polars implementation is intentionally kept for performance. Pandas operations on large GeoDataFrames are significantly slower. Closing without merging.

PR #261, opened 2026-02-25
258 Add notebook for anchor property tech reassignment

### LINKED ISSUE #258: Reassign tech types near anchors as 'communal'
Within some radius of a anchor property we may want to assign surrounding buildings a tech type of communal.

simran-dave:
---

# Description

Notebook to explore reassigning tech type as 'communal solution' based on a building's distance from an anchor property.

Fixes #258 

# Instructions for Reviewer

Run the notebook: `research/exploratory/anchor_property_tech_reassignment.py`

Does the list of anchor properties seem ok?
Does the logic in `assign_communal_solutions` function make sense?

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -0,0 +1,225 @@
...
+from asf_heat_pump_suitability.getters import load_tree_input
+
+# %%
+# load datasets
+# convert to CRS 27700 if necessary
+building_footprints = load_tree_input.load_gdf_os_openmap_local_layer(
+    layer="building", grid_squares="SX"
+)
+anchor_properties = gpd.read_file(
+    "s3://asf-heat-pump-suitability/dump/hack_day/anchor_properties_gb.geojson"
```
crispy-wonton: **suggestion**: Did you make this or me? Can't remember. But either way, it would be good to add a comment about what this is / contains.

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -0,0 +1,225 @@
...
+# %%
+# change all solutions to communal, except district heat network
+
+tech_mapping = {
+    "Individual solution": "Communal solutions",
+    "District heat network": "District heat network",
+    "Individual solution or Networked GSHP": "Communal solutions",
+    "Networked GSHP": "Communal solutions",
+    "Communal solutions": "Communal solutions",
+    "Individual solution or District heat network": "Communal solutions",
```
crispy-wonton: **question:** can you check with Sofia if this is the behaviour we want for this one? Otherwise do we want to retain `District heat network`?

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -0,0 +1,225 @@
...
+).to_crs(27700)
+tech_types = gpd.read_file(
+    "s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_building_most_suitable_tech.geojson"
+).to_crs(27700)
+important_buildings = load_tree_input.load_gdf_os_openmap_local_layer(
+    layer="important_building", grid_squares="SX"
+)
+
+# %%
+# categories in important building list that could be anchors
```
crispy-wonton: **question:** how did you create this list?

**note:** we need to include how we selected these categories in our methodology doc.

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -0,0 +1,225 @@
...
+
+# %%
+# add building footprint data to POI anchor properties so geometry isn't just a point
+anchors_with_footprint = (
+    building_footprints.sjoin(anchor_properties, how="inner", predicate="contains")
+).drop("index_right", axis=1)
+
+# %%
+# add POI and important building lists together and remove duplicate buildings
+# I don't think the removing duplicates is fully working, need to check
```
crispy-wonton: **suggestion:** if the deduplication isn't working, you might first need to [normalise the geometry](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoSeries.normalize.html). Sometimes, the coordinates in the same geometry can appear in a different order and so deduplication won't work. Normalising the geometry reorders the coordinates to allow deduplication.

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -0,0 +1,225 @@
...
+
+# %%
+# function to find distance between a residential building and the nearest anchor + resassign tech type based on this distance
+
+
+def assign_communal_solutions(
+    tech_gdf: gpd.GeoDataFrame, anchors_gdf: gpd.GeoDataFrame, radius: float
+) -> gpd.GeoDataFrame:
+    """
+    Finds nearest anchors and reassigns tech type to 'Communal' if within radius.
```
crispy-wonton: **note:** when refactoring into the pipeline, this docstring will need `args` and `returns` information :)

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -0,0 +1,225 @@
...
+    """
+    # Preserve original geometry for distance calculation after join
+    anchors_gdf["anchor_geometry"] = anchors_gdf.geometry
+
+    # Spatial join to find nearest anchor for every building
+    nearest = tech_gdf.sjoin_nearest(anchors_gdf, how="left")
+    nearest["distance"] = nearest["geometry"].distance(nearest["anchor_geometry"])
```
crispy-wonton:
```suggestion
    # Spatial join to find nearest anchor for every building
    nearest = tech_gdf.sjoin_nearest(anchors_gdf, how="left", max_distance=radius, distance_col="distance_m")
```
**suggestion:** using the `max_distance` parameter will allow you to simplify calculating which buildings are in the anchor property radius. You can also optionally add the `distance_col` parameter if you want to retain the actual distance. You would then need to modify the lines below.

**question:** how is the distance / nearest anchor property calculated specifically from polygons? Is it based on the nearest coordinates on the polygon edges or something else? 

**suggestion:** It would be good to visualise some cases with terraced housing. E.g. if one end of the terrace is at the max distance from anchor property and the rest of the terrace is further away , should the whole terrace be reassigned. Please could you look for any examples of this in your test area and then bring it to a stand up for discussion :)

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -0,0 +1,225 @@
...
+    ]
+
+    return reassigned_tech
+
+
+# %%
+# set anchor_radius as the radius around the anchor you want to change to communal
+anchor_radius = 100
+reassigned_tech = assign_communal_solutions(
+    tech_types, all_anchors_with_footprint, anchor_radius
```
crispy-wonton: **issue:** when using custom functions, I strongly suggest inputting keyword arguments when using them IF they can silently fail. E.g. here, because `tech_types` and `all_anchors_with_footprint` are both GeoDataFrames, if you entered them in the wrong order by mistake and weren't using kwargs, it's possible this function could run without erroring but wouldn't give us the expected results. It would then become much harder to debug.

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -0,0 +1,225 @@
+# %%
```
crispy-wonton: **chore:** please could you add a title and a short description at the top of the notebook to explain what its intended purpose is.

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -0,0 +1,225 @@
...
+        geo_j.add_to(m)
+
+    return m
+
+
+# %%
+# click on the building to see either building type or previous tech type
+dissolve_techs_and_plot_folium(
+    reassigned_tech, anchors_with_footprint, important_buildings_filtered, COLOURS
+)
```
crispy-wonton: **thought:** I think it would be a good idea to save this as a html file and then share on the slack channel with everyone. It's interesting that the whole centre of the city basically gets reassigned to communal tech and I think this is something we might want to double check is a behaviour that we want.

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -0,0 +1,225 @@
...
+
+# %%
+# load datasets
+# convert to CRS 27700 if necessary
+building_footprints = load_tree_input.load_gdf_os_openmap_local_layer(
+    layer="building", grid_squares="SX"
+)
+anchor_properties = gpd.read_file(
+    "s3://asf-heat-pump-suitability/dump/hack_day/anchor_properties_gb.geojson"
+).to_crs(27700)
```
sofiapinto: Where's the code for this?

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -0,0 +1,225 @@
...
+    nearest["distance"] = nearest["geometry"].distance(nearest["anchor_geometry"])
+
+    # Apply reassignment logic
+    reassigned_tech = (nearest[nearest["distance"] < radius]).replace(
+        tech_mapping, inplace=False
+    )
+    # save previous tech type for visualisation
+    reassigned_tech["old_tech"] = nearest[nearest["distance"] < radius][
+        "1st_most_suitable_solution"
+    ]
```
crispy-wonton: **suggestion:** you could use `replace` to create a new column, instead of new df to save some lines of code here.

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -0,0 +1,225 @@
...
+
+    # Create map
+    m = folium.Map(
+        location=[centre_map[0][1], centre_map[0][0]],
+        zoom_start=15,
+        tiles="esri_worldimagery",
+    )
+
+    for _, r in dissolved_4326_gdf.iterrows():
+        colour = colours[r["1st_most_suitable_solution"]]
```
crispy-wonton: **suggestion:** it would also be great to separately map the communal solutions with the colours of their original techs, so we can see what's changed. Alongside this, it would be good to produce some value count & proportion data so that we can see how many of each tech type have been reassigned.

crispy-wonton (commented):
Thanks for this @simran-dave ! It looks good and you've written some very clean and readable code.

I couldn't find any bugs. I made some suggestions.

To answer your questions:

> Does the list of anchor properties seem ok?

They seem reasonable to me but I'm not sure where they came from and I think it's possible some are missing. E.g. certain industrial sites probably make a good anchor property, same with large supermarkets, and shopping centres. Can you elaborate on where the categories came from?

You've already mentioned this previously, but looking at the map, there's some anchor properties which seem quite small so we'd probably want to check with the team whether these should be included. 

> Does the logic in assign_communal_solutions function make sense?

Yes, the logic makes sense. I've made some suggestions to make the code more efficient here which will be especially noticeable on scaling.

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
simran-dave: this file was from you, have added a comment + notebook now prints out categories included for reference

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
simran-dave: yeah makes sense! Have changed to use kwargs

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
simran-dave: I based it on the categories in the `anchor_properties_gb.geojson` file but yeah the list could do with some editing

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
simran-dave: yeah can do that! I am currently plotting all buildings within the anchor property radius, with red being district heat network (which I think is why a lot of the city centre is showing up) and orange being communal tech. But will remove the district heat network properties. Also- this version of the notebook is using a radius of 100m whereas in practice maybe we'd want ~ 50 m ?

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
simran-dave: yeah I wasn't fully sure what we wanted to do in the cases where there were two techs assigned to a a building. Will ask Sofia.

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
simran-dave: I think it's from the nearest edges of the polygon. And yeah, at the moment looks like the whole terrace is reassigned- can show some examples in a stand up

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
simran-dave: have added these in

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
crispy-wonton: Cool, thanks for clarifying - I misunderstood and thought that was all communal.

I think either 50m or 100m is fine for now - we haven't done any work to determine the distance. You can flag this in a sprint review and the methodology document.

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -1,3 +1,8 @@
+# %% [markdown]
+# ## Anchor Property Tech Reassignment
+#
+# notebook to find buildings within a specified distance of an anchor proprty, reassign their tech type as communal and visualise the results
```
crispy-wonton: suggestion: add 'reassign their tech type as communal **(unless assigned district heat network)**

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
crispy-wonton: Cool thanks for clarifying

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -90,6 +105,7 @@
 all_anchors_with_footprint = pd.concat(
     [anchors_with_footprint, important_buildings_filtered]
 )
+all_anchors_with_footprint["geometry"] = all_anchors_with_footprint.normalize()
```
crispy-wonton: question: has this fixed your issue? If so, you can remove the annotation above

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -101,31 +117,38 @@ def assign_communal_solutions(
 ) -> gpd.GeoDataFrame:
     """
     Finds nearest anchors and reassigns tech type to 'Communal' if within radius.
+    Args:
+        tech_gdf (gpd.GeodataFrame): dataframe with most suitable tech types assigned to residential buildings
+        anchors_gdf (gpd.GeodataFrame): dataframe with anchor properties
```
crispy-wonton: suggestion: add that anchor property geometries are point geometries

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -101,31 +117,38 @@ def assign_communal_solutions(
...
+        radius (float): max distance from anchor property to reassign tech types
+    Returns:
+        dataframe of residential buildings that have had their tech type reassigned, with columns for old tech type, corresponding anchor property geometry and distance from anchor added
```
crispy-wonton: chore: missing type hint in returns statement

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -101,31 +117,38 @@ def assign_communal_solutions(
...
 
     # Spatial join to find nearest anchor for every building
-    nearest = tech_gdf.sjoin_nearest(anchors_gdf, how="left")
-    nearest["distance"] = nearest["geometry"].distance(nearest["anchor_geometry"])
-
-    # Apply reassignment logic
-    reassigned_tech = (nearest[nearest["distance"] < radius]).replace(
-        tech_mapping, inplace=False
+    reassigned_tech = tech_gdf.sjoin_nearest(
+        anchors_gdf, how="left", max_distance=radius, distance_col="distance_m"
```
crispy-wonton: suggestion: i think you don't need line 129 now

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
```diff
@@ -27,7 +34,15 @@
...
-# categories in important building list that could be anchors
+# for reference, these are all the anchor property categories in the POI based list
+anchor_properties["main_category"].unique()
+
+# %%
+# these are all possible building type categories in the OS important building layer
+important_buildings["CLASSIFICA"].unique()
+
+# %%
+# now filter to categories in important building layer that could be anchor properties
```
crispy-wonton: thought: we should check in a sprint review, stand up , or slack that there's no categories missing from here.

crispy-wonton (commented):
hi @simran-dave 

these changes look good thank you :) I've left some additional comments, questions, and suggestions.

Additionally, I've realised that the file you're working in should be in its own subdirectory within `/exploratory`. You could create one called something like `anchor_properties` and save it in there. This means that if we do further exploration around anchor properties, we can store it together in the same location.

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
simran-dave: here they are now joined to their building footprints, have added a comment in

asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py
simran-dave: yeah can bring this up

asf_heat_pump_suitability/research/exploratory/anchor_properties/anchor_property_tech_reassignment.py
```diff
@@ -0,0 +1,298 @@
...
+    dissolved_gdf["colour"] = dissolved_gdf[tech_to_plot].map(colours)
+
+    # Convert dissolved tech polygons and buildings to EPSG 4326 for plotting
+    dissolved_4326_gdf = dissolved_gdf.to_crs(epsg=4326)
+    anchor_4326 = anchor_properties_with_footprint.to_crs(4326)
+    important_4326 = important_building_anchors.to_crs(4326)
+
+    # Get centre of boundary to centre map
+    boundary_4326 = dissolved_4326_gdf["geometry"].values[0]
+    centre_map = shapely.get_coordinates(boundary_4326.centroid)
```
crispy-wonton: issue: just noticed here, this is not a boundary. It's the first polygon in the dissolved geodataframe. Annotation and variable names need to be updated.

asf_heat_pump_suitability/research/exploratory/anchor_properties/anchor_property_tech_reassignment.py
```diff
@@ -0,0 +1,298 @@
...
+    tech_gdf: gpd.GeoDataFrame, anchors_gdf: gpd.GeoDataFrame, radius: float
+) -> gpd.GeoDataFrame:
+    """
+    Finds nearest anchors and reassigns tech type to 'Communal' if within radius.
+    Args:
+        tech_gdf (gpd.GeodataFrame): dataframe with most suitable tech types assigned to residential buildings
+        anchors_gdf (gpd.GeodataFrame): dataframe with anchor properties and their building footprints
+        radius (float): max distance from anchor property to reassign tech types
+    Returns:
+        reassigned_tech (gpd.GeoDataFrame): dataframe of residential buildings that have had their tech type reassigned, with columns for old tech type, corresponding anchor property geometry and distance from anchor added
```
crispy-wonton:
```suggestion
        gpd.GeoDataFrame: dataframe of residential buildings that have had their tech type reassigned, with columns for old tech type, corresponding anchor property geometry and distance from anchor added
```
You only need the type hint, not variable name :)

crispy-wonton (changes requested):
Just a couple of minor changes to make.

Also, I'm not sure if this is just github showing me the wrong thing, but it looks like there are now two separate notebook files in different locations. If that's the case, please could you delete the file at the below location, and make sure any changes you push are to the file in the `/anchor_properties` subdir.

To delete: asf_heat_pump_suitability/research/exploratory/anchor_property_tech_reassignment.py

asf_heat_pump_suitability/research/exploratory/anchor_properties/anchor_property_tech_reassignment.py
```diff
@@ -101,7 +101,7 @@
 
 # %%
 # add POI and important building lists together and remove duplicate buildings
-
+# I don't think the removing duplicates is fully working, need to check
```
crispy-wonton: suggestion: I can see that this dataframe also contains building ID. You could use this to deduplicate if it's not working on the geometry. From what I can see, the deduplication seems to be working - there are no duplicate building IDs at this point when running `all_anchors_with_footprint["ID"].duplicated().sum()`. Maybe this annotation was added back in by mistake though.

crispy-wonton (dismissed):
One small point on the deduplication - perhaps it's just an annotation mistake - but otherwise good to merge!

sofiapinto (approved):

PR #262, opened 2026-03-02
216 Domestic Filtering Exploration

### LINKED ISSUE #216: Improve filtering of domestic UPRNs
PR #184 added a pipeline to filter UPRNs to domestic only. There are still buildings retained in the final 'domestic' UPRN dataset which appear to be large commercial units, particularly in the city centres.

We need to:
- Confirm whether there is a bug in the pipeline that means that certain commercial buildings are not being filtered out properly. 
- Explore how we can make improvements to the filtering process to remove more commercial buildings without losing domestic buildings. This could be via implementing further simple rules (e.g. filtering out buildings of a certain size or UPRN density), or by creating a classification model to identify commercial buildings by creating ground truth datasets from the EPC registers and the points of interest datasets, or another idea.

simran-dave:
---

# Description

Notebook added for domestic filtering exploration

Fixes #216 

# Instructions for Reviewer

run `asf_heat_pump_suitability/research/exploratory/domestic_filtering.py`

My recommendations based on this notebook would be:

- filter out properties with footprint > 1000 m^2 per UPRN
- after joining EPC domestic properties to building footprints, join EPC domestic properties to nearest footprint <1m away
- add a step that says if building is in important buildings / POI lists and has exactly 1 UPRN, filter out

Based on the outputs of the notebook, do these seem sensible?

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
+++ b/asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
@@ -1,0 +1,284 @@
+# %% [markdown]
+# ## Domestic Filtering Exploration
+#
+
+# %%
+import numpy as np
+import pandas as pd
+import geopandas as gpd
+import polars as pl
+import matplotlib.pyplot as plt
+import shapely
+import folium
+import os
+
+# %%
+# local imports
+from asf_heat_pump_suitability import PROJECT_DIR
+from asf_heat_pump_suitability import config
+from asf_heat_pump_suitability.getters import (
+    load_tree_input,
+    load_geodata,
+    load_boundaries,
+)
+from asf_heat_pump_suitability.pipeline.transform import uprns, poi
+
+# %%
+# load datasets
+building_footprints_gdf = load_tree_input.load_gdf_os_openmap_local_layer(
+    layer="building", grid_squares="SX"
+)
+anchor_properties_gdf = gpd.read_file(
+    "s3://asf-heat-pump-suitability/dump/hack_day/anchor_properties_gb.geojson"
+).to_crs(27700)
+tech_types_gdf = gpd.read_file(
+    "s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_building_most_suitable_tech.geojson"
+).to_crs(27700)
+important_buildings_gdf = load_tree_input.load_gdf_os_openmap_local_layer(
+    layer="important_building", grid_squares="SX"
+)
+epc_domestic_set = uprns.load_set_valid_epc_uprns(epc_type="domestic")
+epc_commercial_set = uprns.load_set_valid_epc_uprns(epc_type="commercial")
+uprns_df = load_geodata.load_df_osopen_uprn()
+uprns_GB_gdf = uprns.generate_gdf_uprn_coords(uprns_df)
+
+# %%
+# Load POI data
+
+poi_gdf = load_tree_input.load_gdf_poi()
+poi_gdf = poi.transform_gdf_poi(
+    poi_gdf,
+    filter_categories=poi.load_set_non_domestic_poi_categories(),
+)
+
+# %% [markdown]
+# ## EPC Data - Properties that should be domestic
+
+# %%
+# gdf of all uprns in GB that are in the domestic EPC data
+
+uprns_GB_gdf = uprns_GB_gdf[(uprns_GB_gdf["UPRN"].isin(epc_domestic_set))]
+
+# %%
+# filter these to plymouth only
+
+la_boundaries_gdf = load_boundaries.load_gdf_local_authority_boundaries(
+    select_las="Plymouth"
+)
+domestic_uprns_gdf = uprns_GB_gdf.sjoin(
+    la_boundaries_gdf[["LAD23CD", "LAD23NM", "geometry"]],
+    how="inner",
+    predicate="intersects",
+).drop(columns="index_right")
+
+# %%
+# dataframe of domestic EPC points not in a building footprints polygon
+# these are definitely domestic properties but aren't being included in final planning tool because of this
+points_join = gpd.sjoin(
+    domestic_uprns_gdf, building_footprints_gdf, how="left", predicate="intersects"
+)
+
+epc_points_out = (points_join[points_join["index_right"].isna()]).drop(
+    "index_right", axis=1
+)
+
+# %%
+# fraction of points in domestic epc data that aren't being joined to a building footprint
+epc_out_frac = len(epc_points_out) / len(domestic_uprns_gdf) * 100
+print(f"proportion of EPC points not joined to a building: {epc_out_frac:.2f}%")
+
+# %%
+# save building footprint geometry in separate column
+building_footprints_gdf["saved_geom"] = building_footprints_gdf.geometry
+
+# find nearest building to each domestic property that hasn't been assigned a footprint
+nearest_building = epc_points_out.sjoin_nearest(building_footprints_gdf, how="left")
+# save this distance in a column
+nearest_building["distance"] = nearest_building["geometry"].distance(
+    nearest_building["saved_geom"]
+)
+
+# %%
+# fraction of domestic buildings not assigned a footprint that are within X metres of a building
+dist = 1
+dist_fraction = (
+    len(nearest_building[nearest_building["distance"] < dist])
+    / len(nearest_building)
+    * 100
+)
+print(
+    f"proportion of EPC points not joined to a building that are within {dist}m of a building: {dist_fraction:.2f}%"
+)
+
+# %%
+# histogram of distance from nearest building
+h = plt.hist(nearest_building["distance"], bins=200, density=True)
+plt.xlabel("Distance from nearest building [m]")
+plt.ylabel("Fraction of total counts")
+
+
+# %%
+import branca.colormap as cm
+
+
+def plot_nearest_buildings(nearest_building: gpd.GeoDataFrame):
+    """
+    Plot the resulting polygons together for each tech type in Folium.
+
+    Args:
+        gdf (gpd.GeoDataFrame): dataframe with polygons to dissolve and plot and tech labels
+        colours (dict): tech type labels and their corresponding colours for plotting
+
+    Returns Folium map
+    """
+
+    # Convert dissolved tech polygons and buildings to EPSG 4326 for plotting
+    nearest_building_4326_gdf = nearest_building.to_crs(epsg=4326)
+    nearest_building_4326_gdf = nearest_building_4326_gdf.set_geometry("saved_geom")
+    nearest_building_4326_gdf = nearest_building_4326_gdf.to_crs(epsg=4326)
+
+    # Get centre of boundary to centre map
+    boundary_4326 = nearest_building_4326_gdf["geometry"].values[0]
+    centre_map = shapely.get_coordinates(boundary_4326.centroid)
+
+    # Create map
+    m = folium.Map(
+        location=[centre_map[0][1], centre_map[0][0]],
+        zoom_start=15,
+        tiles="esri_worldimagery",
+    )
+
+    # Plot building footprints of nearest buildings
+    for _, r in nearest_building_4326_gdf.iterrows():
+        sim_geo = gpd.GeoSeries(r["saved_geom"])
+        geo_j = sim_geo.to_json()
+        geo_j = folium.GeoJson(
+            data=geo_j,
+            style_function=lambda x: {
+                "color": "blue",
+                "weight": 0.1,
+                "fillOpacity": 0.4,
+            },
+        )
+        geo_j.add_to(m)
+
+    # Plot UPRNS of missed EPC data
+    for _, r in nearest_building_4326_gdf.iterrows():
+        sim_geo = gpd.GeoSeries(r["geometry"])
+        geo_j = sim_geo.to_json()
+        min = np.min(nearest_building_4326_gdf["distance"])
+        max = np.max(nearest_building["distance"])
+        colours = cm.linear.YlOrRd_03.scale(min, max)
+        marker = folium.Circle(
+            radius=5,
+            fill_color=colours(r["distance"]),
+            fill_opacity=0.5,
+            color="black",
+            weight=1,
+        )
+        geo_j = folium.GeoJson(data=geo_j, marker=marker)
+        geo_j.add_to(m)
+
+    return m
+
+
+# %%
+# plot UPRNS that are within 1m of a building, with building footprints in blue
+plot_nearest_buildings(nearest_building)
+
+# %% [markdown]
+# ## Properties labelled domestic but should be commercial
+
+# %%
+COLOURS = {
+    "Individual solution": "#18A48C",
+    "District heat network": "#EA2541",
+    "Individual solution or Networked GSHP": "grey",
+    "Networked GSHP": "#0000FF",
+    "Communal solutions": "#FF6E47",
+    "Individual solution or District heat network": "gray",
+}
+
+
+# %%
+def plot_buildings(buildings_gdf: gpd.GeoDataFrame, colours: dict = COLOURS):
+    """
+    Plot the resulting polygons together for each tech type in Folium.
+
+    Args:
+        gdf (gpd.GeoDataFrame): dataframe with polygons to dissolve and plot and tech labels
+        colours (dict): tech type labels and their corresponding colours for plotting
+
+    Returns Folium map
+    """
+    # Dissolve by tech type
+    dissolved_gdf = buildings_gdf.dissolve(
+        by="1st_most_suitable_solution"
+    ).reset_index()
+    dissolved_gdf["colour"] = dissolved_gdf["1st_most_suitable_solution"].map(colours)
+
+    # Convert dissolved tech polygons and buildings to EPSG 4326 for plotting
+    dissolved_4326_gdf = dissolved_gdf.to_crs(epsg=4326)
+
+    # Get centre of boundary to centre map
+    boundary_4326 = dissolved_4326_gdf["geometry"].values[0]
+    centre_map = shapely.get_coordinates(boundary_4326.centroid)
+
+    # Create map
+    m = folium.Map(
+        location=[centre_map[0][1], centre_map[0][0]],
+        zoom_start=15,
+        tiles="esri_worldimagery",
+    )
+
+    # Plot building footprints
+    for _, r in dissolved_4326_gdf.iterrows():
+        colour = colours[r["1st_most_suitable_solution"]]
+        sim_geo = gpd.GeoSeries(r["geometry"])
+        geo_j = sim_geo.to_json()
+        geo_j = folium.GeoJson(
+            data=geo_j,
+            style_function=lambda x, colour=colour: {
+                "fillColor": colour,
+                "fillOpacity": 0.3,
+            },
+        )
+        folium.Popup(f"UPRNs: {r['UPRN']}").add_to(geo_j)
+        geo_j.add_to(m)
+
+    return m
+
+
+# %%
+# add column with area per UPRN for each building
+
+tech_types_gdf["area_per_UPRN"] = (tech_types_gdf.area) / tech_types_gdf["UPRN"]
+np.average(tech_types_gdf["area_per_UPRN"])  # for reference
+
+# %%
+# buildings with area per UPRN > 1000 m2
+plot_buildings(tech_types_gdf[tech_types_gdf["area_per_UPRN"] > 1000])
+
+# %%
+# join tech types gdf with epc to get buildings with a confirmed domestic property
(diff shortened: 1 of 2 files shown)


asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_filtering.py
crispy-wonton, line 1: todo: please could you move this into its own subdir within `/exploratory`, e.g. called `domestic_properties` or something. This will allow us to add other files here later if we conduct further exploration in the same subject.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_filtering.py
```diff
@@ -0,0 +1,284 @@
+# %% [markdown]
+# ## Domestic Filtering Exploration
+#
```
crispy-wonton, line 3: chore: please could you add a brief summary of the purpose of this notebook

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
```diff
@@ -0,0 +1,284 @@
...
+    how="inner",
+    predicate="intersects",
+).drop(columns="index_right")
+
+# %%
+# dataframe of domestic EPC points not in a building footprints polygon
+# these are definitely domestic properties but aren't being included in final planning tool because of this
+points_join = gpd.sjoin(
+    domestic_uprns_gdf, building_footprints_gdf, how="left", predicate="intersects"
+)
```
crispy-wonton: nitpick: it's recommended by `geopandas` to use `sjoin` as methods rather than functions. This is a note from their d[ocumentation](https://geopandas.org/en/stable/docs/user_guide/mergingdata.html):

>  For historical reasons, both methods are also available as top-level functions [sjoin()](https://geopandas.org/en/stable/docs/reference/api/geopandas.sjoin.html#geopandas.sjoin) and [sjoin_nearest()](https://geopandas.org/en/stable/docs/reference/api/geopandas.sjoin_nearest.html#geopandas.sjoin_nearest). It is recommended to use methods as the functions may be deprecated in the future.

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
```diff
@@ -0,0 +1,284 @@
...
+    Plot the resulting polygons together for each tech type in Folium.
+
+    Args:
+        gdf (gpd.GeoDataFrame): dataframe with polygons to dissolve and plot and tech labels
+        colours (dict): tech type labels and their corresponding colours for plotting
+
+    Returns Folium map
+    """
+
+    # Convert dissolved tech polygons and buildings to EPSG 4326 for plotting
```
crispy-wonton: issue: I dont think there's any dissolving done to the input polygons? Also, think your input is a geodataframe with a UPRN point geometry and a building polygon per row, not tech polygons.

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
```diff
@@ -0,0 +1,284 @@
...
+# %%
+import branca.colormap as cm
+
+
+def plot_nearest_buildings(nearest_building: gpd.GeoDataFrame):
+    """
+    Plot the resulting polygons together for each tech type in Folium.
+
+    Args:
+        gdf (gpd.GeoDataFrame): dataframe with polygons to dissolve and plot and tech labels
```
crispy-wonton: issue: the polygons are not being dissolved in the function

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
```diff
@@ -0,0 +1,284 @@
...
+        colours (dict): tech type labels and their corresponding colours for plotting
```
crispy-wonton: issue: there is no colours param in the function

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_filtering.py
```diff
@@ -0,0 +1,284 @@
...
+        colours (dict): tech type labels and their corresponding colours for plotting
+
+    Returns Folium map
+    """
+
+    # Convert dissolved tech polygons and buildings to EPSG 4326 for plotting
+    nearest_building_4326_gdf = nearest_building.to_crs(epsg=4326)
+    nearest_building_4326_gdf = nearest_building_4326_gdf.set_geometry("saved_geom")
+    nearest_building_4326_gdf = nearest_building_4326_gdf.to_crs(epsg=4326)
+
```
crispy-wonton, line 153: suggestion: it's not necessarily something that really needs to be addressed here as this won't be refactored into the pipeline, but generally I would recommend doing the conversion on the geometries before `sjoin` and then join the converted CRS geometries after.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_filtering.py
```diff
@@ -0,0 +1,284 @@
...
+    # Get centre of boundary to centre map
+    boundary_4326 = nearest_building_4326_gdf["geometry"].values[0]
+    centre_map = shapely.get_coordinates(boundary_4326.centroid)
```
crispy-wonton, line 156: issue: this is just an annotation / variable naming issue but the code is getting the centroid of the first building footprint in the geodataframe, not the boundary.

I know this is from the original annotation and code I wrote and sent you but it needs to be corrected before merge.

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
```diff
@@ -0,0 +1,284 @@
...
+        )
+        geo_j.add_to(m)
+
+    # Plot UPRNS of missed EPC data
+    for _, r in nearest_building_4326_gdf.iterrows():
+        sim_geo = gpd.GeoSeries(r["geometry"])
+        geo_j = sim_geo.to_json()
+        min = np.min(nearest_building_4326_gdf["distance"])
+        max = np.max(nearest_building["distance"])
+        colours = cm.linear.YlOrRd_03.scale(min, max)
```
crispy-wonton: question: are you intending to use two different dfs here to get min / max or should it be the same? I don't think it makes a difference to the distance, it's just inconsistent.

suggestion: also, I believe you could move these three lines outside of the for loop so they don't have to run every time.

crispy-wonton: issue: I think something about the colour map isn't working here. When plotting, all the points look the same colour to me. Do they look different for you?

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
```diff
@@ -0,0 +1,284 @@
...
+            weight=1,
+        )
+        geo_j = folium.GeoJson(data=geo_j, marker=marker)
+        geo_j.add_to(m)
+
+    return m
+
+
+# %%
+# plot UPRNS that are within 1m of a building, with building footprints in blue
```
crispy-wonton: question: please point me to the code if I've missed it, but I don't think this is plotting UPRNs within 1m of a building - I think it's just plotting all UPRNs outside of buildings?

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
```diff
@@ -0,0 +1,284 @@
...
+# %%
+# fraction of domestic buildings not assigned a footprint that are within X metres of a building
+dist = 1
+dist_fraction = (
+    len(nearest_building[nearest_building["distance"] < dist])
+    / len(nearest_building)
+    * 100
+)
+print(
+    f"proportion of EPC points not joined to a building that are within {dist}m of a building: {dist_fraction:.2f}%"
```
crispy-wonton:
```suggestion
    f"proportion of EPC points in Plymouth not joined to a building that are within {dist}m of a building: {dist_fraction:.2f}%"
```

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
```diff
@@ -0,0 +1,284 @@
...
+)
+
+# %%
+# histogram of distance from nearest building
+h = plt.hist(nearest_building["distance"], bins=200, density=True)
+plt.xlabel("Distance from nearest building [m]")
+plt.ylabel("Fraction of total counts")
```
crispy-wonton: suggestion: I'd probably also add the following code:

`nearest_building["distance"].describe()`
This shows that 75% of the UPRNs in Plymouth missing a building are within <3m of a building.

I'd also show another histogram with `density=False` so that we can see the absolute values to get a sense of the impact on actual counts.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_filtering.py
```diff
@@ -0,0 +1,284 @@
...
+# %%
+def plot_buildings(buildings_gdf: gpd.GeoDataFrame, colours: dict = COLOURS):
+    """
+    Plot the resulting polygons together for each tech type in Folium.
+
+    Args:
+        gdf (gpd.GeoDataFrame): dataframe with polygons to dissolve and plot and tech labels
+        colours (dict): tech type labels and their corresponding colours for plotting
+
+    Returns Folium map
```
crispy-wonton, line 231: issue: i think this docstring needs updating again. The main description could do with some more description - it's not clear what it's mapping - and the `buildings_gdf` kwarg description needs to be updated.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_filtering.py
```diff
@@ -0,0 +1,284 @@
...
+        by="1st_most_suitable_solution"
+    ).reset_index()
+    dissolved_gdf["colour"] = dissolved_gdf["1st_most_suitable_solution"].map(colours)
+
+    # Convert dissolved tech polygons and buildings to EPSG 4326 for plotting
+    dissolved_4326_gdf = dissolved_gdf.to_crs(epsg=4326)
+
+    # Get centre of boundary to centre map
+    boundary_4326 = dissolved_4326_gdf["geometry"].values[0]
+    centre_map = shapely.get_coordinates(boundary_4326.centroid)
```
crispy-wonton, line 236: issue: as above, this isn't using the boundary, but the first value in the geometry column of the dissolved gdf.

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
```diff
@@ -0,0 +1,284 @@
...
+        sim_geo = gpd.GeoSeries(r["geometry"])
+        geo_j = sim_geo.to_json()
+        geo_j = folium.GeoJson(
+            data=geo_j,
+            style_function=lambda x, colour=colour: {
+                "fillColor": colour,
+                "fillOpacity": 0.3,
+            },
+        )
+        folium.Popup(f"UPRNs: {r['UPRN']}").add_to(geo_j)
```
crispy-wonton: suggestion: you could also add area of the building footprint or area per UPRN here.

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
```diff
@@ -0,0 +1,284 @@
...
+        geo_j.add_to(m)
+
+    return m
+
+
+# %%
+# add column with area per UPRN for each building
+
+tech_types_gdf["area_per_UPRN"] = (tech_types_gdf.area) / tech_types_gdf["UPRN"]
+np.average(tech_types_gdf["area_per_UPRN"])  # for reference
```
crispy-wonton: issue: I think to be safe, I would not use the tech types dataset as that has gone through further processing after domestic filtering. I would suggest importing the original filtered domestic UPRN data for Plymouth, and then sjoining to building footprints to generate the data for this exploration. If you then need other data from the `tech_types_gdf` dataframe, I would then join it to the main one you've created.

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
```diff
@@ -0,0 +1,284 @@
...
+plot_buildings(tech_types_gdf[tech_types_gdf["area_per_UPRN"] > 1000])
+
+# %%
+# join tech types gdf with epc to get buildings with a confirmed domestic property
+tech_with_epc = domestic_uprns_gdf.sjoin(
+    tech_types_gdf, how="inner", predicate="intersects"
+)
+
+# how many of these have an area per UPRN > 1000 ?
+tech_with_epc[tech_with_epc["area_per_UPRN"] > 1000]
```
crispy-wonton: praise: I think this is a good idea to join to the EPC. 

suggestion: You could take this further to look at the upper end of the distribution of building size and UPRN density in these buildings to try and identify a threshold density, above which we consider buildings to be non-domestic.

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
```diff
@@ -0,0 +1,284 @@
...
+
+# %%
+# join tech types gdf with important buildings
+tech_with_important = tech_types_gdf.sjoin(
+    important_buildings_gdf, how="inner", predicate="intersects"
+)
+
+# important buildings that have just one UPRN. These should then be purely commercial buildings but are getting classed as domestic
+tech_with_important[tech_with_important["UPRN"] == 1]
```
crispy-wonton: praise: really great find and thinking here - this is definitely an oversight of the original pipeline.

note: since there aren't that many buildings in this category and below, I think this warrants some further exploration.

crispy-wonton (changes requested):
hi @simran-dave :)

Thanks for this work and exploration. I think this is a good start with some good ideas and suggestions for improvements we can make to the pipeline. The notebook ran successfully all the way through and the folium plots all were displayed.

I've left some comments with tags according to the [conventional comments](https://conventionalcomments.org/) style to help you filter through the feedback.

I think there is some further exploration needed as the suggestions you've made are good and sensible, but they don't fully address the issue of domestic properties being mislabeled. Your work has given me a clearer idea of what to explore next. The main thing to incorporate into the further exploration, will be the council tax data. I think we can use it as our best source of 'ground truth' for reference. I will discuss this in more detail with you separately.

I would suggest we do further exploration in a **separate PR** though and focus on making changes and merging this work first. This will keep the PRs manageable for review.

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
simran-dave: I have just realised that this is why the colour map wasn't working properly! The min/ max was getting updated in each row. Moved them out of the for loop and now it should work ok

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
simran-dave: yeah you're right, this comment must be old. I've changed it

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
simran-dave: Thanks :) My thinking is that any building in the important buildings or POI datasets (without the assumed non-residential building types filters on them) with exactly 1 UPRN should be removed. I think the important buildings list here is the full list but the POI data has been filtered. I can see what it looks like with no building types filter applied.

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
simran-dave: yeah that makes sense! Will look at the max area per UPRN in the domestic EPC list.

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
simran-dave: sure, have used original domestic UPRN data. I was using tech types as it had a column that had UPRNs per building. I am now making that column myself with the domestic data by finding number of rows with the same building footprint.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_filtering.py
```diff
@@ -28,9 +31,7 @@
 building_footprints_gdf = load_tree_input.load_gdf_os_openmap_local_layer(
     layer="building", grid_squares="SX"
 )
-anchor_properties_gdf = gpd.read_file(
-    "s3://asf-heat-pump-suitability/dump/hack_day/anchor_properties_gb.geojson"
-).to_crs(27700)
+# anchor_properties_gdf = gpd.read_file('s3://asf-heat-pump-suitability/dump/hack_day/anchor_properties_gb.geojson').to_crs(27700)
```
crispy-wonton: nitpick: should this be deleted?

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_filtering.py
```diff
@@ -232,53 +243,85 @@ def plot_buildings(buildings_gdf: gpd.GeoDataFrame, colours: dict = COLOURS):
...
+(residential_with_uprn_counts_gdf["area_per_UPRN"]).describe()
 
 # %%
 # buildings with area per UPRN > 1000 m2
-plot_buildings(tech_types_gdf[tech_types_gdf["area_per_UPRN"] > 1000])
+plot_buildings(
+    residential_buildings_gdf=residential_with_uprn_counts_gdf,
+    boundary_gdf=la_boundaries_gdf,
+    area=1000,
+)
```
crispy-wonton, line 293: praise: nice! i like this clean function!

asf_heat_pump_suitability/research/exploratory/domestic_filtering.py
crispy-wonton: This can now be done with the council tax data in a separate PR as per my message on slack with tasks for this sprint :)

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_filtering.py
```diff
@@ -0,0 +1,327 @@
...
+# (Groupby results often revert to a standard DataFrame)
+residential_with_uprn_counts_gdf = gpd.GeoDataFrame(
+    uprn_counts, geometry="geometry", crs=residential_gdf.crs
+)
+
+# %%
+# add column with area per UPRN for each building
+residential_with_uprn_counts_gdf["area_per_UPRN"] = (
+    residential_gdf.area
+) / residential_gdf["UPRN_count"]
```
crispy-wonton:
```suggestion
residential_with_uprn_counts_gdf["area_per_UPRN"] = (
    residential_with_uprn_counts_gdf.area
 / residential_with_uprn_counts_gdf["UPRN_count"]
)
```
issue: small bug here as it's using the wrong gdf so giving a keyError

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_filtering.py
```diff
@@ -0,0 +1,327 @@
...
+uprns_GB_gdf = uprns.generate_gdf_uprn_coords(uprns_df)
+
+# %%
+residential_uprn_df = base_getters.load_df_from_s3(
+    config["data"]["processed"]["plymouth_residential_uprns"]
+)
+residential_gdf = uprns.generate_gdf_uprn_coords(residential_uprn_df)
+residential_gdf = building_footprints_gdf.sjoin(
+    residential_gdf, how="inner", predicate="intersects"
+)
```
crispy-wonton:
```suggestion
residential_gdf = building_footprints_gdf.sjoin(
    residential_gdf, how="inner", predicate="intersects"
).drop(columns="index_right")
```
issue: this is causing a ValueError in the final cell of the notebook because `'index_right' cannot be a column name in the frames being joined`.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_filtering.py
```diff
@@ -0,0 +1,327 @@
...
+
+# important buildings that have just one UPRN. These should then be purely commercial buildings but are getting classed as domestic
+residential_with_important[residential_with_important["UPRN_count"] == 1]
+
+# %%
+# same as above, with POI data
+
+residential_with_poi = residential_gdf.sjoin(
+    poi_gdf, how="inner", predicate="intersects"
+)
```
crispy-wonton:
```suggestion
residential_with_poi = residential_with_uprn_counts_gdf.sjoin(
    poi_gdf, how="inner", predicate="intersects"
)
```
issue: I think this is using the wrong df. It's giving KeyError even with the `index_right` dropped.

crispy-wonton (commented):
hey @simran-dave !

Thanks for making the changes - there are some small bugs to fix, but then it's good to merge.

crispy-wonton (dismissed):

sofiapinto (approved):

PR #272, opened 2026-03-10
263 productionise clustering

### LINKED ISSUE #263: Add clustering to pipeline
Add initial clustering of buildings by tech type into main pipeline.

crispy-wonton:
Fixes #263 

# Description

**UPDATE**: Additional bugfix now pushed as of 16/03/2026 addressed by the commits shown in the screenshot below. Bug description: after creating the Voronoi cells and overlaying the barriers, there were more Voronoi cells than building footprint polygons remaining. The expected behaviour is that there should be equal number of Voronoi cells as building footprints. The root of the issue was that occasionally a barrier was fragmenting a Voronoi cell into multiple pieces. That meant that, after removing the parts of Voronoi cells that were overlaid with barriers, you ended up with some building footprints which intersected with multiple cell fragments. To solve this, I have removed a 1cm buffer around all the Voronoi polygons, and then dissolved the Voronoi fragments with the building footprints. I have checked the outputs - there are now the same number of Voronoi cells as building footprints (before the final clustering step).

image

Productionise the clustering of the polygons, including reassignment of communal clusters if they neighbour (touching) a networked GSHP cluster. 

Note, @simran-dave is working on a separate branch off this one to add the functions and script updates to reassign technology to buildings depending on their distance from an anchor property.

New files:
- `pipeline/cluster/cluster.py` - add functions to create Voronoi polygons from buildings, separate by physical boundaries, reassign communal tech type to N-GSHP where relevant, and cluster

Modified files:
- `config/base.yaml` - add new Greenspace file location, output save as name, and tech type names for reference
- `getters/load_geodata.py` - move existing function from `load_tree_input.py` and modify to allow loading of greenspace files and surface water files
- `getters/load_tree_input.py` - add filter to POI data to filter to GB only
- `utils/save_utils.py` - update save utils to enable saving geodataframes

# Instructions for Reviewer

To run the script, run the following line in terminal:
`python asf_heat_pump_suitability/pipeline/cluster/cluster.py --tech_gdf s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth/plymouth_buildings_most_suitable_tech.parquet --local_authorities plymouth
`

You can also add the `--save` flag to test whether the saving works for you.

- The order of operations. Does it make sense and then result in the correct clusters?
- The logic of removing commercial voronoi polygons and treating them as physical barriers.
- Are there any obvious physical boundaries missing that should be included? This can also be flagged in the sprint review.

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/config/base.yaml
```diff
@@ -84,11 +86,17 @@ constant:
       - "SJ"
       - "SK"
   target_crs: 27700 # British National Grid
-
+  # TODO update to new names before running the pipeline again
+  tech_types:
+    individual: "Individual solution"
+    N_GSHP: "Networked GSHP"
+    communal: "Communal solutions"
+    heat_networks: "District heat network"
```
crispy-wonton: These are the old names and need to be updated, but they are the names that are currently used in the tech polygons file so to run the main script, they need to use the old versions. To be updated.

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -6,6 +6,8 @@ data:
     gb_os_openmap_local: "s3://asf-heat-pump-suitability/source_data/opmplc_gb.gpkg"
     # OpenMap Local layers for certain grid squares
     grid_square_os_openmap_local: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/v202510_OSOpenMapLocal_geometries_selected/{square}/{square}_{layer}.shp"
+    # OpenMap Green space layers for certain grid squares
+    grid_square_os_openmap_greenspace: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/v202510_OSOpenMapGreenspace_geometries_selected/{square}/{square}_{layer}.shp"
```
sofiapinto, line 10: Is this data at GB level?

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -84,11 +86,17 @@ constant:
       - "SJ"
       - "SK"
   target_crs: 27700 # British National Grid
-
+  # TODO update to new names before running the pipeline again
+  tech_types:
+    individual: "Individual solution"
+    N_GSHP: "Networked GSHP"
```
sofiapinto:
```suggestion
    networked: "Networked GSHP"
```

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -84,11 +86,17 @@ constant:
       - "SJ"
       - "SK"
   target_crs: 27700 # British National Grid
-
+  # TODO update to new names before running the pipeline again
+  tech_types:
```
sofiapinto, line 98: nice! thanks for doing this. With so many changes happening to these names all the time, this becomes increasingly important to have.

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -116,3 +118,84 @@ def load_gdf_spatial_signatures_gb(
...
+    Load specified OS OpenMap Local or Greenspace layer for Great Britain or optionally for a specific grid square. CRS British National Grid (27700).
+
+    Find grid square information at: https://www.ordnancesurvey.co.uk/documents/resources/guide-to-nationalgrid.pdf
+
+    Args:
+        layer (str): name of layer to load. See layer options below.
+        grid_squares (Optional[List[str]]): names of grid squares in OS mapping for regions of Great Britain to be loaded. Default None to load whole GB.
+        **kwargs for geopandas.read_file()
+
+    Layer options:
```
sofiapinto, line 141: To note that I haven't checked if layers are missing from here.

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -116,3 +118,84 @@ def load_gdf_spatial_signatures_gb(
...
+        'surface_water_line',
+        'tidal_boundary',
+        'tidal_water',
+        'woodland'
+
+    Returns:
+        gpd.GeoDataFrame: OS OpenMap Local geometries for specified layer
+    """
+    if not grid_squares:
+        if layer == "greenspace_site":
```
sofiapinto, line 168: what are these greenspace sites?

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -116,3 +118,84 @@ def load_gdf_spatial_signatures_gb(
...
+        if not isinstance(grid_squares, List):
+            grid_squares = [grid_squares]
+
+        # Reformat layer name to how it appears in file name
+        if layer in ["surface_water_area", "surface_water_line"]:
+            layer = "_".join(["SurfaceWater", layer[-4:].title()])
+        else:
+            layer = layer.replace("_", " ").title().replace(" ", "")
+
+        if layer == "GreenspaceSite":
```
sofiapinto, line 190: This one is not in the list above.

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -116,3 +118,84 @@ def load_gdf_spatial_signatures_gb(
...
+        'surface_water_area',
+        'surface_water_line',
+        'tidal_boundary',
+        'tidal_water',
+        'woodland'
+
+    Returns:
+        gpd.GeoDataFrame: OS OpenMap Local geometries for specified layer
+    """
+    if not grid_squares:
```
sofiapinto:
```suggestion
   # Load data for GB
    if not grid_squares:
```

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -116,3 +118,84 @@ def load_gdf_spatial_signatures_gb(
...
+            )
+
+        print(f"Loading OS OpenMap Local - {layer.title()}...")
+        return gpd.read_file(
+            filename=config["data"]["geodata"]["gb_os_openmap_local"],
+            layer=layer,
+            **kwargs,
+        ).drop_duplicates(subset="ID")
+
+    else:
```
sofiapinto, line 180:
```suggestion
    else: # Load data for grid square of list of grid squares
```

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -116,3 +118,84 @@ def load_gdf_spatial_signatures_gb(
...
+
+        files = [file_path.format(square=code, layer=layer) for code in grid_squares]
+
+        gdfs = []
+
+        for file in files:
+            print(f"\nLoading OS OpenMap layer - {layer.title()} file: {file}")
+            gdfs.append(gpd.read_file(file, **kwargs))
+
+        gdf = pd.concat(gdfs)
```
sofiapinto, line 203: The code below might be slightly more memory efficient. I'm not too worried now though, since we typically have a small number of files. Worth thinking about how we would load these in future, with scaling? We can also change it later though.

```suggestion
        gdfs = pd.concat([gpd.read_file(f, **kwargs) for f in files], ignore_index=True)
```

asf_heat_pump_suitability/getters/load_tree_input.py
```diff
@@ -96,4 +96,5 @@ def load_gdf_poi() -> gpd.GeoDataFrame:
         layer="poi_uk",
     ).to_crs("EPSG:4326")
     print(f"POI CRS: {poi.crs}")
-    return poi
+
+    return poi[poi.country == "GB"]
```
sofiapinto: Did you change this because it defaults to UK?

asf_heat_pump_suitability/utils/save_utils.py
```diff
@@ -22,30 +24,52 @@ def save_model_to_pkl_s3(model: BaseEstimator, path: str) -> None:
...
+                f.write(df.to_json(drop_id=True))
+        else:
+            with fsspec.open(path, "wb") as f:
+                df.to_file(f)
     else:
-        raise ValueError(
-            "Save to S3 can only save .parquet or .csv file types."
-            "Please ensure the `path` argument contains one of these file types."
+        with fs.open(path=path, mode="wb") as f:
+            df.to_file(f)
```
sofiapinto: why are we saving if we're raising an error after saying we can't save?

asf_heat_pump_suitability/utils/save_utils.py
```diff
@@ -22,30 +24,52 @@ def save_model_to_pkl_s3(model: BaseEstimator, path: str) -> None:
...
+                "Save to S3 can only save polars DataFrames .parquet or .csv file types."
+                "Please ensure the `path` argument contains one of these file types."
+            )
+    elif isinstance(df, gpd.GeoDataFrame):
+        if file_type == "parquet":
+            df.to_parquet(path)
+        elif file_type == "geojson":
+            df = df.to_crs(epsg=4326)
+            with fsspec.open(path, "w") as f:
+                f.write(df.to_json(drop_id=True))
```
sofiapinto, line 64: why the drop_id?

sofiapinto (commented):
Hey @crispy-wonton , here's the first part of my review, with comments, questions and suggestions related to the following files:

- `base.yaml`
- `load_geodata.py`
- `load_tree_input.py`
- `save_utils.py`

Will now review the actual clustering.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -0,0 +1,402 @@
...
+import numpy as np
+import shapely
+from shapely.geometry import MultiPoint, Polygon, MultiPolygon
+import libpysal
+from asf_heat_pump_suitability import config
+from asf_heat_pump_suitability.utils import save_utils
+from asf_heat_pump_suitability.getters import load_geodata, load_boundaries
+
+N_GSHP = config["constant"]["tech_types"]["N_GSHP"]
+COMMUNAL = config["constant"]["tech_types"]["communal"]
```
sofiapinto: I'd probably load the whole `tech_types` from `config["constant"]` in case the number of categories changes in future.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -0,0 +1,402 @@
...
+    parser.add_argument(
+        "--tech_gdf",
+        help="Path to S3 file containing building footprints with their tech types assigned by the decision tree.",
+        type=str,
+        required=True,
+    )
+
+    parser.add_argument(
+        "--local_authorities",
+        help="Run script for either all of Great Britain; Plymouth only {plymouth}; or Plymouth and 4 similar local authorities {plymouth_similar_cities}; or Plymouth and 5 different local authorities {sampling_areas}. Default to all of GB",
```
sofiapinto: I'd suggest changing this to what i've been using lately:

```suggestion
        help="Local authority or authorities. See base.yaml's `constant` section for options e.g. `plymouth`, `plymouth_similar_cities`, `sampling_areas`, `greater_manchester_las`.",
```

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -0,0 +1,402 @@
...
+    Create ArgumentParser and parse.
+
+    Returns:
+        argparse.Namespace: populated `Namespace`
+    """
+    parser = argparse.ArgumentParser()
+
+    # TODO this is a placeholder and likely to change as the script develops
+    parser.add_argument(
+        "--tech_gdf",
```
sofiapinto: doesn't this `tech_gdf` contain `local_authorities` in the name? if so, we don't need this additional arg.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -0,0 +1,402 @@
...
+    return parser.parse_args()
+
+
+if __name__ == "__main__":
+
+    args = parse_arguments()
+    tech_gdf = (
+        gpd.read_file(args.tech_gdf).to_crs(config["constant"]["target_crs"])
+        # TODO rename column in original dataframe
+        .rename(columns={"1st_most_suitable_solution": "assigned_tech"})
```
sofiapinto: to note that I have done this in the decision tree script.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -0,0 +1,402 @@
...
+    )
+
+    barrier_road_types = [
+        "A Road" "B Road, Collapsed Dual Carriageway",
+        "Minor Road, Collapsed Dual Carriageway",
+        "Primary Road, Collapsed Dual Carriageway",
+        "Motorway",
+        "Motorway, Collapsed Dual Carriageway",
+        "A Road, Collapsed Dual Carriageway",
+    ]
```
sofiapinto: I think we need a doc line above with some indication of the types of roads we're considering.

Also, we should have this in the documentation page when you create one. :)

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -0,0 +1,402 @@
...
+
+    barrier_roads_gdf = roads_gdf[roads_gdf["CLASSIFICA"].isin(barrier_road_types)]
+
+    line_overlays = [barrier_roads_gdf, railways_gdf]
+    line_overlay_gdf = pd.concat([gdf[["geometry"]] for gdf in line_overlays])
+
+    # TODO make more specific for different road types
+    # Add buffer assumed to be width of road / railway (3.5m total - 1.75m either side)
```
sofiapinto, line 403: This is roughly the size of lanes. A and B roads will have widths closer to ~7m, and motorways closer to 30-40m.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -0,0 +1,402 @@
...
+
+
+def extend_edges_gdf(
+    gdf: gpd.GeoDataFrame,
+    boundary: shapely.Polygon | shapely.MultiPolygon,
+    spacing: float = 1.0,
+) -> gpd.GeoDataFrame:
+    """
+    Creates Voronoi polygons around a set of input polygons by extending interpolating additional points along polygon edges
+    to extend Voronoi polygons from.
```
sofiapinto: Let me know if I'm reading this wrong.
```suggestion
    Creates Voronoi polygons around a set of input polygons by interpolating additional points along polygon edges
    to extend Voronoi polygons from.
```

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -0,0 +1,402 @@
...
+        polygon_overlay_gdf (gpd.GeoDataFrame): physical barriers with (Multi)Polygon geometries to separate clusters by.
+
+    Returns:
+        gpd.GeoDataFrame: clusters of building footprints with the same assigned technology, one row per cluster
+    """
+    gdfs = []
+
+    # Create Voronoi polygons and overlay physical barriers
+    for boundary in boundary_gdf["geometry"].unique():
+        voronoi_gdf = extend_edges_gdf(gdf=buildings_gdf, boundary=boundary)
```
sofiapinto, line 91: At this point this `extend_edges_gdf()` is happening at building level right? So what exactly is happening?

I will visualise this separately to have a sense of what's happening, but if you already have a few figures of the before and after, feel free to post them here as I think it's useful.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -0,0 +1,402 @@
...
+
+    Args:
+        gdf (gpd.GeoDataFrame): polygons to create Voronoi polygons around.
+        boundary (shapely.Polygon | shapely.MultiPolygon): boundary to clip Voronoi polygons to.
+        spacing (float): Distance in metres to space interpolating points along polygon edges. Default 1.
+
+    Returns:
+        gpd.GeoDataFrame: Voronoi polygons around the original input polygons. One row per original polygon.
+    """
+    # TODO deal with buildings that cross boundaries
```
sofiapinto, line 159: We could keep those that `intersect`, rather than only those within, to make sure we don't miss any buildings.

We will obviously need to adapt later, depending on how data is displayed on the map (e.g. selecting one LA before visualising vs visualising multiple vs visualising the whole of GB - this last one is very unlikely)

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -0,0 +1,402 @@
...
+    # Create Voronoi polygons and overlay physical barriers
+    for boundary in boundary_gdf["geometry"].unique():
+        voronoi_gdf = extend_edges_gdf(gdf=buildings_gdf, boundary=boundary)
+        cells_gdf = overlay_gdf_physical_barriers(
+            voronoi_gdf=voronoi_gdf,
+            tech_gdf=tech_gdf,
+            line_overlay_gdf=line_overlay_gdf,
+            polygon_overlay_gdf=polygon_overlay_gdf,
+        )
+        gdfs.append(reassign_gdf_communal_networked(cells_gdf))
```
sofiapinto: I'm interested to see how this changes the techs across Plymouth. I think this condition might be too strong, but will visualise it first. The alternative Andy gave yesterday sounds better. Although we need to be careful with all these reassignments, we can easily end up with way too big clusters.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -0,0 +1,402 @@
...
+        line_overlay_gdf (gpd.GeoDataFrame): physical barriers with (Multi)LineString geometries to separate clusters by.
+        polygon_overlay_gdf (gpd.GeoDataFrame): physical barriers with (Multi)Polygon geometries to separate clusters by.
+
+    Returns:
+        gpd.GeoDataFrame: clusters of building footprints with the same assigned technology, one row per cluster
+    """
+    gdfs = []
+
+    # Create Voronoi polygons and overlay physical barriers
+    for boundary in boundary_gdf["geometry"].unique():
```
sofiapinto, line 90: so each boundary would be a local authority? when considering a list of local authorities, wouldn't it make more sense to generate a unique boundary that concatenates all the boundaries?

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -0,0 +1,402 @@
...
+            line_overlay_gdf=line_overlay_gdf,
+            polygon_overlay_gdf=polygon_overlay_gdf,
+        )
+        gdfs.append(reassign_gdf_communal_networked(cells_gdf))
+
+    # Generate final clusters
+    if len(gdfs) > 1:
+        clusters_gdf = pd.concat(gdfs)
+    else:
+        clusters_gdf = gdfs[0]
```
sofiapinto, line 107: i'm guessing this has been coded this way so that we have one gdf per LA? Again, i'd probably go with the approach I noted above.

sofiapinto (commented):
Hi @crispy-wonton,

Beautiful work! Have learnt a lot from reviewing this work.

I've done a first pass review of `cluster.py`. I still need to run the script and also need to visualise how things change across the different steps of the pipeline. Will do that this afternoon or Monday, and will then come back with a final review.

sofiapinto:
A few thoughts that just came to my mind:
- we need to add those lines of code I had in the polygons exploration PR to create a `cluster_id` column at the end
- In the exploration PR, I had added a constain such as: the edges of the voronoi polygon should not be more than 20m away from the closest building, in order to avoid some of the weird clusters that cover outdoor space. I also had a separate one that accounted for the maximum width of a cluster. Curious to see if these are still needed once I look at the final clusters.

asf_heat_pump_suitability/config/base.yaml
sofiapinto, line 288: Flagging that this needs to be removed to avoid confusion. :) we should do a cleanup in a separate issue.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -60,7 +60,8 @@ def generate_gdf_clusters(
             line_overlay_gdf=line_overlay_gdf,
             polygon_overlay_gdf=polygon_overlay_gdf,
         )
-        gdfs.append(reassign_gdf_communal_networked(cells_gdf))
+        # gdfs.append(reassign_gdf_communal_networked(cells_gdf))
```
sofiapinto, line 100: Can you please add a comment for why this is commented out, for future knowledge.

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: yes although in S3 i've currently saved in separate grid squares for the area of interest.

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: agree - let's do this after alpha release. I'd rather not make breaking changes so we can delete this and the scripts that use it at the same time.

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: I'm pretty sure I created this via code and then copied it into here so it should be complete.

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: It can be any of the following:

- Allotments Or Community Growing Space
- Bowling Green
- Cemetery
- Religious Grounds (Populated where there is a significant amount of accessible greenspace; this is defined as more than 500m2 of natural surface within the site.)
- Golf Course
- Other Sports Facility
- Play Space
- Playing Field (Only playing fields that are used by the public at least some of the time are included in the product. School fields, for example, which are entirely enclosed and used only by the school are not included.)
- Public Park Or Garden
- Tennis Court

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: good spot - thank you!

crispy-wonton: oh - actually it's required because `greenspace_site` is in the list, and this is just the same but after the formatting occurs in line 183.

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: The reason I didn't do this is because I wanted to print the loading information for each file. We can change later if it is very slow. Not sure how much speed gains would be as it would still be running a loop.

asf_heat_pump_suitability/getters/load_tree_input.py
crispy-wonton: actually it was unnecessary to change it here because this processing occurs in a later function anyway. I thought I had accidentally removed it from the pipeline before. I'm therefore reverting this line.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: The reason I keep the boundaries separate and process each Local Authority individually is because the boundary is used to define the edges of the voronoi diagram. If the boundaries are too large beyond the LA, then they will create massive Voronoi cells which - a) are possibly more computationally expensive to produce, and b) will start overlapping other UPRNs in other LAs that we didn't include in the loop.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Yes, extend edges is happening at building level. Basically the input here for one row is: a building footprint. The output for the same row is: a single Voronoi cell that was created by expanding the edges out from the building.

E.g. something like this is the output (where the black shapes represent the building footprints):

Image (image taken from [StackOverflow](https://stackoverflow.com/questions/36015521/generating-a-voronoi-diagram-around-2d-polygons))

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Yes, this is so we have one gdf per LA. Another reason to keep it this way is that we risk creating clusters that extend over non-neighbouring LAs if we were to run the code for a subset of LAs that were in different regions of GB.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: you are right

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Yes we could. The reason I haven't done that here yet is because this will generate buildings which belong to two Local Authorities and this is something I want to discuss how to deal with after alpha release when we think about scaling.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: I'm going to come back to this in a separate issue as I looked into the roads documentation further and I need to update this list. Have added a TODO note.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Yes, I wanted to use the minimum road width here so as not to interfere too much with the clusters for the time being. The main purpose of adding roads as a barrier is so that they separate clusters. I had planned to revisit this in the next phase because I think it will require some visual inspection work to make sure a wider buffer wouldn't overlap with buildings.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: This needs to be done in a few places in the repo so I've added a [separate issue](https://github.com/nestauk/asf_heat_pump_suitability/issues/288) to do it all in one go - probably after a lot of the outstanding PRs are merged.

asf_heat_pump_suitability/utils/save_utils.py
crispy-wonton: This just prevents the index being saved into the json file.

asf_heat_pump_suitability/utils/save_utils.py
crispy-wonton: hmm not sure what the heck I was trying to do here...? need to check

crispy-wonton: I think I left this part unfinished for some reason...

asf_heat_pump_suitability/getters/load_geodata.py
sofiapinto: We could add a doc above the if with `#See list of sites here: https://docs.os.uk/os-downloads/products/land-and-terrain-portfolio/os-open-greenspace/os-open-greenspace-technical-specification/code-lists/functionvalue#code-list-functionvalue` or similar

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -0,0 +1,476 @@
...
+- Contains buildings which are assigned the same tech type
+- Contains buildings which are not separated by physical environmental barriers
+
+Contains a script to produce clusters from building footprint polygons with assigned tech types. To run:
+python asf_heat_pump_suitability/pipeline/cluster/cluster.py
+
+Required args:
+--tech_gdf path to geospatial file containing building footprints labelled with assigned tech type
+--local_authorities to specify which local authority / authorities to run the script for
+"""
```
sofiapinto, line 12: Missing info about --save

asf_heat_pump_suitability/pipeline/cluster/cluster.py
sofiapinto: I think this is now the only script where this hasn't been actioned, so if I were you i'd action the change here. But feel free to do in a separate PR. But I'd say it's useful that we do it asap, because for someone less familiar with the code, they wouldn't know which file to pass as an argument for the tech_gdf param. But even us, we can easily forget which one to pass here/or pass the wrong one (old names of data files...)

It's also a fairly simple one, so should only be a couple of lines change:
- Removing tech_gdf from the parse_arguments()
- Adding buildings tech gdf to base.yaml. Should actually already live in dev's base.yaml:
```
output:
  dataset:
    .....
    # Residential buildings with their most suitable low carbon heat technology group
    buildings_most_suitable_tech: "s3://asf-heat-pump-suitability/local_heat_planning/outputs/{local_authorities}/{local_authorities}_buildings_most_suitable_tech.parquet"

```
- Read path from config rather than from args in cluster.py.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
sofiapinto: Fair. We need to add a caveat (if not already there) to the methodology doc saying that we might be missing buildings at the edge of the LA.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
sofiapinto: I know we can effectively run it for any number of LAs in a list, but i'd say the latest pipelines have been designed to be run for one LA or a group of neighbouring LAs/ LAs within the same CA.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
sofiapinto: Okay. We just need to make sure this is appropriately caveated in the documentation,

sofiapinto (commented):
Hey @crispy-wonton ,

I have run 
`python asf_heat_pump_suitability/pipeline/cluster/cluster.py --local_authorities "plymouth" --tech_gdf "s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth/plymouth_buildings_most_suitable_tech.parquet" --save`

and it works well.

I have a few additional comments about documentation.

Looking at the differences between this and the older approach: this new one creates fewer clusters but it creates a much heavier file.
Old: 43.9 MB (4502 clusters)
New: 519.4 MB (1811 clusters)

When I try to open the new one I also get the error below, meaning I'm not even able to load the file.

```
DataSourceError: Failed to read GeoJSON data; At line 1, character 118731745: GeoJSON object too complex/large. You may define the OGR_GEOJSON_MAX_OBJ_SIZE configuration option to a value in megabytes to allow for larger features, or 0 to remove any size limit.
```

My guess is that the polygons are too detailed and need to be simplified.

I know you were keen to merge this PR, since it's impacting a lot of other PRs being merged. I don't oppose to doing that after you action the simple requests I left throughout + dealing with conflicts with dev. But we do need to investigate this further afterwards (e.g. experimenting with simplifying polygons).

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: Good idea

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: thanks!

crispy-wonton:
@sofiapinto I have addressed your comments and added the caveat to the methodology document where I've tagged you for review.

I have looked into the file size issue where the productionised pipeline is creating files more than 10x the size of the [notebook version](https://github.com/nestauk/asf_heat_pump_suitability/blob/dev/asf_heat_pump_suitability/research/exploratory/voronoi_polygons_exploration/first_pass_merging_polygons.py). I'm unable to find any bugs in the pipeline. I have tested running the script and it runs all the way through well for Plymouth. 

What I can see is that there appear to be some very large complex clusters created with many holes. I believe the complexity created by these interior holes is responsible for the large file size. I've tried the following to address this:
- Use all roads instead of only A, B roads and motorways as environmental barriers. This made the clusters smaller and created more clusters - ~1800 -> 4800. However, this did not significantly reduce the file size (550MB -> 510MB).
- In addition, I simplified the final clusters to a tolerance of 5 metres. This combination reduced the file size to ~20% of the original ~550MB, i.e. ~150MB. However, this is still over 3x larger than the file produced by the notebook.
- I thought the main complexity remaining is coming from the inclusion in this pipeline of a step to exclude non-domestic buildings from the cluster space. I.e. we are overlaying non-domestic buildings and removing the area they cover from cluster polygons. This was not in the original notebook implementation. After a test run to exclude this step and use all road types as barriers, but WITHOUT simplification of polygons, the resulting file is 412MB so that didn't really help, although perhaps with simplification it would be the smallest file. However, I didn't test this and I also don't know if we want our cluster polygons to cover non-domestic buildings.

We can explore reducing this further in a separate PR if you're happy with the current status.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -161,22 +165,61 @@ def extend_edges_gdf(
...
+    """
+    # Create a buffered polygon for all polygons
+    buffered_gdf = polygon_gdf[[id_col, "geometry"]].copy()
+    buffered_gdf["geometry"] = buffered_gdf.geometry.buffer(buffer)
+
+    # Clip Voronoi cells to the defined buffer area if they are larger than it by calculating intersections
+    clipped_gdf = voronoi_gdf.overlay(buffered_gdf, how="intersection")
+    # Filter clipped cells (intersections) to ensure each polygon's Voronoi cell is clipped only by its own buffer
+    return (
+        clipped_gdf[clipped_gdf[f"{id_col}_1"] == clipped_gdf[f"{id_col}_2"]]
```
sofiapinto, line 258: I don't really understand this step. Could you clarify?

sofiapinto:
Hey @crispy-wonton , thank you for all of this work on this PR so far. I have reviewed the latest code & also ran it.

> 
> 
> @sofiapinto I have addressed your comments and added the caveat to the methodology document where I've tagged you for review.
> 
> I have looked into the file size issue where the productionised pipeline is creating files more than 10x the size of the [notebook version](https://github.com/nestauk/asf_heat_pump_suitability/blob/dev/asf_heat_pump_suitability/research/exploratory/voronoi_polygons_exploration/first_pass_merging_polygons.py). I'm unable to find any bugs in the pipeline. I have tested running the script and it runs all the way through well for Plymouth.

Same. When I did review it before this, I also didn't find any bugs. I think this is all due increased complexity of the polygons.

> What I can see is that there appear to be some very large complex clusters created with many holes. I believe the complexity created by these interior holes is responsible for the large file size. I've tried the following to address this:
> 
> Use all roads instead of only A, B roads and motorways as environmental barriers. This made the clusters smaller and created more clusters - ~1800 -> 4800. However, this did not significantly reduce the file size (550MB -> 510MB).

I know we've been thinking about roads as barrier types, in terms of feasibility. But maybe we should think about them as what defines a neighbourhood?

I don't know the exact definition of the road names below, but thinking of the neighbourhood perspective, which ones should we consider as barriers? All of them except "Minor road"? Worth checking if we leave minor roads out of the barriers - do we still get those massive clusters?

```
[
         "A Road", 
         "B Road, Collapsed Dual Carriageway",
         "Minor Road, Collapsed Dual Carriageway",
         "Primary Road, Collapsed Dual Carriageway",
         "Motorway",
         "Motorway, Collapsed Dual Carriageway",
         "A Road, Collapsed Dual Carriageway",
    ]
```

I think in my opinion, at this point, I'd rather have more clusters that users can merge and let us know which ones they'd like to merge for us to then optimise the road types, than have huge clusters that span multiple neighbourhoods and feel hard to action up on. 

But we can sense check with the rest of the team, especially Andy.

> In addition, I simplified the final clusters to a tolerance of 5 metres. This combination reduced the file size to ~20% of the original ~550MB, i.e. ~150MB. However, this is still over 3x larger than the file produced by the notebook.

Great! I have also checked that the simplification and tolerance value make sense for the specific CRS we're using and it does.

We need to let Simon know that this is the case, as he is currently applying a simplification on the browser when the data loads. We can also ask Simon about which tolerance value he is using - maybe he was even using a higher value and it might be fine?

> I thought the main complexity remaining is coming from the inclusion in this pipeline of a step to exclude non-domestic buildings from the cluster space. I.e. we are overlaying non-domestic buildings and removing the area they cover from cluster polygons. This was not in the original notebook implementation. After a test run to exclude this step and use all road types as barriers, but WITHOUT simplification of polygons, the resulting file is 412MB so that didn't really help, although perhaps with simplification it would be the smallest file. However, I didn't test this and I also don't know if we want our cluster polygons to cover non-domestic buildings.

I can't see where in the code we're actually doing this?

```
# Load and transform physical barriers for clusters
    line_overlay_gdf = load_tranform_gdf_linestring_barriers(grid_squares)
    polygon_overlay_gdf = load_transform_gdf_polygon_barriers(grid_squares)

    # Generate clusters
    clusters_gdf = generate_gdf_clusters(
        buildings_gdf=buildings_gdf,
        boundary_gdf=boundary_gdf,
        tech_gdf=tech_gdf,
        line_overlay_gdf=line_overlay_gdf,
        polygon_overlay_gdf=polygon_overlay_gdf,
    )
```

As I don't think `polygon_overlay_gdf` includes the polygons of non-domestic buildings. Can you point me to the correct part of the code where this is happening?

If this is indeed included in the code, I think this might be the cause of the increased file size as I can imagine it will add a lot of small wholes to clusters?

My two takes are:
- I think more importantly than creating wholes where non domestic buildings are, is **1)** us removing as many UPRNs from the list of domestic buildings (which you are doing in a separate PR) and **2)** not counting non-domestic UPRNs as part of the cluster (which we should be doing too, even if don't have wholes where the domestic buildings are). In essence, I don't think it's that problematic that the polygons cover non domestic buildings, as long as we don't count them in the number of UPRNs/remaining contextual features. As I have told you before, when we think of anchor loads specifically, I think it's actually helpful to have the anchor load as part of the cluster (inside the polygon, not as part of the counts) - but this is my take. I think this is one to also check with the rest of the team.
        - I also don't have a sense of how much this is happening at the moment though? My sense is that with adding more roads as barriers, then each clusters spans a small area and so this should be happening less?
- We should check with Simon about the file sizes, so we can then handle it from our side, as in what would be the ideal maximum.

> We can explore reducing this further in a separate PR if you're happy with the current status.

I am not currently able to open the file generated by this though?

```
local_authorities = "plymouth"
tolerance = 5
new_version = gpd.read_file(
f"s3://asf-heat-pump-suitability/local_heat_planning/outputs/{local_authorities}/{local_authorities}_clustered_tech_polygons_simplified_{tolerance}m.geojson",
        driver="GeoJSON",
    )
```

Am I doing anything wrong when loading?

```
DataSourceError: '/vsis3/asf-heat-pump-suitability/local_heat_planning/outputs/plymouth/plymouth_clustered_tech_polygons_simplified_5m.geojson' does not exist in the file system, and is not recognized as a supported dataset name.
```

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: This is the same as in your original code. It is required to ensure that the buffer of building A is only used to clip the Voronoi cell of building A. I.e. building B's buffer will not clip building A's Voronoi cell.

sofiapinto:
Update: I was able to load the file!

asf_heat_pump_suitability/pipeline/cluster/cluster.py
sofiapinto: ah! got it! thanks

crispy-wonton:
> I know we've been thinking about roads as barrier types, in terms of feasibility. But maybe we should think about them as what defines a neighbourhood?

I think this is a good idea! I like the way you've put this :) let's use that framing from now on.
 
> I don't know the exact definition of the road names below, but thinking of the neighbourhood perspective, which ones should we consider as barriers? All of them except "Minor road"? Worth checking if we leave minor roads out of the barriers - do we still get those massive clusters?
> 
> ```
> [
>          "A Road", 
>          "B Road, Collapsed Dual Carriageway",
>          "Minor Road, Collapsed Dual Carriageway",
>          "Primary Road, Collapsed Dual Carriageway",
>          "Motorway",
>          "Motorway, Collapsed Dual Carriageway",
>          "A Road, Collapsed Dual Carriageway",
>     ]
> ```

The list you've posted here is the list of barrier roads that I added in this PR originally to separate clusters. However, I have now commented this part out as, per my comment & findings, filtering the roads to these types only made the clusters too big. Removing 'minor roads' will just make the clusters even bigger. Since commenting this part out, we now use all roads as barriers. As discussed in the stand up today, we can filter this to a more specific set of roads if desired. I will post in slack about it. 

> > I thought the main complexity remaining is coming from the inclusion in this pipeline of a step to exclude non-domestic buildings from the cluster space. I.e. we are overlaying non-domestic buildings and removing the area they cover from cluster polygons. This was not in the original notebook implementation. After a test run to exclude this step and use all road types as barriers, but WITHOUT simplification of polygons, the resulting file is 412MB so that didn't really help, although perhaps with simplification it would be the smallest file. However, I didn't test this and I also don't know if we want our cluster polygons to cover non-domestic buildings.
> 
> I can't see where in the code we're actually doing this?

This happens in [line 245-247](https://github.com/nestauk/asf_heat_pump_suitability/blob/17bcd57754dc7c8ae927c423686d2753935e6d82/asf_heat_pump_suitability/pipeline/cluster/cluster.py#L245) of `cluster.py` where the Voronoi cells are filtered to those containing domestic properties only.

> If this is indeed included in the code, I think this might be the cause of the increased file size as I can imagine it will add a lot of small wholes to clusters?

I did test commenting out this section and running the script, plus using all roads as barriers. As per my comment above, the resulting file is 412MB so that didn't really help. However, I was surprised by that so it's possible I overlooked another change I should have made elsewhere. I didn't use simplification in that instance.
 
I see you can open the file - that's great! :) Have also tested again and it works for me too :)

Let me know if you are happy to merge and then we can do further exploration on the file size issue in a separate PR.

sofiapinto (dismissed):

sofiapinto (approved):

PR #274, opened 2026-03-11
270 refactor anchor notebook

### LINKED ISSUE #270: Refactor anchor properties notebook to clustering pipeline
Refactor key functions from the anchor properties notebook into the `/pipeline/cluster/cluster.py ` module and apply the functions in the script at the bottom of the module to reassign tech type in buildings surrounding anchor properties.

simran-dave:
---

# Description

Add function to reassign tech type based on anchor property proximity in the `pipeline/cluster/cluster.py` script.

Fixes #270 

# Instructions for Reviewer

run  `python asf_heat_pump_suitability/pipeline/cluster/cluster.py --tech_gdf s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_building_most_suitable_tech.geojson --local_authorities plymouth`

I think the reassignment logic is working correctly as the same proportion of buildings is being reassigned as was in the exploratory notebook, but maybe double check. In lines 117 - 119 I then replaced the old assigned tech column with this new column, which contained the old tech types for buildings outside the radius and the new for buildings in the radius. I did this using building ID since the length of my dataframe was slightly longer than the length of the voronoi polygons dataframe. Does this part of the code make sense? It's the bit I'm most unsure about. 

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/config/base.yaml
```diff
@@ -92,6 +92,7 @@ constant:
     N_GSHP: "Networked GSHP"
     communal: "Communal solutions"
     heat_networks: "District heat network"
+  radius: 50
```
crispy-wonton: suggestion: to rename to something a bit more descriptive e.g. `anchor_radius`. Also, it's probably good to add an annotation to indicate that this is a distance in metres

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -26,25 +26,63 @@
...
+
+tech_mapping = {
+    "Individual solution": "Communal solutions",
+    "District heat network": "District heat network",
+    "Individual solution or Networked GSHP": "Communal solutions",
+    "Networked GSHP": "Communal solutions",
+    "Communal solutions": "Communal solutions",
+    "Individual solution or District heat network": "District heat network",
+}
+
```
crispy-wonton: todo: please can we use upper case for global variables, i.e. `ANCHOR_RADIUS`, `ANCHOR_CATEGORIES` etc.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -26,25 +26,63 @@
...
+]
+
+tech_mapping = {
+    "Individual solution": "Communal solutions",
+    "District heat network": "District heat network",
+    "Individual solution or Networked GSHP": "Communal solutions",
+    "Networked GSHP": "Communal solutions",
+    "Communal solutions": "Communal solutions",
+    "Individual solution or District heat network": "District heat network",
+}
```
crispy-wonton: issue: please can you update these to use the names from the config. You can define `INDIVIDUAL` and `HEAT_NETWORK` global variables above where I've defined `COMMUNAL` and `N_GSHP` and then use those to recreate the dictionary here. The reason for this is that the tech names have already changed twice. If they change again, this means we can just update the config variables and the names will be updated across all scripts.

issue: we also need to check if the two combined categories should be in here.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -26,25 +26,63 @@
...
+
 
 def generate_gdf_clusters(
     buildings_gdf: gpd.GeoDataFrame,
     boundary_gdf: gpd.GeoDataFrame,
     tech_gdf: gpd.GeoDataFrame,
     line_overlay_gdf: gpd.GeoDataFrame,
     polygon_overlay_gdf: gpd.GeoDataFrame,
+    poi_gdf: gpd.GeoDataFrame,
+    important_building_gdf: gpd.GeoDataFrame,
```
crispy-wonton: suggestion: I think it would be better if we have one single anchor property gdf which is called here. So I would suggest creating a separate function which combines them to create the anchor property building gdf, and then input the output of that function here.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -26,25 +26,63 @@
...
+    radius: float,
 ) -> gpd.GeoDataFrame:
     """
     Generate clusters of building footprints, where one cluster:
     - Contains buildings which are assigned the same tech type
     - Contains buildings which are not separated by physical environmental barriers
+    - Buildings within a given radius of an anchor as assigned a tech type of 'Communal solutions', unless in a district heat network
```
```suggestion
    Buildings within a given radius of an anchor are assigned a tech type of 'Communal solutions', unless in a district heat network.
```

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -26,25 +26,63 @@
...
 
     Args:
         buildings_gdf (gpd.GeoDataFrame): all building footprint polygons for area of interest, including domestic and non-domestic.
         boundary_gdf (gpd.GeoDataFrame): boundaries of Local Authorities to generate clusters for.
         tech_gdf (gpd.GeoDataFrame): domestic building footprints with assigned tech types.
         line_overlay_gdf (gpd.GeoDataFrame): physical barriers with (Multi)LineString geometries to separate clusters by.
         polygon_overlay_gdf (gpd.GeoDataFrame): physical barriers with (Multi)Polygon geometries to separate clusters by.
+        poi_gdf (gpd.GeoDataFrame): anchor properties dataframe taken from POI data, with pount geometries
```
```suggestion
        poi_gdf (gpd.GeoDataFrame): anchor properties dataframe taken from POI data, with point geometries
```

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -26,25 +26,63 @@
...
+        important_buildings_gdf (gpd.GeoDataFrame): important building footprint polygons
+        radius (float): radius around anchor property within which communal solutions should be assigned
```
crispy-wonton: todo: specify that the radius is in meters. 

thought: Also, should we rename this `distance`? I'm thinking that it's not technically a radius, it's just the distance from the edge of the building footprint. Either way, perhaps the docstring can be updated to reflection this.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -68,6 +106,18 @@ def generate_gdf_clusters(
...
+        tech_gdf=tech_gdf,
+        building_footprint_gdf=buildings_gdf,
+        poi_gdf=poi_gdf,
+        important_building_gdf=important_building_gdf,
+        radius=radius,
+    )
+    clusters_gdf["assigned_tech"] = clusters_gdf.ID.map(
+        dict(reassigned_gdf[["ID", "assigned_tech"]].values)
+    )
+
```
crispy-wonton: issue: let's reassign directly to the original `clusters_gdf` dataframe here. Since the function will just return `clusters_gdf` alone, there's not much point in assigning to a separate gdf first.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -331,6 +381,69 @@ def reassign_gdf_communal_networked(
     )
 
 
+def reassign_gdf_anchor_properties(
```
crispy-wonton: nitpick: perhaps the function name could reflect that we are reassigning other buildings based on the anchor properties, rather than reassigning something to do with the anchor properties themselves. E.g. `reassign_gdf_near_anchor_properties` or something better.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -331,6 +381,69 @@ def reassign_gdf_communal_networked(
...
+    building_footprint_gdf: gpd.GeoDataFrame,
+    poi_gdf: gpd.GeoDataFrame,
+    important_building_gdf: gpd.GeoDataFrame,
+    radius: float,
+) -> gpd.GeoDataFrame:
+    """
+    Reassign building tech type to communal if within a given radius of an anchor load property, unless in district heat network
+
+    Args:
+        tech_gdf (gpd.GeoDataFrame): dataframe with most suitable tech types assigned to residential buildings
```
```suggestion
        tech_gdf (gpd.GeoDataFrame): domestic building footprints with assigned tech types.
```
nitpick: for consistency with the existing docstrings.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -331,6 +381,69 @@ def reassign_gdf_communal_networked(
...
+        building_footprint_gdf (gpd.GeoDataFrame): dataframe with building footprint polygons
```
```suggestion
        buildings_gdf (gpd.GeoDataFrame): dataframe with building footprint polygons
```
nitpick: for consistency with existing variable names. Will need to be updated below.

issue: can you also specify in the docstring which building footprints these are - i.e. domestic only or all? For reference, my docstring above is: `all building footprint polygons for area of interest, including domestic and non-domestic.`

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -331,6 +381,69 @@ def reassign_gdf_communal_networked(
...
+        poi_gdf (gpd.GeoDataFrame): dataframe with anchor properties and their point geometries
+        important_building_gdf (gpd.GeoDataFrame): dataframe with important buildings and their polygon geometries
+        radius (float): radius around an achor, within which buildings will be assigned tech type of 'communal solutions'
```
```suggestion
        radius (float): radius around an anchor, within which buildings will be assigned tech type of 'communal solutions' unless they are already assigned to district heat networks
```

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -331,6 +381,69 @@ def reassign_gdf_communal_networked(
...
+    Returns:
+        reassigned_tech (gpd.GeoDataFrame): dataframe with assigned_tech column now reading communal if property is in radius of anchor, unless in district heat network
```
```suggestion
        gpd.GeoDataFrame: dataframe with `assigned_tech` column now reading communal if property is in radius of an anchor property, unless in district heat network
```
issue: only return type required here

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -331,6 +381,69 @@ def reassign_gdf_communal_networked(
...
+    ).drop("index_right", axis=1)
+
+    # add POI and important building lists together and remove duplicate buildings
+    all_anchors_with_footprint = pd.concat(
+        [anchors_with_footprint, important_buildings_filtered]
+    )
+    all_anchors_with_footprint["geometry"] = all_anchors_with_footprint.normalize()
+    all_anchors_with_footprint = all_anchors_with_footprint.drop_duplicates(
+        ["geometry"]
+    )
```
crispy-wonton, line 487: issue: I would put all of this into a separate function. I might even suggest you use a `load_transform` function so that the function loads the required datasets directly and then transforms and combines them. This will make it easier for us to debug any issues as it keeps the separate tasks in discrete functions.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -331,6 +381,69 @@ def reassign_gdf_communal_networked(
...
+        radius (float): radius around an achor, within which buildings will be assigned tech type of 'communal solutions'
+    Returns:
+        reassigned_tech (gpd.GeoDataFrame): dataframe with assigned_tech column now reading communal if property is in radius of anchor, unless in district heat network
+    """
+
+    # select anchors out of important building gdf using anchor_categories list
+    # anchor categories list is defined at start of script
+    important_buildings_filtered = important_building_gdf[
+        important_building_gdf["CLASSIFICA"].isin(anchor_categories)
+    ]
```
crispy-wonton: suggestion: when creating a new function with this code, I would recommend setting `anchor_properties` as an argument and then using the global variable as a default. This will make it more flexible to use in other scripts and notebooks if we want to do any exploration of different anchor property types.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -331,6 +381,69 @@ def reassign_gdf_communal_networked(
...
+        poi_gdf (gpd.GeoDataFrame): dataframe with anchor properties and their point geometries
+        important_building_gdf (gpd.GeoDataFrame): dataframe with important buildings and their polygon geometries
+        radius (float): radius around an achor, within which buildings will be assigned tech type of 'communal solutions'
+    Returns:
+        reassigned_tech (gpd.GeoDataFrame): dataframe with assigned_tech column now reading communal if property is in radius of anchor, unless in district heat network
+    """
+
+    # select anchors out of important building gdf using anchor_categories list
+    # anchor categories list is defined at start of script
+    important_buildings_filtered = important_building_gdf[
```
```suggestion
    important_building_gdf = important_building_gdf[
```
suggestion: i would just assign back to same gdf to save on memory. These datasets could be quite big if we are loading them for more than one local authority.
 
I would generally suggest reusing variable names where you no longer need the original variable in this function (and generally) to save on memory. E.g. below, you can use `tech_gdf` instead of `reassigned_tech_gdf`.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -331,6 +381,69 @@ def reassign_gdf_communal_networked(
...
+
+    # add building footprint data to POI anchor properties so geometry isn't just a point
+    anchors_with_footprint = (
+        building_footprint_gdf.sjoin(poi_gdf, how="inner", predicate="contains")
+    ).drop("index_right", axis=1)
+
+    # add POI and important building lists together and remove duplicate buildings
+    all_anchors_with_footprint = pd.concat(
+        [anchors_with_footprint, important_buildings_filtered]
+    )
```
crispy-wonton: suggestion: you might want to retain only common and required columns here.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -331,6 +381,69 @@ def reassign_gdf_communal_networked(
...
+
+    # Spatial join to find nearest anchor for every building
+    reassigned_tech = tech_gdf.sjoin_nearest(
+        all_anchors_with_footprint[["geometry"]],
+        how="left",
+        max_distance=radius,
+        distance_col="distance_m",
+    ).drop("index_right", axis=1)
+    reassigned_tech["old_tech"] = reassigned_tech["assigned_tech"]
```
crispy-wonton: nitpick: I think we probably don't want to keep this. @sofiapinto what do you think? Do you want to retain the original assigned technology after reassignment, or just replace directly with the new one?

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -331,6 +381,69 @@ def reassign_gdf_communal_networked(
...
+
+    # distance_m column now reads a number (distance from anchor) for all buildings within radius of anchor, and NaN for all outside of that radius
+    # if distance column is not NaN (i.e. building is within the radius of an anchor), reassign tech type according to the map
+    reassigned_tech[(reassigned_tech["distance_m"]).notna()] = reassigned_tech[
+        (reassigned_tech["distance_m"]).notna()
+    ].replace({"assigned_tech": tech_mapping})
+
+    # add column with True if near anchor, False if not
+    reassigned_tech["near_anchor"] = np.where(
```
```suggestion
    reassigned_tech["within_{radius}m_from_anchor_property"] = np.where(
```
nitpick: it would be helpful to make this more descriptive as it will be in a bigger dataframe at the end.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -331,6 +381,69 @@ def reassign_gdf_communal_networked(
...
+        max_distance=radius,
+        distance_col="distance_m",
+    ).drop("index_right", axis=1)
+    reassigned_tech["old_tech"] = reassigned_tech["assigned_tech"]
+
+    # distance_m column now reads a number (distance from anchor) for all buildings within radius of anchor, and NaN for all outside of that radius
+    # if distance column is not NaN (i.e. building is within the radius of an anchor), reassign tech type according to the map
+    reassigned_tech[(reassigned_tech["distance_m"]).notna()] = reassigned_tech[
+        (reassigned_tech["distance_m"]).notna()
+    ].replace({"assigned_tech": tech_mapping})
```
crispy-wonton: question: is it possible to use `np.where` here as well? Otherwise you're reassigning on a subframe of the dataframe.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -380,6 +493,14 @@ def parse_arguments() -> argparse.Namespace:
         layer="building", grid_squares=grid_squares
     )
 
+    poi_gdf = gpd.read_file(
+        "s3://asf-heat-pump-suitability/dump/hack_day/anchor_properties_gb.geojson"
```
crispy-wonton: issue: please can you move this file to this folder on S3 and update the file naming according to the style guide: `s3://asf-heat-pump-suitability/local_heat_planning/inputs/processed/`. Then, please could you put the file path into `config/base.yaml` under `data` `geodata` `processed` to then call here.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -68,6 +106,18 @@ def generate_gdf_clusters(
...
+    # anchor property tech reassignment
+    reassigned_gdf = reassign_gdf_anchor_properties(
+        tech_gdf=tech_gdf,
+        building_footprint_gdf=buildings_gdf,
+        poi_gdf=poi_gdf,
+        important_building_gdf=important_building_gdf,
+        radius=radius,
+    )
+    clusters_gdf["assigned_tech"] = clusters_gdf.ID.map(
+        dict(reassigned_gdf[["ID", "assigned_tech"]].values)
```
```suggestion
        reassigned_gdf.set_index("ID").to_dict()["assigned_tech"])
```
makes it slightly more readable

crispy-wonton (changes requested):
hi @simran-dave :)

Thanks for putting this together so quickly! I think your code is readable and easy to follow.

I have left some comments, the main important ones are about making the code more flexible and modular - e.g. using the config more and extracting separate tasks into discrete functions. There is also some suggestions to help conserve memory.

Other comments are smaller and more nitpicky to adhere to style guidelines and have consistency within the repository.

Regarding your comment in the PR description, I've checked the logic of the way you assign the technology. I think it makes sense, I can't see an issue. I can see there ends up being some duplicate building IDs and also that, as you've said, some buildings are lost. I've looked into this and isolated the cause to the `overlay_gdf_physical_barriers` function in my code. Thanks for checking this and flagging - it highlighted an issue I didn't expect or realise was there. Basically there seems to be some overlay barrier polygons which actually overlay buildings...strange. No action required from you on that, I'll look further into it.

asf_heat_pump_suitability/config/base.yaml
sofiapinto: In addition to what @crispy-wonton says above, I'd add an indication of how this is being used as we're adding it to `base.yaml` which covers the whole repo.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -26,25 +26,63 @@
...
+    "Higher or University Education",
+    "Special Needs Education",
+    "Medical Care Accommodation",
+    "Non State Primary Education",
+    "Non State Secondary Education",
+    "Art Gallery",
+    "Police Station",
+    "Hospice",
+    "Airport",
+]
```
sofiapinto: I believe this is used across other parts of the pipeline, like the domestic properties identification? 

I'll leave it up to @crispy-wonton to decide what to do, but if we do have this across multiple pipelines, then it should live somewhere more general like `base.yaml`. So that we don't have different versions of what we consider to be anchor properties across the repo.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -26,25 +26,63 @@
...
+
+tech_mapping = {
+    "Individual solution": "Communal solutions",
+    "District heat network": "District heat network",
+    "Individual solution or Networked GSHP": "Communal solutions",
```
sofiapinto: you can remove this line, we deal with it earlier in the pipeline

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -26,25 +26,63 @@
...
+    "Networked GSHP": "Communal solutions",
+    "Communal solutions": "Communal solutions",
+    "Individual solution or District heat network": "District heat network",
```
sofiapinto: you can remove this, we deal with it earlier in the pipeline

asf_heat_pump_suitability/pipeline/cluster/cluster.py
sofiapinto: the combined categories don't need to be here, they are dealt with earlier in the pipeline - I'll put the script out for review soon, but it's just after applying the decision tree.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -68,6 +106,18 @@ def generate_gdf_clusters(
...
         clusters_gdf = gdfs[0]
 
+    # anchor property tech reassignment
+    reassigned_gdf = reassign_gdf_anchor_properties(
+        tech_gdf=tech_gdf,
+        building_footprint_gdf=buildings_gdf,
+        poi_gdf=poi_gdf,
+        important_building_gdf=important_building_gdf,
+        radius=radius,
+    )
```
sofiapinto: As discussed in a previous sprint review, I believe this should happen prior to the clustering. So when we consider the radius around an anchor property, we should be looking for buildings close to that property, rather than clusters.

sofiapinto: Do let me know if I misunderstood this and it's actually happening at building level, but I believe it's not? The clustering happens in the previous lines.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
sofiapinto: I think we can replace directly with the new one. But we do need to see the impact of reassignments in changes of tech. We can do that separately though.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
sofiapinto: where's the code that generated this geojson?

sofiapinto (commented):
Hi @simran-dave , thanks for putting this together so quickly!

I haven't yet been able to run the code or visualise how this addition affects the clusters, but added a few additional comments based on reading the code.

My main takes are:
- The reassignment should be happening prior to clustering (the documentation says so, but by looking at the code it seems to be happening after?)
- I agree with @crispy-wonton on creating a few additional functions, it will increase readability of the code.

I will come back to this later to run the code and explore/visualise how this affects the technologies.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
simran-dave: It's slightly different to the list used for domestic property identification. I think this particular list is only used in this script

asf_heat_pump_suitability/pipeline/cluster/cluster.py
simran-dave: yeah makes sense, have rewritten using np.where

asf_heat_pump_suitability/pipeline/cluster/cluster.py
simran-dave: I'm not sure where the code itself is but this is a file Roisin made for the hack day, it's the POI buildings filtered to a list of suitable anchor properties using (I believe) chatGPT, similar to the process used to filter to non-domestic POI buildings

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -25,10 +25,12 @@
...
 N_GSHP = config["constant"]["tech_types"]["N_GSHP"]
 COMMUNAL = config["constant"]["tech_types"]["communal"]
+INDIVIDUAL = config["constant"]["tech_types"]["individual"]
+HEAT_NETWORK = config["constant"]["tech_types"]["heat_networks"]
 
-radius = config["constant"]["radius"]
+ANCHOR_RADIUS = config["constant"]["anchor_radius"]
 
-anchor_categories = [
+ANCHOR_CATEGORIES = [
```
crispy-wonton, line 28: TODO: we need to update these based on our conversation yesterday. I think let's do it in a separate PR. Can you please make an issue?

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -64,23 +64,22 @@ def generate_gdf_clusters(
...
-    important_building_gdf: gpd.GeoDataFrame,
+    combined_anchor_gdf: gpd.GeoDataFrame,
     radius: float,
 ) -> gpd.GeoDataFrame:
     """
     Generate clusters of building footprints, where one cluster:
     - Contains buildings which are assigned the same tech type
     - Contains buildings which are not separated by physical environmental barriers
-    - Buildings within a given radius of an anchor as assigned a tech type of 'Communal solutions', unless in a district heat network
+    - Buildings within a given radius of an anchor are assigned a tech type of 'Communal solutions', unless in a district heat network.
```
crispy-wonton: issue: it's now just buildings which were assigned N-GSHP solutions right?

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -64,23 +64,22 @@ def generate_gdf_clusters(
...
     Args:
         buildings_gdf (gpd.GeoDataFrame): all building footprint polygons for area of interest, including domestic and non-domestic.
         boundary_gdf (gpd.GeoDataFrame): boundaries of Local Authorities to generate clusters for.
         tech_gdf (gpd.GeoDataFrame): domestic building footprints with assigned tech types.
         line_overlay_gdf (gpd.GeoDataFrame): physical barriers with (Multi)LineString geometries to separate clusters by.
         polygon_overlay_gdf (gpd.GeoDataFrame): physical barriers with (Multi)Polygon geometries to separate clusters by.
-        poi_gdf (gpd.GeoDataFrame): anchor properties dataframe taken from POI data, with pount geometries
+        poi_gdf (gpd.GeoDataFrame): anchor properties dataframe taken from POI data, with point geometries
         important_buildings_gdf (gpd.GeoDataFrame): important building footprint polygons
         radius (float): radius around anchor property within which communal solutions should be assigned
```
crispy-wonton: issue: missing specification that radius is in metres

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -381,67 +435,86 @@ def reassign_gdf_communal_networked(
...
-    important_building_gdf: gpd.GeoDataFrame,
-    radius: float,
+def load_transform_anchor_property_gdfs(
+    buildings_gdf: gpd.GeoDataFrame,
+    grid_squares: Optional[List[str]],
+    anchor_categories=ANCHOR_CATEGORIES,
 ) -> gpd.GeoDataFrame:
     """
-    Reassign building tech type to communal if within a given radius of an anchor load property, unless in district heat network
+    Load data from POI and important buildings lists, select buildings using anchor proerty categoriesm, and combine the resultant dataframes
```
crispy-wonton: todo: fix typo in 'categories'

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -381,67 +435,86 @@ def reassign_gdf_communal_networked(
...
+        Find grid square information at: https://www.ordnancesurvey.co.uk/documents/resources/guide-to-nationalgrid.pdf
+        anchor_categories (Optional[List[str]]): list of anchor properties to filter important buildings list by. Defaults to ANCHOR_CATEGORIES
     """
-
     # select anchors out of important building gdf using anchor_categories list
     # anchor categories list is defined at start of script
-    important_buildings_filtered = important_building_gdf[
+
+    poi_gdf = gpd.read_file(
+        "s3://asf-heat-pump-suitability/local_heat_planning/inputs/processed/poi_anchor_properties_gb.geojson"
```
crispy-wonton: todo: please could you move this file path to the `config` and call it here.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -381,67 +435,86 @@ def reassign_gdf_communal_networked(
...
+    return combined_anchor_gdf
+
 
+def reassign_gdf_near_anchor_properties(
+    tech_gdf: gpd.GeoDataFrame,
+    combined_anchor_gdf: gpd.GeoDataFrame,
+    radius: float,
+) -> gpd.GeoDataFrame:
+    """
+    Reassign building tech type to communal if within a given radius of an anchor load property, unless in district heat network
```
crispy-wonton: todo: this needs to be updated to say that only N-GSHP tech types are reassigned

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -381,67 +435,86 @@ def reassign_gdf_communal_networked(
...
+
+    Args:
+        tech_gdf (gpd.GeoDataFrame): domestic building footprints with assigned tech types.
+        combined_anchor_gdf (gpd.GeoDataFrame): combined anchor property lists from important buildings and POI data, with building footprints
+        radius (float): distance in metres around an anchor, within which buildings will be assigned tech type of 'communal solutions' unless they are already assigned to district heat networks.
```
crispy-wonton: todo: same as above, needs updating here to N-GSHP only

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -381,67 +435,86 @@ def reassign_gdf_communal_networked(
...
+    Returns:
+        gpd.GeoDataFrame: dataframe with `assigned_tech` column now reading communal if property is in radius of an anchor property, unless in district heat network.
```
crispy-wonton: todo: same here

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -381,67 +435,86 @@ def reassign_gdf_communal_networked(
...
+
+    tech_gdf = tech_gdf.sjoin_nearest(
+        combined_anchor_gdf[["geometry"]],
         how="left",
         max_distance=radius,
         distance_col="distance_m",
     ).drop("index_right", axis=1)
-    reassigned_tech["old_tech"] = reassigned_tech["assigned_tech"]
-
+    tech_gdf["old_tech"] = tech_gdf["assigned_tech"]
```
crispy-wonton: suggestion: I think we don't need to retain the 'old_tech' column here. This can be deleted

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -381,67 +435,86 @@ def reassign_gdf_communal_networked(
...
     # distance_m column now reads a number (distance from anchor) for all buildings within radius of anchor, and NaN for all outside of that radius
     # if distance column is not NaN (i.e. building is within the radius of an anchor), reassign tech type according to the map
-    reassigned_tech[(reassigned_tech["distance_m"]).notna()] = reassigned_tech[
-        (reassigned_tech["distance_m"]).notna()
-    ].replace({"assigned_tech": tech_mapping})
-
+    tech_gdf["assigned_tech"] = np.where(
+        tech_gdf["distance_m"].notna(),
+        tech_gdf["assigned_tech"].replace(TECH_MAPPING),
```
crispy-wonton, line 521: suggestion: since you are only replacing N-GSHP with communal solutions, I don't think you need the whole mapping dict here. You should just be able to use {"N_GSHP": "COMMUNAL"} (please double check that works before implementation)

crispy-wonton (changes requested):
hey @simran-dave !

Looks good - just a few minor comments to clean up the code before merge - namely updating some of the docstrings to represent the changes in the logic of reassigning N-GSHP only to communal solutions, and to move source file paths to config.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
simran-dave: yep tested this on my notebook and looks like it works

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -441,7 +438,7 @@ def load_transform_anchor_property_gdfs(
     anchor_categories=ANCHOR_CATEGORIES,
 ) -> gpd.GeoDataFrame:
     """
-    Load data from POI and important buildings lists, select buildings using anchor proerty categoriesm, and combine the resultant dataframes
+    Load data from POI and important buildings lists, select buildings using anchor proerty categories, and combine the resultant dataframes
```
crispy-wonton: nitpick: typo in `property` here

crispy-wonton (approved):
one minor typo to fix but then good to merge!

PR #275, opened 2026-03-13
221 refactor decision tree code

### LINKED ISSUE #221: Refactor decision tree and maps code

sofiapinto:
---

# Description

Refactoring of the decision tree code.

We apply the decision tree to UPRN level data and then perform aggregations at building level. Both UPRN and building level data are saved.

Fixes #221 

# Instructions for Reviewer

@crispy-wonton could you please
- check the logic is in line with the discussions in sprint review 10?
- review the code and run it to check all works well?
- let me know if you'd like the decision tree function to be implemented in polars. i think the dictionary format is more readable and less prone to errors, but happy to change.

Thanks!


asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -0,0 +1,463 @@
...
+    gdf = gdf.sjoin(
+        hn_zones_gdf[["geometry", "desnz_hn_zone_geometry"]],
+        how="left",
+        predicate="within",
+    )
+
+    return gdf
+
+
+def identify_dict_most_suitable_tech(
```
sofiapinto, line 91: I could write this one in polars but I think it might become less readable. Let me know what you would prefer @crispy-wonton .

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto, line 1: The location of this script & name can be changed later. I can imagine this code will likely be added to another script.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+        solutions_per_footprint_gdf["building_in_hn_zone"],
+        TECH_TYPES["heat_network"],
+        solutions_per_footprint_gdf["assigned_tech"],
+    )
+
+    # Drop unecessary columns before saving outputs to S3
+    uprns_gdf = uprns_gdf.drop(columns=["in_hn_zone"])
+    solutions_per_footprint_gdf = solutions_per_footprint_gdf.drop(
+        columns=["building_in_hn_zone"]
+    )
```
sofiapinto: This is temporary. We should have a clearer view on what to do on Monday, after I go through the slides from sprint review 10 that we didn't have time for.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
+"""
+This script codes the decision tree that outputs the most suitable tech for each UPRN given the following inputs:
+- whether it is in a city centre or planned HN zone
+- contigous outdoor space size
```
crispy-wonton:
```suggestion
- maximum contiguous outdoor space size
```

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -0,0 +1,463 @@
...
+
+To run the script:
+
+python asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py --local_authorities LOCAL_AUTHORITIES
+
+LOCAL AUTHORITIES can be one of the following, as defined in the `constant` section of base.yaml:
+- plymouth
+- plymouth_similar_cities
+- sampling_areas
+- greater_manchester_las
```
crispy-wonton, line 17: praise: super helpful! love this doc

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
crispy-wonton: suggestion: I'd probably move this to `/pipeline/transform` and then rename the file to `decision_tree.py`.

issue: generally, I want to avoid having any functions defined within the `/run` subdir modules and want to keep these as scripts only.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+        action="store_true",
+    )
+
+    return parser.parse_args()
+
+
+def load_df_uprn_data(local_authorities: str) -> gpd.GeoDataFrame:
+    """
+    TODO: change after all columns exist within the same df.
+    Temporary function to join UPRN features including:
```
crispy-wonton: question: what do you mean by temporary here? will the whole function be deleted at some point?

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+    )
+    gdf["in_city_centre_or_hn_zone"] = gdf["in_city_centre"] | gdf["in_hn_zone"]
+
+    gdf.reset_index(inplace=True)
+
+    return gdf
+
+
+def extend_gdf_building_footprints(
+    gdf: gpd.GeoDataFrame, building_footprints: gpd.GeoDataFrame
```
crispy-wonton:
```suggestion
    gdf: gpd.GeoDataFrame, buildings_gdf: gpd.GeoDataFrame
```
nitpick: for consistency across the repo. This is what I've been naming it across functions.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+) -> gpd.GeoDataFrame:
+    """
+    Extends the GeoDataFrame with building footprints information.
+
+    Args:
+        gdf (gpd.GeoDataFrame): GeoDataFrame with UPRN data.
+        building_footprints (gpd.GeoDataFrame): GeoDataFrame with building footprints.
+
+    Returns:
+        gpd.GeoDataFrame: GeoDataFrame with building footprints information.
```
crispy-wonton:
```suggestion
        gpd.GeoDataFrame: GeoDataFrame with building footprint polygons.
```

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+    gdf = gdf.sjoin(
+        building_footprints[["geometry", "building_geometry"]],
+        how="left",
+        predicate="within",
+    ).drop(columns=["index_right"])
+
+    return gdf
+
+
+def extend_gdf_hn_zone_geometry(
```
crispy-wonton: question: isn't this covered by the HN zone work that's already done? I'm confused why this is here.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
crispy-wonton: I see what you mean about readability. I'd rather not use `apply` if we can avoid it, due to speed, so that's my only reservation. I think let's leave it for now (assuming it's not extremely slow) and then we can refactor into polars when scaling if we need speed gains. This also means we'll refactor it after the logic is (hopefully) finalised and so we'd less likely to introduce errors because of editing.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+    in_block_of_flats: bool, outdoor_space: float, city_centre_or_hnz: bool
+) -> dict:
+    """
+    Defines the decision tree to identify:
+    - most suitable low carbon heating solutions for each UPRN.
+    - the path taken in the decision tree.
+
+    Args:
+        in_block_of_flats (bool): Whether UPRN is in a block of flats.
+        outdoor_space (float): Outdoor space in squared meters.
```
crispy-wonton:
```suggestion
        outdoor_space (float): Maximum contiguous outdoor space in metres squared.
```

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+        city_centre_or_hnz (bool): Whether the UPRN is in the city centre or in a planned heat network zone.
+
+    Returns:
+        dict: A dictionary with the first and second most suitable heating solutions and the path taken in the decision tree.
```
crispy-wonton:
```suggestion
        dict: A dictionary with the most suitable heating solution and the path taken in the decision tree.
```

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+                "decision_tree_path": "1. blocks of flats, in HNZ/ city centre",
+            }
+        else:
+            return {
+                "assigned_tech": TECH_TYPES["communal"],
+                "decision_tree_path": "2. blocks of flats, not in  HNZ/ city centre",
+            }
+    else:
+        if city_centre_or_hnz:
+            if pd.isnull(outdoor_space):
```
crispy-wonton:
```suggestion
            if not outdoor_space:
```

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+                    "assigned_tech": TECH_TYPES["individual"],
+                    "decision_tree_path": "3. not in blocks of flats, in city centre, large outdoor space",
+                }
+            else:
+                return {
+                    "assigned_tech": TECH_TYPES["heat_network"],
+                    "decision_tree_path": "4. not in blocks of flats, in city centre, small or no outdoor space",
+                }
+        else:
+            if pd.isnull(outdoor_space):
```
crispy-wonton:
```suggestion
            if not outdoor_space:
```

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+            return {
+                "assigned_tech": TECH_TYPES["communal"],
+                "decision_tree_path": "2. blocks of flats, not in  HNZ/ city centre",
+            }
+    else:
+        if city_centre_or_hnz:
+            if pd.isnull(outdoor_space):
+                return {
+                    "assigned_tech": f"{TECH_TYPES['individual']} or {TECH_TYPES['heat_network']}",
+                    "decision_tree_path": "Unknown outdoor space in city centre",
```
crispy-wonton:
```suggestion
                    "decision_tree_path": "Not in block of flats. Unknown outdoor space in city centre",
```
or something similar

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+            else:
+                return {
+                    "assigned_tech": TECH_TYPES["heat_network"],
+                    "decision_tree_path": "4. not in blocks of flats, in city centre, small or no outdoor space",
+                }
+        else:
+            if pd.isnull(outdoor_space):
+                return {
+                    "assigned_tech": f"{TECH_TYPES['individual']} or {TECH_TYPES['networked']}",
+                    "decision_tree_path": "Unknown outdoor space not in city centre/ HN zone",
```
crispy-wonton:
```suggestion
                    "decision_tree_path": "Not in block of flats. Unknown outdoor space not in city centre/ HN zone",
```

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+                    "decision_tree_path": "6. not in blocks of flats, not in city centre/ HN zone, small/no outdoor space",
+                }
+
+
+def identify_df_building_most_suitable_tech(
+    uprns_gdf: gpd.GeoDataFrame,
+) -> gpd.GeoDataFrame:
+    """
+    Identifies sets of most suitable tech per building footprint, and assigns a unique solution for each building
+    based on the combination of solutions in the set (for the UPRNs in the building).
```
crispy-wonton: suggestion: you could make this clearer by explicitly stating that in some cases a building footprint contains UPRNs with different solutions.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+
+    Args:
+        uprnsuprns_gdf_gdf (gpd.GeoDataFrame): GeoDataFrame with UPRN data.
```
crispy-wonton:
```suggestion
        uprns_gdf (gpd.GeoDataFrame): GeoDataFrame with UPRN data.
```
issue: I think this needs some further explanation. E.g. it looks like the input gdf contains features and assigned tech columns. I would specify the names of the columns that are required in this df. 

nitpick: Also, please could we rename to `tech_gdf` or something similar. Again, for consistency, as I've used `uprns_gdf` throughout the repository to refer to UPRN data only, not joined to any features.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+
+    Returns:
+       gpd.GeoDataFrame: building footprint with assigned most suitable tech.
```
crispy-wonton:
```suggestion
       gpd.GeoDataFrame: building footprints with assigned most suitable tech.
```
question: correct me if I'm wrong, but it seems like at this point you have a subset of the buildings - only those with more than one solution. If that's the case, we should add it to docstring.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+    """
+
+    # Convert building geometry to WKB format before converting to Polars DataFrame
+    uprns_gdf["building_geometry"] = uprns_gdf["building_geometry"].to_wkb()
```
crispy-wonton: suggestion: alternatively you could use building ID

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+            uprns_gdf[
+                [
+                    "UPRN",
+                    "assigned_tech",
+                    "building_geometry",
+                    "max_contiguous_outdoor_space_area_m2",
+                    "in_hn_zone",
+                ]
+            ]
+        ).copy()
```
crispy-wonton: suggestion: I don't think you need the `.copy()` here because you are converting to `polars` anyway

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+            n_properties_available_outdoor_space_data=pl.col(
+                "max_contiguous_outdoor_space_area_m2"
+            )
+            .drop_nulls()
+            .count(),
+            n_properties=pl.col("UPRN").count(),
+            building_in_hn_zone=pl.col("in_hn_zone").any(),
+        )
+        # Filter for footprints with more than 1 solution
+        .with_columns(n_solutions=pl.col("assigned_tech").list.len())
```
crispy-wonton:
```suggestion
            building_in_hn_zone=pl.col("in_hn_zone").any(),
            n_solutions=pl.col("assigned_tech").n_unique(),
        )
        # Filter for footprints with more than 1 solution
```
you could also do this

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+    )
+
+    # Convert back to GeoDataFrame
+    solutions_per_footprint_df = solutions_per_footprint_df.to_pandas()
+    solutions_per_footprint_df["building_geometry"] = solutions_per_footprint_df[
+        "building_geometry"
+    ].apply(wkb.loads)
+    solutions_per_footprint_gdf = gpd.GeoDataFrame(
+        solutions_per_footprint_df, geometry="building_geometry", crs="EPSG:27700"
+    )
```
crispy-wonton: suggestion: if you do decide to use building ID, then it simplifies this to a join.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -0,0 +1,463 @@
...
+        "building_geometry"
+    ].apply(wkb.loads)
+    solutions_per_footprint_gdf = gpd.GeoDataFrame(
+        solutions_per_footprint_df, geometry="building_geometry", crs="EPSG:27700"
+    )
+
+    return solutions_per_footprint_gdf
+
+
+def assign_df_unique_solution(solutions_per_footprint_df: pl.DataFrame) -> pl.DataFrame:
```
crispy-wonton, line 272: note: just a note to say I've carefully checked this logic against the slides from sprint review 10 and it all looks correct!

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+            .when(
+                pl.col("assigned_tech").list.contains(
+                    f"{TECH_TYPES['individual']} or {TECH_TYPES['networked']}"
+                )
+            )
+            .then(
+                pl.when(
+                    (pl.col("perc_properties_available_outdoor_space_data") >= 50)
+                    & (
+                        pl.col("median_contiguous_outdoor_space_area_m2") > 30
```
crispy-wonton: suggestion: just thinking for this 30m threshold and the 70m threshold below, it might be worth putting these in the config and setting them as global variables in this file. Just in case we decide to change it at some point because I think it has a few usages across the repo.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+            .otherwise(pl.lit("Unexpected combination"))
+        )
+    )
+
+    return solutions_per_footprint_df
+
+
+def identify_most_suitable_tech_uprn_and_building(
+    local_authorities: str, save: bool
+) -> None:
```
crispy-wonton: suggestion: I think it would be worth returning the final df as well in case this was used in a notebook or another script at some point

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+        local_authorities (str): Local authority or authorities.
+        save (bool): Whether to save outputs to S3.
+    """
+
+    # Load and process data
+    hn_zones_gdf = load_gdf_heat_network_zones(local_authority=local_authorities)
+    building_footprints = load_gdf_os_openmap_local_layer(
+        layer="building",
+        grid_squares=config["constant"][local_authorities]["grid_squares"],
+    )
```
crispy-wonton: suggestion: ideally, I'd take these in as function parameters. It makes the function more modular and means that we can save on loading the same data multiple times if/when we combine this with other scripts.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -0,0 +1,463 @@
...
+        axis=1,
+    )
+
+    # Create new columns for assigned tech and decision tree path based on the most suitable solutions dictionary
+    uprns_gdf["assigned_tech"] = [
+        x["assigned_tech"] for x in uprns_gdf["most_suitable_solutions"]
+    ]
+    uprns_gdf["decision_tree_path"] = [
+        x["decision_tree_path"] for x in uprns_gdf["most_suitable_solutions"]
+    ]
```
crispy-wonton:
```suggestion
    # Identify most suitable tech for each UPRN
    suitable_tech_dicts = list(uprns_gdf.apply(
        lambda x: identify_dict_most_suitable_tech(
            x["in_block_of_flats"],
            x["max_contiguous_outdoor_space_area_m2"],
            x["in_city_centre_or_hn_zone"],
        ),
        axis=1,
    ))

    # Create new columns for assigned tech and decision tree path based on the most suitable solutions
    uprns_gdf = pd.concat([uprns_gdf, pd.DataFrame(suitable_tech_dicts)], axis=1)
```
not tested, but does this work?

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
crispy-wonton: ok will wait for your changes before commenting on this part

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -0,0 +1,463 @@
...
+
+    # Identify set of most suitable tech per building
+    solutions_per_footprint_gdf = identify_df_building_most_suitable_tech(uprns_gdf)
+
+    # Re-assign most suitable tech to each UPRN based on decision for each building
+    uprns_gdf = uprns_gdf.drop(columns=["assigned_tech"]).merge(
+        solutions_per_footprint_gdf[["building_geometry", "assigned_tech"]],
+        on="building_geometry",
+        how="left",
+    )
```
crispy-wonton, line 424: question: doesn't this mean we lose assigned tech data for any buildings that only have 1 solution? due to dropping the `assigned_tech` column first.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -0,0 +1,463 @@
...
+
+    if args.save:
+        uprns_gdf.to_parquet(
+            uprns_gdf.to_parquet(
+                f"s3://asf-heat-pump-suitability/local_heat_planning/outputs/{local_authorities}/{local_authorities}_uprns_most_suitable_tech.parquet"
+            )
+        )
+        solutions_per_footprint_gdf.to_parquet(
+            f"s3://asf-heat-pump-suitability/local_heat_planning/outputs/{local_authorities}/{local_authorities}_building_most_suitable_tech.parquet"
+        )
```
crispy-wonton, line 452: issue: let's remove hard-coding and put these file paths in the config yaml. I've been trying to implement this in my scripts as well so it's easier for us to update them later. Please flag if you see me hard-coding file paths going forward!

If there are files that are for temporary use, let's save them in a `temporary` section in the config.

crispy-wonton (changes requested):
hey @sofiapinto 

Thanks for refactoring this! i've mostly left some suggestions, questions, and nitpicks. I have tagged my comments using the [Conventional Comments](https://conventionalcomments.org/) conventions to hopefully help prioritise feedback. Let me know if it's helpful or not.

I think you've applied the logic of the decision tree correctly in your functions but I think there is a bug in the code. When you are generating the most suitable solution for each building, you drop any buildings which only have 1 solution across all UPRNs in them and then don't concat these buildings back on at the end of the function.

I think this is why the results look weird - I ran it, and only 2074 buildings are in the final dataset of buildings with solutions. And only 2307 UPRNs have an assigned tech, the rest are NaN. It also seems odd that all of the UPRNs are assigned 'communal solution' and no other technology, although I don't know what the cause of this specifically is.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
sofiapinto: Yes. If all these columns live within the same file, we will just need to import it. I think you already implemented this in the meantime, so I've removed most of what existed within this function already.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
sofiapinto: I can see why it's confusing. This is legacy from the notebook. It's not needed to identify the most suitable tech, it was used solely for the purpose of visualisation. I have deleted for now, but we might also eventually want to add it back on with some additional code to save intermediate plots.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
sofiapinto: Yes, this was a good shout. I like the new docstring much more. Let me know if it's clearer now from your perspective.

I have also improve the docstring of `assign_df_unique_solution()`.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
sofiapinto: I have now renamed  it to `tech_gdf`, but I find it less intuitive. When I see `tech_gdf` I wonder at what level? UPRN level? building level?

I think it's fine though, people read the docstring and they'll understand.

sofiapinto: Have also specified the required columns.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: yes, good flag. I had forgotten to copy some of the code from the notebook into the the script. I have now hopefully fixed it.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
sofiapinto: It was wrong, I have now hopefully fixed it. It should now output all of them.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
sofiapinto: I did this instead.

```
    # Identify most suitable tech for each UPRN
    decision_tree_outputs = uprns_gdf.apply(
        lambda x: identify_dict_most_suitable_tech(
            x["in_block_of_flats"],
            x["max_contiguous_outdoor_space_area_m2"],
            x["in_city_centre_or_hn_zone"],
        ),
        axis=1,
        result_type="expand"
    )

    # Create new columns for `assigned_tech` and `decision_tree_path` based on the most suitable solutions dictionary
    uprns_gdf = pd.concat([uprns_gdf, decision_tree_outputs], axis=1)
```

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
sofiapinto: have moved, let me know if you like the location & names etc.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
sofiapinto: local authorities is already a function parameter. What else did you want to have as a function parameter? grid squares?

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: I'd like to discuss this one at our next meeting :)

sofiapinto:
Hey @crispy-wonton , thank you for your review!

I think i've now fixed the issues , answered your questions and addressed all your suggestions except two - let's discuss those in our next meeting.

I still need to move & rename this file, just wanted to avoid all lines of code appearing as new so will do it after your review.


A note on plymouth data:
I think we might need to run add_features.py for Plymouth again, ahead of running this script for Plymouth. The name of the file changed a few times.

I have tried running it for cardiff and it worked well with:


python asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py --local_authorities "cardiff"

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -68,10 +66,12 @@ def parse_arguments() -> argparse.Namespace:
 
 def load_df_uprn_data(local_authorities: str) -> gpd.GeoDataFrame:
     """
-    TODO: change after all columns exist within the same df.
-    Temporary function to join UPRN features including:
-    - UPRNs and their geometries
-    - Block of flats information
+    Loads UPRN level data with relevant features for the decision tree, including:
```
crispy-wonton: todo: this function should be called `load_gdf_uprn_data`. I'd also probably name it `load_gdf_uprn_features` or something.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -80,94 +80,47 @@ def load_df_uprn_data(local_authorities: str) -> gpd.GeoDataFrame:
...
 def extend_gdf_building_footprints(
-    gdf: gpd.GeoDataFrame, building_footprints: gpd.GeoDataFrame
+    gdf: gpd.GeoDataFrame, buildings_gdf: gpd.GeoDataFrame
 ) -> gpd.GeoDataFrame:
     """
-    Extends the GeoDataFrame with building footprints information.
+    Extends the GeoDataFrame with building footprint polygons
 
     Args:
         gdf (gpd.GeoDataFrame): GeoDataFrame with UPRN data.
```
crispy-wonton, line 72: suggestion: perhaps specify that the UPRNs would have point geometries

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -83,8 +83,13 @@ constant:
...
       - "SJ"
       - "SK"
+  cardiff:
+    la_names: "Cardiff"
+    grid_squares: "ST"
   target_crs: 27700 # British National Grid
-
+  outdoor_space_threshold_m2:
+    within_hn_zone: 30
+    outside_hn_zone: 70
```
crispy-wonton: issue: i think these are the wrong way round. Within HN zone should be 70m2.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -232,25 +185,54 @@ def identify_dict_most_suitable_tech(
...
+
+    - Identifies the different solutions for each building (the set of solutions for UPRNs within the same building footprint)
+    - Where only 1 solution exists for a building footprint, assign that solution as the most suitable tech for the building footprint
+    - Where multiple solutions exist for a building footprint, assign a unique solution based on the combination of solutions in the set and the median outdoor space of properties within the building footprint.
+    This function aggregates the most suitable tech from individual UPRNs to the building level.
+
+    Multiple solutions within the same building might occur when a building contains multiple UPRNs and the decision tree assigns different technologies to at least 2 of them.
+    A few examples of where this happens:
+    - a building that sits at the edge of a HN zones, might have properties in the building assigned district heat network and others networked GSHP.
+    - a building footprint consisting of a row of terraced houses, where some properties have large outdoor space and others small outdoor space.
```
crispy-wonton, line 168: nice to include these examples!

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -232,25 +185,54 @@ def identify_dict_most_suitable_tech(
 
 
 def identify_df_building_most_suitable_tech(
```
crispy-wonton, line 154: nitpick: this function should be called `identify_gdf_building_most_suitable_tech` since it returns a gdf.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
```diff
@@ -232,25 +185,54 @@ def identify_dict_most_suitable_tech(
...
+        "in_hn_zone",
+    ]:
+        if col not in tech_gdf.columns:
+            raise ValueError(
+                "Input GeoDataFrame must contain the following columns: ['UPRN', 'assigned_tech', 'building_geometry', 'max_contiguous_outdoor_space_area_m2', 'in_hn_zone']."
+            )
+
     # Convert building geometry to WKB format before converting to Polars DataFrame
-    uprns_gdf["building_geometry"] = uprns_gdf["building_geometry"].to_wkb()
+    tech_gdf["building_geometry"] = tech_gdf["building_geometry"].to_wkb()
```
crispy-wonton: I still think i'd rather use building ID here and join the geometry back on than do this conversion. It makes it quicker and cleaner. Then below you can group by building ID. They should be unique per geometry.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
crispy-wonton: issue: this needs updating with the params from config.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
crispy-wonton: is this still awaiting changes?

crispy-wonton (commented):
minor comments / questions and some params need updating with config.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
sofiapinto: I deleted the code for now. We have decided we want to show both the tool outputs and, separately, the HN zone - so no need to allow the HN to take precedence over the pipeline outputs. 

The mock ups on the UI are still happening, but hopefully this shouldn't influence going back in our decision. If so, we'll add it back in.

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
sofiapinto: from chat with roisin: hn_zones_gdf and building_footprints being loaded outsidse and then having them as parameters

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
sofiapinto: After our chat this morning about getters, I have actually deleted the function and am instead loading the data inside `if __name__ == "__main__":`

sofiapinto: We will create the getters alter and can then update this - have left as TODO

asf_heat_pump_suitability/pipeline/run/identify_most_suitable_tech.py
sofiapinto: Apologies, had completely missed this suggestions. I agree & have now implemented!

sofiapinto:
Hi @crispy-wonton thank you for the second review - the script has improved substantially with your suggestions!

I have now:

- removed the code where HN took precedence over pipeline outputs
- fixed the thresholds: both the values & the calls within the script
- added building_footprints and uprns as parameters to `identify_gdf_tuple_most_suitable_tech_uprn_and_building()` to avoid loading dfs multiple times
- moved file paths to base.yaml
- clarified the `extend_gdf_building_footprints()` docstring
- add clear return types to function names (e.g. `identify_gdf_tuple_most_suitable_tech_uprn_and_building()` was originally missing the return type)
- used building ID for the groupby's and joins instead of the geometry
- moved and renamed the script

Can you please make one final check to see if function and variable names are all as you wish? And that I haven't missed any of your suggestions?

You can now run the code with:
`python asf_heat_pump_suitability/pipeline/transform/decision_tree.py --local_authorities "cardiff"`

Have also added a few temporary prints to try and debug a few numbers. It seems like some building footprints have ID but no geometry. Wanted to check if this is expected?

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -88,15 +88,18 @@ constant:
...
     individual: "Individual solution"
     networked: "Networked heat pump"
     communal: "Communal solution"
     heat_network: "District heat network"
 
 output:
+  uprns_with_features: "s3://asf-heat-pump-suitability/local_heat_planning/outputs/{local_authority}/{local_authority}_with_features.parquet"
+  uprns_most_suitable_tech: "s3://asf-heat-pump-suitability/local_heat_planning/outputs/{local_authorities}/{local_authorities}_uprns_most_suitable_tech.parquet"
+  buildings_most_suitable_tech: "s3://asf-heat-pump-suitability/local_heat_planning/outputs/{local_authorities}/{local_authorities}_buildings_most_suitable_tech.parquet"
   save_as:
```
sofiapinto, line 109: have put the above outside of save_as because I think we could probably get rid of save_as in the future? let me know if there's a different structure you would prefer

crispy-wonton: I think it makes sense to not be under `save_as`. I would probably still have something to distinguish it as a data output. so `output` `dataset` for example. And then, as this could get quite full with various paths, it would be great to add some clear annotation about what each file contains.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -64,40 +62,16 @@ def parse_arguments() -> argparse.Namespace:
...
-
-
 def extend_gdf_building_footprints(
     gdf: gpd.GeoDataFrame, buildings_gdf: gpd.GeoDataFrame
 ) -> gpd.GeoDataFrame:
     """
-    Extends the GeoDataFrame with building footprint polygons
+    Extends the GeoDataFrame with building footprint geometries by performing a
+    spatial join to add building footprint geometry to the UPRN geodataframe based on
+    whether the UPRN point is within the building footprint polygon.
```
crispy-wonton: suggestion: you could simplify this to:
'Extends the GeoDataFrame with the building footprint geometries the UPRNs are located within.'

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -204,12 +178,11 @@ def identify_df_building_most_suitable_tech(
     tech_gdf (gpd.GeoDataFrame): GeoDataFrame containing UPRN-level most suitable tech. Must contain the following columns:
         * `UPRN`: Unique identifier for the property.
         * `assigned_tech`: The most suitable tech assigned by the decision tree.
-        * `building_geometry`: Geometry of the building footprint.
+        * `ID`: Unique identifier for the building footprint.
```
crispy-wonton: suggestion: wherever i've used this building `ID` column, I've added an arg to the function called `id_col` and then defined it in the scripts in case it ever changes (possible as the datasets are updated every month I think). We could also define ID col in the config and call it in the scripts or as a default.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -261,7 +231,6 @@ def identify_df_building_most_suitable_tech(
             .drop_nulls()
             .count(),
             n_properties=pl.col("UPRN").count(),
-            building_in_hn_zone=pl.col("in_hn_zone").any(),
```
crispy-wonton: question: why are we removing this? don't we still need it for the decision tree?

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -397,21 +367,34 @@ def identify_most_suitable_tech_uprn_and_building(
 
     Args:
         local_authorities (str): Local authority or authorities.
+        building_footprints_gdf (gpd.GeoDataFrame): GeoDataFrame with building footprints.
```
crispy-wonton: nitpick: rename to `buildings_gdf` just for consistency again

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -397,21 +367,34 @@ def identify_most_suitable_tech_uprn_and_building(
...
+        len(uprns_gdf[uprns_gdf["building_geometry"].isnull()]),
+    )
+    print(
+        "Number of UPRNs with ID but no building geometry: ",
+        len(
+            uprns_gdf[
+                (uprns_gdf["building_geometry"].isnull()) & (uprns_gdf["ID"].notnull())
+            ]
+        ),
+    )
```
crispy-wonton, line 398: this should not happen

crispy-wonton (commented):
hey @sofiapinto !

I have left some minor comments.

My main question is because I got a bit lost and confused in the changes to using the `in_hn_zone` flag. Don't we still need the flag for the decision tree? Why is it no longer aggregated per building footprint?

asf_heat_pump_suitability/config/base.yaml
sofiapinto: Do you mean?
```
    output:
         dataset:
              ....
          model:
             ....
```

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: I know, but it is. Hence why I left it here, so you could see it when running the script.

> You can now run the code with:
> python asf_heat_pump_suitability/pipeline/transform/decision_tree.py --local_authorities "cardiff"
> 
> Have also added a few temporary prints to try and debug a few numbers. It seems like some building footprints have ID but no geometry. Wanted to check if this is expected?

Do you know why this could be happening?

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: are you proposing I add `id_col` as parameter to this function then?

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: Notice that this is happening inside `identify_df_building_most_suitable_tech()`, i.e. where we're identifying the unique solution for each building. To identify the unique solution for each building all we need is:
- the set of solutions in that building (considering all UPRNs)
- the average outdoor space & number of properties with non-missing outdoor space data (so we can decide what to assign when a building footprint has some properties with missing outdoor space)

We previously had `building_in_hn_zone=pl.col("in_hn_zone").any()` because we wanted to be able to ensure, later in the pipeline, that the HN zone could take precedence (and replace the any outputs with "HN", if the buildings wee in HN zones). Because we're now separating the outputs of our modelling from the HN zones, this could be removed. Let me know if this isn't clear, happy to chat through it on a call.

But in `identify_dict_most_suitable_tech()`, the HN zone info is still used, as it should, as that's where we're defining the decision tree.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
crispy-wonton: yes

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
crispy-wonton: Makes sense thanks

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -0,0 +1,482 @@
...
+
+    # Combine the dataframes of building footprints with multiple solutions and single solution to get the final dataframe with a unique assigned solution for each building footprint
+    solutions_per_footprint_df = pl.concat(
+        [buildings_with_multiple_solutions_df, buildings_with_single_solution_df]
+    )
+
+    # Convert back to GeoDataFrame
+    solutions_per_footprint_df = solutions_per_footprint_df.to_pandas()
+
+    return solutions_per_footprint_df
```
crispy-wonton: Just realised, I don't think this is converting back to a geodataframe, just a pandas df. Why not just retain it as a polars df?

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
crispy-wonton: Have looked through your code here. I think it's because you are joining the final building footprints back on with a spatial join. However, you should just be using the building `ID` column from the features dataframe to do a merge join as the building ID has already been assigned to the UPRNs in the features dataset.

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: yep exactly

crispy-wonton (commented):
Responded to your comments and left a final comment.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: the .to_pandas() was there by mistake, have removed.

sofiapinto: Actually, it wasn't a mistake. It's so that we can do the `merge`'s after applying identify_df_building_most_suitable_tech(). The merges don't work if solutions_per_footprint_df is a polars df.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: It was due to the join in the beginning in `extend_gdf_building_footprints()`, likely because in add_features.py you did the sjoin in a slightly different way to the one I was doing here. Using the building ID sorted it.

This is what we're getting now for Plymouth:

```
Number of building IDs:  32944
Number of building geometries:  32944
Number of UPRNs without building geometry:  2248
Number of UPRNs without building ID:  2248
Number of UPRNs with ID but no building geometry:  0
Number of building IDS  32944
Number of building geometries:  32944
```

So we still have UPRNs without a matching ID/geometry. I'm guessing these UPRNs in buildings at the edge of the LA boundary? Haven't visualised them though.

sofiapinto:
Hey @crispy-wonton ,

Thanks for the review.

- I've fixed the join that was causing the issues. But please see my comment here: https://github.com/nestauk/asf_heat_pump_suitability/pull/275#discussion_r3064685330
- Have reorganised the outputs in base.yaml
- Answered your Qs

Still need to remove the prints after you have a final look.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
crispy-wonton: The UPRNs without building geometry should just be the ones that are in the domestic EPC register but are located outside of building footprints. You can cross-check with the EPC if you want to be sure.

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -97,9 +97,15 @@ constant:
...
-  buildings_most_suitable_tech: "s3://asf-heat-pump-suitability/local_heat_planning/outputs/{local_authorities}/{local_authorities}_buildings_most_suitable_tech.parquet"
+  dataset:
+    # Residential UPRNs and their point geometries
+    residential_uprns: "s3://asf-heat-pump-suitability/local_heat_planning/outputs/{local_authority}/{local_authority}_residential_uprns.parquet"
+    # Residential UPRNs with all features for modelling, mapping and aggregation
+    residential_uprns_with_features: "s3://asf-heat-pump-suitability/local_heat_planning/outputs/{local_authority}/{local_authority}_with_features.parquet"
+    # Residential UPRNs with their most suitable low carbon heat technology group
+    uprns_most_suitable_tech: "s3://asf-heat-pump-suitability/local_heat_planning/outputs/{local_authorities}/{local_authorities}_uprns_most_suitable_tech.parquet"
+    # Residential buildings with their most suitable low carbon heat technology group
+    buildings_most_suitable_tech: "s3://asf-heat-pump-suitability/local_heat_planning/outputs/{local_authorities}/{local_authorities}_buildings_most_suitable_tech.parquet"
```
crispy-wonton, line 108: nice!

crispy-wonton (approved):
Changes look good - do you want me to run anything before merging?

PR #277, opened 2026-03-16
276 handle cardiff data + enable pipelines to be ran for any LA (even when HN zones are missing)

### LINKED ISSUE #276: Handle Cardiff data
Handle cardiff data in a bespoke way.

sofiapinto:
---

# Description

Hey there @crispy-wonton ,

I started working on this branch to handle LAs outside of Plymouth. As you know I also did this when we ran the pipelines for Greater Manchester, but a few new things came up when running the data for Cardiff:

- **Loading correct building footprints**:`add_features.py` to load correct building footprints independently of the LA. `SX` was hardcoded into the script.
- **Enabling one or lists of LAs:** Enabling a list of LAs or a unique LA to be loaded to get the inspire file names. Was previously erroring.
- **Dealing with no available HN data**: DESNZ HN zones cover only a number of LAs in England. Our code was not prepared for this data not to be available, so I had to change it to account for this data not being available. Probably not the nicest way of dealing with it, but works for now.

A few additional things:
- **Folders for outputs in each LA**: You previously mentioned you wanted to save the data for each LA in its own folder, so I updated the paths. Do note that I haven't defined the paths in base.yaml as per your review in a separate PR, so will still need to do that.
- **Spatial signatures framework level of detail**: Allowing the user to define the level of detail of the spatial signatures framework to load when identify city centres. This is useful, because the script using the full detail data takes a long time to run and it's not needed for prototypes.
- **Adding cardiff info to base.yaml**: I know we will eventually move away from this approach, but for now have added it to base.yaml.
- **Changing arguments of `add_features.py`**: swapping path to UPRNs with the local authority names. This makes it so much easier to run the pipelines each time.

Fixes #276 

# Instructions for Reviewer

Hey there @crispy-wonton 

It would be really great to get your review on this ahead of you merging your [PR 279](https://github.com/nestauk/asf_heat_pump_suitability/pull/279), so that you can bring changes from dev into your branch and account for them.
Thanks so much in advance!

# Checklist:

- [x] I have tested the code


asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -67,11 +65,20 @@ def parse_arguments() -> argparse.Namespace:
 
     args = parse_arguments()
     las = args.local_authorities.lower()
+    # TODO check if we want to import HNZ data separately or move code into this script.
+    uprns_path = f"s3://asf-heat-pump-suitability/local_heat_planning/outputs/{las}/{las}_residential_uprns_with_hn_zones_city_centres.parquet"
```
crispy-wonton: question: why are you loading this specific file instead of just the residential UPRNs?

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
```diff
@@ -41,25 +41,27 @@ def parse_arguments() -> argparse.Namespace:
...
-
     parser.add_argument(
         "--save",
         help="If --save is set, it saves outputs to S3.",
         required=False,
         action="store_true",
     )
 
+    parser.add_argument(
+        "--spatial_signatures_detail_level",
```
```suggestion
        "--detail",
```
suggest changing to something much shorter and allowing the docstring to explain.

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
```diff
@@ -80,39 +81,63 @@ def parse_arguments() -> argparse.Namespace:
...
-                id_col,  # zone unique identifier
-            ],
-        )
-
-        # Clean up columns
-        hn_zone_uprn_df = hn_zone_uprn_df.select(
-            ["UPRN", "LAD23NM", "X_COORDINATE", "Y_COORDINATE", "in_hn_zone", id_col]
-        ).rename({id_col: "HNZoneID"})
+        try:
+            hn_zones_gdf = load_geodata.load_gdf_heat_network_zones(local_authority=las)
```
crispy-wonton: issue: I just noticed this won't work if `las` is a list, only if it's a string.

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
```diff
@@ -80,39 +81,63 @@ def parse_arguments() -> argparse.Namespace:
...
+                [
+                    "UPRN",
+                    "LAD23NM",
+                    "X_COORDINATE",
+                    "Y_COORDINATE",
+                    "in_hn_zone",
+                    id_col,
+                ]
+            ).rename({id_col: "HNZoneID"})
+        except ValueError:
```
crispy-wonton: suggestion: could the `except` block execute earlier? I'm worries that something else could trigger it in this block of code and then we could falsely end up with 'no heat network zone data' flag when in fact, there is data.

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
crispy-wonton, line 1: note: as i've deleted this whole file in my [PR 279](https://github.com/nestauk/asf_heat_pump_suitability/pull/279), you will need to copy these changes into the `add_features.py` script later.

crispy-wonton (commented):
Couple of minor questions and comments.

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
sofiapinto: That's why I was hoping we could merge this one first, so you could bring these changes into add_features.py in your PR.

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
sofiapinto: I think that's a really good idea.

asf_heat_pump_suitability/pipeline/run/add_features.py
sofiapinto: I can reverse this change. This was just so that the city centres/HN flags were added to the data generated by add_features.py

It was just a temp thing since you said you were working on combining the heat neworks script code into add_features.py

asf_heat_pump_suitability/pipeline/run/heat_network_city_centres.py
sofiapinto: It depends. For Greater Manchester LAs for example it did, because we had the HN zones file for the whole GM area, rather than the individual LAs. But I have now allowed the code for checking both (the combination of LAs + the individual LAs)

sofiapinto:
Hey there @crispy-wonton,

I have now changed the code according to the latest changes in dev. A reminder of the summary of changes:

- **Loading correct building footprints:** `add_features.py` to load correct building footprints independently of the LA. SX was hardcoded into the script.
- **Enabling one or lists of LAs**: Enabling a list of LAs or a unique LA to be loaded to get the inspire file names & to get HN zones
- **Dealing with no available HN data**: DESNZ HN zones cover only a number of LAs in England. Our code was not prepared for this data not to be available, so I had to change it to account for this data not being available. 
- **Folders for outputs in each LA**: We're saving the data for each LA in its own folder. Paths are defined in base.yaml, instead of hardcoded into the scripts.
- **Spatial signatures framework level of detail:** Allowing the user to define the level of detail of the spatial signatures.
- **Adding cardiff info to base.yaml**: I know we will eventually move away from this approach, but for now have added it to base.yaml.
- **Changing arguments of add_features.py**: swapping path to UPRNs with the local authority names. This makes it so much easier to run the pipelines each time.

I have also taken your suggestions into account.

I have also run both the uprns.py and the add_features.py pipelines for "plymouth", "cardiff" and "greater_manchester_las" and inspected the resulting dfs.

To test whether things run properly, you can run the simplified versions without saving to S3:
- `python asf_heat_pump_suitability/pipeline/run/add_features.py --local_authorities "cardiff" --detail "simplified"`
- `python asf_heat_pump_suitability/pipeline/run/add_features.py --local_authorities "plymouth" --detail "simplified"`
- `python asf_heat_pump_suitability/pipeline/run/add_features.py --local_authorities "greater_manchester_las" --detail "simplified"`

Can you please also briefly inspect the resulting dfs?

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -139,29 +162,55 @@ def parse_arguments() -> argparse.Namespace:
...
+            hn_zones_gdf = pd.concat(
+                [
+                    hn_zones_gdf,
+                    load_geodata.load_gdf_heat_network_zones(local_authority=la),
+                ],
+                ignore_index=True,
+            )
+        except ValueError:
+            print(
+                f"No heat network zone geodata found for {la}. Assuming no UPRNs are in heat network zones in this Local Authority."
```
```suggestion
                f"No heat network zone geodata found for {la}. All UPRNs will be labelled as 'outside heat network zone' in this Local Authority."
```
Feel free to change the wording but the current wording makes it sounds like a methodological assumption rather than how we treat missing data. Same below.

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -139,29 +162,55 @@ def parse_arguments() -> argparse.Namespace:
...
+            )
+
+    # Check if data is available for all LAs in the list, and if not, check if there is data for the whole set of LAs (e.g. Greater Manchester as a whole instead of individual LAs)
```
crispy-wonton: question: I'm not understanding this sentence. Could it be reworded more clearly?

asf_heat_pump_suitability/pipeline/transform/outdoor_space.py
```diff
@@ -33,7 +33,7 @@ def match_series_files_land_building(
 
 def generate_gdf_building_intersections(
     land_parcels_gdf: gpd.GeoDataFrame,
-    building_footprints_gdf: gpd.GeoDataFrame,
+    buildings_gdf: gpd.GeoDataFrame,
```
crispy-wonton, line 36: suggestion: I think this might be a breaking change in the old scripts - please could you check and update those too if necessary. Here's the scripts and lines to look at:
```
asf_heat_pump_suitability/pipeline/flows/run_calculate_garden_size_flow.py:154:                intersection_gdf = outdoor_space.generate_gdf_building_intersections(
asf_heat_pump_suitability/pipeline/run_scripts/run_calculate_garden_size.py:142:        intersection_gdf = outdoor_space.generate_gdf_building_intersections(
```

crispy-wonton (commented):
Minor comments but mostly looking good!

Much cleaner with the args updated this way!

Have run for Cardiff and Plymouth and briefly checked the outputs - looks good!

asf_heat_pump_suitability/pipeline/run/add_features.py
sofiapinto: In some cases we have HN zones data for individual LAs. But in other cases we have data for groups of LAs: an example of that is Greater Manchester. The data exists in one file for the Greater Manchester Combined Authority, rather individual files for the individual LAs in the combined authority.

I think we'll want to do some work in the future to put all zones into one file and then read from there, but for now this code allows for checking if:
- there's HN zones for each LA individually
- HN zones for the combined authority (or list of LAs)
- there's no HN zone data at all.

Happy to discuss more at the stand up.

asf_heat_pump_suitability/pipeline/transform/outdoor_space.py
sofiapinto: I have checked and changed, although I don't think we should be worrying about this. As I see it, anyone wanting to run the old flows should be doing so using **v1 of the repo**, not this v2. As far as I'm aware, we don't currently need to run the two scripts you highlighted in our pipelines - if we do, we should move them to the appropriate new folders within the repo. We'll need to action the removal of the old files quite soon.

asf_heat_pump_suitability/pipeline/run/add_features.py
sofiapinto: And this is why I test the script with the 3 different areas:
- Cardiff: no HN zone
- Plymouth: HN zone (individual LA)
- Greater Manchester: HN zone for the whole set of LAs rather than individual LAs

sofiapinto: Have now added documentation that should make it easier to understand the above, but please flag if you'd still want to clarify anything

asf_heat_pump_suitability/pipeline/flows/run_calculate_garden_size_flow.py
```diff
@@ -153,7 +153,7 @@ def estimate_garden_size(self):
                 # Get intersection of building footprint polygons and land polygons
                 intersection_gdf = outdoor_space.generate_gdf_building_intersections(
                     land_parcels_gdf=land_parcels_gdf,
-                    building_footprints_gdf=building_footprints_gdf,
+                    building_gdf=building_footprints_gdf,
```
crispy-wonton: should be `buildings_gdf`

asf_heat_pump_suitability/pipeline/transform/outdoor_space.py
crispy-wonton: I agree that we should remove old files and that's my hope to prioritise cleaning up the repository in the beginning of the next phase.

I do think it is generally bad practice to merge breaking changes as we go along, rather than removing files which no longer function. Especially as we are not explicitly documenting these breakages in the repo. This is why, for now, I prefer to fix breakages and then remove the files in the next phase.

asf_heat_pump_suitability/pipeline/run/add_features.py
crispy-wonton: Much clearer thank you!

crispy-wonton (commented):
Typo in updated script to fix but the other changes look good - thanks for adding the new documentation , it's really clear!

crispy-wonton (approved):

sofiapinto:
Hey @crispy-wonton , just to confirm that I have re-run the following:

**Plymouth**
- `python asf_heat_pump_suitability/pipeline/transform/uprns.py --local_authorities "plymouth" --save`
-  python asf_heat_pump_suitability/pipeline/run/add_features.py --local_authorities "plymouth" --save (which will run this with "full" level of detail)

**Cardiff**
- `python asf_heat_pump_suitability/pipeline/transform/uprns.py --local_authorities "cardiff" --save`
- python asf_heat_pump_suitability/pipeline/run/add_features.py --local_authorities "cardiff" --save (which will run this with "full" level of detail)

**Greater Manchester Combined Authority**
- `python asf_heat_pump_suitability/pipeline/transform/uprns.py --local_authorities "greater_manchester_las" --save`
- `python asf_heat_pump_suitability/pipeline/run/add_features.py --local_authorities "greater_manchester_las" --save` (which will run this with "full" level of detail)

So will now merge this PR :) 

Thanks for the thorough reviews as always.

PR #284, opened 2026-03-30
282 implement additions domestic filtering

### LINKED ISSUE #282: Implement recommended improvements to domestic filtering pipeline
Implement the following to the domestic filtering pipeline:

- if building intersects with an important buildings / POI lists (any category) and has exactly 1 UPRN, filter the building out. Note, the building should be retained if the 1 UPRN is also in the domestic EPC register.
- After joining EPC domestic properties to building footprints, for any domestic EPC UPRNs which have not been joined to a building, join them to the nearest footprint <1m away. Note, make the 1m an argument of the function so it can be changed easily.

simran-dave:
---

# Description
Implements the following changes to the domestic filtering pipeline:

- Add a step to the domestic filtering that says if building intersects with an important buildings / POI lists (any category) and has exactly 1 UPRN, filter the building out. This happens in `non_residential_entities.py` in the function `generate_gdf_non_residential_buildings`. I take the unfiltered POI and important buildings lists, join them to the UPRNs and find buildings that have 1 UPRN in them. I think add these to the non residential gdf. UPRNs in the domestic EPC data are then removed in the step after. This then gets implemented in `uprns.py`

- After joining EPC domestic properties to building footprints, for any domestic EPC UPRNs which have not been joined to a building, join them to the nearest footprint <1m away. This happens in `uprns.py` in the function `map_dict_uprns_to_building_id`. I've added a step which uses a mask of EPC UPRNs to select domestic UPRNs that are in / not in the EPC list. I then `sjoin` UPRNs not in EPC list to building footprints if they are contained within them. I `sjoin.nearest` the UPRNs in the EPC list to buildings < 1m away. Note for most of them, this should be the building they are contained within. This function is then implemented in `add_features.py` . 

Fixes #282 

# Instructions for Reviewer

In order to test the code in this PR you need to run `uprns.py` and `add_features.py`. 

I think that having the EPC buildings joined to buildings < 1m away in `add_features.py` is the right place to do it but let me know if there is anywhere else in the pipeline this should be implemented

# Checklist:

- [x] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
+++ b/asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
@@ -6,6 +6,7 @@ import geopandas as gpd
 import pandas as pd
 
 from asf_heat_pump_suitability.pipeline.transform import uprns
+from asf_heat_pump_suitability import config
 
 # TODO these may need refinement for large cities where overlap with residential is possible
 NO_RESIDENTIAL_OVERLAP_BUILDING_TYPES = [
@@ -33,7 +34,7 @@ NO_RESIDENTIAL_OVERLAP_BUILDING_TYPES = [
 ]
 
 
-def generate_gdf_all_uprns_non_res(
+def _generate_gdf_fully_commercial_buildings(
     important_building_gdf: gpd.GeoDataFrame,
     poi_gdf: gpd.GeoDataFrame,
     uprns_gdf: gpd.GeoDataFrame,
@@ -50,31 +51,58 @@ def generate_gdf_all_uprns_non_res(
         gpd.GeoDataFrame: geometries of buildings which are in the important buildings list and have 1 UPRN in them, or in the POI list and have same number of UPRNs and POI
     """
     # join POI data to building footprints
-    poi_gdf = building_gdf.sjoin(poi_gdf, how="inner", predicate="contains").drop(
-        "index_right", axis=1
-    )
+    poi_buildings_gdf = building_gdf.sjoin(
+        poi_gdf, how="inner", predicate="contains"
+    ).drop("index_right", axis=1)
 
-    # count instances where multiple POI are located inside a building
-    poi_gdf = poi_gdf.groupby("geometry").size().reset_index(name="POI_count")
+    # count instances where multiple POI are located inside a building and convert result to gdf
+    poi_buildings_df = (
+        poi_buildings_gdf.groupby("ID")
+        .agg(POI_count=("ID", "count"), geometry=("geometry", "first"))
+        .reset_index()
+    )
+    poi_buildings_gdf = gpd.GeoDataFrame(
+        poi_buildings_df,
+        geometry=poi_buildings_df["geometry"],
+        crs=config["constant"]["target_crs"],
+    )
 
     # join important buildings and POI with all UPRNs
     important_building_gdf = important_building_gdf.sjoin(
         uprns_gdf, how="left", predicate="contains"
     ).drop("index_right", axis=1)
 
-    poi_gdf = poi_gdf.sjoin(uprns_gdf, how="left", predicate="contains").drop(
-        "index_right", axis=1
+    poi_buildings_gdf = poi_buildings_gdf.sjoin(
+        uprns_gdf, how="left", predicate="contains"
+    ).drop("index_right", axis=1)
+
+    # find number of UPRNs per building for important buildings and convert result to gdf
+    important_building_df = (
+        important_building_gdf.groupby("ID")
+        .agg(UPRN_count=("UPRN", "count"), geometry=("geometry", "first"))
+        .reset_index()
+    )
+    important_building_gdf = gpd.GeoDataFrame(
+        important_building_df,
+        geometry=important_building_df["geometry"],
+        crs=config["constant"]["target_crs"],
     )
 
-    # find number of UPRNs per building for important buildings
-    important_building_gdf = (
-        important_building_gdf.groupby("geometry").size().reset_index(name="UPRN_count")
+    # find number of UPRNs per building for POI buildings and convert result to gdf
+    poi_buildings_df = (
+        poi_buildings_gdf.groupby("ID")
+        .agg(
+            UPRN_count=("UPRN", "count"),
+            POI_count=("POI_count", "first"),
+            geometry=("geometry", "first"),
+        )
+        .reset_index()
     )
-    # find number of UPRNs per building for POI buildings
-    uprn_counts = poi_gdf.groupby("geometry").size().reset_index(name="UPRN_count")
-
-    # join uprn counts as a column to POI dataframe
-    poi_gdf = poi_gdf.sjoin(uprn_counts, how="left", predicate="intersects")
+    poi_buildings_gdf = gpd.GeoDataFrame(
+        poi_buildings_df,
+        geometry=poi_buildings_df["geometry"],
+        crs=config["constant"]["target_crs"],
+    )
 
     # select building footprints from important building data where UPRN count = 1
     important_building_gdf = important_building_gdf.loc[
@@ -82,11 +110,16 @@ def generate_gdf_all_uprns_non_res(
     ]
 
     # select building footprints from POI data where UPRN count = POI count
-    poi_gdf = poi_gdf.loc[poi_gdf["UPRN_count"] == poi_gdf["POI_count"]]
+    poi_buildings_gdf = poi_buildings_gdf.loc[
+        poi_buildings_gdf["UPRN_count"] == poi_buildings_gdf["POI_count"]
+    ]
 
-    non_res_uprns = pd.concat([important_building_gdf, poi_gdf])
+    # concat gdfs and keep only geometry column
+    non_domestic_buildings_gdf = pd.concat(
+        [important_building_gdf[["geometry"]], poi_buildings_gdf[["geometry"]]]
+    )
 
-    return non_res_uprns
+    return non_domestic_buildings_gdf
 
 
 def generate_gdf_non_residential_buildings(
@@ -153,11 +186,12 @@ def generate_gdf_non_residential_buildings(
         building_gdf, how="inner", predicate="within"
     )
 
-    # creating list of buildings from all important buildings and POI data with exactly 1 UPRN in them
-    all_uprns_non_res_gdf = generate_gdf_all_uprns_non_res(
+    # creating list of buildings from all important buildings and POI data with exactly 1 UPRN in them- these are likely to be fully commercial units
+    fully_commercial_buildings_gdf = _generate_gdf_fully_commercial_buildings(
         important_building_gdf=important_building_gdf,
         poi_gdf=poi_gdf,
         uprns_gdf=uprns_gdf,
+        building_gdf=building_gdf,
     )
 
     # add this to list of buildings to exclude
@@ -166,7 +200,7 @@ def generate_gdf_non_residential_buildings(
             exclude_buildings_gdf[["geometry"]],
             railway_station_gdf[["geometry"]],
             non_domestic_poi_gdf[["geometry"]],
-            all_uprns_non_res_gdf[["geometry"]],
+            fully_commercial_buildings_gdf[["geometry"]],
         ]
     )
 
--- a/asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
+++ b/asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
@@ -6,7 +6,6 @@ import geopandas as gpd
 import pandas as pd
 
 from asf_heat_pump_suitability.pipeline.transform import uprns
-from asf_heat_pump_suitability import config
 
 # TODO these may need refinement for large cities where overlap with residential is possible
 NO_RESIDENTIAL_OVERLAP_BUILDING_TYPES = [
@@ -34,7 +33,7 @@ NO_RESIDENTIAL_OVERLAP_BUILDING_TYPES = [
 ]
 
 
-def _generate_gdf_fully_commercial_buildings(
+def _generate_gdf_fully_non_domestic_buildings(
     important_building_gdf: gpd.GeoDataFrame,
     poi_gdf: gpd.GeoDataFrame,
     uprns_gdf: gpd.GeoDataFrame,
@@ -45,7 +44,7 @@ def _generate_gdf_fully_commercial_buildings(
     Args:
         important_building_gdf (gpd.GeoDataFrame): OS OpenMap Local important building footprints in area of interest
         poi_gdf (gpd.GeoDataFrame): All Points of Interest geopoints in area of interest
-        uprns_gdf (gpd.GeoDataFrame): UPRNs with point geometries in area of interest
+        uprns_gdf (gpd.GeoDataFrame): all UPRNs with point geometries in area of interest
         building_gdf (gpd.GeoDataFrame): all building footprints in area of interest
     Returns:
         gpd.GeoDataFrame: geometries of buildings which are in the important buildings list and have 1 UPRN in them, or in the POI list and have same number of UPRNs and POI
@@ -64,7 +63,7 @@ def _generate_gdf_fully_commercial_buildings(
     poi_buildings_gdf = gpd.GeoDataFrame(
         poi_buildings_df,
         geometry=poi_buildings_df["geometry"],
-        crs=config["constant"]["target_crs"],
+        crs=poi_buildings_gdf.crs,
     )
 
     # join important buildings and POI with all UPRNs
@@ -85,7 +84,7 @@ def _generate_gdf_fully_commercial_buildings(
     important_building_gdf = gpd.GeoDataFrame(
         important_building_df,
         geometry=important_building_df["geometry"],
-        crs=config["constant"]["target_crs"],
+        crs=important_building_gdf.crs,
     )
 
     # find number of UPRNs per building for POI buildings and convert result to gdf
@@ -101,7 +100,7 @@ def _generate_gdf_fully_commercial_buildings(
     poi_buildings_gdf = gpd.GeoDataFrame(
         poi_buildings_df,
         geometry=poi_buildings_df["geometry"],
-        crs=config["constant"]["target_crs"],
+        crs=poi_buildings_gdf.crs,
     )
 
     # select building footprints from important building data where UPRN count = 1
@@ -111,7 +110,7 @@ def _generate_gdf_fully_commercial_buildings(
 
     # select building footprints from POI data where UPRN count = POI count
     poi_buildings_gdf = poi_buildings_gdf.loc[
-        poi_buildings_gdf["UPRN_count"] == poi_buildings_gdf["POI_count"]
+        poi_buildings_gdf["UPRN_count"] <= poi_buildings_gdf["POI_count"]
     ]
 
     # concat gdfs and keep only geometry column
@@ -141,7 +140,7 @@ def generate_gdf_non_residential_buildings(
         non_domestic_poi_gdf (gpd.GeoDataFrame): non-domestic Points of Interest geopoints in area of interest, unlikely to be in mixed-use buildings
         poi_gdf (gpd.GeoDataFrame): All Points of Interest geopoints in area of interest
         building_gdf (gpd.GeoDataFrame): all building footprints in area of interest
-        uprns_gdf (gpd.GeoDataFrame): UPRNs with point geometries in area of interest
+        uprns_gdf (gpd.GeoDataFrame): all UPRNs with point geometries in area of interest
 
     Returns:
         gpd.GeoDataFrame: geometries of buildings which are unlikely to contain residential properties
@@ -187,7 +186,7 @@ def generate_gdf_non_residential_buildings(
     )
 
     # creating list of buildings from all important buildings and POI data with exactly 1 UPRN in them- these are likely to be fully commercial units
-    fully_commercial_buildings_gdf = _generate_gdf_fully_commercial_buildings(
+    fully_non_domestic_buildings_gdf = _generate_gdf_fully_non_domestic_buildings(
         important_building_gdf=important_building_gdf,
         poi_gdf=poi_gdf,
         uprns_gdf=uprns_gdf,
@@ -200,7 +199,7 @@ def generate_gdf_non_residential_buildings(
             exclude_buildings_gdf[["geometry"]],
             railway_station_gdf[["geometry"]],
             non_domestic_poi_gdf[["geometry"]],
-            fully_commercial_buildings_gdf[["geometry"]],
+            fully_non_domestic_buildings_gdf[["geometry"]],
         ]
     )
 
--- a/asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
+++ b/asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
@@ -37,6 +37,7 @@ def generate_gdf_non_residential_buildings(
     important_building_gdf: gpd.GeoDataFrame,
     railway_station_gdf: gpd.GeoDataFrame,
     poi_gdf: gpd.GeoDataFrame,
+    poi_unfiltered_gdf: gpd.GeoDataFrame,
     building_gdf: gpd.GeoDataFrame,
     uprns_gdf: gpd.GeoDataFrame,
 ) -> gpd.GeoDataFrame:
@@ -49,6 +50,7 @@ def generate_gdf_non_residential_buildings(
         important_building_gdf (gpd.GeoDataFrame): OS OpenMap Local important building footprints in area of interest
         railway_station_gdf (gpd.GeoDataFrame): OS OpenMap Local railway station point geometries in area of interest
         poi_gdf (gpd.GeoDataFrame): non-domestic Points of Interest geopoints in area of interest
+        poi_unfiltered_gdf (gpd.GeoDataFrame): All Points of Interest geopoints in area of interest
         building_gdf (gpd.GeoDataFrame): all building footprints in area of interest
         uprns_gdf (gpd.GeoDataFrame): UPRNs with point geometries in area of interest
 
@@ -63,6 +65,7 @@ def generate_gdf_non_residential_buildings(
                 important_building_gdf.crs,
                 railway_station_gdf.crs,
                 poi_gdf.crs,
+                poi_unfiltered_gdf.crs,
                 building_gdf.crs,
                 uprns_gdf.crs,
             }
@@ -86,6 +89,9 @@ def generate_gdf_non_residential_buildings(
         important_building_gdf[col].isin(NO_RESIDENTIAL_OVERLAP_BUILDING_TYPES)
     ]
     poi_buildings_gdf = building_gdf.sjoin(poi_gdf, how="inner", predicate="contains")
+    poi_unfiltered_gdf = building_gdf.sjoin(
+        poi_unfiltered_gdf, how="inner", predicate="contains"
+    ).drop("index_right", axis=1)
 
     # Get buildings which are railway stations (railway stations are only given as point geometries)
     railway_station_gdf = railway_station_gdf.sjoin(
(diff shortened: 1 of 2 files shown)


asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -49,6 +50,7 @@ def generate_gdf_non_residential_buildings(
         important_building_gdf (gpd.GeoDataFrame): OS OpenMap Local important building footprints in area of interest
         railway_station_gdf (gpd.GeoDataFrame): OS OpenMap Local railway station point geometries in area of interest
         poi_gdf (gpd.GeoDataFrame): non-domestic Points of Interest geopoints in area of interest
+        poi_unfiltered_gdf (gpd.GeoDataFrame): All Points of Interest geopoints in area of interest
```
crispy-wonton: question: have you checked the POI dataset and know that it doesn't contain any domestic POI? Just thinking, there's a possibility it has something in it that represents domestic only. Worth double checking - there are a lot of categories so you could use gemini.

suggestion: as `poi_unfiltered_gdf` represents the full POI set, I would just call this `poi_gdf` and then rename the non-domestic points one as something like `non_domestic_poi_gdf` and update the docstring to clarify that the non-domestic POI represents POI which are unlikely to be in mixed-use buildings. Even better would be to just have a single arg with the full POI dataset and then filter it inside this function.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -100,6 +106,39 @@ def generate_gdf_non_residential_buildings(
         ]
     )
 
+    # creating list of buildings from all important buildings and POI data with exactly 1 UPRN in them
```
crispy-wonton: suggestion: this function is quite big now, so I'd probably move this into a separate private function and then call it here.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -100,6 +106,39 @@ def generate_gdf_non_residential_buildings(
...
+    )
+    poi_unfiltered_gdf = (
+        poi_unfiltered_gdf.groupby("geometry").size().reset_index(name="UPRN_count")
+    )
+
+    # select buildings with 1 UPRN
+    important_building_gdf = important_building_gdf.loc[
+        important_building_gdf["UPRN_count"] == 1
+    ]
+    poi_unfiltered_gdf = poi_unfiltered_gdf.loc[poi_unfiltered_gdf["UPRN_count"] == 1]
```
crispy-wonton: suggestion: you could concatenate `important_building_gdf` and `poi_unfiltered_gdf` first so that you don't have to repeat each processing step twice.

crispy-wonton: suggestion: there's probably some building footprints which have multiple POI joined to them. This function could be improved further by removing POI buildings with an equal number of POIs and UPRNs.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -100,6 +106,39 @@ def generate_gdf_non_residential_buildings(
         ]
     )
```
crispy-wonton: suggestion: you can move this down and concatenate all the gdfs in one go. Let me know if I'm missing something and there's a reason to concat first tho!

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -100,6 +106,39 @@ def generate_gdf_non_residential_buildings(
...
+    poi_unfiltered_gdf = poi_unfiltered_gdf.loc[poi_unfiltered_gdf["UPRN_count"] == 1]
+
+    # add this to list of buildings to exclude
+    exclude_buildings_gdf = pd.concat(
+        [
+            exclude_buildings_gdf[["geometry"]],
+            important_building_gdf[["geometry"]],
+            poi_unfiltered_gdf[["geometry"]],
+        ]
+    )
```
crispy-wonton, line 204: suggestion: I imagine there's probably some duplicates at this point. It doesn't matter too much but for speed reasons it might be worth dropping them here.

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -162,19 +164,40 @@ def map_dict_uprns_to_building_id(
...
-        uprns_gdf.sjoin(buildings_gdf, how="inner", predicate=predicate)
+    epc_residential_uprns = load_set_valid_epc_uprns(epc_type="domestic")
+
+    # set of residential UPRNs not in the domestic epc list
+    uprns_non_epc_gdf = uprns_gdf[~(uprns_gdf["UPRN"].isin(epc_residential_uprns))]
+
+    # set of residential UPRNs in the domestic epc list
+    epc_gdf = uprns_gdf[(uprns_gdf["UPRN"].isin(epc_residential_uprns))]
+
+    # join non-epc domestic UPRNs to building footprints if they are located within them
```
crispy-wonton:
```suggestion
    # join non-epc domestic UPRNs to building footprints if they intersect with / are located within them
```

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -162,19 +164,40 @@ def map_dict_uprns_to_building_id(
...
+    uprns_non_epc_gdf = (
```
crispy-wonton:
```suggestion
    non_epc_uprn_building_dict = (
```
This is returning a dict

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -162,19 +164,40 @@ def map_dict_uprns_to_building_id(
...
+        uprns_non_epc_gdf.sjoin(buildings_gdf, how="inner", predicate=predicate)
+        .set_index("UPRN")
+        .to_dict()[id_col]
+    )
+
+    # join epc domestic UPRNs to nearest building footprint < max_distance (default 1m) away. They are usually inside a building
+    epc_gdf = (
```
crispy-wonton:
```suggestion
    epc_building_dict = (
```
This is returning a dict

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -162,19 +164,40 @@ def map_dict_uprns_to_building_id(
...
+        epc_gdf.sjoin_nearest(buildings_gdf, how="inner", max_distance=max_distance)
         .set_index("UPRN")
         .to_dict()[id_col]
     )
 
+    return gpd.GeoDataFrame(
+        pd.concat([uprns_non_epc_gdf, epc_gdf]), crs=config["constant"]["target_crs"]
+    )
+
```
crispy-wonton:
```suggestion
    return non_epc_uprn_building_dict | epc_building_dict

```
To merge the final dicts.

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -162,19 +164,40 @@ def map_dict_uprns_to_building_id(
...
+    epc_gdf = uprns_gdf[(uprns_gdf["UPRN"].isin(epc_residential_uprns))]
+
+    # join non-epc domestic UPRNs to building footprints if they are located within them
+    uprns_non_epc_gdf = (
+        uprns_non_epc_gdf.sjoin(buildings_gdf, how="inner", predicate=predicate)
+        .set_index("UPRN")
+        .to_dict()[id_col]
+    )
+
+    # join epc domestic UPRNs to nearest building footprint < max_distance (default 1m) away. They are usually inside a building
```
crispy-wonton: question: what happens in `sjoin_nearest` when the UPRN is located inside a building footprint?

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -162,19 +164,40 @@ def map_dict_uprns_to_building_id(
...
+        max_distance (float): max distance (metres) from which to join EPC UPRNs to a building footprint. Default 1m.
 
     Returns:
         dict: mapping of UPRNs to building IDs
     """
     uprns_gdf = geo_utils.verify_gdf_crs(uprns_gdf)
     buildings_gdf = geo_utils.verify_gdf_crs(buildings_gdf)
 
-    return (
-        uprns_gdf.sjoin(buildings_gdf, how="inner", predicate=predicate)
```
crispy-wonton, line 173: suggestion: I think it might give us marginal speed gains here if we do a left join of UPRNs to buildings, and then for all those which don't have a building, `sjoin_nearest` to the nearest building within `max_distance`.

thought: looking at this code now, I also think we don't need to specify that it must be EPC only. Where we use this function in the pipeline, it should only be EPC UPRNs which don't have a building footprint assigned to them - please could you check this is true? If that's the case, I think we can just `sjoin_nearest` for any UPRN outside of a building here, rather than just the EPC ones. This would also give us significant speed gains as we wouldn't have to load the EPC dataset.

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -240,6 +263,10 @@ def parse_arguments() -> argparse.Namespace:
         ).drop(columns="index_right")
 
     poi_gdf = load_tree_input.load_gdf_poi()
+    poi_unfiltered_gdf = poi.transform_gdf_poi(
+        poi_gdf,
+        filter_categories=None,
+    )
     poi_gdf = poi.transform_gdf_poi(
         poi_gdf,
         filter_categories=poi.load_set_non_domestic_poi_categories(),
```
crispy-wonton: suggestion: same comment regarding the naming of these dataframes as above.

crispy-wonton (changes requested):
hey @simran-dave !

Thanks for working on this code.

I've found one bug which is that the mapping UPRN to building ID function should output a dict rather than a geodataframe. 

I also want you to check the POI dataset (if you haven't already), to make sure there aren't any categories that could indicate a domestic building. See more detail about both of these in the comments. 

I've also left some questions and suggestions.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
simran-dave: I've had a look through the categories using Gemini. I think, even though some buildings have `main_category` as something that looks residential, the `alternate_category` looks to be non-residential. E.g. a building listed as an apartment is actually a holiday let. So I think using the full list is ok

asf_heat_pump_suitability/pipeline/transform/uprns.py
simran-dave: As far as I am aware, the distance is 0 and that building footprint is matched to the UPRN as it will be the nearest

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
simran-dave: I've changed the code to now find POI buildings with the same number of POIs and UPRNs in a building footprint- let me know if the logic makes sense.

simran-dave: Now that I am looking for the same number of POI and UPRNs in a building footprint, I won't add this as the processing step will now be different for the POI and important building lists.

asf_heat_pump_suitability/pipeline/transform/uprns.py
simran-dave: yep makes sense it would speed things up. Have now split into two steps. Are you able to clarify the second point? I thought that here we are joining all residential UPRNs to building footprints. And that we only want to join <1m away the confirmed domestic buildings- so that would be EPC only.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: Great thank you for checking!

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: Makes sense!

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -33,11 +33,67 @@
...
+) -> gpd.GeoDataFrame:
+    """
+    Function to create dataframe of buildings from the important buildings with exactly 1 UPRN in them, and buildings from POI data with equal number of POI and UPRNs. These are unlikely to be residential buildings
+    Args:
+        important_building_gdf (gpd.GeoDataFrame): OS OpenMap Local important building footprints in area of interest
+        poi_gdf (gpd.GeoDataFrame): All Points of Interest geopoints in area of interest
+        uprns_gdf (gpd.GeoDataFrame): UPRNs with point geometries in area of interest
+        building_gdf (gpd.GeoDataFrame): all building footprints in area of interest
+    Returns:
+        gpd.GeoDataFrame: geometries of buildings which are in the important buildings list and have 1 UPRN in them, or in the POI list and have same number of UPRNs and POI
```
crispy-wonton, line 50: praise: great docstring! really clear and succinct

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -33,11 +33,67 @@
...
+    """
+    # join POI data to building footprints
+    poi_gdf = building_gdf.sjoin(poi_gdf, how="inner", predicate="contains").drop(
+        "index_right", axis=1
+    )
```
crispy-wonton: suggestion: to rename the variable here as `poi_buildings_gdf` so that it's clear.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -33,11 +33,67 @@
...
+
+    # find number of UPRNs per building for important buildings
+    important_building_gdf = (
+        important_building_gdf.groupby("geometry").size().reset_index(name="UPRN_count")
+    )
+    # find number of UPRNs per building for POI buildings
+    uprn_counts = poi_gdf.groupby("geometry").size().reset_index(name="UPRN_count")
+
+    # join uprn counts as a column to POI dataframe
+    poi_gdf = poi_gdf.sjoin(uprn_counts, how="left", predicate="intersects")
```
crispy-wonton: suggestion: I would use a `.groupby("ID").agg()` here so that you don't have to join the UPRN counts back on. It's also easier to understand what the code is doing this way.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -33,11 +33,67 @@
...
+    poi_gdf = poi_gdf.sjoin(uprns_gdf, how="left", predicate="contains").drop(
+        "index_right", axis=1
+    )
+
+    # find number of UPRNs per building for important buildings
+    important_building_gdf = (
+        important_building_gdf.groupby("geometry").size().reset_index(name="UPRN_count")
+    )
+    # find number of UPRNs per building for POI buildings
+    uprn_counts = poi_gdf.groupby("geometry").size().reset_index(name="UPRN_count")
```
crispy-wonton: issue: not sure if this will actually happen, but as you are doing left joins above, I suppose you could end up here with an edge case of a building footprint containing no UPRNs, but the building footprint would be retained, giving you a `size` of 1. This is another reason to use `.groupby().agg()` here as outlined below.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -33,11 +33,67 @@
...
+    important_building_gdf = important_building_gdf.loc[
+        important_building_gdf["UPRN_count"] == 1
+    ]
+
+    # select building footprints from POI data where UPRN count = POI count
+    poi_gdf = poi_gdf.loc[poi_gdf["UPRN_count"] == poi_gdf["POI_count"]]
+
+    non_res_uprns = pd.concat([important_building_gdf, poi_gdf])
+
+    return non_res_uprns
```
crispy-wonton: question: have you checked that this is returning a `gdf` and not a `df`? Sometimes the `groupby` manipulations can convert a `gdf` to `df` and it may need to be reinstated.

nitpick: as this is returning buildings not UPRNs, I would call this something like `non_domestic_buildings_gdf`. I would also suggest only returning the geometry (as a GeoDataFrame, not GeoSeries) and not the other columns.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -33,11 +33,67 @@
 ]
 
 
+def generate_gdf_all_uprns_non_res(
```
crispy-wonton: suggestion: to rename this function to `_generate_gdf_non_domestic_buildings` or smth similar to be more accurate. The underscore before the name indicates it is a private function, i.e. a helper function that is only intended to be called from within another function.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -88,57 +144,36 @@ def generate_gdf_non_residential_buildings(
...
-        uprns_gdf, how="left", predicate="contains"
-    ).drop("index_right", axis=1)
-
-    # find number of UPRNs per building
-    important_building_gdf = (
-        important_building_gdf.groupby("geometry").size().reset_index(name="UPRN_count")
-    )
-    poi_unfiltered_gdf = (
-        poi_unfiltered_gdf.groupby("geometry").size().reset_index(name="UPRN_count")
+    all_uprns_non_res_gdf = generate_gdf_all_uprns_non_res(
```
crispy-wonton: suggestion: to rename this variable similar to the new function name.

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -156,47 +156,45 @@ def map_dict_uprns_to_building_id(
...
-    uprns_non_epc_gdf = uprns_gdf[~(uprns_gdf["UPRN"].isin(epc_residential_uprns))]
+    # first find domestic UPRNs inside buildings to speed up computation time.
+    uprns_inside_buildings = uprns_gdf.sjoin(
+        buildings_gdf, how="inner", predicate="intersects"
+    )
 
-    # set of residential UPRNs in the domestic epc list
-    epc_gdf = uprns_gdf[(uprns_gdf["UPRN"].isin(epc_residential_uprns))]
+    # find domestic UPRNs not joined to a buildings.
+    unmatched_uprns = uprns_gdf[~uprns_gdf.index.isin(uprns_inside_buildings.index)]
```
crispy-wonton: suggestion: to use the actual UPRNs here rather than the index.

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -156,47 +156,45 @@ def map_dict_uprns_to_building_id(
...
+        max_distance (float): max distance (metres) from which to join UPRNs to a building footprint. Default 1m.
 
     Returns:
         dict: mapping of UPRNs to building IDs
     """
     uprns_gdf = geo_utils.verify_gdf_crs(uprns_gdf)
     buildings_gdf = geo_utils.verify_gdf_crs(buildings_gdf)
 
-    epc_residential_uprns = load_set_valid_epc_uprns(epc_type="domestic")
+    # join domestic UPRNs to either the building footprint they are located within or building < max_distance (default 1m) away. They are usually inside a building.
```
crispy-wonton, line 175: praise: great idea to do it this way!

crispy-wonton (commented):
Have left some suggestions but the changes look great - thank you so much!

Question: have you run the `uprns.py` file to check that it still works with your changes?

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
simran-dave: I've gone with `_generate_gdf_fully_commercial_buildings` does that work?

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
simran-dave: have changed it to use the above, let me know if it looks ok

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
simran-dave: now that I'm using `.groupby().agg()` I imagine this won't be an issue?

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
simran-dave: have changed it to return a gdf instead of a df, and now it should only return the geometry column

simran-dave:
Checked whether `uprns.py` runs and found a bug where I wasn't passing the `building_gdf` to the function `_generate_gdf_fully_commercial_buildings()` - this has now been resolved and the code runs

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -50,43 +51,75 @@ def generate_gdf_all_uprns_non_res(
...
+    # count instances where multiple POI are located inside a building and convert result to gdf
+    poi_buildings_df = (
+        poi_buildings_gdf.groupby("ID")
+        .agg(POI_count=("ID", "count"), geometry=("geometry", "first"))
+        .reset_index()
+    )
+    poi_buildings_gdf = gpd.GeoDataFrame(
+        poi_buildings_df,
+        geometry=poi_buildings_df["geometry"],
+        crs=config["constant"]["target_crs"],
```
crispy-wonton: suggestion: I'd probably set `crs=poi_buildings_gdf.crs` here so that it's definitely using the correct CRS here, in case the original is different from the target. In the `uprns.py` pipeline it should be redundant because the poi gdf has been converted to the target CRS, but this function could be used outside the pipeline at some point for exploration or in another script.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
crispy-wonton: This is definitely a nitpick which I apologise for but can we use 'non-domestic' rather than 'commercial'. The reason I say this is because I was using the term 'commercial' myself previously when referring to these kinds of buildings but it actually caused confusion in the team because some non-domestic buildings are not technically 'commercial' e.g. churches, schools etc.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -33,9 +34,98 @@
...
+    # find number of UPRNs per building for important buildings and convert result to gdf
+    important_building_df = (
+        important_building_gdf.groupby("ID")
+        .agg(UPRN_count=("UPRN", "count"), geometry=("geometry", "first"))
+        .reset_index()
+    )
+    important_building_gdf = gpd.GeoDataFrame(
+        important_building_df,
+        geometry=important_building_df["geometry"],
+        crs=config["constant"]["target_crs"],
```
crispy-wonton: suggestion: as above, I'd probably refer to the original `important_building_gdf.crs` here.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -33,9 +34,98 @@
...
+            UPRN_count=("UPRN", "count"),
+            POI_count=("POI_count", "first"),
+            geometry=("geometry", "first"),
+        )
+        .reset_index()
+    )
+    poi_buildings_gdf = gpd.GeoDataFrame(
+        poi_buildings_df,
+        geometry=poi_buildings_df["geometry"],
+        crs=config["constant"]["target_crs"],
```
crispy-wonton: suggestion: same as above re the CRS

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -33,9 +34,98 @@
...
+
+    # select building footprints from important building data where UPRN count = 1
+    important_building_gdf = important_building_gdf.loc[
+        important_building_gdf["UPRN_count"] == 1
+    ]
+
+    # select building footprints from POI data where UPRN count = POI count
+    poi_buildings_gdf = poi_buildings_gdf.loc[
+        poi_buildings_gdf["UPRN_count"] == poi_buildings_gdf["POI_count"]
+    ]
```
crispy-wonton, line 114: suggestion: let me know if you don't agree with my logic here but I'm thinking if there's 2 POI and one UPRN, for example, then it should still be selected because it probably means the POI is duplicated for some reason. What do you think? Can you think of a reason not to do this that I've missed?

```suggestion
    # select building footprints from POI data where UPRN count = POI count
    poi_buildings_gdf = poi_buildings_gdf.loc[
        poi_buildings_gdf["UPRN_count"] <= poi_buildings_gdf["POI_count"]
    ]
```

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -33,9 +34,98 @@
...
+    poi_gdf: gpd.GeoDataFrame,
+    uprns_gdf: gpd.GeoDataFrame,
+    building_gdf: gpd.GeoDataFrame,
+) -> gpd.GeoDataFrame:
+    """
+    Function to create dataframe of buildings from the important buildings with exactly 1 UPRN in them, and buildings from POI data with equal number of POI and UPRNs. These are unlikely to be residential buildings
+    Args:
+        important_building_gdf (gpd.GeoDataFrame): OS OpenMap Local important building footprints in area of interest
+        poi_gdf (gpd.GeoDataFrame): All Points of Interest geopoints in area of interest
+        uprns_gdf (gpd.GeoDataFrame): UPRNs with point geometries in area of interest
```
crispy-wonton:
```suggestion
        uprns_gdf (gpd.GeoDataFrame): all UPRNs with point geometries in area of interest
```

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
```diff
@@ -153,11 +186,12 @@ def generate_gdf_non_residential_buildings(
         building_gdf, how="inner", predicate="within"
     )
 
-    # creating list of buildings from all important buildings and POI data with exactly 1 UPRN in them
-    all_uprns_non_res_gdf = generate_gdf_all_uprns_non_res(
+    # creating list of buildings from all important buildings and POI data with exactly 1 UPRN in them- these are likely to be fully commercial units
+    fully_commercial_buildings_gdf = _generate_gdf_fully_commercial_buildings(
```
crispy-wonton: Have put this comment in response to the original thread about this, but posting it here as well as the original thread is outdated and you may not see it.

This is definitely a nitpick which I apologise for but can we use 'non-domestic' rather than 'commercial'. The reason I say this is because I was using the term 'commercial' myself previously when referring to these kinds of buildings but it actually caused confusion in the team because some non-domestic buildings are not technically 'commercial' e.g. churches, schools etc.

crispy-wonton (commented):
Hey @simran-dave !

Great changes thank you - the function you've written to generate non-domestic buildings is really neat and easy to understand.

I've left a few suggestions, mainly on CRS setting and a small suggestion for the logic of identifying non-domestic buildings from the POI.

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
simran-dave: yeah makes sense!

asf_heat_pump_suitability/pipeline/transform/non_residential_entities.py
simran-dave: yeah that's fair! have changed it

crispy-wonton (approved):

PR #287, opened 2026-04-02
216 council tax domestic filtering

### LINKED ISSUE #216: Improve filtering of domestic UPRNs

simran-dave:
---

# Description

Added a notebook that compares council tax data to UPRNs identified as domestic in the pipeline. Follows the steps outlined in the domestic filtering exploration tasks document.

Fixes #216 

# Instructions for Reviewer

In order to test the code in this PR you need to run asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
+++ b/asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
@@ -247,9 +247,23 @@ plot_buildings(
 
 # %%
 # join council tax data to building footprints. These are all buildings with at least one comfirmed domestic property
+"""
 council_with_buildings = building_footprints_gdf.sjoin(
     council_tax_gdf, how="inner", predicate="contains"
 ).drop("index_right", axis=1)
+"""
+
+# %%
+# joining council tax points to nearest building < 3m away
+# for some reason using the max_distance parameter causes an error here so using this manual workaround for now
+council_with_buildings = council_tax_gdf.sjoin_nearest(
+    building_footprints_gdf, how="inner", distance_col="dist"
+)
+council_with_buildings = council_with_buildings[council_with_buildings["dist"] <= 3]
+council_with_buildings = council_with_buildings.drop(["index_right", "dist"], axis=1)
+
+# %%
+council_with_buildings
 
 # %%
 # aggregate features by building ID
@@ -364,9 +378,10 @@ def plot_pipeline_mixed_buildings(
                 "fillOpacity": 0.3,
             },
         )
-        folium.Popup(f"pipeline UPRNs: {r['pipeline_UPRN_count']}").add_to(geo_j)
-        folium.Popup(f"domestic UPRNs: {r['council_UPRN_count']}").add_to(geo_j)
-        folium.Popup(f"total UPRNs: {r['total_UPRN_count']}").add_to(geo_j)
+        folium.Popup(
+            f"pipeline UPRNs: {r['pipeline_UPRN_count']} <br> domestic UPRNs: {r['council_UPRN_count']} <br> total UPRNs: {r['total_UPRN_count']}",
+            max_width=200,
+        ).add_to(geo_j)
         geo_j.add_to(m)
 
     return m
@@ -451,6 +466,7 @@ def plot_wrong_uprn_count(
                     "fillOpacity": 0.3,
                 },
             )
+
         if r["pipeline_UPRN_count"] < r["council_UPRN_count"]:
             geo_j = folium.GeoJson(
                 data=geo_j,
@@ -460,6 +476,9 @@ def plot_wrong_uprn_count(
                     "fillOpacity": 0.3,
                 },
             )
+        folium.Popup(
+            f"pipeline UPRNs minus council tax UPRNs: {r['UPRN_difference']}"
+        ).add_to(geo_j)
         geo_j.add_to(m)
     return m
 
@@ -468,6 +487,61 @@ def plot_wrong_uprn_count(
 # blue = we are identifying more UPRNs than in council tax data
 # red = we are identifying fewer UPRNs than in council tax data
 plot_wrong_uprn_count(wrong_uprn_number, la_boundaries_gdf)
+
+# %%
+import branca.colormap as cm
+
+
+def plot_heat_map_wrong_uprn_count(
+    buildings_gdf: gpd.GeoDataFrame, boundary_gdf: gpd.GeoDataFrame
+):
+    """
+    Plot the building footprint polygons where we are identiifying the wrong number of UPRNs (not including cases where the whole building is wrong)
+
+    Args:
+        residential_buildings_gdf (gpd.GeoDataFrame): dataframe of buildings assigned as residential, joined to building footprint polygons
+        boundary_gdf (gpd.GeoDataFrame): dataframe with LA boundaries
+        area (float): area per UPRN above which you want to plot
+
+    Returns Folium map
+    """
+    buildings_gdf_4326 = buildings_gdf.to_crs(epsg=4326)
+    # Get centre of boundary to centre map
+    boundary_4326 = boundary_gdf.to_crs(epsg=4326)["geometry"].values[0]
+    centre_map = shapely.get_coordinates(boundary_4326.centroid)
+
+    # Create map
+    m = folium.Map(
+        location=[centre_map[0][1], centre_map[0][0]],
+        zoom_start=15,
+        tiles="esri_worldimagery",
+    )
+
+    # Plot building footprints
+    min = np.min(buildings_gdf_4326["UPRN_difference"])
+    max = np.max(buildings_gdf_4326["UPRN_difference"])
+    colours = cm.linear.YlOrRd_03.scale(min, max)
+    for _, r in buildings_gdf_4326.iterrows():
+        sim_geo = gpd.GeoSeries(r["geometry"])
+        geo_j = sim_geo.to_json()
+        geo_j = folium.GeoJson(
+            data=geo_j,
+            style_function=lambda x, colour=colours(r["UPRN_difference"]): {
+                "fillColor": colour,
+                "weight": 0.1,
+                "fillOpacity": 0.7,
+            },
+        )
+        folium.Popup(
+            f"pipeline UPRNs minus council tax UPRNs: {r['UPRN_difference']}"
+        ).add_to(geo_j)
+        geo_j.add_to(m)
+    return m
+
+
+# %%
+# heat map of UPRN difference
+plot_heat_map_wrong_uprn_count(wrong_uprn_number, la_boundaries_gdf)
 
 # %% [markdown]
 # ## Identify features to classify as domestic / non-domestic
--- a/asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
+++ b/asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
@@ -22,13 +22,10 @@ from asf_heat_pump_suitability.getters import (
     load_boundaries,
     base_getters,
 )
-from asf_heat_pump_suitability.pipeline.transform import uprns, poi
-from asf_heat_pump_suitability.pipeline.transform import outdoor_space
+from asf_heat_pump_suitability.pipeline.transform import uprns
 
 # %%
-council_tax_gdf = gpd.read_file(
-    config["data"]["geodata"]["council_tax_data"]["plymouth"]
-)
+council_tax_df = pd.read_csv(config["data"]["geodata"]["council_tax_data"]["plymouth"])
 pipeline_domestic_gdf = base_getters.load_df_from_s3(
     config["data"]["processed"]["plymouth_residential_uprns"]
 )
@@ -38,42 +35,56 @@ building_footprints_gdf = load_tree_input.load_gdf_os_openmap_local_layer(
 la_boundaries_gdf = load_boundaries.load_gdf_local_authority_boundaries(
     select_las="Plymouth"
 )
+uprns_df = load_geodata.load_df_osopen_uprn()
+uprns_gdf = uprns.generate_gdf_uprn_coords(uprns_df)
 
 # %%
+building_footprints_gdf = building_footprints_gdf.sjoin(
+    la_boundaries_gdf[["LAD23CD", "LAD23NM", "geometry"]],
+    how="inner",
+    predicate="intersects",
+).drop(columns="index_right")
+
+# %%
+council_tax_df
+
+# %%
+# total number of council tax records (before empty UPRN/ coordinate rows are removed)
+n_council_tax_records = len(council_tax_df)
+
 # remove empty UPRN rows
-council_tax_gdf = council_tax_gdf[council_tax_gdf["UPRN"] != ""]
+council_tax_df = council_tax_df[council_tax_df["UPRN"] != ""]
 # remove empty coordinate rows
-council_tax_gdf = council_tax_gdf[council_tax_gdf["EASTING"] != ""]
-council_tax_gdf = council_tax_gdf[council_tax_gdf["NORTHING"] != ""]
+council_tax_df = council_tax_df[council_tax_df["EASTING"] != ""]
+council_tax_df = council_tax_df[council_tax_df["NORTHING"] != ""]
 
 # %%
+# convert council tax data to gdf
 council_tax_gdf = gpd.GeoDataFrame(
-    council_tax_gdf,
-    geometry=gpd.points_from_xy(
-        council_tax_gdf["EASTING"], council_tax_gdf["NORTHING"]
-    ),
+    council_tax_df,
+    geometry=gpd.points_from_xy(council_tax_df["EASTING"], council_tax_df["NORTHING"]),
     crs="EPSG:27700",
-)
+).drop_duplicates("UPRN")
+
+# convert pipeline UPRN data to gdf
 pipeline_domestic_gdf = uprns.generate_gdf_uprn_coords(pipeline_domestic_gdf)
+pipeline_domestic_gdf = pipeline_domestic_gdf.drop_duplicates("UPRN")
 
 # %% [markdown]
 # ## How many UPRNs are in each dataset
 
 # %%
-council_tax_gdf
+# total number of council tax records (including empty UPRN/ coordinate rows)
+print(f"total number of records in council tax data: {n_council_tax_records}")
 
 # %%
-council_tax_gdf = council_tax_gdf[council_tax_gdf["UPRN"] != ""]
-
-# %%
-pipeline_domestic_gdf
-
-# %%
+# number of UPRNs with a coordinate in council tax data
 print(
     f"number of unique URPNs in council tax data: {(council_tax_gdf['UPRN']).nunique()}"
 )
 
 # %%
+# number of UPRNs identified as domestic in pipeline data
 print(
     f"number of unique URPNs in pipeline data: {(pipeline_domestic_gdf['UPRN']).nunique()}"
 )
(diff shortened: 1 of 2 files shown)


asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -0,0 +1,611 @@
...
+from asf_heat_pump_suitability import PROJECT_DIR
+from asf_heat_pump_suitability import config
+from asf_heat_pump_suitability.getters import (
+    load_tree_input,
+    load_geodata,
+    load_boundaries,
+    base_getters,
+)
+from asf_heat_pump_suitability.pipeline.transform import uprns, poi
+from asf_heat_pump_suitability.pipeline.transform import outdoor_space
```
crispy-wonton:
```suggestion
from asf_heat_pump_suitability.pipeline.transform import uprns
```
Note that `outdoor_space` and `poi` modules are not currently accessed. But I appreciate this is a WIP

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -0,0 +1,611 @@
...
+
+# %%
+council_tax_gdf = gpd.read_file(
+    config["data"]["geodata"]["council_tax_data"]["plymouth"]
+)
```
crispy-wonton: Issue: since this is geodata but doesn't have a geometry column, I would recommend reading this with `pandas` instead of `geopandas`, then convert to a `GeoDataFrame` after, as you have done with your code below.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -0,0 +1,611 @@
...
+pipeline_domestic_gdf = uprns.generate_gdf_uprn_coords(pipeline_domestic_gdf)
+
+# %% [markdown]
+# ## How many UPRNs are in each dataset
+
+# %%
+council_tax_gdf
+
+# %%
+council_tax_gdf = council_tax_gdf[council_tax_gdf["UPRN"] != ""]
```
crispy-wonton: suggestion: you can remove this line as this was already done above

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -0,0 +1,611 @@
...
+
+# %%
+print(
+    f"number of unique URPNs in council tax data: {(council_tax_gdf['UPRN']).nunique()}"
+)
+
+# %%
+print(
+    f"number of unique URPNs in pipeline data: {(pipeline_domestic_gdf['UPRN']).nunique()}"
+)
```
crispy-wonton, line 90: suggestion: for comparison against our pipeline, I would also compare the total number of council tax records - not just the number of UPRNs.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -0,0 +1,611 @@
...
+# %%
+council_in_pipeline
+
+# %%
+print(
+    f"number of domestic UPRNs in council tax data not picked up by pipeline: {len(council_tax_gdf) - len(council_in_pipeline)}"
+)
+print(
+    f"proportion of domestic UPRNs in council tax data not picked up by pipeline: {(len(council_tax_gdf) - len(council_in_pipeline))*100/ len(council_tax_gdf):.2f}%"
+)
```
crispy-wonton, line 112: suggestion: in general throughout this notebook I'd probably use `set`s here and set differences, just in case there are things like duplicates. You don't need to change it all here but just for future reference. However, I would still add a couple of lines to deduplicate the UPRNs in both the council tax and pipeline datasets in case there are any.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -0,0 +1,611 @@
...
+
+# %% [markdown]
+# ## Pipeline domestic UPRNs not in council tax data
+# These are probably not actually domestic
```
crispy-wonton: thought: there's a note here that because there are some council tax records without a matching UPRN, it's possible that some of the UPRNs in the pipeline are truly domestic but not being verified here. I would add a note about that here. Alternatively, there could be more true domestic UPRNs that are NOT being picked up by the pipeline, but we're unable to identify those for the same reason.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -0,0 +1,611 @@
...
+
+# %% [markdown]
+# ## Pipeline domestic building footprints with no council tax UPRNs in them
+#
+# Fully commercial units
+
+# %%
+council_with_buildings = building_footprints_gdf.sjoin(
+    council_tax_gdf, how="inner", predicate="contains"
+).drop("index_right", axis=1)
```
crispy-wonton: note: when joining both the council tax UPRNs and the pipeline UPRNs to the building footprints, we lose a portion of the UPRNs. About 2.5% of council tax UPRNs (in addition to the ones already lost from not having a UPRN) and about 1.5% of pipeline UPRNs. The pipeline ones must all be EPC records. But the council tax ones is interesting because it gives us a sense of how many we might lose in our pipeline from the step in the filtering that says domestic UPRNs must be located within building footprints.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -0,0 +1,611 @@
...
+# %%
+# list of building footprints we identify as domestic with at least 1 UPRN in council tax data
+pipeline_buildings_in_council_tax = pipeline_with_buildings[
+    pipeline_with_buildings["UPRN"].isin(council_tax_gdf["UPRN"].tolist())
+]
+
+# %%
+print(
+    f"proportion of buildings footprints with at least 1 council tax UPRN: {100*len(pipeline_buildings_in_council_tax.drop_duplicates('geometry'))/len(pipeline_with_buildings.drop_duplicates('geometry')):.2f}%"
+)
```
crispy-wonton: suggestion: generally it's better to avoid deduplication on geometry columns where possible because the same geometry can be differently ordered and thus not register as a duplicate. I'd use building `ID` here and other relevant ID columns where you've deduplicated in other parts of the notebook. This goes for other manipulations as well e.g. the `groupby` etc. Generally I recommend using geometry for these kind of manipulations only when there is no unique ID.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -0,0 +1,611 @@
...
+    f"Proportion of building footprints with no council tax UPRNs in them: {len(uprn_counts)*100/len(pipeline_with_buildings.drop_duplicates('geometry')):.2f}%"
+)
+
+
+# %%
+def plot_buildings(
+    pipeline_buildings_gdf: gpd.GeoDataFrame, boundary_gdf: gpd.GeoDataFrame
+):
+    """
+    Plot the building footprint polygons with area per UPRN > area specified in function
```
crispy-wonton: issue: this function doesn't take area as an arg

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -0,0 +1,611 @@
...
+
+    Args:
+        residential_buildings_gdf (gpd.GeoDataFrame): dataframe of buildings assigned as residential, joined to building footprint polygons
+        boundary_gdf (gpd.GeoDataFrame): dataframe with LA boundaries
+        area (float): area per UPRN above which you want to plot
```
crispy-wonton: issue: `residential_buildings_gdf` and `area` aren't args in this function.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -0,0 +1,611 @@
...
+# %%
+pipeline_buildings_no_uprn = pipeline_buildings_no_uprn.drop_duplicates("geometry")
+
+# %%
+uprn_counts = (
+    pipeline_buildings_no_uprn.groupby("geometry").size().reset_index(name="UPRN_count")
+)
+pipeline_buildings_no_uprn = gpd.GeoDataFrame(
+    uprn_counts, geometry="geometry", crs="EPSG:27700"
+)
```
crispy-wonton: thought: something that strikes me as odd here is that they all contain only 1 UPRN...I can't see any errors in the code causing this but would be worth you double checking as well because I would expect some of them - especially the larger buildings to have multiple UPRNs. Is it possible you've grouped a groupby?

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -0,0 +1,611 @@
...
+plot_buildings(pipeline_buildings_no_uprn, la_boundaries_gdf)
+
+# %% [markdown]
+# ## Mixed-use buildings
+
+# %%
+# buildings with at least 1 UPRN in council tax data (domestic) and at least 1 UPRN not in council tax data (commercial)
+mixed_use_gdf = pipeline_buildings_not_in_council_tax.sjoin(
+    pipeline_buildings_in_council_tax, how="inner", predicate="intersects"
+)
```
crispy-wonton: issue: because of the variable names here, this is quite hard to understand. It seems like you're intersecting two gdfs which should be mutually exclusive. I would suggest either renaming variables to indicate that they are UPRNs, not just buildings, or just redoing the spatial joins here to create them in the first place.

In any case, I would suggest here not to create the mixed use gdf like this, but instead by doing the following:

1. Join council tax UPRNs to footprints to get buildings containing domestic properties
2. Get count of domestic UPRNs per building
3. Join all UPRNs from OS Open UPRN in Plymouth to building footprints and get a total UPRN count per footprint
4. Merge the dfs so that you get count of domestic UPRNs and count of all UPRNs per building footprint
5. Filter to footprints where N domestic UPRNs != N total UPRNs

I think this gives us a more reliable picture of mixed-use buildings rather than using any information from our own pipeline, especially because we have removed UPRNs appearing in the non-domestic EPC register from our pipeline UPRNs.

suggestion: there's also a lot of duplicated columns here with `_left` and `_right` suffixes. To clean it up, I would suggest only joining together the columns you need in analysis.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -0,0 +1,611 @@
...
+
+# %%
+mixed_use_gdf.drop_duplicates("geometry")
```
crispy-wonton:
```suggestion
# Drop duplicate building footprints to get one row per building
mixed_use_gdf = mixed_use_gdf.drop_duplicates("ID_left")
```
suggestion: or are you not intending to save this change to the df?

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -0,0 +1,611 @@
...
+
+# %%
+mixed_use_with_uprn_counts
+
+# %%
+mixed_use_with_uprn_counts[
+    mixed_use_with_uprn_counts["pipeline_UPRN_count"]
+    == mixed_use_with_uprn_counts["council_tax_UPRN_count"]
+]
+# why would we have some with the same number ???? unless the UPRNs are wrong ? or something here is wrong
```
crispy-wonton: thought: I think something is probably going wrong earlier with the creation of the `pipeline_buildings_in_council_tax` or `pipeline_buildings_not_in_council_tax` dataframes. I don't know what it is as I've found it a bit difficult to follow all the manipulations through the notebook.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -0,0 +1,611 @@
...
+            )
+        geo_j.add_to(m)
+
+    return m
+
+
+# %%
+# mixed use buildings
+# blue = we are identifying more UPRNs as domestic than are in council tax data
+# red = we are identifying fewer UPRNs as domestic than are in council tax data
```
crispy-wonton: suggestion: I remember you mentioned that in most cases for red buildings, the number of UPRNs in the pipeline was only marginally less than in the council tax data. I would add some documentation here explaining that - and code to show it - with the speculation that this is likely because of mapping errors meaning that our pipeline misses UPRNs just outside of building footprints.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -0,0 +1,611 @@
...
+# orange = same number (not sure why this is happening)
+plot_mixed_use_buildings(mixed_use_with_uprn_counts, la_boundaries_gdf)
+
+# %%
+h = plt.hist(
+    mixed_use_with_uprn_counts["proportion_domestic_UPRN"], bins=100, range=(0, 5)
+)
+
+# %% [markdown]
+# ## Identify features to classify as domestic / non-domestic
```
crispy-wonton, line 547: thought: It would be useful to see how these results change if we focus on the subset of buildings where our model is failing on. Does it help us remove additional non-domestic buildings, or are we already removing them?

crispy-wonton (commented):
Thanks so much for this work @simran-dave ! Some great findings in here to give us more confidence in our existing model and help identify where we could make improvements.

It would be great to have the findings from this analysis for each question written up into a slidedeck or in the original exploration document.

In some places this became a bit hard to follow due to long variable names and little annotation. I think adding a lot more annotation of what you're doing when you're transforming a dataframe would be really helpful. 

I've left further comments and suggestions on the code.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
simran-dave: Oh, interesting. I can have a look at how many council tax UPRNs are within 1m of a building for a verification of that step

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
simran-dave: yeah sorry I've not explained this bit well! Here I am finding buildings that have at least 1 UPRN in the council tax data and at least 1 UPRN not in the council tax data. And yeah, this is just looking at mixed-use buildings within the pipeline but makes sense to look at all of the data

simran-dave: I've changed the code to use this method to identify mixed-use buildings and then look at pipeline buildings that are part of this mixed-use list. Let me know if the updated logic makes sense.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
simran-dave: yeah that makes sense, I've changed all the `groupby` to use ID instead

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
simran-dave: yeah I think something must have been going wrong here. I'm now I'm using the updated logic for the mixed-use buildings so this these dataframes aren't being created any more

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
simran-dave: Now that I've changed the code a bit this is not happening any more

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
simran-dave: have added some lines in looking at this

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
simran-dave: added some code in to show this

simran-dave:
Thanks for the review @crispy-wonton ! The suggestions were really helpful- I've made quite a few changes in the way the dataframes are being made and added more comments to hopefully make things clearer. I've also changed the mixed-use gdfs to use the method you suggested here.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
crispy-wonton: Cool - this might also be useful to include in the methodology documentation again to contextualise our error rate. Might be too much detail but it could be useful to add.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -231,99 +228,196 @@ def plot_buildings(
...
-pipeline_buildings_not_in_council_tax
+pipeline_mixed_use_gdf
 
 # %%
-mixed_use_with_uprn_counts = pipeline_buildings_in_council_tax.sjoin(
-    pipeline_buildings_not_in_council_tax, how="inner", predicate="intersects"
-).drop("index_right", axis=1)
+print(
+    f"proportion of pipeline building footprints which are mixed use: {100*len(pipeline_mixed_use_gdf)/len(pipeline_with_buildings):.2f}%"
+)
```
crispy-wonton, line 342: Nice that this is giving us very similar proportions to the council tax data!

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -231,99 +228,196 @@ def plot_buildings(
...
+        geo_j = folium.GeoJson(
+            data=geo_j,
+            style_function=lambda x, colour="blue": {
+                "fillColor": colour,
+                "fillOpacity": 0.3,
+            },
+        )
+        folium.Popup(f"pipeline UPRNs: {r['pipeline_UPRN_count']}").add_to(geo_j)
+        folium.Popup(f"domestic UPRNs: {r['council_UPRN_count']}").add_to(geo_j)
+        folium.Popup(f"total UPRNs: {r['total_UPRN_count']}").add_to(geo_j)
```
crispy-wonton: issue (non-blocking): I think you're overwriting the pop up each time here and need to enter the pop up info as one function. I've plotted the map and only the `total UPRNs` are appearing.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -231,99 +228,196 @@ def plot_buildings(
...
+        geo_j.add_to(m)
+
+    return m
+
 
 # %%
-mixed_use_with_uprn_counts.drop_duplicates("geometry")
+# mixed use buildings in the pipeline
+plot_pipeline_mixed_buildings(pipeline_mixed_use_gdf, la_boundaries_gdf)
```
crispy-wonton, line 392: thought: really interesting to map this out. It's worth showing this map in a sprint review when you're next presenting. It's interesting that the mixed use buildings tend to be those towards the city centre (which we would expect I guess) - but specifically this means that the accuracy of our UPRN data is spatially variable!

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -231,99 +228,196 @@ def plot_buildings(
...
+
+# %% [markdown]
+# ## Where are we identifying the wrong number of UPRNs?
+# Not including cases where the whole building is misidentified
```
crispy-wonton, line 396: thought: I'm wondering if in the next phase we can use the census counts of households per area to improve our model - e.g. some kind of distribution of the census counts across buildings that our pipeline estimates as containing domestic with the goal of reducing the overestimation of N domestic UPRNs.

Not to address now, just noting for future ideas.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -366,31 +460,14 @@ def plot_mixed_use_buildings(
...
-        if r["pipeline_UPRN_count"] == r["council_tax_UPRN_count"]:
-            geo_j = folium.GeoJson(
-                data=geo_j,
-                style_function=lambda x, colour="orange": {
-                    "fillColor": colour,
-                    "weight": 0.1,
-                    "fillOpacity": 0.3,
-                },
-            )
         geo_j.add_to(m)
```
crispy-wonton, line 482: suggestion: would be great to add a pop-up with the diff counts here.
 
I just added the following line above here and it works:
`folium.Popup(f"pipeline UPRNs minus council tax UPRNs: {r['UPRN_difference']}").add_to(geo_j)`

thought: This doesn't need to be done now unless you know how to do it quickly but it would also be interesting to plot a heat map of this. I.e. to see if there's a spatial pattern to how inaccurate the UPRN count is. We could do this as an addition in the next phase.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -604,8 +649,61 @@ def plot_building_feature(
...
+
+# %%
+# for these UPRNs, find the nearest building footprint
+council_outside_buildings = council_outside_buildings.sjoin_nearest(
+    building_footprints_gdf, how="inner", distance_col="distance"
+)
+
+# %%
+# what is the distribution of distance to nearest footprint for these UPRNs
+council_outside_buildings["distance"].describe()
```
crispy-wonton, line 776: thought: I think within 3-5m of a building it would be acceptable to join the council tax UPRNs to a building for this analysis. 82% of these UPRNs outside buildings get assigned to one within 3m, versus 92% within 5m. I'd consider updating the earlier `sjoin` to create your council tax buildings for the analysis to an `sjoin_nearest` with one of these distances and then rerun the notebook to see how it changes the results.

```
print(len(council_outside_buildings[council_outside_buildings["distance"] <=3]) / len(council_outside_buildings))
print(len(council_outside_buildings[council_outside_buildings["distance"] <=5]) / len(council_outside_buildings))
```

crispy-wonton (commented):
Thanks so much @simran-dave the changes are really great! The analysis is very clear and readable and I think you looked into some really interesting things here. Well done!

I have left a couple of thoughts for your consideration, though nothing major. The main one is a suggestion to test out changing the original `sjoin` between council tax UPRNs and buildings to an `sjoin_nearest` with a max distance of up to 5m. This might give us results that are closer to the ground truth. What do you think? Maybe i've missed a reason not to do this.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
simran-dave: yeah can show this in a review! I think it is nice that the mixed-use buildings in the pipeline follow the expected behaviour of more being in the city centre. But yeah, I guess this means the accuracy of any UPRN based data decreases in city centres.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
simran-dave: This sounds like a good idea!

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
simran-dave: I've added a first attempt at at heat map

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
simran-dave: have changed to join council tax UPRNs to footprints < 3m away

asf_heat_pump_suitability/research/exploratory/domestic_filtering/council_tax_domestic_filtering.py
```diff
@@ -364,9 +378,10 @@ def plot_pipeline_mixed_buildings(
                 "fillOpacity": 0.3,
             },
         )
-        folium.Popup(f"pipeline UPRNs: {r['pipeline_UPRN_count']}").add_to(geo_j)
-        folium.Popup(f"domestic UPRNs: {r['council_UPRN_count']}").add_to(geo_j)
-        folium.Popup(f"total UPRNs: {r['total_UPRN_count']}").add_to(geo_j)
+        folium.Popup(
+            f"pipeline UPRNs: {r['pipeline_UPRN_count']} <br> domestic UPRNs: {r['council_UPRN_count']} <br> total UPRNs: {r['total_UPRN_count']}",
```
crispy-wonton, line 382: oh nice! I never figured out how to make the line breaks work - great :D

crispy-wonton (approved):
Thanks @simran-dave ! The changes look great - thanks for taking the time to make the heat map! It's interesting to see.

All good to merge!

PR #292, opened 2026-04-13
add new contextual features

### LINKED ISSUE #283: Add new contextual features:  listed buildings, number of off-gas properties, coastline, conservation area

sofiapinto:
---

# Description

This PR adds the following new contextual features to .add_features.py:
- SOLAR_WATER_HEATING_FLAG
- ENERGY_CONSUMPTION_CURRENT and ENERGY_CONSUMPTION_POTENTIAL
- Listed building flag
- Off-gas flag
- Anchor load flag
- Near coastline flag
- Conservation area flag

Fixes #283 

# Instructions for Reviewer

Hey @crispy-wonton ,

Could you please focus on whether the logic holds in the following code? Don’t worry about how the code is organised for now, as I haven’t had time of focus on that yet (moving functions/scripts/refactoring code etc). Thanks!

For now, I just need someone else to check the logic. Do note that: listed buildings & near coastline parts aren’t working properly, but I haven’t had time to investigate why.

Data/code for the features:
- SOLAR_WATER_HEATING_FLAG: EPC. Looking at EPC columns, I think this is the only one that makes sense?
- ENERGY_CONSUMPTION_CURRENT and ENERGY_CONSUMPTION_POTENTIAL: from EPC, to be used in the absence of heat demand - have asked Andy if any of these are suitable, and waiting for him to come back
- Listed building flag: used your existing code
- Off-gas flag: used your existing code
- Anchor load flag: I used some of yours and Simran!
- Near coastline flag: used your existing code
- Conservation area flag: used your existing code, but had to collect National Statistics UPRN Lookup data (https://geoportal.statistics.gov.uk/datasets/4e0b4b3fbc2540caae27e7be532e61be/about)

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/config/base.yaml
```diff
@@ -32,6 +32,7 @@ data:
   processed:
     non_domestic_poi_categories: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/processed/non_residential_poi_categories.csv"
     plymouth_residential_uprns: "s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_residential_uprns.parquet"
+    poi_anchor_properties: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/processed/poi_anchor_properties_gb.geojson"
```
sofiapinto, line 52: copied from from Simran's PR

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -146,3 +148,83 @@ def load_gdf_poi() -> gpd.GeoDataFrame:
     ).to_crs("EPSG:4326")
     print(f"POI CRS: {poi.crs}")
     return poi
+
+
+def load_gdf_os_openmap_layer(
```
sofiapinto, line 190: copied from Roisin's PR

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -274,18 +275,264 @@ def parse_arguments() -> argparse.Namespace:
...
+
+    # bucket_name = 'asf-heat-pump-suitability'
+    # prefix = 'local_heat_planning/inputs/geodata/NSUL_DEC_2025/'
+
+    # response = s3_client.list_objects_v2(Bucket=bucket_name, Prefix=prefix)
+    # files = [
+    #     f"s3://{bucket_name}/{obj['Key']}"
+    #     for obj in response.get('Contents', [])
+    #     if obj['Key'].endswith('.csv')
+    # ]
```
sofiapinto: The data has been slowly uploading to S3, hence why this is currently commented out. The local version is below

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -274,18 +275,264 @@ def parse_arguments() -> argparse.Namespace:
...
+
+    # uprn_to_country_df = pd.concat([pd.read_csv(file, usecols=["UPRN", "PCDS", "ctry25cd"]) for file in files], ignore_index=True)
+
+    uprn_to_country_df = pd.DataFrame()
+    import os
+
+    folder_path = "/Users/anasofiapinto/Downloads/NSUL_DEC_2025/Data"
```
sofiapinto: replace with your local path while data is uploading to S3. Data can be downloaded from: https://geoportal.statistics.gov.uk/datasets/4e0b4b3fbc2540caae27e7be532e61be/about

asf_heat_pump_suitability/pipeline/transform/anchor_loads.py
sofiapinto, line 1: Created this, with only the categories for now. Simran's code could be moved here.

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -274,18 +275,264 @@ def parse_arguments() -> argparse.Namespace:
...
+
+    features_df = features_df.join(
+        off_gas_df.select(["UPRN", "off_gas"]),
+        how="left",
+        on="UPRN",
+    )
+
+    print(features_df["off_gas"].value_counts())
+
+    # Add near anchor load boolean flag
```
sofiapinto: Mostly copied from Simran's code

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: Have not reviewed as it's copied - let me know if I should review anything if you made changes.

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -88,6 +89,7 @@ constant:
   cardiff:
     la_names: "Cardiff"
     grid_squares: "ST"
+  anchor_load_radius: 50 # threshold for proximity to anchor load properties in meters
```
crispy-wonton: nitpick: 'metres' - as 'meters' is american english :D

asf_heat_pump_suitability/pipeline/transform/anchor_loads.py
crispy-wonton: Ok - in general you don't need to add any anchor load flag in your PR as Simran's work is doing that already. I'd rather merge her code into `dev` than copy code across as it makes it much easier to keep track of.

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -274,18 +274,264 @@ def parse_arguments() -> argparse.Namespace:
...
+    # ADD EPC FEATURES - EPC RATING, ATTACHMENT, TENURE, SOLAR PV and HEAT DEMAND
     epc_df = pl.read_parquet(
         config["data"]["epc"]["domestic"],
-        columns=["UPRN", "TENURE", "BUILT_FORM", "CURRENT_ENERGY_RATING"],
+        columns=[
+            "UPRN",
+            "TENURE",
+            "BUILT_FORM",
+            "CURRENT_ENERGY_RATING",
+            "SOLAR_WATER_HEATING_FLAG",
```
crispy-wonton, line 257: suggestion: you could also process the `PHOTO_SUPPLY` column here. Not sure it will add much as previously I've looked at this data and there was mostly nulls but it could be useful to process.

From what I remember, there was a lot of different possible values so it might require some more complex code to effectively identify houses with PV from this column.

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -274,18 +274,264 @@ def parse_arguments() -> argparse.Namespace:
...
+            "ENERGY_CONSUMPTION_CURRENT",
+            "ENERGY_CONSUMPTION_POTENTIAL",
```
crispy-wonton: note: if we do end up going with these we will need to list it as a caveat in the methodology that we are using total energy demand rather than just for heating.

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -274,18 +274,264 @@ def parse_arguments() -> argparse.Namespace:
...
+        ],
     )
+
+    # Add column to indicate whether EPC data is available for the property, to distinguish between False and unknown for features derived from EPC data
+    epc_df = epc_df.with_columns(pl.lit(True).alias("epc_data_available"))
```
crispy-wonton: question: how will this help distinguish? If the EPC data is empty for a feature, I would assume that's unknown rather than False. Is there a case where this assumption doesn't apply?

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -274,18 +274,264 @@ def parse_arguments() -> argparse.Namespace:
...
+            "CURRENT_ENERGY_RATING",
+            "SOLAR_WATER_HEATING_FLAG",
+            "ENERGY_CONSUMPTION_CURRENT",
+            "ENERGY_CONSUMPTION_POTENTIAL",
+        ],
+    )
+
+    features_df = features_df.with_columns(
+        pl.col("SOLAR_WATER_HEATING_FLAG")
+        .replace({"unknown": None})
```
crispy-wonton: question: does this cover all cases of `unknown` string? Asking because I've seen mixed case usage in EPC before e.g. Title case, upper case etc.

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -274,18 +274,264 @@ def parse_arguments() -> argparse.Namespace:
...
+        how="left",
+        on="UPRN",
+    )
+
+    # Add listed building boolean flag
+    from asf_heat_pump_suitability.pipeline.prepare_features import listed_buildings
+
+    uprns_listed_buildings_df = listed_buildings.generate_df_epc_listed_buildings(
+        epc_df=epc_with_geometries_df
+    ).with_columns(pl.lit(True).alias("in_listed_building"))
```
crispy-wonton: issue 1: we should identify listed building flag for all UPRNs, not just those in the EPC data. I believe the docstrings of the `listed_building` functions are misleading, and the functions should actually work for any dataset with UPRN, X_COORDINATE, and Y_COORDINATE columns. Please double check as I've only had a brief look, but if you agree - the docstrings, function names, and input argument names can be generalised. This misnomer stems from the fact that the EPC data was the baseline dataset in the previous iteration of the pipeline.
This can cut out some of the code below as you should be able to input `features_df` into the function directly.

issue 2: there is a potential bug in the `listed_building` code that results in not all listed buildings being joined correctly. See the initial exploration conducted on this [here](https://github.com/nestauk/asf_heat_pump_suitability/pull/172).

issue 3: I believe the listed building dataset may need to be updated as well as it's potentially incomplete and outdated. Also I think we were using GB-wide datasets, but we know there are Local-Authority-specific ones so you may want to find one for Plymouth specifically - I think this is less of an issue though as it will be much more complex to scale.

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -274,18 +274,264 @@ def parse_arguments() -> argparse.Namespace:
...
+    code_point_df["POSTCODE"] = code_point_df["postcode"].str.replace(" ", "")
+
+    nearest_postcode_df = pl.from_pandas(
+        uprns_gdf.sjoin_nearest(
+            code_point_df[["POSTCODE", "geometry"]],
+            how="left",
+            max_distance=1000,
+            distance_col="distance_to_postcode_m",  # distance in metres
+        ).drop(columns="index_right")[["UPRN", "POSTCODE", "distance_to_postcode_m"]]
+    )
```
crispy-wonton: issue: I would use something like the methodology below to assign postcodes to UPRNs:
1. Use EPC postcode for UPRNs where available
2. Use EPC postcode of nearest UPRN in the same building if available
3. Use EPC postcode of nearest UPRN / nearest code point (whichever is closer) as a last resort. Within 1km or whatever specified distance is best.

My concern is that there is a risk here that the current methodology mislabels a lot of postcodes, and we don't have any sense of the error rate (unless you have looked at this separately).

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -274,18 +274,264 @@ def parse_arguments() -> argparse.Namespace:
...
+    anchor_gdf = anchor_gdf.drop_duplicates(["geometry"])
+    anchor_gdf["near_anchor_load"] = True
+
+    uprns_gdf = uprns_gdf.sjoin(anchor_gdf, how="left", predicate="within")
+
+    features_df = features_df.join(
+        pl.from_pandas(uprns_gdf[["UPRN", "near_anchor_load"]]),
+        how="left",
+        on="UPRN",
+    ).with_columns(pl.col("near_anchor_load").fill_null(False))
```
crispy-wonton: question: what's the purpose of duplicating this work here? Is it just so you can get the UPRN counts per cluster of UPRNs near anchor load? If that's the case, I would just present a final boolean flag for the whole cluster in the opportunity page rather than each UPRN. Is there a reason to do it this granularly?

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -274,18 +274,264 @@ def parse_arguments() -> argparse.Namespace:
...
+
+    # Add near coastline boolean flag
+    coast_gdf = gpd.read_file(
+        "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/Countries_December_2024_Boundaries_UK_BFC_6983126662299524946/CTRY_DEC_2024_UK_BFC.shp"
+    )
+
+    # Simplify coastline boundaries by 150m and buffer by 1500m to create a 'near coastline' area
+    coast_gdf["simplified_geometry"] = coast_gdf["geometry"].apply(
+        lambda x: x.simplify(tolerance=150).buffer(1500)
```
crispy-wonton: suggestion: I would suggest not using `lambda` here and instead use the built-ins in separate lines. These geospatial functions can be quite slow anyway so I would not add further speed detriment with the `apply`.

note: tolerance of 150 was the tolerance I found that worked well when adding the buffer (i.e. gave me sufficient speed gains). I didn't test for an optimal threshold so there may be a better one that creates less simplification, and it's worth noting that this will create some inaccuracies.

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -274,18 +274,264 @@ def parse_arguments() -> argparse.Namespace:
...
+    )
+    coast_gdf = coast_gdf.set_geometry("simplified_geometry")
```
crispy-wonton: suggestion: why not just overwrite the original geometry in the line above and skip this step.

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -274,18 +274,264 @@ def parse_arguments() -> argparse.Namespace:
...
+    coast_gdf["near_coastline"] = True
+
+    uprns_gdf.drop(columns=["index_right"], inplace=True)
+    uprns_gdf = uprns_gdf.sjoin(
+        coast_gdf[["near_coastline", "simplified_geometry"]],
+        how="left",
+        predicate="within",
+    )
```
```suggestion
    uprns_gdf = uprns_gdf.drop(columns=["index_right"]).sjoin(
        coast_gdf[["near_coastline", "simplified_geometry"]],
        how="left",
        predicate="within",
    )
```

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -274,18 +274,264 @@ def parse_arguments() -> argparse.Namespace:
...
+            {
+                "E": "England",
+                "W": "Wales",
+                "S": "Scotland",
+            }
+        )
+    )
+    uprns_gdf = uprns_gdf.merge(
+        uprn_to_country_df[["UPRN", "COUNTRY"]], on="UPRN", how="left"
+    ).drop(columns=["index_right"])
```
crispy-wonton: suggestion: you could create a function to output a mapping dict with UPRN to country. Then you could map a new column instead of merging dfs.

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -274,18 +274,264 @@ def parse_arguments() -> argparse.Namespace:
...
+
+    uprns_conservation_areas_df = (
+        protected_areas.load_transform_df_uprn_in_protected_area(gdf=uprns_gdf)
+    )
+
+    uprns_conservation_areas_df = uprns_conservation_areas_df.with_columns(
+        pl.lit(True).alias("in_conservation_area")
+    )
```
crispy-wonton: suggestion: I don't think you need this step as the function above already creates a new column called `in_protected_areas`. They are referred to as `protected` because it combines World Heritage Sites and building conservation areas.

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -274,18 +274,264 @@ def parse_arguments() -> argparse.Namespace:
...
+
+    features_df = features_df.join(
+        uprns_conservation_areas_df.select(["UPRN", "in_conservation_area"]),
+        how="left",
+        on="UPRN",
+    ).with_columns(pl.col("in_conservation_area").fill_null(False))
```
crispy-wonton: question: are there nulls here? again, the function you're using should already handle the nulls so there might be something going wrong in the function if you need to fill them.

crispy-wonton (changes requested):
hi @sofiapinto !

Thanks for your work on this PR - great to get the initial code into the pipeline!

As requested, I have focussed my review on the logic of the code rather than the structure. I've left some questions and suggestions to improve the methodology.

My main suggestion is an update to the methodology you're using to assign postcodes, and my main question to think about is whether we need to duplicate the anchor load work here. I think it's better to keep it to where it's already implemented (rather than duplicated and then having to update code in two places) but perhaps I'm missing some context. I've left more detail about both of these things in comments on the relevant parts of the code.

asf_heat_pump_suitability/pipeline/transform/anchor_loads.py
sofiapinto: In general I agree that should be the approach we take. However, I had to take a different approach here & i'll explain why below.

- Unless I didn't understand it correctly, Simran's code is about reassigning categories during the clustering/ aggregation process. However, it is not adding the anchor load flag to the contextual features script/dataset. Hence why I copied her code her (temporarily!). 

- I had to branch off `276_handle_cardiff_data` and, hence, couldn't branch off Simran's work, so copied over her work for now. Not for you to review it twice, but as a placeholder. My suggestion is: now that her code is needed in two different parts of our pipeline, instead of living within the clustering.py script, it should live in this new script trasnform/anchor_loads.py. She could make those changes in a separate PR, after all the reassignments during the clustering are merged to dev.

I can also remove all the code for now & add it back on in a separate PR once this one has been merged & Simran's PR has been merged. But we effectively need the anchor load flag as it's one of the contextual features displayed in the tool.

Happy to discuss the best way forward in tomorrow's stand up.

asf_heat_pump_suitability/getters/load_geodata.py
sofiapinto: No, only copied so that the code could work. No review needed.

asf_heat_pump_suitability/pipeline/transform/anchor_loads.py
sofiapinto: Discussed in the stand up - Simran creates a flag during the clustering that can be used when aggregating contextual features to cluster level.

asf_heat_pump_suitability/pipeline/run/add_features.py
sofiapinto: This is a good idea. Both of them have high missingness, but I've used them to create two different functions.

**Photo supply**: Percentage of photovoltaic area as a percentage of total roof area. 0% indicates that a Photovoltaic Supply is not present in the property. This is a float and it takes values between 0 and 100 or `None`. This is what allows us to actually identify PV presence.

**Solar water heating flag**: Indicates whether the heating in the property is solar powered. Takes values "True", "False" and "unknown".

So I've created two functions to create two flags: one to identify whether the property has solar pv and another to identify whether it's solar powered (either solar pv or solar water heating present). We can discuss with the team which one to move forward with.

asf_heat_pump_suitability/pipeline/run/add_features.py
sofiapinto: Looking at the data, it does. Dealing with the different unknowns should be part of the daps pipeline!

asf_heat_pump_suitability/pipeline/run/add_features.py
sofiapinto: Having thought about this I think I'm going to only keep ENERGY_CONSUMPTION_CURRENT for now, and making it explicit that this is not heat demand, rather estimated total energy consumption for the property in a 12 month period (kWh/m2).

sofiapinto: We can then gather feedback on whether the features displayed should change.

asf_heat_pump_suitability/pipeline/run/add_features.py
sofiapinto: Thanks for all of this context @crispy-wonton 🙏

asf_heat_pump_suitability/pipeline/run/add_features.py
sofiapinto: Thanks for the context @crispy-wonton, it makes sense - because this code had already been used separately, I thought this has been evaluated properly before. My bad. I will change the methodology, & agree with the steps above.

asf_heat_pump_suitability/pipeline/run/add_features.py
sofiapinto: In the tool we've been calling them conservation areas? But it seems like we should be calling the protected areas?

asf_heat_pump_suitability/pipeline/run/add_features.py
sofiapinto: no longer using this :)

asf_heat_pump_suitability/pipeline/run/add_features.py
sofiapinto: I've moved this code into the protected_areas.py file. I do need to fill nulls at one point, but let me know if you still think there's something wrong.

sofiapinto:
hey @crispy-wonton thank you for the review.

Hopefully I've now fixed the features that needed fixing. I've also refactored the code (move code from the add_features.py to the respective pipeline files). But i'd really appreciate if you could do a careful review of the code.

Summary:
- `has_solar_pv` feature: I created two new functions in epc.py, extend_df_solar_pv_flag() and extend_df_is_solar_powered() but only one is used.
- `in_listed_building`: I created new code as found it quite hard to use existing code. Can you please review thoroughly to see if there are any issues with it? I didn't collect new data. We can do that as part of phase 2. I found it was more pressing to have working code, even if the list of listed buildings is incomplete. This deals with listed building geometries which are points and polygons.
- `off_gas`: this now uses the approach you describe above. Using EPC postcode of nearest UPRN in the same building if available and then using EPC postcode of nearest code point within a specified distance. I set a max distance of 500m to nearest code point to be conservative. **This needs to be further evaluated later (in a following PR).** For now i've checked and the number of off gas properties in Plymouth we're getting is below the correct number - meaning we might need to increase this threshold in the future. I guess, for now, i'd rather we are conservative as I haven't been able to properly evaluate this.
- `protected_areas`: do you want to call this protected areas since  it contains World Heritage Sites and conservation areas? We've been calling it conservation areas in the tool.
- `within_1500m_coastline`: I've used the same 150m tolerance and 1.5km distance threshold for now. We'll have it as an assumption for now, and we can check with supply chain that the right distance is. We can make it explicit on the tool hat we're using 1.5km.

crispy-wonton:
Just responding to a couple of key points here and then will review:

> * `off_gas`: this now uses the approach you describe above. Using EPC postcode of nearest UPRN in the same building if available and then using EPC postcode of nearest code point within a specified distance. I set a max distance of 500m to nearest code point to be conservative. **This needs to be further evaluated later (in a following PR).** For now i've checked and the number of off gas properties in Plymouth we're getting is below the correct number - meaning we might need to increase this threshold in the future. 

What ground truth are you comparing the data to to know that we are getting below the correct number?

> * `protected_areas`: do you want to call this protected areas since  it contains World Heritage Sites and conservation areas? We've been calling it conservation areas in the tool.

Yes, I think we will need to change this in the tool as it's relevant for scaling to Scotland. In England and Wales we use building conservation areas, but based on previous conversations internally in the previous version of the heat pump suitability map, World Heritage Sites are more relevant to planning and implementation of LCH technologies in Scotland.

asf_heat_pump_suitability/pipeline/run/add_features.py
crispy-wonton: Yes have left a general comment on the PR about this :)

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -30,6 +30,15 @@ data:
...
 
+    # OpenMap Code Point GB data for postcode lookup
+    # TODO: move this file to a different S3 location as it's not just Plymouth data - it's GB data
+    gb_code_point_data: "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/codepo_gb.gpkg"
+
+    # GB coastline boundaries
+    gb_coast_boundaries: "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/Countries_December_2024_Boundaries_UK_BFC_6983126662299524946/CTRY_DEC_2024_UK_BFC.shp"
+
+    # Path to folder with UPRN to country mapping files for GB
+    gb_uprn_country_mapping: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/NSUL_DEC_2025/"
```
crispy-wonton: nitpick: for neatness and readability I'd move this above the `boundaries` key. Can be done in a separate PR when we clean up the repo tho.

asf_heat_pump_suitability/getters/get_datasets.py
```diff
@@ -245,12 +245,12 @@ def get_df_spa_offgasgrid() -> pl.DataFrame:
     return df
 
 
-def load_gdf_listed_buildings(nation: str, **kwargs) -> gpd.GeoDataFrame:
+def load_gdf_listed_buildings(nation: str = "GB", **kwargs) -> gpd.GeoDataFrame:
```
crispy-wonton, line 249: TODO: we need to move this into `load_geodata.py` at some point. If not doing in this PR, please could you add a TODO note so we can do it when refactoring the repo.

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -274,3 +275,78 @@ def load_gdf_poi() -> gpd.GeoDataFrame:
     print(f"POI CRS: {poi.crs}")
 
     return poi
+
+
+def load_code_point_data() -> gpd.GeoDataFrame:
+    """
+    Load GB code point geodataframe for postcode lookup and clean postcode column by removing spaces.
```
crispy-wonton, line 345: nitpick: please could you add the CRS of the loaded data to the docstring.

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -274,3 +275,78 @@ def load_gdf_poi() -> gpd.GeoDataFrame:
...
+        config["data"]["geodata"]["gb_code_point_data"],
+        layers="codepoint",
+    )
+    code_point_df["POSTCODE"] = code_point_df["postcode"].str.replace(" ", "")
+    return code_point_df
+
+
+def load_gb_coast_boundaries():
+    """
+    Load GB coastline boundaries geodataframe and dissolve into a single geometry.
```
crispy-wonton, line 366: nitpick: Same here re CRS

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -274,3 +275,78 @@ def load_gdf_poi() -> gpd.GeoDataFrame:
     print(f"POI CRS: {poi.crs}")
 
     return poi
+
+
+def load_code_point_data() -> gpd.GeoDataFrame:
```
crispy-wonton: nitpick. here and below we are missing the return type in the function name. I.e. `load_gdf_etc`.

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -274,3 +275,78 @@ def load_gdf_poi() -> gpd.GeoDataFrame:
...
+    # Dissolve coastline boundaries into a single geometry
+    coast_gdf = gpd.GeoDataFrame(
+        geometry=[coast_gdf.geometry.union_all()], crs=coast_gdf.crs
+    )
+    return coast_gdf
+
+
+def load_transform_dict_uprn_to_country_mapping() -> dict:
+    """
+    Load and transform the UPRN to country mapping data from S3.
```
crispy-wonton, line 390: nitpick: Same here re CRS

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -274,3 +275,78 @@ def load_gdf_poi() -> gpd.GeoDataFrame:
...
+
+    Returns:
+        dict: A dictionary mapping UPRN to corresponding country information.
+    """
+    s3_client = boto3.client("s3")
```
crispy-wonton, line 397:
```suggestion
    """
    print("Loading UPRN to country mapping...")
    s3_client = boto3.client("s3")
```

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,23 +29,89 @@ def generate_df_epc_listed_buildings(
...
+    listed_buildings_gdf: gpd.GeoDataFrame,
+) -> pl.DataFrame:
+    """
+    Add boolean column to features_df indicating whether UPRN is in a listed building.
+
+    Args:
+        features_df (pl.DataFrame): dataframe with one row per UPRN and UPRN column.
+        uprns_gdf (gpd.GeoDataFrame): GeoDataFrame with point geometries for each UPRN.
+        buildings_gdf (gpd.GeoDataFrame): GeoDataFrame of building footprints with geometry column.
+        listed_buildings_gdf (gpd.GeoDataFrame): GeoDataFrame of listed buildings with geometry column.
```
crispy-wonton, line 62:
```suggestion
        listed_buildings_gdf (gpd.GeoDataFrame): GeoDataFrame of listed buildings with geometry column. Can be point or polygon geometries.
```

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,23 +29,89 @@ def generate_df_epc_listed_buildings(
...
+    ]
+
+    # Join for polygons: We want buildings that touch/overlap listed building polygons
+    joined_polys_gdf = gpd.sjoin(
+        buildings_gdf, listed_polygons_gdf, how="inner", predicate="intersects"
+    )
+
+    # Join for points: We want buildings that contain the listed building points
+    joined_pts_gdf = gpd.sjoin(
+        buildings_gdf, listed_points_gdf, how="inner", predicate="contains"
```
crispy-wonton: nitpick: per `geopandas` docs regarding `sjoin` and `sjoin_nearest` functions:

> For historical reasons, both methods are also available as top-level functions [sjoin()](https://geopandas.org/en/v1.1.2/docs/reference/api/geopandas.sjoin.html#geopandas.sjoin) and [sjoin_nearest()](https://geopandas.org/en/v1.1.2/docs/reference/api/geopandas.sjoin_nearest.html#geopandas.sjoin_nearest). It is recommended to use methods as the functions may be deprecated in the future.

```suggestion
    # Join for polygons: We want buildings that touch/overlap listed building polygons
    joined_polys_gdf =  buildings_gdf.sjoin(
       listed_polygons_gdf, how="inner", predicate="intersects"
    )

    # Join for points: We want buildings that contain the listed building points
    joined_pts_gdf = buildings_gdf.sjoin(
       listed_points_gdf, how="inner", predicate="contains"
```

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,23 +29,89 @@ def generate_df_epc_listed_buildings(
...
+
+    # Identify listed buildings that have point geometries
+    listed_points_gdf = listed_buildings_gdf[
+        listed_buildings_gdf.geometry.type.isin(["Point", "MultiPoint"])
+    ]
+
+    # Join for polygons: We want buildings that touch/overlap listed building polygons
+    joined_polys_gdf = gpd.sjoin(
+        buildings_gdf, listed_polygons_gdf, how="inner", predicate="intersects"
+    )
```
crispy-wonton, line 85: TODO (minor): I suppose we might want to double check that this isn't labelling buildings which touch a listed building on one edge as 'listed'. We can do this separately but might be worth adding a TODO note here.

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,23 +29,89 @@ def generate_df_epc_listed_buildings(
...
+    joined_pts_gdf = gpd.sjoin(
+        buildings_gdf, listed_points_gdf, how="inner", predicate="contains"
+    )
+
+    # Combine joined polygons and points
+    joined_gdf = pd.concat([joined_polys_gdf, joined_pts_gdf], ignore_index=True)
+
+    # Spatial join UPRNs with listed buildings to label UPRNs in listed buildings.
+    uprns_gdf = (
+        gpd.sjoin(
```
crispy-wonton: nitpick: same here re the `sjoin`.

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
```diff
@@ -29,23 +29,89 @@ def generate_df_epc_listed_buildings(
...
+
+    # Identify listed buildings that have polygon geoemetries
+    listed_polygons_gdf = listed_buildings_gdf[
+        listed_buildings_gdf.geometry.type.isin(["Polygon", "MultiPolygon"])
+    ]
+
+    # Identify listed buildings that have point geometries
+    listed_points_gdf = listed_buildings_gdf[
+        listed_buildings_gdf.geometry.type.isin(["Point", "MultiPoint"])
+    ]
```
crispy-wonton: suggestion (non-blocking): I'd probably also select the columns that you want to keep here as the `listed_buildings_gdf` has a lot of columns.

asf_heat_pump_suitability/pipeline/prepare_features/off_gas.py
```diff
@@ -28,3 +29,94 @@ def add_off_gas_feature(df: pl.DataFrame, off_gas_postcodes: List[str]) -> pl.Da
...
+
+def extend_df_off_gas(
+    features_df: pl.DataFrame,
+    uprns_gdf: gpd.GeoDataFrame,
+    code_point_df: gpd.GeoDataFrame,
+    off_gas_list: list,
+    max_distance_m: int = 500,
+) -> pl.DataFrame:
+    """
+    Add boolean column to features_df indicating whether UPRN is in an off-gas postcode.
```
crispy-wonton, line 45: suggestion: could you add a brief line on what happens when the UPRN doesn't have it's own postcode assigned from EPC.

asf_heat_pump_suitability/pipeline/prepare_features/off_gas.py
```diff
@@ -28,3 +29,94 @@ def add_off_gas_feature(df: pl.DataFrame, off_gas_postcodes: List[str]) -> pl.Da
...
+        uprns_gdf (gpd.GeoDataFrame): GeoDataFrame with point geometries for each UPRN.
+        code_point_df (gpd.GeoDataFrame): GeoDataFrame of postcode centroids with geometry column and POSTCODE column.
+        off_gas_list (list): list of postcodes that are off-gas.
+        max_distance_m (int): maximum distance in metres to assume a UPRN is associated with a postcode if it doesn't have its own postcode. Defaults to 500 metres.
+
+    Returns:
+        pl.DataFrame: input features_df with new boolean column `off_gas` indicating whether UPRN is in an off-gas postcode.
+    """
+    # Step 0: Use EPC postcode for UPRNs where available
+    # Done previously, as we features_df already contains the EPC postcode column
```
crispy-wonton:
```suggestion
    # Done previously, as features_df already contains the EPC postcode column
```

asf_heat_pump_suitability/pipeline/prepare_features/off_gas.py
```diff
@@ -28,3 +29,94 @@ def add_off_gas_feature(df: pl.DataFrame, off_gas_postcodes: List[str]) -> pl.Da
...
+    uprns_gdf: gpd.GeoDataFrame,
+    code_point_df: gpd.GeoDataFrame,
+    off_gas_list: list,
+    max_distance_m: int = 500,
+) -> pl.DataFrame:
+    """
+    Add boolean column to features_df indicating whether UPRN is in an off-gas postcode.
+
+    Args:
+        features_df (pl.DataFrame): dataframe with one row per UPRN, UPRN column and POSTCODE column.
```
crispy-wonton: question: looks like this should also contain building ID column?
please could you change hard-coding `ID` to `id_col` that can be set as an argument. I think it's possible we'll change the name of this as it's currently not descriptive.

asf_heat_pump_suitability/pipeline/prepare_features/off_gas.py
```diff
@@ -28,3 +29,94 @@ def add_off_gas_feature(df: pl.DataFrame, off_gas_postcodes: List[str]) -> pl.Da
...
+    # Step 0: Use EPC postcode for UPRNs where available
+    # Done previously, as we features_df already contains the EPC postcode column
+
+    # Step 1: Use EPC postcode of nearest UPRN in the same building if available
+
+    # create mapping between ID and POSTCODE when POSTCODE is not null
+    id_postcode_mapping_df = (
+        features_df.filter(pl.col("POSTCODE").is_not_null())
+        .select(["ID", "POSTCODE"])
+        .rename({"POSTCODE": "MAPPED_POSTCODE"})
```
crispy-wonton: nitpick: why not just keep the name as `POSTCODE` as you are renaming back to `POSTCODE` below anyway.

asf_heat_pump_suitability/pipeline/prepare_features/off_gas.py
```diff
@@ -28,3 +29,94 @@ def add_off_gas_feature(df: pl.DataFrame, off_gas_postcodes: List[str]) -> pl.Da
...
+    Returns:
+        pl.DataFrame: input features_df with new boolean column `off_gas` indicating whether UPRN is in an off-gas postcode.
+    """
+    # Step 0: Use EPC postcode for UPRNs where available
+    # Done previously, as we features_df already contains the EPC postcode column
+
+    # Step 1: Use EPC postcode of nearest UPRN in the same building if available
+
+    # create mapping between ID and POSTCODE when POSTCODE is not null
+    id_postcode_mapping_df = (
```
crispy-wonton: question: have you checked if there are building footprints with more than one postcode?

asf_heat_pump_suitability/pipeline/prepare_features/off_gas.py
```diff
@@ -28,3 +29,94 @@ def add_off_gas_feature(df: pl.DataFrame, off_gas_postcodes: List[str]) -> pl.Da
...
+
+    # Create dataframe of UPRNs with their mapped POSTCODE from the same building where available
+    postcodes_df = (
+        features_df.select(["UPRN", "ID"])
+        .join(id_postcode_mapping_df, on="ID", how="left")
+        .rename({"MAPPED_POSTCODE": "POSTCODE"})
+    )
+
+    print(
+        "Number of UPRNs with POSTCODE after step 1:",
```
crispy-wonton:
```suggestion
        "Number of UPRNs with POSTCODE after mapping postcodes from same building:",
```

can we have this for step 0 as well?

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -275,18 +271,115 @@ def parse_arguments() -> argparse.Namespace:
...
+            "CURRENT_ENERGY_RATING",
+            "SOLAR_WATER_HEATING_FLAG",
+            "ENERGY_CONSUMPTION_CURRENT",
+            "PHOTO_SUPPLY",
+            "POSTCODE",
+        ],
+    )
+
+    print(features_df["has_solar_pv"].value_counts())
+    print(features_df["ENERGY_CONSUMPTION_CURRENT"].mean())
```
crispy-wonton: nitpick: are these legacy `print` statements to be removed? there's also more below.

asf_heat_pump_suitability/pipeline/prepare_features/protected_areas.py
```diff
@@ -149,3 +149,30 @@ def generate_df_conservation_area_data_availability(
...
+    Returns:
+        pl.DataFrame: Input dataframe extended with boolean column indicating whether each UPRN is within a protected area.
+    """
+
+    features_df = (
+        features_df.join(
+            protected_areas_df.select(["UPRN", "in_protected_area"]),
+            how="left",
+            on="UPRN",
+        ).with_columns(pl.col("in_protected_area").fill_null(False))
```
crispy-wonton: TODO: we need to double check that filling with `False` is appropriate in all cases. We previously did an analysis on this to basically assess where we needed to label it as `unknown` instead due to missing data. Please could you add a TODO note as this is important in the scaling.

asf_heat_pump_suitability/pipeline/transform/epc.py
```diff
@@ -122,3 +125,52 @@ def clean_df_tenure(df: pl.DataFrame) -> pl.DataFrame:
...
+    no_pv = pl.col("PHOTO_SUPPLY") == 0
+
+    return df.with_columns(
+        pl.when(has_pv)
+        .then(True)
+        .when(no_pv)
+        .then(False)
+        .otherwise(None)
+        .alias("has_solar_pv")
+    )
```
crispy-wonton, line 150: praise: love how clean this is!

crispy-wonton (commented):
Hello ! :D this looks really clean and nice thank you!

I have left mostly nitpicks with a couple of suggestions and comments. I am happy to implement the nitpicks and add the TODO notes next week to save you time.

I didn't find any bugs per se but I do have one important question about the postcode methodology. Have you checked if there are any building footprints with multiple postcodes? I suppose this could possibly happen if the footprint is large? E.g. thinking of a long terrace. Not sure but worth checking cuz if that's the case we would end up with UPRNs with multiple postcodes.

sofiapinto:
> Just responding to a couple of key points here and then will review:
> 
> > * `off_gas`: this now uses the approach you describe above. Using EPC postcode of nearest UPRN in the same building if available and then using EPC postcode of nearest code point within a specified distance. I set a max distance of 500m to nearest code point to be conservative. **This needs to be further evaluated later (in a following PR).** For now i've checked and the number of off gas properties in Plymouth we're getting is below the correct number - meaning we might need to increase this threshold in the future.
> 
> What ground truth are you comparing the data to to know that we are getting below the correct number?

Here you go:
https://www.gov.uk/government/statistics/sub-national-estimates-of-households-not-connected-to-the-gas-network

> > * `protected_areas`: do you want to call this protected areas since  it contains World Heritage Sites and conservation areas? We've been calling it conservation areas in the tool.
> 
> Yes, I think we will need to change this in the tool as it's relevant for scaling to Scotland. In England and Wales we use building conservation areas, but based on previous conversations internally in the previous version of the heat pump suitability map, World Heritage Sites are more relevant to planning and implementation of LCH technologies in Scotland.

Cool! Noted.

asf_heat_pump_suitability/pipeline/transform/epc.py
sofiapinto: thank you queen!!👑

asf_heat_pump_suitability/config/base.yaml
sofiapinto: Makes sense!

asf_heat_pump_suitability/pipeline/prepare_features/off_gas.py
sofiapinto: I was actually going to remove these prints, but I can keep them if you think they're helpful?

asf_heat_pump_suitability/config/base.yaml
sofiapinto: Done!

asf_heat_pump_suitability/getters/get_datasets.py
sofiapinto: I think we've spoken about this at the beginning of this PR and you said you'd prefer for us to move these functions + all that lives within pipeline/prepare_features/ to pipeline/transform in a separate PR. Have added a TODO note.

asf_heat_pump_suitability/getters/load_geodata.py
sofiapinto: Apologies!! Have added.

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
sofiapinto: good shout :)

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
sofiapinto: Have selected the `geometry` and `in_listed_building` columns only :)

asf_heat_pump_suitability/getters/load_geodata.py
sofiapinto: This is reading tabular attribute data, no CRS.

asf_heat_pump_suitability/pipeline/run/add_features.py
sofiapinto: Yep, I left them as they're useful when I was running the script. But will remove now.

asf_heat_pump_suitability/pipeline/prepare_features/listed_buildings.py
sofiapinto: Very good shout, have added a todo.

sofiapinto: An idea would be to do the join and after we could filter out rows where the intersection area is negligible. But there might be better ways for checking it.

asf_heat_pump_suitability/pipeline/prepare_features/off_gas.py
sofiapinto: Very important, thank you.

asf_heat_pump_suitability/pipeline/prepare_features/off_gas.py
sofiapinto: good shout!

asf_heat_pump_suitability/pipeline/prepare_features/off_gas.py
sofiapinto: Have added prints for each step.

asf_heat_pump_suitability/pipeline/prepare_features/protected_areas.py
sofiapinto: Interesting, that makes sense! Want to hear more about how you'd go about distinguishing unknowns from False's for this.

sofiapinto: Have added a todo.

sofiapinto:
Hi @crispy-wonton ,

- I made the changes as per your suggestions. 

- I then merged dev into this branch and ran the script. I had to fix a bug introduced by HNZoneD that was breaking the pipeline.

- Then I noticed that there were some parts of the codebase introducing duplicate UPRNs. Hopefully I fixed that, but please run add_features.py and check if the that's the case.

To answer your question:

> question: have you checked if there are building footprints with more than one postcode?

This was a good shout. Hopefully I fixed it, but do check. This function is now way too complicated. If it works that's okay, but might be one of those that we might want to revisit.

**Update:** I think I introduced a bug actually (a bracket in the wrong place). Lines 90 to 104 of off_gas.py should be instead:
  ```
 # Create geodataframe of UPRNs with their mapped closest POSTCODE from the same building, where available
    postcodes_gdf = (
        features_gdf.groupby(id_col, group_keys=False)
        .apply(
            lambda x: gpd.sjoin_nearest(
                x,
                id_postcode_mapping_gdf[id_postcode_mapping_gdf[id_col] == x.name],
                how="left",
                distance_col="distance_to_nearest_postcode_m",
            ))
            .sort_values("distance_to_nearest_postcode_m", ascending=True)
            .drop_duplicates(subset=["UPRN_left"])
        [["UPRN_left", id_col + "_left", "POSTCODE"]]
        .rename(columns={"UPRN_left": "UPRN", id_col + "_left": id_col})
    )
```
Haven't had time to add a commit. sorry!

Please merge when you're happy with this. When you do so, please remove the branch so that the next PR merges into dev.

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: my bad!

crispy-wonton (dismissed):

crispy-wonton (dismissed):

crispy-wonton (approved):

PR #296, opened 2026-04-15
294 update barrier roads

### LINKED ISSUE #294: Update barrier roads in clustering
Two things to do:
1. Update the source of the roads data to use [OS Open Roads](https://www.ordnancesurvey.co.uk/products/os-open-roads) instead of OS OpenMap Local Roads layer. I think this is a more granular dataset than what we were using before.
2. Update the types of roads which act as barriers in `asf_heat_pump_suitability/pipeline/cluster/cluster.py`.

Road types which are NOT barriers:
B Road
Guided Busway Carriageway
Local Access Road
Local Road
Minor Road
Primary Road (wouldn't be desirable but doable.)  # We might need to change this if the clusters get too big
Restricted Local Access Road
Shared Use Carriageway

Road types which ARE barriers:
Motorway
A Road

simran-dave:
---

# Description

Updated roads data to use [OS OpenRoads ](https://docs.os.uk/os-downloads/products/transport-network-portfolio/os-open-roads). Road categories classed as barrier roads are now limited to A Roads and Motorways

Fixes #294

# Instructions for Reviewer

run `cluster.py`

I have added a new getter in `load_geodata.py` which loads the OS OpenRoad data for specified grid squares or for all of GB. I am a bit unsure about the implementation for all of GB. Overlaying the barrier roads in `cluster.py` is exactly the same as before but with the new data, and only A Roads and Motorways as the barrier road categories. 

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -199,3 +200,42 @@ def load_gdf_os_openmap_layer(
...
+    if not grid_squares:
+        fs = s3fs.S3FileSystem()
+        file_path = config["data"]["geodata"]["gb_os_openroad"]
+        files = fs.glob(f"{file_path}*_RoadLink.shp")
+        gdfs = []
+        for file in files:
+            print(f"\nLoading OS OpenRoad file: {file}")
+            gdfs.append(gpd.read_file(f"s3://{file}"))
+
+        gdf = pd.concat(gdfs)
```
crispy-wonton, line 232: praise: I haven't fully tested running this as I imagine it will take a while to load all the files, but I have run the lines below and it looks like it's picking up all the correct files :) thanks - really neat code and nice use of `glob`. 
```
fs = s3fs.S3FileSystem()
file_path = config["data"]["geodata"]["gb_os_openroad"]
files = fs.glob(f"{file_path}*_RoadLink.shp")
```

thought: the thing that could cause this to fail later in the pipeline, is if different files have different column names. I don't expect this to happen as the OS has good data quality and are generally consistent across grid square files. I think this should work well but we will find out for sure when scaling!

thought: the OS also has single files for all of GB so there should be a whole GB open roads file. Potentially it could give us speed gains to use a single file if we run the pipeline for all of GB in one go. I'm just noting this here for information. I think let's leave the implementation as you've written it because we might want both options and this works well anyway!

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -199,3 +200,42 @@ def load_gdf_os_openmap_layer(
...
+    else:
+        file_path = config["data"]["geodata"]["grid_square_os_openroad"]
+        if not isinstance(grid_squares, List):
+            grid_squares = [grid_squares]
+        files = [file_path.format(square=code) for code in grid_squares]
+
+        gdfs = []
+
+        for file in files:
+            gdfs.append(gpd.read_file(file))
```
crispy-wonton, line 243:
```suggestion
        for file in files:
            print(f"\nLoading OS OpenRoad file: {file}")
            gdfs.append(gpd.read_file(file))
```
Can we add this again here so we can see what's being loaded.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -290,23 +290,15 @@ def load_tranform_gdf_linestring_barriers(
...
 
-    barrier_road_types = [
-        "A Road" "B Road, Collapsed Dual Carriageway",
-        "Minor Road, Collapsed Dual Carriageway",
-        "Primary Road, Collapsed Dual Carriageway",
-        "Motorway",
-        "Motorway, Collapsed Dual Carriageway",
-        "A Road, Collapsed Dual Carriageway",
-    ]
+    barrier_road_types = ["A Road", "Motorway"]
```
crispy-wonton: wow! so much nicer :D

note: I've cross-checked the types and the different column headers and I think this should cover all A Road and Motorway instances. I.e. I don't think they should appear as 'unclassified' or 'unknown' but be discernible from another column.

crispy-wonton (commented):
Thanks so much @simran-dave ! This looks great - lovely clean code which is very readable and easy to follow. I don't see anything that would cause a bug.

Have just left a very minor suggestion and a couple of thoughts, but even now this is good to merge (although let's hold off until PR #272 is ready for merge). Thanks for this work!

Fyi I have run the `cluster.py` script using the following line to check that the script ran all the way through and it did:
`python asf_heat_pump_suitability/pipeline/cluster/cluster.py --tech_gdf s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth/plymouth_buildings_most_suitable_tech.parquet --local_authorities plymouth`

asf_heat_pump_suitability/getters/load_geodata.py
simran-dave: I've had a go at loading all the files and looks like the column names are consistent.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
simran-dave: yeah from [this bit of the documentation](https://docs.os.uk/os-downloads/products/transport-network-portfolio/os-open-roads/os-open-roads-technical-specification/code-lists-and-enumerations/roadclassificationvalue) I think all the motorways and A roads are labelled as such

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: Awesome thanks for checking - out of interest , do you know roughly how long it took to load all the files?

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Ah yes - agree! Great use of docs

crispy-wonton (dismissed):
Excellent work - let's wait til #272 is ready before merging but feel free to already merge changes from the base branch into this one and fix any conflicts.

asf_heat_pump_suitability/getters/load_geodata.py
simran-dave: yeah took about 10 mins

crispy-wonton: wow fast!

crispy-wonton (approved):

Issue #290, opened 2026-04-09
Add new criterion to domestic filtering


crispy-wonton:
Improve the domestic filtering with a new threshold as a continuation of the exploration work by @simran-dave in #287 .

PR #298, opened 2026-04-15
290 Identify domestic density threshold

crispy-wonton:
Identifies the threshold required for #290. Will be implemented in the pipeline separately. 

# Description

This is an exploratory script, plus some additional utils, which aims to identify a new feature to use in the domestic UPRN identification pipeline and a threshold value to use for binary classification of buildings into 'contains domestic' and not.

This builds off the initial work and findings by @simran-dave in #287 .

New files:
- [...y/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py](https://github.com/nestauk/asf_heat_pump_suitability/compare/216_council_tax_domestic_filtering...290_domestic_density_threshold#diff-6c366e5b506822e876a5d7d0734fff6f563fd7d4d0a2f849c4841cdc9c50e657) - main script to identify a threshold and see the effect of it on the pipeline.
- [asf_heat_pump_suitability/utils/mapping_utils.py](https://github.com/nestauk/asf_heat_pump_suitability/compare/216_council_tax_domestic_filtering...290_domestic_density_threshold#diff-345103a29291337b73f9d8d8ce2886483f385516fbb43470c60b7fa731790114) - new mapping utils
- [asf_heat_pump_suitability/utils/plotting_utils.py](https://github.com/nestauk/asf_heat_pump_suitability/compare/216_council_tax_domestic_filtering...290_domestic_density_threshold#diff-a0578cf0d02ec5ea41472e74ea2ff19e0faaba25bec467ce50788910a7dad8bc) - new plotting utils

# Instructions for Reviewer

Run the following line in your terminal:
`python -i asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
`
You can use the interactive `-i` flag optionally to enter interactive mode and interact with the variables.

The script will save 4 outputs to your local `asf_heat_pump_suitability/outputs/` folder. Please could you inspect them and flag if you notice anything strange-looking. The expected outputs are as follows, the maps are the most important to look at - don't worry too much about the other two:
- `asf_heat_pump_suitability/outputs/maps/plymouth_false_positive_domestic_buildings.html`
- `asf_heat_pump_suitability/outputs/maps/plymouth_threshold_effect_on_labelling.html`  
- `asf_heat_pump_suitability/outputs/figures/plymouth_feature_distribution_for_domestic_vs_non_domestic_buildings.png`
- `asf_heat_pump_suitability/outputs/figures/plymouth_decision_tree_nobalance.png`

- The subset of data that I have generated the threshold from. I.e. I have filtered the data to only those buildings which we have currently predicted as containing a domestic property in our pipeline. My reasoning is that the goal of this work is to add an extra filtering step into the existing pipeline in order to remove buildings we're mislabelling as domestic when they are truly non-domestic. Please let me know if you think a different subset makes more sense.
- The `plot_folium_threshold_effect_on_labelling` function. Please could you check this carefully and let me know if you think it's accurately assessing the impact of the threshold on the labelling, if we were to implement it as an additional step in the pipeline. I got a bit confused here trying to account for the existing pipeline steps. I think I just need to filter out the domestic EPC UPRNs because we should never label them as non-domestic but would be great for you to double check and let me know what you think.
- The script will print some results to the terminal. Please could you review these and flag if you think anything looks weird.

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
```diff
@@ -0,0 +1,481 @@
...
+    epc_uprns = uprns.load_set_valid_epc_uprns(epc_type="domestic")
+    epc_gdf = uprns_gdf[uprns_gdf["UPRN"].isin(epc_uprns)][["UPRN", "geometry"]]
+    labelled_gdf = (
+        labelled_gdf.sjoin(epc_gdf, how="left", predicate="contains")
+        .drop_duplicates(subset="ID")
+        .astype({"UPRN": "bool"})
+    )
+
+    # Remove buildings containing a domestic EPC as these will be retained regardless of threshold
+    labelled_gdf = labelled_gdf[~labelled_gdf["UPRN"]]
```
simran-dave: This line is returning an empty gdf for me, which then means this plotting function doesn't plot any buildings. I think it's because the line above is setting all the entries in the UPRN column to be True. Removing .`astype({"UPRN": "bool"})` from the above code on line 306 and then changing this line (310) to `labelled_gdf["UPRN"] = labelled_gdf["UPRN"].notna()` fixed it for me. I don't know if that's the best way to do it

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
```diff
@@ -0,0 +1,481 @@
...
+
+    Args:
+        buildings_gdf (gpd.GeoDataFrame): building footprints for area of evaluation.
+        uprn_gdf (gpd.GeoDataFrame): all UPRNs and point geometries for area of interest.
+        council_tax_gdf (gpd.GeoDataFrame): processed council tax data with empty UPRN rows removed and with geopoints per UPRN.
+        pipeline_gdf (gpd.GeoDataFrame): domestic UPRNs identified by current pipeline.
+        boundary (shapely.Polygon | shapely.MultiPolygon): boundary of area of interest.
+
+    Returns:
+        gpd.GeoDataFrame: buildings in area of interest labelled with actual and predicted domestic UPRN counts and boolean
```
simran-dave: I think it's also labelling total UPRN count

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
```diff
@@ -0,0 +1,481 @@
...
+
+    # UPRN density measures
+    labelled_gdf["predicted_UPRN_per_m2"] = (
+        labelled_gdf["predicted_UPRN_count"] / labelled_gdf["footprint_area_m2"]
+    )
+    labelled_gdf["total_UPRN_per_m2"] = (
+        labelled_gdf["total_UPRN_count"] / labelled_gdf["footprint_area_m2"]
+    )
+
+    # UPRN density measures
```
simran-dave:
```suggestion
    # Area per UPRN measures
```

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
```diff
@@ -0,0 +1,481 @@
...
+
+
+def generate_df_threshold_evaluation(
+    model_df: pd.DataFrame, features: List[str]
+) -> pd.DataFrame:
+    """
+    For each feature in `features`, test a set of candidate thresholds to identify the best threshold for that feature
+    for binary classification of buildings into 'contains domestic' or not. Candidate thresholds for each feature are
+    the values at each percentile in that feature. The best threshold is selected by maximising for Matthew's Correlation
+    Coefficient (MCC).
```
simran-dave, line 189: How come we are using MCC now instead of Youden's J? Do you think it would be worth trying different test statistics to maximise for since the threshold that ends up getting recommended here is different to the one from Youden's J

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
```diff
@@ -0,0 +1,481 @@
...
+) -> DecisionTreeClassifier:
+    """
+    Train a decision tree classifier model to label buildings as 'contains domestic' or not.
+
+    Args:
+        model_df (pd.DataFrame): dataframe with target variable and all features of interest
+        features (List[str]): features of interest
+        save_as (str): Path to save a plot of the decision tree to. Optional.
+    """
+    clf = DecisionTreeClassifier(max_leaf_nodes=5, class_weight=None)
```
simran-dave, line 512: Did you perform a search to get the best value for `max_leaf_nodes ` ? I don't know how much of a difference it would make

asf_heat_pump_suitability/utils/plotting_utils.py
```diff
@@ -0,0 +1,54 @@
...
+    Returns:
+        plt.Figure
+    """
+    nrows = 3
+    ncols = math.ceil(len(features) / nrows)
+
+    fig, axs = plt.subplots(nrows, ncols, figsize=(15, 8))
+
+    for ax, feature in zip(axs.ravel(), features):
+        ax.hist(df.filter(pl.col(target))[feature], bins=40, alpha=0.5, density=density)
```
simran-dave, line 41: minor but the plot is missing an x and y label. Maybe x could be a shared label saying 'feature' and y could be 'count' if density = False and density if True or something like that

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
```diff
@@ -0,0 +1,481 @@
...
+    labelled_gdf = label_gdf_buildings_domestic_bool(
+        buildings_gdf=buildings_gdf,
+        uprn_gdf=uprns_gdf,
+        council_tax_gdf=council_tax_uprns_gdf,
+        pipeline_gdf=pipeline_domestic_uprns_gdf,
+        boundary=plymouth_boundary,
+    )
+
+    # We're only interested in buildings we've predicted to be containing domestic
+    # This is because threshold will be applied AFTER the current pipeline
```
simran-dave, line 577: I think using this subset of data makes sense. I guess the only advantage of using all the data would be including more data for the training steps. But if we still have enough after removing these buildings then I think this is better.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
```diff
@@ -0,0 +1,481 @@
...
+    plot_folium_threshold_effect_on_labelling(
+        boundary=plymouth_boundary,
+        labelled_gdf=labelled_gdf,
+        feature=feature,
+        threshold=threshold,
+        uprns_gdf=uprns_gdf,
+    )
+
+    # Sense check results with basic decision tree classifier
+    clf = train_model_decision_tree_classifier(
```
simran-dave, line 624: Are we thinking of using a decision tree in the pipeline? Looks like the MCC score is a bit better than using a singular feature for the threshold but maybe it adds too much complexity

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
```diff
@@ -0,0 +1,481 @@
...
+    Returns:
+        folium.Map
+    """
+    # Get domestic EPC UPRNs and join to the building footprints to identify buildings containing domestic EPC records
+    epc_uprns = uprns.load_set_valid_epc_uprns(epc_type="domestic")
+    epc_gdf = uprns_gdf[uprns_gdf["UPRN"].isin(epc_uprns)][["UPRN", "geometry"]]
+    labelled_gdf = (
+        labelled_gdf.sjoin(epc_gdf, how="left", predicate="contains")
+        .drop_duplicates(subset="ID")
+        .astype({"UPRN": "bool"})
```
simran-dave, line 444: I think renaming the column `"UPRN"` to `"EPC"` or something similar could be helpful as it took me a while to understand this isn't a column with UPRNs in it

simran-dave (commented):
hey @crispy-wonton ! 

- I've run` python -i asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py` and the threshold identification steps look as expected. 
- The code produced 3 of the 4 outputs you mentioned, the map `asf_heat_pump_suitability/outputs/maps/plymouth_false_positive_domestic_buildings.html` was not produced and I can't find anywhere in the code that would save a file with this name. The other outputs seem ok. 
- I think I found a bug in the function `plot_folium_threshold_effect_on_labelling` where the function was producing an empty dataframe which meant it wasn't plotting any buildings. I've commented on the lines of code where the bug is and what I did to get it to work for me. 
- I agree with the subset of data that you are using for the threshold identification, I think if we used all data the risk is we could end up with a threshold that mainly removes buildings we were already removing anyway.
- Aside from the bug, the logic in the function `plot_folium_threshold_effect_on_labelling` makes sense to me. From my understanding, we are looking only at buildings that were already being labelled as domestic by the pipeline, removing any that are labelled domestic because of EPC data. And then we are looking at the remainder.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
crispy-wonton: Good question. I shall add some documentation about why I selected MCC to the code.

I tried this with Youden's J statistic again initially by refactoring your code into a function, and the ROC AUC score was okay for the best feature (0.72). However, when I looked at the effects of implementing the threshold in the pipeline, there were a lot of false negatives - i.e. lots of new buildings being labelled as non-domestic when they are actually domestic compared to the original pipeline (it was around 2300, compared to 223 for the best MCC feature and threshold). 

I then looked at the class balance and saw that the classes were extremely imbalanced and figured that using the ROC AUC score (and Youden's J statistic) was less effective due to it having less sensitivity to class imbalance, resulting in overly optimistic results of classification performance.

After researching alternatives, I selected Matthew's Correlation Coefficient as a metric to optimise for as it is considered to be a robust metric for imbalanced data.

According to [this paper](https://link.springer.com/article/10.1186/s12864-019-6413-7): 'MCC is the only binary classification rate that generates a high score only if the binary predictor was able to correctly predict the majority of positive data instances and the majority of negative data instances [[80](https://link.springer.com/article/10.1186/s12864-019-6413-7#ref-CR80), [97](https://link.springer.com/article/10.1186/s12864-019-6413-7#ref-CR97)].'

crispy-wonton: To address this question I have:
1. Added a docstring at the top of the script to explain the analysis and the decisions.
2. Added the function to calculate ROC AUC score and Youden's J statistic back into the file so that you can see the comparison against MCC.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
crispy-wonton: Great catch thank you! The `np.nan` behaviour caught me out here as I expected it to evaluate to `False`.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
crispy-wonton: Good idea - I have renamed!

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
crispy-wonton: I did not because I was just using the classifier to compare our final threshold against. Since we're not implementing the classifier in the pipeline, I don't think it's necessary to conduct hyperparameter tuning. I set `max_leaf_nodes` as 5 here to use a small value to avoid overfitting. To tune this properly, we would need to split the data into a train and test set as increasing the leaf nodes will just result in more and more overfitting.

To be honest, I'm not sure how useful it is to include it as a comparison without tuning but I just wanted to compare the threshold output to a bare-bones classifier to see if it made a big difference. Of course, maybe it _would_ make a big difference if we put effort into training it properly. This will be something to explore in the next phase of the project.

Let me know your thoughts.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
crispy-wonton: Thanks for thinking about this. I agree with the points you've made here and in your main comment.

My main concern was also this (taken from your [main comment](https://github.com/nestauk/asf_heat_pump_suitability/pull/298#pullrequestreview-4140554029)): 

> if we used all data the risk is we could end up with a threshold that mainly removes buildings we were already removing anyway.

So I think even though it would give us more training data technically, it wouldn't really be useful training data for the purpose of the selecting a threshold to apply after the main pipeline. It would, however, be the full dataset we can use if we end up building a general classifier model to replace our whole filtering criteria.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
crispy-wonton: For now I've opted not to implement the classifier in the pipeline because I'd want to do further evaluation and hyperparameter tuning on it before implementing which we don't have time for at the moment. I also think the improvement is not significant enough to warrant adding it without further evaluation/tuning. Basically, I feel more confident implementing a new filtering threshold at the last minute, rather than an untuned model.

In general, I would also want us to put our efforts into testing out replacing the whole filtering criteria with a model, instead of using a combination of both. 

It is something that I can flag in a stand up though.

asf_heat_pump_suitability/utils/plotting_utils.py
crispy-wonton: I'll leave off the x label as it's covered by the title of each plot, but have added the y label :)

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
```diff
@@ -397,7 +497,7 @@ def train_model_decision_tree_classifier(
 
     # Pipeline outputs
     pipeline_domestic_uprns_df = base_getters.load_df_from_s3(
-        config["data"]["processed"]["plymouth_residential_uprns"]
+        "s3://asf-heat-pump-suitability/local_heat_planning/static/domestic_identification/plymouth_residential_uprns.parquet"
```
crispy-wonton, line 544: Note: the reason I've changed this to a static line is because when we update the pipeline in `dev` to add the new threshold, the file that is called from the config will be different, which will create different results when re-running this analysis. To preserve the original results, I have saved a copy of this file to a different location in S3.

crispy-wonton:
hi @simran-dave - thanks so much for your review. I thought it was excellent and thoughtful, with some great questions that prompted me to think about the methodology more.

I have responded to your comments and questions and pushed changes I've made based on your comments. Thank you for spotting the bugs, which I think I have fixed now - please could you double check on the second review and check the map is now working for you. I have successfully run the script through and created the plots so hopefully it works for you now to. Thanks for running the script and checking the outputs and that everything works for you!

Responding to a couple of your bullet points here:
> * The code produced 3 of the 4 outputs you mentioned, the map `asf_heat_pump_suitability/outputs/maps/plymouth_false_positive_domestic_buildings.html` was not produced and I can't find anywhere in the code that would save a file with this name. The other outputs seem ok.

That's my bad - I think the missing output was a legacy plot that I had created in an earlier commit and then removed the code for. You're correct that there should only be 3 plots.


> * Aside from the bug, the logic in the function `plot_folium_threshold_effect_on_labelling` makes sense to me. From my understanding, we are looking only at buildings that were already being labelled as domestic by the pipeline, removing any that are labelled domestic because of EPC data. And then we are looking at the remainder.

Great! That is what I was aiming to filter to.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
simran-dave: Yeah that all makes sense!

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
simran-dave: Yeah in the updates I've made to the council tax filtering branch I've also seen that the threshold was removing a lot of truly domestic properties. Makes sense that the classes are imbalanced- I assume there are a lot more domestic buildings than non-domestic? The docstring is really helpful, thanks!!

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
simran-dave: yeah, thats fair! I think it's an interesting comparison anyway.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
simran-dave: Yes, I think if we were to include a decision tree it would make sense to try replace all the filtering steps with one. Or at least test replacing the whole thing vs replacing just this step

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
simran-dave, line 273: now that you are also printing ROC AUC scores could be worth printing 'MCC percentages removed for each feature' or something similar

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
```diff
@@ -258,6 +275,87 @@ def generate_df_threshold_evaluation(
...
+        pl.DataFrame: best feature with Youden's J statistic for each threshold tested with some summary statistics
+    """
+    # Find best feature for thresholding from ROC AUC score
+    scores = {
+        feature: roc_auc_score(model_df["actual_domestic"], model_df[feature])
+        for feature in features
+    }
+    print("ROC AUC scores for each feature:")
+    print(pl.DataFrame(scores))
+    best_feature = max(scores, key=scores.get)
```
simran-dave: For some of the features, the ROC AUC score is < 0.5 . This could be wrong but I think this is since ROC AUC assumes a higher value of the feature for True and lower for False if you just feed it the raw feature values. If this is true, using max might not actually find the best score. You would need to do 1 - score for any that have a score < 0.5. But maybe double check if that is all true.

simran-dave (commented):
hey @crispy-wonton ! I've run the code again and all works as expected, and the folium map is working for me now :) 
Thanks for your response to my questions- I think using MCC makes sense to me and the docstring/ your reply cleared up why we would use this.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
crispy-wonton: Good to know that you also saw a similar effect in your analysis - I saw it on my review!
Yes, the dataset is like 97% domestic.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
crispy-wonton: Yes you are correct! This was my mistake. I have fixed this by determining the direction of the threshold as well as the threshold itself. Please check I have fixed it correctly :) if so, then it hasn't made a difference to the ultimate feature selection using ROC AUC and thus the results are the same.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
crispy-wonton: Please could you explain what you mean here? I'm not sure what this means :D

crispy-wonton:
Thanks @simran-dave for flagging the bug about the ROC AUC score - have fixed!

I also realised that implementation of my MCC threshold exploration was wrong. Similar to the ROC / AUC score issue, I was not looking at the absolute MCC value to account for reverse predictions. Actually this was deliberate on my part as I originally thought it didn't make a difference to the results - I could see that the absolute values of the negative MCC scores were not larger than the MCC score for the selected feature. HOWEVER, I forgot to account for the fact that the final MCC we see is only the 'best' for that feature, so it was possible that there was a greater absolute value hiding in the for loop.

After correcting for this, I also realised there were two redundant features which were just the reverse of other features, so I removed them.

I have corrected for this as well now. Ultimately, none of my changes have affected the final feature selection and results. Please do have a look at the new logic and let me know if there's any new issues as I did get a bit confused implementing this!

There's still one final thing TODO which I have added in a note for correcting.

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
simran-dave: sorry I just meant to print a statement to terminal before you print these scores saying that these are the MCC scores. As afterwards you print the ROC AUC scores to the terminal as well

crispy-wonton: ohhh I see - makes sense will add!

crispy-wonton:
@simran-dave right, I have cleaned up the two functions to generate the thresholds optimising for MCC and ROC AUC score. They were getting really messy and hard to read so I refactored with some help from ✨ gemini ✨ which introduced me to the `operator` package.

I've run the code and it's giving the same results as before (just now with the fixed `N_removed_true_positives` values in the function to threshold with ROC AUC which was what the TODO was related to). So I'm fairly confident I haven't introduced new bugs but please do check.

I think this should be the final major change!

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
```diff
@@ -0,0 +1,628 @@
...
+            "N_removed_true_positives",
+        ]
+    )
+
+    print(f"Youden's score for each threshold tested for {best_feature}...")
+    print(youdens_df)
+    return youdens_df
+
+
+def extract_tuple_best_feature_threshold(df: pd.DataFrame) -> dict:
```
simran-dave:
```suggestion
def extract_dict_best_feature_threshold(df: pd.DataFrame) -> dict:
```

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
```diff
@@ -0,0 +1,628 @@
...
+    """
+    Exract the best feature and corresponding threshold for implementation by maximising for Matthew's Correlation Coefficient.
+
+    Args:
+        df (pd.DataFrame): dataframe of features, best thresholds, and maximum MCC values
+
+    Returns:
+        tuple: feature name and best threshold value
```
simran-dave:
```suggestion
        dict: feature name, best threshold value and direction of threshold
```

asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
```diff
@@ -0,0 +1,628 @@
...
+        model_df=features_df, features=features
+    )
+
+    # Find best thresholds for each feature using ROC AUC score and Youden's J statistic
+    youdens_thresholds_df = generate_df_threshold_evaluation_roc_auc(
+        model_df=features_df, features=features
+    )
+
+    # Get best feature, threshold, and direction
+    best_combo = extract_tuple_best_feature_threshold(mcc_thresholds_df)
```
simran-dave:
```suggestion
    best_combo = extract_dict_best_feature_threshold(mcc_thresholds_df)
```

simran-dave (changes requested):
hey @crispy-wonton ! I ran the code and I am also getting the same results as before. The new logic with the `operator` package makes sense to me. There's just one function where we are now outputting a dict instead of a tuple and the name / docstring hasn't been updated.

crispy-wonton:
@simran-dave great spot thank you! Have committed your suggested changes :D

simran-dave (approved):

PR #299, opened 2026-04-16
297 add uprn density threshold

### LINKED ISSUE #290: Add new criterion to domestic filtering

crispy-wonton:
Fixes #290 

# Description

Based on the analysis in PR #298 , this PR adds a new step to the domestic UPRN identification pipeline which removes buildings not containing domestic EPC records and which have a `m2_per_predicted_UPRN` value above the threshold identified in the analysis. These buildings are thought to be mostly non-domestic.

Modified files:
- asf_heat_pump_suitability/pipeline/transform/uprns.py - add new logic to exclude additional buildings as above.
- asf_heat_pump_suitability/config/base.yaml - add new threshold to config and move existing thresholds into the new `threshold` key.
- asf_heat_pump_suitability/pipeline/transform/decision_tree.py - update existing threshold to use new config key

# Instructions for Reviewer

run the following line in your terminal:
`python -i asf_heat_pump_suitability/pipeline/transform/uprns.py --local_authorities plymouth`
You can then interact with the output dataframe `df` to inspect it. Please could you take a look and flag if you notice anything strange. You can compare it to the current `dev` outputs of the UPRN filtering:
`original_df = pl.read_parquet("s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth/plymouth_residential_uprns.parquet")`

If you prefer to do it in a notebook instead of interactive mode, you can also load the updated pipeline outputs from this file:
`s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth/plymouth_residential_uprns_test.parquet`

I did my own comparison and will send you the results on slack.

- The logic of the newly implemented filtering step. Is it applied correctly and in the right place?

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/config/base.yaml
+++ b/asf_heat_pump_suitability/config/base.yaml
@@ -34,6 +34,8 @@ data:
     plymouth_residential_uprns: "s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_residential_uprns.parquet"
 
 constant:
+  threshold:
+    m2_per_predicted_UPRN: 261.608136 # Identified by analysis in /research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
   plymouth:
     la_names: "Plymouth"
     grid_squares: "SX"
--- a/asf_heat_pump_suitability/config/base.yaml
+++ b/asf_heat_pump_suitability/config/base.yaml
@@ -34,8 +34,13 @@ data:
     plymouth_residential_uprns: "s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_residential_uprns.parquet"
 
 constant:
+  id:
+    building: "ID"
   threshold:
     m2_per_predicted_UPRN: 261.608136 # Identified by analysis in /research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
+    outdoor_space_threshold_m2:
+      within_hn_zone: 70
+      outside_hn_zone: 30
   plymouth:
     la_names: "Plymouth"
     grid_squares: "SX"
@@ -91,9 +96,6 @@ constant:
     la_names: "Cardiff"
     grid_squares: "ST"
   target_crs: 27700 # British National Grid
-  outdoor_space_threshold_m2:
-    within_hn_zone: 70
-    outside_hn_zone: 30
   tech_types:
     individual: "Individual solution"
     networked: "Networked heat pump"
--- a/asf_heat_pump_suitability/pipeline/transform/uprns.py
+++ b/asf_heat_pump_suitability/pipeline/transform/uprns.py
@@ -177,7 +177,7 @@ def _generate_gdf_non_domestic_buildings_by_density(
     domestic_uprns_gdf: gpd.GeoDataFrame,
     buildings_gdf: gpd.GeoDataFrame,
     id_col: str,
-    threshold: float = config["threshold"]["m2_per_predicted_UPRN"],
+    threshold: float = config["constant"]["threshold"]["m2_per_predicted_UPRN"],
 ) -> gpd.GeoDataFrame:
     """
     Generate a GeoDataFrame containing footprints of buildings which contain a UPRN predicted to be domestic and have a
--- a/asf_heat_pump_suitability/pipeline/transform/uprns.py
+++ b/asf_heat_pump_suitability/pipeline/transform/uprns.py
@@ -138,7 +138,7 @@ def filter_gdf_domestic_uprns(
     )
 
     # Get valid residential UPRNs
-    epc_residential_uprns = load_set_valid_epc_uprns(epc_type="domestic")
+    domestic_epc_uprns = load_set_valid_epc_uprns(epc_type="domestic")
 
     domestic_uprn_gdf = uprn_gdf[
         (
@@ -147,7 +147,7 @@ def filter_gdf_domestic_uprns(
             & (uprn_gdf["UPRN"].isin(uprns_in_buildings))
         )
         # Or UPRNs which are in domestic EPC register
-        | (uprn_gdf["UPRN"].isin(epc_residential_uprns))
+        | (uprn_gdf["UPRN"].isin(domestic_epc_uprns))
     ]
 
     # TODO this could be updated to a classification model and scaled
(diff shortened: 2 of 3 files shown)


asf_heat_pump_suitability/config/base.yaml
```diff
@@ -34,6 +34,13 @@ data:
     plymouth_residential_uprns: "s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_residential_uprns.parquet"
 
 constant:
+  id:
+    building: "ID"
+  threshold:
+    m2_per_predicted_UPRN: 261.608136 # Identified by analysis in /research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
```
simran-dave, line 40: This isn't relevant now but for the future, are we thinking of using one threshold for every area or identifying a different threshold for each?

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -134,18 +138,103 @@ def filter_gdf_residential_uprns(
...
+    domestic_uprns_gdf: gpd.GeoDataFrame,
+    buildings_gdf: gpd.GeoDataFrame,
+    epc_uprns: set,
+    id_col: str,
+    threshold: float = config["constant"]["threshold"]["m2_per_predicted_UPRN"],
+) -> gpd.GeoDataFrame:
+    """
+    Generate a GeoDataFrame containing footprints of buildings which contain a UPRN predicted to be domestic and have a
+    `m2_per_predicted_UPRN` value above the defined threshold. The aim of this function is to remove some of the true
+    non-domestic buildings mislabelled as domestic by removing those with large building footprints and low UPRN density.
```
simran-dave: could add that we are still retaining any building with at a UPRN in the domestic EPC register, regardless of size

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -98,24 +98,28 @@ def load_set_valid_epc_uprns(epc_type: str) -> set:
...
     buildings_gdf: gpd.GeoDataFrame,
     non_residential_buildings_gdf: gpd.GeoDataFrame,
+    local_authority: str,
+    id_col: str = config["constant"]["id"]["building"],
 ) -> gpd.GeoDataFrame:
     """
-    Filter UPRNs to residential UPRNs only by retaining UPRNs which appear in domestic EPC register, OR are located within
+    Filter UPRNs to domestic UPRNs only by retaining UPRNs which appear in domestic EPC register, OR are located within
     a building footprint AND are not in the commercial EPC register and / or a building type that is unlikely to contain
     residential properties, e.g. hospital, train station, museum etc.
```
simran-dave:
```suggestion
    residential properties, e.g. hospital, train station, museum etc, AND are in a building with 'm2_per_predicted_UPRN' below a defined threshold.
```

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -134,18 +138,103 @@ def filter_gdf_residential_uprns(
...
+) -> gpd.GeoDataFrame:
+    """
+    Generate a GeoDataFrame containing footprints of buildings which contain a UPRN predicted to be domestic and have a
+    `m2_per_predicted_UPRN` value above the defined threshold. The aim of this function is to remove some of the true
+    non-domestic buildings mislabelled as domestic by removing those with large building footprints and low UPRN density.
+    The threshold selected as default (set in config) was determined through analysis that can be found in
+    asf_heat_pump_suitability/research/exploratory/domestic_filtering/domestic_building_identification_threshold_selection.py
+
+    Args:
+        domestic_uprns_gdf (gpd.GeoDataFrame): UPRNs which are assumed to represent domestic properties with their point geometries.
```
simran-dave: maybe instead of 'assumed' could say 'predicted by the pipeline' or something like that so it's clearer where this assumption comes from

simran-dave (commented):
hey @crispy-wonton ! Looks to me like the logic is working so only left some minor comments :)

- I ran `python -i asf_heat_pump_suitability/pipeline/transform/uprns.py --local_authorities plymouth` and everything ran ok, the output looks as expected.
- I had a look at the new df and compared to the old one and got the same results as you. 
- I also took a look at m2 per predicted UPRN for the buildings that had been removed and they are all above the threshold so looks like everything is working as expected.
- The logic of the newly implemented step looks good to me and I agree this is the best place to implement it, after the rest of the filtering pipeline has been applied.

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: Yes, great point!

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: Yeah good question. 

I guess you saw this when you went through the rest of the code but I've set the code up with the [snippet here](https://github.com/nestauk/asf_heat_pump_suitability/blob/15012ffdfa7857f48893b99c6d99be4a381b0782/asf_heat_pump_suitability/pipeline/transform/uprns.py#L154) so that this whole additional filtering step with the threshold only triggers for Plymouth local authority. We would need to test this on other areas before implementing more broadly - I suspect the same threshold would not be appropriate for every LA. In general, we definitely need to think about how we scale the pipelines so that they are generic enough to work for different areas but not so much that they significantly reduce the performance on individual LAs.

We will have to explore what exactly we do here in the next phase. It's possible that we will update this whole part of the pipeline to a classifier model.

crispy-wonton:
hey @simran-dave , thanks so much for your review. I really appreciate you checking and comparing the results so carefully! Good idea to check the buildings that got filtered out to make sure the threshold is implemented correctly! This prompted me to also ensure that all the domestic UPRNs are still retained in the final set and it looks like they are. For reference, to check this I ran the pipeline in interactive mode, then ran:

```
# Get all EPC records domestic
epc_uprns = load_set_valid_epc_uprns(epc_type="domestic")
# Filter to those in Plymouth LA (uprns_gdf == all UPRNs in Plymouth LA)
epc_gdf = uprns_gdf[uprns_gdf["UPRN"].isin(epc_uprns)].copy()
# Check if any EPC UPRNs in Plymouth are not in the final df (df == final output of uprns.py script)
len(set(epc_gdf["UPRN"]).difference(set(df["UPRN"])))
```
len of the set difference was 0.

simran-dave (approved):
hey @crispy-wonton , good to know about all the EPC UPRNs being retained- thanks for checking!

PR #303, opened 2026-04-23
302 Reduce cluster file size

### LINKED ISSUE #302: Reduce file size of final cluster output
Since productionising clustering, the final file size is very large compared with the original notebook outputs. This needs to be fixed.

crispy-wonton:
Fixes #302 

# Description

This fixes the file size problem related to the clustering. The new file size with these changes is 48MB.

The issue was caused by a line in the function to handle Voronoi cell fragments that was adding a buffer to the Voronoi polygons of 1cm. This step was approx doubling the total number of vertices in the geodataframe. The buffering process can add loads of points at the corners of a polygon through densification to ensure rounded corners. This can be addressed by using buffer with `join_style=2` which gives mitre corners and by simplifying the resulting buffered polygons with a very low tolerance (e.g. 1mm).

However, I applied a new methodology entirely to remove use of the buffer. As well as size reduction of the final file, the new methodology is simpler and therefore gives significant speed gains.

Modified files:
- `asf_heat_pump_suitability/pipeline/cluster/cluster.py` - update the function to handle cell fragments to use a series of `sjoin`s to handle fragments, rather than a buffer.

# Instructions for Reviewer

run the following line in your terminal:
`python asf_heat_pump_suitability/pipeline/cluster/cluster.py --local_authorities plymouth`

I have run it and saved the results to s3 and it all worked.

Please could you review the resulting geodataframe and let me know if you see anything strange. I have inspected and plotted it and it looks ok to me but have not plotted with folium to look in detail at the clusters.

```
import geopandas as gpd
clusters_gdf = gpd.read_file("s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth/plymouth_clustered_tech_polygons.geojson")
colour_mapping = {
    "Communal solution" : "red",
    "District heat network" : "green",
    "Individual solution" : "blue",
    "Networked heat pump" : "purple",
}
clusters_gdf["colour"] = clusters_gdf["assigned_tech"].map(colour_mapping)
clusters_gdf.plot(color=clusters_gdf['colour'])
```

The logic of the new function. Let me know if the purpose of the function is not clear from the documentation and I can explain further what it's supposed to be doing.

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/pipeline/cluster/cluster.py
+++ b/asf_heat_pump_suitability/pipeline/cluster/cluster.py
@@ -249,7 +249,9 @@ def _clip_gdf_voronoi_cells_polygon_buffer(
     """
     # Create a buffered polygon for all polygons
     buffered_gdf = polygon_gdf[[id_col, "geometry"]].copy()
-    buffered_gdf["geometry"] = buffered_gdf.geometry.buffer(buffer)
+    buffered_gdf["geometry"] = buffered_gdf.geometry.buffer(
+        buffer, join_style=2, mitre_limit=2
+    ).simplify(0.001)
 
     # Clip Voronoi cells to the defined buffer area if they are larger than it by calculating intersections
     clipped_gdf = voronoi_gdf.overlay(buffered_gdf, how="intersection")
@@ -320,28 +322,25 @@ def _handle_gdf_fragmented_cells(
     Returns:
         gpd.GeoDataFrame: domestic building cells with overlapping physical barriers removed and cell fragments handled
     """
-    # Reduce the polygons by 1cm to avoid unions of touching cells
-    # Then union cells with building footprints
-    union = pd.concat(
-        [cells_gdf["geometry"].buffer(-0.01), tech_gdf["geometry"].buffer(-0.01)]
-    ).unary_union
+    # Add a temporary ID for each building
+    id_col = "_internal_building_id"
+    tech_gdf[id_col] = np.arange(len(tech_gdf))
+    cols = [id_col, "geometry"]
 
-    # Explode union to get one unionised polygon per building
-    union_gdf = gpd.GeoDataFrame(
-        geometry=gpd.GeoSeries(union).explode(index_parts=False),
-        crs=config["constant"]["target_crs"],
+    # Keep only cell fragments that intersect with a building and label with the ID of the building
+    cells_gdf = cells_gdf.sjoin(
+        tech_gdf[cols], how="inner", predicate="intersects"
+    ).drop(columns=["index_right"])
+
+    # Dissolve building footprints and their cell fragments together
+    union_gdf = (
+        pd.concat([cells_gdf[cols], tech_gdf[cols]])
+        .dissolve(by=id_col)
+        .rename(columns={"geometry": "unionised_geometry"})
     )
-
-    # Add 1cm buffer back so that dissolving works later
-    union_gdf["geometry"] = union_gdf["geometry"].buffer(0.01)
-
-    # Retain original unionised cell geometry - this will be the final geometry per building unless it's missing
-    union_gdf["unionised_geometry"] = union_gdf["geometry"]
 
     # Join unionised cells back to original buildings
-    cells_gdf = tech_gdf.sjoin(union_gdf, how="left", predicate="intersects").drop(
-        columns=["index_right"]
-    )
+    cells_gdf = tech_gdf.merge(union_gdf, how="left", on=id_col)
 
     # If building is missing a Voronoi cell (due to overlay operation), then assign it the building footprint geometry
     # This happens in some edge cases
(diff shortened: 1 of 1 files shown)


asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -640,14 +643,9 @@ def parse_arguments() -> argparse.Namespace:
...
-        # Simplify geometry for file size to tolerance of 5m
-        tolerance = 5
-        clusters_gdf["geometry"] = clusters_gdf["geometry"].simplify(
-            tolerance=tolerance
-        )
         save_utils.save_to_s3(
             clusters_gdf,
             config["output"]["dataset"]["tech_clusters"].format(
-                local_authorities=args.local_authorities, tolerance=tolerance
+                local_authorities=args.local_authorities
```
sofiapinto: Hey @crispy-wonton I don't think we want to remove this simplification at the end. I know we added it in the previous PR as it helped us with the file size, but this is currently happening on the browser regardless, when people first open the tool, and we want to move it to our end.

I haven't reviewed the code so I might have missed this happening elsewhere?

crispy-wonton: True! Will add back in!

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -320,28 +326,25 @@ def _handle_gdf_fragmented_cells(
...
+    # Keep only cell fragments that intersect with a building and label with the ID of the building
+    cells_gdf = cells_gdf.sjoin(
+        tech_gdf[cols], how="inner", predicate="intersects"
+    ).drop(columns=["index_right"])
 
-    # Retain original unionised cell geometry - this will be the final geometry per building unless it's missing
-    union_gdf["unionised_geometry"] = union_gdf["geometry"]
+    # Dissolve building footprints and their cell fragments together
+    union_gdf = (
+        pd.concat([cells_gdf[cols], tech_gdf[cols]])
```
simran-dave, line 341: I think I'm not fully following why we concat the two gdfs here. From my understanding we have already sjoined `cells_gdf `and `tech_gdf` so I don't follow why we need to concat `tech_gdf` here again

simran-dave (commented):
hey @crispy-wonton ! Looks good to me, I've run `python asf_heat_pump_suitability/pipeline/cluster/cluster.py --local_authorities plymouth` and it ran as expected. I also made a folium map using the following and the clusters looked good : 
```
def plot_clusters(clusters: gpd.GeoDataFrame):

    clusters_4326_gdf = clusters.to_crs(epsg=4326)

    # Get centre of first polygon in gdf to centre map
    centre_4326 = clusters_4326_gdf["geometry"].values[0]
    centre_map = shapely.get_coordinates(centre_4326.centroid)

    # Create map
    m = folium.Map(location=[centre_map[0][1], centre_map[0][0]], zoom_start=15, tiles="esri_worldimagery")


    for _, r in clusters_4326_gdf.iterrows():
        colour = r["colour"]
        sim_geo = gpd.GeoSeries(r["geometry"])
        geo_j = sim_geo.to_json()
        geo_j = folium.GeoJson(data=geo_j, style_function=lambda x, colour=colour: {"fillColor": colour, 'weight':0.1, "fillOpacity": 0.66})
        geo_j.add_to(m)
    return m
```
I just had one question on a step I don't quite follow, which I have left as a comment

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Let me know if you think this is not correct. Here's my intention with the logic in this function:

Step 1: filter the cell fragments. This step should basically filter the cells_gdf to only those rows which intersect with a building footprint, while retaining the cells_gdf geometries. According to `geopandas` [docs](https://geopandas.org/en/v1.1.2/docs/reference/api/geopandas.sjoin.html#geopandas-sjoin), the left_gdf geometries are retained when using an `inner` `sjoin`. The purpose of this step is (a) to filter the cell fragments to relevant ones, and (b) to join the `_internal_building_id` to the cells.
```
# Keep only cell fragments that intersect with a building and label with the ID of the building
    cells_gdf = cells_gdf.sjoin(
        tech_gdf[cols], how="inner", predicate="intersects"
    ).drop(columns=["index_right"])
```

Step 2: concat the building footprints contained in `tech_gdf` to the `cell_gdf`. This should give us a geodataframe with (a) cell fragment geometries with corresponding `_internal_building_id` and (b) building footprint geometries with corresponding `_internal_building_id`.

Step 3: dissolve geometries by `_internal_building_id` so we're flattening all the geometries related to the same building ID into one polygon. This will give one polygon per building. The polygon should be the geospatial union of Voronoi cell fragments + building footprint. In most cases, this will just be the original voronoi cell, but for edge cases, some of the building footprint will be visible.

simran-dave: ah I see - I wasn't understanding that we wanted to keep both the cell and building geometries. This makes more sense to me now, thanks :)

simran-dave (approved):

PR #304, opened 2026-04-24
300 Append Heat Network zones to clusters_gdf

### LINKED ISSUE #300: Append HN zones to the cluster data

### LINKED ISSUE #301: Update HN zones data version for Plymouth

sofiapinto:
---

# Description

Append HN zones to clusters_gdf and change the source data for the Plymouth HN zones.

Fixes #300 and #301

Main changes to the codebase include:
- `asf_heat_pump_suitability/config/base.yaml`: source data for Plymouth planned HN zone
- `asf_heat_pump_suitability/getters/load_geodata.py`: I moved some of the code that lived in `add_features.py` to `load_geodata.load_gdf_heat_network_zones()` as I was going to have to repeat it in `cluster.py`. `load_geodata.load_gdf_heat_network_zones()` now allows for loading HN zones for a specific LA, CA or group of LAs.
- `asf_heat_pump_suitability/pipeline/run/add_features.py`: update to this script, to reflect changes in how we load HN zones.
- `asf_heat_pump_suitability/pipeline/cluster/cluster.py`: Appending heat network zones to clusters_gdf, if they exist

I have also run the following:

`add_features.py` for 3 different areas to ensure results still hold:
- `python asf_heat_pump_suitability/pipeline/run/add_features.py --local_authorities "plymouth"`
- `python asf_heat_pump_suitability/pipeline/run/add_features.py --local_authorities "cardiff"`
- `python asf_heat_pump_suitability/pipeline/run/add_features.py --local_authorities "greater_manchester_las" --detail "simplified"`

`cluster.py`:
- python asf_heat_pump_suitability/pipeline/run/add_features.py --local_authorities "plymouth"

# Instructions for Reviewer

The changes ended up being more than just a few lines, due to moving some code to `load_geodata.load_gdf_heat_network_zones()`. Could you please ensure the logic holds up and that the documentation is clear?

Can you also please check I used the correct version of the HN zones for Plymouth?

### DIFF
--- a/asf_heat_pump_suitability/config/base.yaml
+++ b/asf_heat_pump_suitability/config/base.yaml
@@ -17,7 +17,7 @@ data:
 
     heat_network_zones:
       # Plymouth Heat Network zones
-      plymouth: "s3://asf-heat-pump-suitability/local_heat_planning/plymouth_inputs/NMR3 Heat Network Zones (Post-Ref).gpkg"
+      plymouth: "s3://asf-heat-pump-suitability/heat_network_desnz_data/heat-network-zone-map-Plymouth.gpkg"
       greater_manchester_las: "s3://asf-heat-pump-suitability/heat_network_desnz_data/heat-network-zone-map-Greater-Manchester.gpkg"
 
     gb_spatial_signatures:
--- a/asf_heat_pump_suitability/getters/load_geodata.py
+++ b/asf_heat_pump_suitability/getters/load_geodata.py
@@ -50,7 +50,7 @@ def load_gdf_heat_network_zones(local_authority: str, **kwargs) -> gpd.GeoDataFr
     Args:
         local_authority (str): Local Authority or Local Authorities to load Heat Network zone polygons for.
         e.g. `plymouth` for Plymouth Local Authority; `greater_manchester_las` for Greater Manchester Combined Authority (all 10 LAs in Greater Manchester).
-        See config/base.yaml for options.
+        See config/base.yaml under the `constant` key for options.
 
         **kwargs for `gpd.read_file()`
 
@@ -58,7 +58,8 @@ def load_gdf_heat_network_zones(local_authority: str, **kwargs) -> gpd.GeoDataFr
         gpd.GeoDataFrame: polygons of heat network zones in given Local Authority.
     """
 
-    # TODO: this will currently only work for HN zone files defined in config/base.yaml. We need to change this to make it work for all other HN zone files,
+    # TODO: this will currently only work for HN zone files defined in config/base.yaml.
+    # We need to change this to make it work for all other HN zone files,
     # for example by concatenating all HN zone files and checking if the geometries intersect with the local authority boundary.
 
     gdf = gpd.GeoDataFrame()
@@ -72,17 +73,23 @@ def load_gdf_heat_network_zones(local_authority: str, **kwargs) -> gpd.GeoDataFr
             path=config["data"]["geodata"]["heat_network_zones"][local_authority],
             **kwargs,
         )
+        # Assume first column with `ID` substring is the zone ID column
+        # Note original ID column retained in case of erroneous ID assignment
+        id_col = [col for col in gdf.columns if "ID" in col][0]
+        gdf["HNZoneID"] = gdf[id_col]
+        print(f"Using Heat Network Zone {id_col} column as ID")
     except (ValueError, KeyError):
         print(f"No heat network zone geodata found for {local_authority}.")
 
-    # Get list of LAs (e.g. for `greater_manchester_las` this means getting a list of all indivividual LAs) to attempt loading heat network zone geodata for each LA individually if no geodata found for the whole group of LAs.
-    la_names = config["constant"][local_authority]["la_names"]
-    list_las = la_names if isinstance(la_names, list) else [la_names]
-    del la_names
+    # Get list of LAs (e.g. for `greater_manchester_las` this means getting a list of all individual LAs) to attempt
+    # loading heat network zone geodata for each LA individually if no geodata found for the whole group of LAs.
+    list_las = config["constant"][local_authority]["la_names"]
+    list_las = list_las if isinstance(list_las, list) else [list_las]
 
     # If gdf is still empty and `local_authority` represents a group of LAs
     if gdf.empty and len(list_las) > 1:
-        # Check if heat network zone geodata is available for each LA in the list, and if so, load it and concatenate it to a single geodataframe.
+        # Check if heat network zone geodata is available for each LA in the list, and if so, load it and concatenate
+        # it to a single geodataframe.
         for la in list_las:
             try:
                 gdf = pd.concat(
@@ -97,9 +104,10 @@ def load_gdf_heat_network_zones(local_authority: str, **kwargs) -> gpd.GeoDataFr
                 )
                 # Deal with different ID column names in different geodataframes by renaming the ID column to "ZoneID"
                 id_col = [col for col in gdf.columns if "ID" in col][0]
-                gdf.rename(columns={id_col: "ZoneID"}, inplace=True)
+                gdf["HNZoneID"] = gdf[id_col]
+                print(f"Using Heat Network Zone {id_col} column as ID")
             except (ValueError, KeyError):
-                print(f"No heat network zone geodata found for individual {la}.")
+                print(f"No heat network zone geodata found for Local Authority: {la}.")
 
     if len(gdf) > 0:
         gdf.set_geometry("geometry", inplace=True)
--- a/asf_heat_pump_suitability/getters/load_geodata.py
+++ b/asf_heat_pump_suitability/getters/load_geodata.py
@@ -79,41 +79,43 @@ def load_gdf_heat_network_zones(local_authority: str, **kwargs) -> gpd.GeoDataFr
         gdf["HNZoneID"] = gdf[id_col]
         print(f"Using Heat Network Zone {id_col} column as ID")
     except (ValueError, KeyError):
-        print(f"No heat network zone geodata found for {local_authority}.")
+        # Get list of LAs (e.g. for `greater_manchester_las` this means getting a list of all individual LAs) to attempt
+        # loading heat network zone geodata for each LA individually if no geodata found for the whole group of LAs.
+        list_las = config["constant"][local_authority]["la_names"]
+        list_las = list_las if isinstance(list_las, list) else [list_las]
 
-    # Get list of LAs (e.g. for `greater_manchester_las` this means getting a list of all individual LAs) to attempt
-    # loading heat network zone geodata for each LA individually if no geodata found for the whole group of LAs.
-    list_las = config["constant"][local_authority]["la_names"]
-    list_las = list_las if isinstance(list_las, list) else [list_las]
-
-    # If gdf is still empty and `local_authority` represents a group of LAs
-    if gdf.empty and len(list_las) > 1:
-        # Check if heat network zone geodata is available for each LA in the list, and if so, load it and concatenate
-        # it to a single geodataframe.
-        for la in list_las:
-            try:
-                gdf = pd.concat(
-                    [
-                        gdf,
-                        base_getters.get_gdf_from_gpkg_s3_path(
-                            path=config["data"]["geodata"]["heat_network_zones"][la],
-                            **kwargs,
-                        ),
-                    ],
-                    ignore_index=True,
-                )
-                # Deal with different ID column names in different geodataframes by renaming the ID column to "ZoneID"
-                id_col = [col for col in gdf.columns if "ID" in col][0]
-                gdf["HNZoneID"] = gdf[id_col]
-                print(f"Using Heat Network Zone {id_col} column as ID")
-            except (ValueError, KeyError):
-                print(f"No heat network zone geodata found for Local Authority: {la}.")
-
-    if len(gdf) > 0:
-        gdf.set_geometry("geometry", inplace=True)
-        print(
-            f"Heat network zone geodataframe successfully loaded for {local_authority} with CRS {gdf.crs}."
-        )
+        # If gdf is still empty and `local_authority` represents a group of LAs
+        if gdf.empty and len(list_las) > 1:
+            # Check if heat network zone geodata is available for each LA in the list, and if so, load it and concatenate
+            # it to a single geodataframe.
+            for la in list_las:
+                try:
+                    gdf = pd.concat(
+                        [
+                            gdf,
+                            base_getters.get_gdf_from_gpkg_s3_path(
+                                path=config["data"]["geodata"]["heat_network_zones"][
+                                    la
+                                ],
+                                **kwargs,
+                            ),
+                        ],
+                        ignore_index=True,
+                    )
+                    # Deal with different ID column names in different geodataframes by renaming the ID column to "ZoneID"
+                    id_col = [col for col in gdf.columns if "ID" in col][0]
+                    gdf["HNZoneID"] = gdf[id_col]
+                    print(f"Using Heat Network Zone {id_col} column as ID")
+                except (ValueError, KeyError):
+                    print(
+                        f"No heat network zone geodata found for Local Authority: {la}."
+                    )
+    finally:
+        if len(gdf) > 0:
+            gdf.set_geometry("geometry", inplace=True)
+            print(
+                f"Heat network zone geodataframe successfully loaded for {local_authority} with CRS {gdf.crs}."
+            )
     return gdf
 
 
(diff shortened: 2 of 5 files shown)


asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -639,6 +642,20 @@ def parse_arguments() -> argparse.Namespace:
...
+    )
+    if len(hn_zones_gdf) > 0:
+        id_col = [col for col in hn_zones_gdf.columns if "ID" in col][0]
+        hn_zones_gdf = hn_zones_gdf.rename(columns={id_col: "cluster_id"}).assign(
+            assigned_tech="DESNZ_HNZ",
+            cluster_id=lambda df: "DESNZ_HNZ_"
+            + df["cluster_id"].astype(str).str.replace("-", "_"),
+        )[["cluster_id", "geometry"]]
+
+        clusters_gdf = pd.concat([clusters_gdf, hn_zones_gdf], ignore_index=True)
```
sofiapinto: Forgot to create a function for this above!

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -17,7 +17,7 @@ data:
 
     heat_network_zones:
       # Plymouth Heat Network zones
-      plymouth: "s3://asf-heat-pump-suitability/local_heat_planning/plymouth_inputs/NMR3 Heat Network Zones (Post-Ref).gpkg"
+      plymouth: "s3://asf-heat-pump-suitability/heat_network_desnz_data/heat-network-zone-map-Plymouth.gpkg"
```
crispy-wonton, line 24: note: have confirmed this is the correct file.

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -48,29 +48,64 @@ def load_gdf_heat_network_zones(local_authority: str, **kwargs) -> gpd.GeoDataFr
     Load GeoDataFrame with heat network zone polygons in given Local Authority.
 
     Args:
-        local_authority (str): Local Authority to load Heat Network zone polygons for.
+        local_authority (str): Local Authority or Local Authorities to load Heat Network zone polygons for.
+        e.g. `plymouth` for Plymouth Local Authority; `greater_manchester_las` for Greater Manchester Combined Authority (all 10 LAs in Greater Manchester).
+        See config/base.yaml for options.
```
crispy-wonton:
```suggestion
        See config/base.yaml under the `constant` key for options.
```

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -48,29 +48,64 @@ def load_gdf_heat_network_zones(local_authority: str, **kwargs) -> gpd.GeoDataFr
...
+    # Load heat network zone geodata for the specified local authority or group of local authorities.
+    try:
+        gdf = base_getters.get_gdf_from_gpkg_s3_path(
+            path=config["data"]["geodata"]["heat_network_zones"][local_authority],
+            **kwargs,
+        )
+    except (ValueError, KeyError):
+        print(f"No heat network zone geodata found for {local_authority}.")
+
+    # Get list of LAs (e.g. for `greater_manchester_las` this means getting a list of all indivividual LAs) to attempt loading heat network zone geodata for each LA individually if no geodata found for the whole group of LAs.
```
crispy-wonton:
```suggestion
    # Get list of LAs (e.g. for `greater_manchester_las` this means getting a list of all individual LAs) to attempt loading heat network zone geodata for each LA individually if no geodata found for the whole group of LAs.
```
question: how come the code below doesn't just trigger in the `except` block? what am i missing?

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -48,29 +48,64 @@ def load_gdf_heat_network_zones(local_authority: str, **kwargs) -> gpd.GeoDataFr
...
+    la_names = config["constant"][local_authority]["la_names"]
+    list_las = la_names if isinstance(la_names, list) else [la_names]
+    del la_names
```
crispy-wonton:
```suggestion
    list_las = config["constant"][local_authority]["la_names"]
    list_las = list_las if isinstance(list_las, list) else [list_las]
```

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -48,29 +48,64 @@ def load_gdf_heat_network_zones(local_authority: str, **kwargs) -> gpd.GeoDataFr
...
+                            **kwargs,
+                        ),
+                    ],
+                    ignore_index=True,
+                )
+                # Deal with different ID column names in different geodataframes by renaming the ID column to "ZoneID"
+                id_col = [col for col in gdf.columns if "ID" in col][0]
+                gdf.rename(columns={id_col: "ZoneID"}, inplace=True)
+            except (ValueError, KeyError):
+                print(f"No heat network zone geodata found for individual {la}.")
```
crispy-wonton:
```suggestion
                print(f"No heat network zone geodata found for Local Authority: {la}.")
```

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -639,6 +642,20 @@ def parse_arguments() -> argparse.Namespace:
         radius=ANCHOR_RADIUS,
     )
 
+    # Add heat network zones to clusters_gdf, if they exist
+    hn_zones_gdf = load_geodata.load_gdf_heat_network_zones(
+        local_authority=local_authorities
+    )
+    if len(hn_zones_gdf) > 0:
+        id_col = [col for col in hn_zones_gdf.columns if "ID" in col][0]
```
crispy-wonton: question: won't `id_col` always be `ZoneID` now because we manually changed it to that in the `load_gdf_heat_network_zones` function?

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: shall i refactor into a function?

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -639,6 +642,20 @@ def parse_arguments() -> argparse.Namespace:
...
+        hn_zones_gdf = hn_zones_gdf.rename(columns={id_col: "cluster_id"}).assign(
+            assigned_tech="DESNZ_HNZ",
+            cluster_id=lambda df: "DESNZ_HNZ_"
+            + df["cluster_id"].astype(str).str.replace("-", "_"),
+        )[["cluster_id", "geometry"]]
```
```suggestion
        )[["assigned_tech", "geometry", "cluster_id"]]
```
The `assigned_tech` column is currently showing `nan` values because this is missing.

crispy-wonton (commented):
hey @sofiapinto ! thanks for these updates.

I have left some questions and minor suggestions - didn't find any issues with the logic in general. As discussed, I'm happy to implement any changes. Please could you take a look at my suggestions and let me know if there's any you don't want me to implement. Secondly, please do take a look at my questions for clarity.

The new HNZ file you've used is the correct one.

I've also rerun `add_features.py` and `cluster.py`. I have compared the results to the current outputs on S3.

For `add_features.py` I've noticed a difference in the results in terms of counts in HNZ. Since we changed the HNZ input file this is expected and the numbers are not drastically different - the proportions are very similar. 25% in HNZ with the new changes, compared to 23% in the original pipeline.

I inspected `clusters_gdf` from the clustering script and can see that the HNZ have been added successfully. After excluding those, I wanted to compare the clusters. I can see there is one new individual solution cluster in the new clusters file when compared to the `dev` outputs. I'm not sure why that is...If you also don't know then I will try re-running both in case I introduced an error into the results somehow. It could also come from the conversion to `geojson`.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
sofiapinto: yes please! I only noticed after publishing the PR for review.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
sofiapinto: Upsi! If you could add it back that would be great.

asf_heat_pump_suitability/getters/load_geodata.py
sofiapinto: Not sure I understand what you mean. If the error is either ValueError or KeyError, the try block stops and the except block runs, and script will continue to run afterwards.

Examples:

- Plymouth: the first try/except block works. The second one doesn't need to run because  gdf isn't empty.
- Greater Manchester: the first try/except block works. The second one doesn't need to run because  gdf isn't empty.
- Cardiff: both try blocks stop and except blocks run, script will continue.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
sofiapinto: No - not necessarily. We would need to add 

```
id_col = [col for col in gdf.columns if "ID" in col][0]
gdf.rename(columns={id_col: "ZoneID"}, inplace=True)
```

to the first try block. And now that you raise it, I think we should do it! 

The reason for adding it to the second one is, in case different LA files had different ID columns, the concat would great different ID columns, which we don't want.

sofiapinto:
Hey @crispy-wonton - thanks for the great questions and quick thorough review!

> hey @sofiapinto ! thanks for these updates.
> 
> I have left some questions and minor suggestions - didn't find any issues with the logic in general. As discussed, I'm happy to implement any changes. Please could you take a look at my suggestions and let me know if there's any you don't want me to implement. Secondly, please do take a look at my questions for clarity.

Have done and replied to all of them - thank you!

> The new HNZ file you've used is the correct one.

Wonderful!

> I've also rerun `add_features.py` and `cluster.py`. I have compared the results to the current outputs on S3.
> 
> For `add_features.py` I've noticed a difference in the results in terms of counts in HNZ. Since we changed the HNZ input file this is expected and the numbers are not drastically different - the proportions are very similar. 25% in HNZ with the new changes, compared to 23% in the original pipeline.

Yes, as expected.

> I inspected `clusters_gdf` from the clustering script and can see that the HNZ have been added successfully. After excluding those, I wanted to compare the clusters. I can see there is one new individual solution cluster in the new clusters file when compared to the `dev` outputs. I'm not sure why that is...If you also don't know then I will try re-running both in case I introduced an error into the results somehow. It could also come from the conversion to `geojson`.

I don't know why that's happening - would you be happy to inspect it? I guess it might be because we're running things from different branches? So might just be worth re-running dev & then running this and inspecting if there really are differences?

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: cool will do!

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: Yeh sorry so I meant this: if the `gdf` is empty, that means the exception got triggered. Therefore, doesn't it make sense that the code below (except for the final `if` statement where you are setting geometry) is located within the `except` block? Or is it possible that the `try` block passes but the `gdf` can still be empty?

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Oh yes you are right - I misremembered and thought it was happening earlier.

asf_heat_pump_suitability/getters/load_geodata.py
sofiapinto: You mean this code could live within the first except block? Yep, I think that makes sense.

```
la_names = config["constant"][local_authority]["la_names"]
list_las = la_names if isinstance(la_names, list) else [la_names]
del la_names
```


Let me know, I might be missing your point

crispy-wonton: yes and this code also:

```
# If gdf is still empty and `local_authority` represents a group of LAs
    if gdf.empty and len(list_las) > 1:
        # Check if heat network zone geodata is available for each LA in the list, and if so, load it and concatenate it to a single geodataframe.
        for la in list_las:
            try:
                gdf = pd.concat(
                    [
                        gdf,
                        base_getters.get_gdf_from_gpkg_s3_path(
                            path=config["data"]["geodata"]["heat_network_zones"][la],
                            **kwargs,
                        ),
                    ],
                    ignore_index=True,
                )
                # Deal with different ID column names in different geodataframes by renaming the ID column to "ZoneID"
                id_col = [col for col in gdf.columns if "ID" in col][0]
                gdf.rename(columns={id_col: "ZoneID"}, inplace=True)
            except (ValueError, KeyError):
                print(f"No heat network zone geodata found for individual {la}.")
```

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -72,40 +73,49 @@ def load_gdf_heat_network_zones(local_authority: str, **kwargs) -> gpd.GeoDataFr
             path=config["data"]["geodata"]["heat_network_zones"][local_authority],
             **kwargs,
         )
+        # Assume first column with `ID` substring is the zone ID column
+        # Note original ID column retained in case of erroneous ID assignment
```
sofiapinto, line 78: If were afraid of erroneous ID assignment why don't we check whether multiple IDs exist within the file instead? There will be no breaking changes from this so it will be the same as not keeping the original one.

We could instead print the existing ID columns for whoever runs the scripts to check.

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -72,40 +73,49 @@ def load_gdf_heat_network_zones(local_authority: str, **kwargs) -> gpd.GeoDataFr
...
+                            base_getters.get_gdf_from_gpkg_s3_path(
+                                path=config["data"]["geodata"]["heat_network_zones"][
+                                    la
+                                ],
+                                **kwargs,
+                            ),
+                        ],
+                        ignore_index=True,
+                    )
+                    # Deal with different ID column names in different geodataframes by renaming the ID column to "ZoneID"
```
sofiapinto, line 105: You've changed the code so we're no longer renaming, you're creating a new column. I think this might be problematic though, e.g. in case in some files the ID column is actually named HNZoneID, but some other files it's not. 

I think it would be safer to rename, instead of creating a new one, while printing the existing ID columns or even throwing an exception if multiple ID columns are available.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -576,6 +576,33 @@ def reassign_gdf_near_anchor_properties(
     return tech_gdf
 
 
+def append_gdf_heat_network_zone_layer(
```
sofiapinto, line 570: thanks for moving the code into this function, looks beautiful! ✨

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
```diff
@@ -63,19 +63,20 @@ def generate_dict_heat_network_zone_uprns(
         print(f"hn_zone_gdf reprojected to target CRS: {target_crs}")
 
     # Assume first column with `ID` substring is the zone ID column
-    id_col = [col for col in hn_zone_gdf.columns if "ID" in col][0]
-    print(f"Using Heat Network Zone {id_col} column as ID")
+    if "HNZoneID" not in hn_zone_gdf.columns:
+        id_col = [col for col in hn_zone_gdf.columns if "ID" in col][0]
+        print(f"Using Heat Network Zone {id_col} column as ID")
+        hn_zone_gdf = hn_zone_gdf.rename(columns={id_col: "HNZoneID"})
```
sofiapinto: Do we need this at all? Wouldn't it always be the case HNZoneID will be the name of the ID column given the changes you've now done to load_gdf_heat_network_zones()?

crispy-wonton: Yes we shouldn't need it if we load heat network zones by the getter but I've included it here in case the `getter` isn't used first and the zones are loaded directly.

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: Hmm ok didn't realise it was called `HNZoneID` in some files. I didn't want to rename as this is a `getter` function so it should be loading the raw files. I don't want to rename the `id_col` here in case it's the wrong column, given that we can't explicitly select the ID column.

sofiapinto: I don't know if it ever is - but if was, it could cause problems.

asf_heat_pump_suitability/pipeline/transform/heat_network_zones.py
crispy-wonton: I've changed this to just state `HNZoneID` is a required column in the docstring.

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: Have added a print to show all the ID col options.

sofiapinto (commented):
I've looked at the changes you've made. Thanks for adding `_extend_gdf_hn_zone_id()`, it makes everything clear and bug free :)

I can't actually approve the PR, because I was the one creating it. So i think you need to approve it yourself?

crispy-wonton (approved):

PR #311, opened 2026-04-30
fix bug in saving rfc

### LINKED ISSUE #310: Bug in saving classification model caused by config

crispy-wonton:
Fixes #310 

# Description

Fix small bug in saving the trained classifier model because of changes to the config.


sofiapinto (approved):
Thanks for this @crispy-wonton :)

PR #323, opened 2026-05-07
migrate to new S3 bucket

### LINKED ISSUE #322: Migrate to new S3 bucket

simran-dave:
---

# Description

Changed paths in `base.yaml ` to the new s3 bucket and moved the data. Can you check that the data is being stored in the appropriate location?

Fixes #322

### DIFF
--- a/asf_heat_pump_suitability/config/base.yaml
+++ b/asf_heat_pump_suitability/config/base.yaml
@@ -1,60 +1,60 @@
 data:
   geodata:
     # OpenMap UPRN with coordinates
-    uk_osopen_uprn: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/osopenuprn_202510_csv.zip"
+    uk_osopen_uprn: "s3://asf-local-heat-planning-tool/inputs/geodata/osopenuprn_202510_csv.zip"
     # OpenMap Local Geopackage
-    gb_os_openmap_local: "s3://asf-heat-pump-suitability/source_data/opmplc_gb.gpkg"
+    gb_os_openmap_local: "s3://asf-local-heat-planning-tool/inputs/geodata/opmplc_gb.gpkg"
     # OpenMap Local layers for certain grid squares
-    grid_square_os_openmap_local: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/v202510_OSOpenMapLocal_geometries_selected/{square}/{square}_{layer}.shp"
+    grid_square_os_openmap_local: "s3://asf-local-heat-planning-tool/inputs/geodata/v202510_OSOpenMapLocal_geometries_selected/{square}/{square}_{layer}.shp"
     # OpenMap Green space layers for certain grid squares
-    grid_square_os_openmap_greenspace: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/v202510_OSOpenMapGreenspace_geometries_selected/{square}/{square}_{layer}.shp"
+    grid_square_os_openmap_greenspace: "s3://asf-local-heat-planning-tool/inputs/geodata/v202510_OSOpenMapGreenspace_geometries_selected/{square}/{square}_{layer}.shp"
     # OpenRoad Local road data for GB
-    gb_os_openroad: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/oproad_essh_gb/"
+    gb_os_openroad: "s3://asf-local-heat-planning-tool/inputs/geodata/oproad_essh_gb/"
     # OpenRoad Local road data for certain grid squares
-    grid_square_os_openroad: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/oproad_essh_gb/{square}_RoadLink.shp"
+    grid_square_os_openroad: "s3://asf-local-heat-planning-tool/inputs/geodata/oproad_essh_gb/{square}_RoadLink.shp"
     # UK Points of Interest
-    UK_poi_locations: "s3://asf-heat-pump-suitability/source_data/poi_uk.gpkg"
+    UK_poi_locations: "s3://asf-local-heat-planning-tool/inputs/geodata/poi_uk.gpkg"
     # OpenMap Code Point GB data for postcode lookup
     # TODO: move this file to a different S3 location as it's not just Plymouth data - it's GB data
-    gb_code_points: "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/codepo_gb.gpkg"
+    gb_code_points: "s3://asf-local-heat-planning-tool/inputs/geodata/codepo_gb.gpkg"
     # GB coastline boundaries
-    gb_coast_boundaries: "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/Countries_December_2024_Boundaries_UK_BFC_6983126662299524946/CTRY_DEC_2024_UK_BFC.shp"
+    gb_coast_boundaries: "s3://asf-local-heat-planning-tool/inputs/geodata/Countries_December_2024_Boundaries_UK_BFC_6983126662299524946/CTRY_DEC_2024_UK_BFC.shp"
     # Path to folder with UPRN to country mapping files for GB
-    gb_uprn_country_mapping: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/NSUL_DEC_2025/"
+    gb_uprn_country_mapping: "s3://asf-local-heat-planning-tool/inputs/geodata/NSUL_DEC_2025/"
 
     boundaries:
       # UK LA boundaries
-      UK_ons_lad_bounds: "s3://asf-heat-pump-suitability/source_data/Local_Authority_Districts_December_2023_Boundaries_UK_BFE_-2600600853110041429/LAD_DEC_2023_UK_BFE.shp"
+      UK_ons_lad_bounds: "s3://asf-local-heat-planning-tool/inputs/geodata/Local_Authority_Districts_December_2023_Boundaries_UK_BFE_-2600600853110041429/LAD_DEC_2023_UK_BFE.shp"
 
     heat_network_zones:
       # Plymouth Heat Network zones
-      plymouth: "s3://asf-heat-pump-suitability/heat_network_desnz_data/heat-network-zone-map-Plymouth.gpkg"
-      greater_manchester_las: "s3://asf-heat-pump-suitability/heat_network_desnz_data/heat-network-zone-map-Greater-Manchester.gpkg"
+      plymouth: "s3://asf-local-heat-planning-tool/inputs/geodata/heat_network_desnz_data/heat-network-zone-map-Plymouth.gpkg"
+      greater_manchester_las: "s3://asf-local-heat-planning-tool/inputs/geodata/heat_network_desnz_data/heat-network-zone-map-Greater-Manchester.gpkg"
 
     gb_spatial_signatures:
       # UrbanGrammarAI Spatial Signatures GB simplified
-      simplified: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/v202301_UrbanGrammarAI_SpatialSignatures_GB/spatial_signatures_GB_simplified.gpkg"
+      simplified: "s3://asf-local-heat-planning-tool/inputs/geodata/v202301_UrbanGrammarAI_SpatialSignatures_GB/spatial_signatures_GB_simplified.gpkg"
       # UrbanGrammarAI Spatial Signatures GB detailed
-      full: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/v202301_UrbanGrammarAI_SpatialSignatures_GB/spatial_signatures_GB.gpkg"
+      full: "s3://asf-local-heat-planning-tool/inputs/geodata/v202301_UrbanGrammarAI_SpatialSignatures_GB/spatial_signatures_GB.gpkg"
 
     council_tax_data:
       # plymouth council tax data
-      plymouth: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/PLYMOUTH_CTBANDS_ONSUD_202512.csv"
+      plymouth: "s3://asf-local-heat-planning-tool/inputs/geodata/PLYMOUTH_CTBANDS_ONSUD_202512.csv"
 
     # Names of inspire file bounds for England, Wales and Scotland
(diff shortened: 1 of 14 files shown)


asf_heat_pump_suitability/config/base.yaml
```diff
@@ -1,60 +1,60 @@
...
     # UK Points of Interest
-    UK_poi_locations: "s3://asf-heat-pump-suitability/source_data/poi_uk.gpkg"
+    UK_poi_locations: "s3://asf-local-heat-planning-tool/inputs/geodata/poi_uk.gpkg"
     # OpenMap Code Point GB data for postcode lookup
     # TODO: move this file to a different S3 location as it's not just Plymouth data - it's GB data
-    gb_code_points: "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/codepo_gb.gpkg"
+    gb_code_points: "s3://asf-local-heat-planning-tool/inputs/geodata/codepo_gb.gpkg"
     # GB coastline boundaries
-    gb_coast_boundaries: "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/Countries_December_2024_Boundaries_UK_BFC_6983126662299524946/CTRY_DEC_2024_UK_BFC.shp"
+    gb_coast_boundaries: "s3://asf-local-heat-planning-tool/inputs/geodata/Countries_December_2024_Boundaries_UK_BFC_6983126662299524946/CTRY_DEC_2024_UK_BFC.shp"
```
crispy-wonton: suggestion: we can put this into the `/inputs/geodata/boundaries` subdir

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -1,60 +1,60 @@
...
     # Path to folder with UPRN to country mapping files for GB
-    gb_uprn_country_mapping: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/NSUL_DEC_2025/"
+    gb_uprn_country_mapping: "s3://asf-local-heat-planning-tool/inputs/geodata/NSUL_DEC_2025/"
 
     boundaries:
       # UK LA boundaries
-      UK_ons_lad_bounds: "s3://asf-heat-pump-suitability/source_data/Local_Authority_Districts_December_2023_Boundaries_UK_BFE_-2600600853110041429/LAD_DEC_2023_UK_BFE.shp"
+      UK_ons_lad_bounds: "s3://asf-local-heat-planning-tool/inputs/geodata/Local_Authority_Districts_December_2023_Boundaries_UK_BFE_-2600600853110041429/LAD_DEC_2023_UK_BFE.shp"
```
crispy-wonton: suggestion: we can put this whole folder (`Local_Authority_Districts_December_2023_Boundaries_UK_BFE_-2600600853110041429` ) into the `/inputs/geodata/boundaries` subdir. Also, in the `boundaries` subdir, there is already some loose files which i think are the same as those contained in the `Local_Authority_Districts_December_2023_Boundaries_UK_BFE_-2600600853110041429` folder. Please could you check if these are the same, then if they are and if these are referenced anywhere in the repo, and delete them / refactor them out to replace with the full folder.

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -1,60 +1,60 @@
...
-      simplified: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/v202301_UrbanGrammarAI_SpatialSignatures_GB/spatial_signatures_GB_simplified.gpkg"
+      simplified: "s3://asf-local-heat-planning-tool/inputs/geodata/v202301_UrbanGrammarAI_SpatialSignatures_GB/spatial_signatures_GB_simplified.gpkg"
       # UrbanGrammarAI Spatial Signatures GB detailed
-      full: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/v202301_UrbanGrammarAI_SpatialSignatures_GB/spatial_signatures_GB.gpkg"
+      full: "s3://asf-local-heat-planning-tool/inputs/geodata/v202301_UrbanGrammarAI_SpatialSignatures_GB/spatial_signatures_GB.gpkg"
 
     council_tax_data:
       # plymouth council tax data
-      plymouth: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/PLYMOUTH_CTBANDS_ONSUD_202512.csv"
+      plymouth: "s3://asf-local-heat-planning-tool/inputs/geodata/PLYMOUTH_CTBANDS_ONSUD_202512.csv"
```
crispy-wonton: suggestion: we could put this in a subdir called `/inputs/geodata/council_tax`. If we add more of these files, then I think it would be better to have them in a separate folder.

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -1,60 +1,60 @@
...
 
     # Names of inspire file bounds for England, Wales and Scotland
-    inspire_file_names: "s3://asf-heat-pump-suitability/outputs/2023Q4/gardens/inspire_file_bounds_EWS.geojson"
+    inspire_file_names: "s3://asf-local-heat-planning-tool/inputs/inspire/inspire_file_bounds_EWS.geojson"
```
crispy-wonton: suggestion: I think this file (and the `_EW` one should be located in the `/inputs/processed/` folder because it was created by me. Related, perhaps this should live under the `processed` key in the config.

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -1,60 +1,60 @@
...
+      EW: "s3://asf-local-heat-planning-tool/inputs/epc/v092025_EPC_NonDomestic_EW.csv"
+      S: "s3://asf-local-heat-planning-tool/inputs/epc/v122025_EPC_NonDomestic_S.csv"
 
   # Preprocessed data
   processed:
-    non_domestic_poi_categories: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/processed/non_residential_poi_categories.csv"
-    plymouth_residential_uprns: "s3://asf-heat-pump-suitability/local_heat_planning/outputs/plymouth_residential_uprns.parquet"
-    poi_anchor_properties: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/processed/poi_anchor_properties_gb.geojson"
+    non_domestic_poi_categories: "s3://asf-local-heat-planning-tool/inputs/processed/non_residential_poi_categories.csv"
+    plymouth_residential_uprns: "s3://asf-local-heat-planning-tool/outputs/plymouth_residential_uprns.parquet"
```
crispy-wonton: suggestion: I think this file can be deleted from S3 and the config - please do check if it's used anywhere in the pipeline and if so, please could you refactor it out. It shouldn't be used anymore as it's a path that's no longer being updated.

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -153,19 +153,19 @@ constant:
 output:
   dataset:
     # Residential UPRNs and their point geometries
-    residential_uprns: "s3://asf-heat-pump-suitability/local_heat_planning/outputs/{local_authority}/{local_authority}_residential_uprns.parquet"
+    residential_uprns: "s3://asf-local-heat-planning-tool/outputs/{local_authority}/{local_authority}_residential_uprns.parquet"
```
crispy-wonton:
```suggestion
    residential_uprns: "s3://asf-local-heat-planning-tool/outputs/{local_authority}/{local_authority}_domestic_uprns.parquet"
```
I think we should change this to `domestic` going forward. You can update the key if you want to (and the one below) or we can do it in a separate PR if it has a lot of references.

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -153,19 +153,19 @@ constant:
 output:
   dataset:
```
crispy-wonton, line 154: suggestion: for all the output datasets below (not model) we could save them into `/outputs/data/` - what do you think? just cuz at the high level subdir, you would see `models` or `data` so it's easier to understand what we're looking at.

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -1,60 +1,60 @@
...
     # OpenMap Code Point GB data for postcode lookup
     # TODO: move this file to a different S3 location as it's not just Plymouth data - it's GB data
-    gb_code_points: "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/codepo_gb.gpkg"
+    gb_code_points: "s3://asf-local-heat-planning-tool/inputs/geodata/codepo_gb.gpkg"
     # GB coastline boundaries
-    gb_coast_boundaries: "s3://asf-heat-pump-suitability/exploration/spatial_clustering_plymouth/Countries_December_2024_Boundaries_UK_BFC_6983126662299524946/CTRY_DEC_2024_UK_BFC.shp"
+    gb_coast_boundaries: "s3://asf-local-heat-planning-tool/inputs/geodata/Countries_December_2024_Boundaries_UK_BFC_6983126662299524946/CTRY_DEC_2024_UK_BFC.shp"
     # Path to folder with UPRN to country mapping files for GB
-    gb_uprn_country_mapping: "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/NSUL_DEC_2025/"
+    gb_uprn_country_mapping: "s3://asf-local-heat-planning-tool/inputs/geodata/NSUL_DEC_2025/"
```
crispy-wonton: suggestion: we could put this into a new subdir called `inputs/lookups` as I think we will probably end up with a few of these. It shouldn't have been under `geodata` before as I don't think it contains geodata? although correct me if i'm wrong.

crispy-wonton (commented):
Thanks so much for this work @simran-dave ! This is waaay better and cleaner!

I have reviewed both the PR changes and cross-checked against the S3 bucket.

I have left a couple of suggestions to organise the data further in S3, but couldn't see any issues.

Suggestion: I can see there is a `inputs/reference` folder containing 2 files. I think one is a duplicate of the `non_residential_poi_categories.csv` in the `inputs/processed` subdir. I would say to delete the older one unless there was a reason to keep both? You can also move the manually labelled block of flats classification dataset in that subdir to `inputs/processed` as well.

Question: have you searched for the original file paths in the config with `grep` or otherwise? If not, it would be worth doing that for each file changed (sorry) and update any references to point to the config instead of a hard-coded path. I think it's possible that this occurs in some places, e.g. exploratory notebooks, and it would be worth fixing those bugs now while the new structure is fresh, instead of doing it later. E.g. I think the `/inputs/geodata/Plymouth_Land_Registry_Cadastral_Parcels.gml` file is hard-coded. Let me know if you want to split this work and I can search for a subset of them.

asf_heat_pump_suitability/config/base.yaml
simran-dave: I've changed the keys, may be worth double checking I have updated all the references correctly

asf_heat_pump_suitability/config/base.yaml
simran-dave: yeah that makes sense!

simran-dave:
I've gone through the config and looked for references to those files in the repo. I've changed where these file paths are hardcoded to point to the config instead. Let me know if I missed any out/ if there are other hardcoded file paths that are not currently in the config

asf_heat_pump_suitability/research/exploratory/block_of_flats_classification_model/build_classification_model_block_of_flats.py
```diff
@@ -58,7 +58,9 @@
     "s3://asf-heat-pump-suitability/local_heat_planning/inputs/v042025_OSOpenMapLocal_geometries_selected/SX/SX_Building.shp"
 )
 oct_buildings_plymouth_gdf = gpd.read_file(
-    "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/v202510_OSOpenMapLocal_geometries_selected/SX/SX_Building.shp"
+    config["data"]["geodata"]["grid_square_os_openmap_local"].format(
+        layer="Building", grid_squares="SX"
+    )
```
crispy-wonton: this one, I'd revert to the original hard coding, just because it makes it more explicit to use the October 2025 version of the file which is very important here. I.e. if we update the config to use a newer version of the OS OpenMap local file, then we will load the wrong dataset. I suppose this is located in the new bucket now so might need to update the original path.

asf_heat_pump_suitability/research/exploratory/sample_buildings_for_labelling.py
```diff
@@ -49,7 +49,9 @@
 
 # Load October 2025 building footprints
 oct_buildings_gdf = gpd.read_file(
-    "s3://asf-heat-pump-suitability/local_heat_planning/inputs/geodata/v202510_OSOpenMapLocal_geometries_selected/SX/SX_Building.shp"
+    config["data"]["geodata"]["grid_square_os_openmap_local"].format(
+        layer="Building", grid_squares="SX"
+    )
```
crispy-wonton: same as above, because the October 2025 version is really important here, I would revert to hard-coding in this instance so we don't lose the connection to the original file.

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: Have checked for any references to `\"residential_uprns\"` and `\"residential_uprns_with_features\"` and nothing was returned. I think you've removed them all successfully! :)

crispy-wonton (commented):
Amazing work @simran-dave ! Thanks for taking the time to do this carefully and properly.

I've just left two comments on rare instances where I think hard-coding is preferred.

I haven't checked for every reference to each file but trust your searching. If we have missed any, they will throw us errors anyway when we run the pipeline so we'll just have to bear in mind that this migration could be the cause if we get any file not found errors in the short term.

I have looked through the five main pipeline scripts for hard-coding and have found one remaining instance in `add_features.py` in line 141. We could move this to the config under `data/processed/`. There's other hard-coding in the repo but I don't think it's worth spending the time to remove it now, especially as this PR is mainly to migrate to a new bucket. We can address these other instances by:
- Removing further hard-coding in the scripts as we work on them in other PRs
- Removing dead files where there is a lot of hard-coding
- Reducing / eliminating hard-coding where appropriate going forward.

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -138,7 +138,7 @@ def parse_arguments() -> argparse.Namespace:
 
     # TODO make this robust by automatically loading same labelled data used to train model
     labelled_df = pl.read_parquet(
-        "s3://asf-heat-pump-suitability/local_heat_planning/inputs/processed/manually_labelled_block_of_flats.parquet"
+        config["data"]["processed"]["manually_labelled_block_of_flats"]
```
crispy-wonton, line 141: the file just needs to be added to the config now i think

crispy-wonton (approved):

PR #336, opened 2026-05-12
fix for Glasgow City being called Glasgow in INSPIRE data

### LINKED ISSUE #335: glasgow city INSPIRE file called glasgow

simran-dave:
---

# Description

Please include a summary of the changes.

Fixes #335 

Added a function in `outdoor_space.py` that looks for the name in the INSPIRE dataframe that best matches the LA name given. I've tested this in a notebook and works for Glasgow City. Hopefully it doesn't introduce some accidental matches though.

### DIFF
--- a/asf_heat_pump_suitability/pipeline/run/add_features.py
+++ b/asf_heat_pump_suitability/pipeline/run/add_features.py
@@ -188,7 +188,12 @@ if __name__ == "__main__":
     )
 
     # Find closest LA name in the INSPIRE data
-    list_las = outdoor_space.get_inspire_file_match(list_las)
+    list_las = [
+        outdoor_space.get_inspire_file_match(
+            la_name=la, inspire_file_names=inspire_file_names
+        )
+        for la in list_las
+    ]
 
     # Find INSPIRE file path that matches this LA name
     inspire_file_names = inspire_file_names[
--- a/asf_heat_pump_suitability/pipeline/run/add_features.py
+++ b/asf_heat_pump_suitability/pipeline/run/add_features.py
@@ -186,6 +186,11 @@ if __name__ == "__main__":
     inspire_file_names = gpd.read_file(
         config["data"]["processed"]["inspire_file_names"]
     )
+
+    # Find closest LA name in the INSPIRE data
+    list_las = outdoor_space.get_inspire_file_match(list_las)
+
+    # Find INSPIRE file path that matches this LA name
     inspire_file_names = inspire_file_names[
         (inspire_file_names["LAD23NM"].isin(list_las))
         | (inspire_file_names["registration_county"].isin(list_las))
--- a/asf_heat_pump_suitability/pipeline/transform/outdoor_space.py
+++ b/asf_heat_pump_suitability/pipeline/transform/outdoor_space.py
@@ -6,6 +6,7 @@ import geopandas as gpd
 import logging
 import pandas as pd
 import polars as pl
+import difflib
 from asf_heat_pump_suitability.getters import load_geodata
 from asf_heat_pump_suitability.utils import geo_utils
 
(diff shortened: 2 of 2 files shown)


asf_heat_pump_suitability/pipeline/transform/outdoor_space.py
```diff
@@ -228,3 +229,40 @@ def sjoin_df_uprn_to_outdoor_space(
...
     )
+
+
+def get_inspire_file_match(la_name, inspire_file_names, threshold=0.7):
+    """
+    Finds the closest INSPIRE file for a given LA name, if there is a match above a specified similarity threshold.
+
+    Args:
+        la_name (str): LA name to find an INSPIRE file match for.
+        file_list (gdf): dataframe of INSPIRE file names for England, Wales and Scotland with LAD23NM and registration_county columns.
```
crispy-wonton: issue: the name of this variable doesn't match what's used in the function

asf_heat_pump_suitability/pipeline/transform/outdoor_space.py
```diff
@@ -228,3 +229,40 @@ def sjoin_df_uprn_to_outdoor_space(
...
+        .dropna()
+        .tolist()
+    )
+
+    # make INSPIRE file names lower case for better matching, but keep original name to return
+    combined_map = {name.lower(): name for name in combined_list}
+
+    # get closest INSPIRE file match
+    match = difflib.get_close_matches(
+        la_name.lower(), combined_map.keys(), n=1, cutoff=threshold
```
crispy-wonton:
```suggestion
        la_name.lower(), combined_map.keys(), n=3, cutoff=threshold
```
suggestion: I'd probably use the default of `n=3` here in case there is a true match with a lower score. I would then suggest adding to the `print` statement below to print any other possible matches, but still returning the top match.

asf_heat_pump_suitability/pipeline/transform/outdoor_space.py
```diff
@@ -228,3 +229,40 @@ def sjoin_df_uprn_to_outdoor_space(
...
+        )
+        .dropna()
+        .tolist()
+    )
+
+    # make INSPIRE file names lower case for better matching, but keep original name to return
+    combined_map = {name.lower(): name for name in combined_list}
+
+    # get closest INSPIRE file match
+    match = difflib.get_close_matches(
```
crispy-wonton: question: what algorithm is this using? I couldn't find an obvious answer from the docs.

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -186,6 +186,11 @@ def parse_arguments() -> argparse.Namespace:
     inspire_file_names = gpd.read_file(
         config["data"]["processed"]["inspire_file_names"]
     )
+
+    # Find closest LA name in the INSPIRE data
+    list_las = outdoor_space.get_inspire_file_match(list_las)
```
crispy-wonton: issue: I believe in some cases, `list_las` will be a list of strings, rather than a single string. It doesn't look like this function / script can handle that at the moment.

crispy-wonton (changes requested):
Thanks @simran-dave ! Good use of `difflib`, never seen that before.

I have a couple of questions and I think there's an issue with handling lists of strings rather than strings.

Also, I think this is a really key function to test, so as you're writing, you can already start thinking about how you would unit test this and the edge cases you would need to test on. It would be good to document that somewhere for when we get to the testing - e.g. it could be in on Asana on the backlog, and include your ideas for details of the tests in the description.

asf_heat_pump_suitability/pipeline/transform/outdoor_space.py
simran-dave: `get_close_matches` works using difflib's `SequenceMatcher,` which is described at the top of [this doc](https://docs.python.org/3/library/difflib.html). It says the algorithm is similar to Gestalt pattern matching which gives a score of 2.0*M / T where M = number of matching characters and T = total number of characters in the two strings, plus some extra bits on top to handle whitespace/ other junk elements.

crispy-wonton: Cool, it will need further testing when we scale up to ensure it works well for other areas. Perhaps could you add a TODO note in the code to note that.

Secondly, what is the score for Glasgow vs Glasgow City?

asf_heat_pump_suitability/pipeline/transform/outdoor_space.py
```diff
@@ -256,13 +256,14 @@ def get_inspire_file_match(la_name, inspire_file_names, threshold=0.7):
...
+        la_name.lower(), combined_map.keys(), n=3, cutoff=threshold
     )
-    if not match:
+    if not matches:
         print(f"couldn't find an INSPIRE filename for {la_name}")
     else:
-        match = match[0]
+        match = matches[0]
         print(f"found an INSPIRE file match for {la_name}: {combined_map[match]}")
+        print(f"other possible matches for {la_name}: {matches[1:]}")
```
crispy-wonton: `print(f"All possible matches found for {la_name}: {matches}")`
suggestion: I'd probably just print all in case there is only one match.

asf_heat_pump_suitability/pipeline/transform/outdoor_space.py
```diff
@@ -256,13 +256,14 @@ def get_inspire_file_match(la_name, inspire_file_names, threshold=0.7):
...
-        la_name.lower(), combined_map.keys(), n=1, cutoff=threshold
+    matches = difflib.get_close_matches(
+        la_name.lower(), combined_map.keys(), n=3, cutoff=threshold
     )
-    if not match:
+    if not matches:
         print(f"couldn't find an INSPIRE filename for {la_name}")
     else:
-        match = match[0]
+        match = matches[0]
```
crispy-wonton, line 265: question: have you confirmed that the best match is always returned first in the list?

crispy-wonton (commented):
Thanks! Have left a couple of final comments :)

asf_heat_pump_suitability/pipeline/transform/outdoor_space.py
simran-dave: yep it returns them in order of the score, with the best being first

asf_heat_pump_suitability/pipeline/transform/outdoor_space.py
simran-dave: yeah agreed about the tests. The score is 0.74 for Glasgow and Glasgow City

PR #338, opened 2026-05-13
add south lanarkshire config references

### LINKED ISSUE #337: Add config references for South Lanarkshire

simran-dave:
---

# Description

I think the grid squares for South Lanarkshire are mainly NT and NS, with a little bit in NX but maybe double check.

Fixes #337

### DIFF
--- a/asf_heat_pump_suitability/config/base.yaml
+++ b/asf_heat_pump_suitability/config/base.yaml
@@ -132,6 +132,12 @@ constant:
   glasgow:
     la_names: "Glasgow City"
     grid_squares: "NS"
+  south_lanarkshire:
+    la_names: "South Lanarkshire"
+    grid_squares:
+      - "NT"
+      - "NS"
+      - "NX"
   target_crs: 27700 # British National Grid
   tech_types:
     individual: "Individual solution"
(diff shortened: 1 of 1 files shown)


crispy-wonton (commented):
Thanks @simran-dave !

I don't see any issues in the code but have a couple of questions:

1. Have you added the relevant `NX` grid square files to S3? If not, let me know if you need help to know which files to add.
2. Have you loaded the data, e.g. OS OpenMap Local Buildings layer for these squares and compared to the South Lanarkshire LA boundary to check that there's no areas missing data? It does look like South Lanarkshire has not got a lot of buildings in the south, so it might even be worth loading the actual grid square boundaries themselves and comparing against the South Lanarkshire boundary if you haven't already.

simran-dave:
Have checked and looks like the boundary file doesn't overlap with NX after all. It looks like it does when you look up south lanarkshire boundaries and compare to BNG grid squares but I guess this could be different to what's in the file

crispy-wonton:
Cool, yeh let's go with what's in the boundary file!

crispy-wonton (approved):

PR #342, opened 2026-05-13
inspire file name mismatch fix

### LINKED ISSUE #341: scotland INSPIRE filenames are not LAs

simran-dave:
---

# Description
Now using sjoin to join UPRNs to inspire file bounds. I am not sure why but using `normalize().drop_duplicates()` wasn't working so I had to split this over two lines and it works that way ?
Fixes #341

### DIFF
--- a/asf_heat_pump_suitability/pipeline/run/add_features.py
+++ b/asf_heat_pump_suitability/pipeline/run/add_features.py
@@ -182,12 +182,10 @@ if __name__ == "__main__":
     # ESTIMATE OUTDOOR SPACE
     # TODO scale beyond Plymouth. This is a temporary fix to working with multiple LAs
     print("Loading land registry data...")
-    inspire_file_names = gpd.read_file(
-        config["data"]["processed"]["inspire_file_names"]
-    )
+    inspire_file_gdf = gpd.read_file(config["data"]["processed"]["inspire_file_names"])
     inspire_file_names = uprns_gdf.sjoin(
-        inspire_file_names, how="inner", predicate="intersects"
-    )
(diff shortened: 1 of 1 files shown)


asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -185,10 +185,9 @@ def parse_arguments() -> argparse.Namespace:
     inspire_file_names = gpd.read_file(
```
```suggestion
    inspire_file_gdf = gpd.read_file(
```
to make it more readable. Would then need updating below as well.

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -185,10 +185,9 @@ def parse_arguments() -> argparse.Namespace:
...
         config["data"]["processed"]["inspire_file_names"]
     )
-    inspire_file_names = inspire_file_names[
-        (inspire_file_names["LAD23NM"].isin(list_las))
-        | (inspire_file_names["registration_county"].isin(list_las))
-    ]["inspire_file_name"].unique()
+    inspire_file_names = uprns_gdf.sjoin(
+        inspire_file_names, how="inner", predicate="intersects"
+    )
```
```suggestion
    )["inspire_file_name"].unique()
```
I think you still need this part to get the actual names right?

crispy-wonton (changes requested):
Minor change requested which I think is required to extract the actual INSPIRE file names.

Separately, if you haven't done so yet, it might be worth commenting out Lines 109-181 and everything after line 202 and running the script (without the `--save` flag) to check the output is as you expect.

crispy-wonton (approved):

PR #349, opened 2026-05-19
Change values from True/False to Yes/No for better user experience and add new percentage features

### LINKED ISSUE #348: Add new percentage features and change True/False to Yes/No
In `compute_contextual_features.py`:
- Change booleans to Yes/No for a better user experience
- Create % solar PV and % off gas grid

sofiapinto:
---

# Description

In `compute_contextual_features.py:`

- Change booleans to Yes/No rather than True/False for a better user experience
- Create % solar PV and % off gas grid
- Fixed Unknowns that were being assigned as False for the all flag features
- Fixed the 0/1 issue for the `f"within_{ANCHOR_LOAD_RADIUS}m_from_anchor_load"`. This was happening in Plymouth only, where we had missing values due to appending the HN zones as clusters at the end of cluster.py. When we save a GeoPandas df with missing data in boolean columns as a geojson, the values in these columns are cast to float or integer (0 / 1 / NaN). The fix was moving to saving tech_clusters as a geoParquet. This way only the final output that is saved as a geojson.

I also added some of thresholds to base.yaml that were previously hardcoded. I do think we need do clean the base.yaml (make sure thresholds are in the correct place, rename some of them - but I think we should do that in a separate PR).

Fixes #348 

# Instructions for Reviewer

hey @crispy-wonton ,

I re-run `cluster.py` for Plymouth with --save after these changes. I also re-run `compute_contextual_features.py` with -save for Plymouth. If you're happy with the changes I'll remove the prints and will re-run this for the remaining areas.

Can you please check changes to the code and the new outputs? I've left some prints for now. Can remove them once you've had a look.


asf_heat_pump_suitability/config/base.yaml
```diff
@@ -143,13 +143,15 @@ constant:
...
     communal: "Communal solution"
     heat_networks: "District heat network"
-  # radius in metres around anchor properties within which we are reassigning N-GSHP to communal
-  anchor_radius: 50
   tech_type_codes:
     individual: "IND"
     networked: "NHP"
     communal: "COM"
     heat_networks: "DHNZ"
+  anchor_radius: 50 # radius in metres around anchor properties within which we are reassigning N-GSHP to communal
```
sofiapinto, line 118: I just moved the tech_type_codes above this, haven't really changed the name of this one.

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -143,13 +143,15 @@ constant:
...
+  coastline:
+    distance_from_coastline_threshold_m: 1500 # distance in metres from coastline threshold
+    simplify_tolerance_m: 150 # tolerance in m (metres) used to simplify coastline geometries
```
sofiapinto, line 121: Have added these two. I think you might want to move them to a different location or rename them, but I would do that in a separate PR for the whole base.yaml.

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -143,13 +143,15 @@ constant:
...
   ID:
     building: "ID"
```
sofiapinto, line 123: For example, there are two building ID locations. Here and in lines 60 and 61.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -134,6 +136,14 @@ def extend_df_contextual_features(
...
         uprns_df.group_by("cluster_id")
         .agg(
+            # n_uprns
+            pl.col("UPRN").n_unique().alias("n_UPRNs"),
+            # n_uprns_listed_building
+            pl.col("in_listed_building").sum().alias("n_uprns_in_listed_building"),
+            # n_uprns_solar_pv
+            pl.col("has_solar_pv").sum().alias("n_uprns_solar_pv"),
+            # n_uprns_off_gas
+            pl.col("off_gas").sum().alias("n_uprns_off_gas"),
```
sofiapinto, line 133: I just moved these above, haven't really changed them.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -186,15 +206,32 @@ def extend_df_contextual_features(
...
 
+    # Switching from booleans to Yes/No/Unknown `within_{ANCHOR_LOAD_RADIUS}m_from_anchor_load`
+    clusters_df = clusters_df.with_columns(
+        pl.when(pl.col(f"within_{ANCHOR_LOAD_RADIUS}m_from_anchor_load").is_null())
+        .then(pl.lit("Unknown"))
+        .when(pl.col(f"within_{ANCHOR_LOAD_RADIUS}m_from_anchor_load"))
+        .then(pl.lit("Yes"))
+        .otherwise(pl.lit("No"))
+        .alias(f"within_{ANCHOR_LOAD_RADIUS}m_from_anchor_load")
+    )
```
crispy-wonton, line 209: issue: I realised there is an issue with the way that column is aggregated per cluster in `cluster.py`. I think the way we currently do it means that we select a random value from the cluster because we don't specify. We need to add `agg` to the dissolve function I think.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -186,15 +206,32 @@ def extend_df_contextual_features(
...
+        "n_uprns_off_gas",
+    ]
     clusters_df = clusters_df.with_columns(
         [
-            (pl.col(col) / pl.col("n_UPRNs") * 100).alias("perc_" + col)
-            for col in tenure_cols
+            (pl.col(col) / pl.col("n_UPRNs") * 100).alias(
+                "perc_" + col if col in tenure_cols else "perc_" + col.split("n_")[1]
+            )
+            for col in perc_cols
```
crispy-wonton, line 228: suggestion: for solar PV, I think it would also be good to have the percentage of unknown values because this will be very high i imagine.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
crispy-wonton: I have addressed this in my commit to `cluster.py` - please review :)

crispy-wonton (commented):
hi @sofiapinto ! Thanks for this work :D it's looking good and readable

I left a comment about the issue in the `cluster.py` aggregation / dissolve with anchor load flag and then pushed a commit to fix this. Please do review that this works.

My only other comment was about the solar PV to suggest adding a percent unknown as well as percent of properties with PV. Otherwise, due to the high missingness, I think the solar PV % value could be a bit misleading.

On results: I checked the outputs and I think they look ok. I double checked that all the unknown anchor load radius properties are coming from the DESNZ heat network zones so that's all good. However, I noticed there are 32 unknown values in the other 

Lastly, I ran the `cluster.py` file again (without saving) after I pushed my commit to see how it changed the results. It does change the value for a lot of clusters compared to the previous run so it was affecting a lot of clusters (although I don't think this is a big issue because it only affects a single contextual label in the final output). However, the previous run is not conducted from `dev` but from this branch.
 
```
>>> clusters_gdf["within_50m_from_anchor_load"].value_counts()
within_50m_from_anchor_load
True     1547
False     963
Name: count, dtype: int64
>>> old_gdf["within_50m_from_anchor_load"].value_counts()
within_50m_from_anchor_load
False    2137
True      373
Name: count, dtype
```

Something else I noticed which I'm not sure of the cause is that in the re-run of `cluster.py` I got the following assertion error when checking for equality between the new clusters and the old clusters: `AssertionError: 2468 out of 2522 geometries are not equal.` I.e. most of the geometries changed since the previous run. Again, this is not comparing to `dev` so I'm not sure what the changes are here that cause this (e.g. is this branch any commits behind `dev`?) - generally we should only runs with `--save` from `dev` going forwards. I think we need to re-run from `dev` then compare the changes from this branch and see if we are still getting a lot of different geometries.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
sofiapinto: I agree :) have added

sofiapinto: I have created `n_uprns_missing_listed_building_flag`, `n_uprns_missing_solar_pv_flag` and `n_uprns_missing_off_gas_flag`. And then the percentages too for solar pv and off gas as we're using these for the filtering. Please check I haven't made any mistakes when calculating these please.

sofiapinto:
Thanks for the thorough review @crispy-wonton!

> I left a comment about the issue in the `cluster.py` aggregation / dissolve with anchor load flag and then pushed a commit to fix this. Please do review that this works.

Reviewed! Thanks for adding it, it looks good :)

> My only other comment was about the solar PV to suggest adding a percent unknown as well as percent of properties with PV. Otherwise, due to the high missingness, I think the solar PV % value could be a bit misleading.

Thanks for pointing this out. I think this is really important, especially when people use this in the tool (I don't think this filter is particularly useful at the moment, given the missingness for this variable, and agree it's actually quite misleading!)

I have created n_uprns_missing_listed_building_flag, n_uprns_missing_solar_pv_flag and n_uprns_missing_off_gas_flag. And then the percentages too for solar pv and off gas as we're using these for the filtering. 

> On results: I checked the outputs and I think they look ok. I double checked that all the unknown anchor load radius properties are coming from the DESNZ heat network zones so that's all good. However, I noticed there are 32 unknown values in the other

But this is expected right? We know there are Unknowns across the other variables. But for anchor load this should only be happening for the desnz heat network zones.

> Lastly, I ran the `cluster.py` file again (without saving) after I pushed my commit to see how it changed the results. It does change the value for a lot of clusters compared to the previous run so it was affecting a lot of clusters (although I don't think this is a big issue because it only affects a single contextual label in the final output). However, the previous run is not conducted from `dev` but from this branch.
> 
> ```
> >>> clusters_gdf["within_50m_from_anchor_load"].value_counts()
> within_50m_from_anchor_load
> True     1547
> False     963
> Name: count, dtype: int64
> >>> old_gdf["within_50m_from_anchor_load"].value_counts()
> within_50m_from_anchor_load
> False    2137
> True      373
> Name: count, dtype
> ```

We know it was definitely wrong before.

> Something else I noticed which I'm not sure of the cause is that in the re-run of `cluster.py` I got the following assertion error when checking for equality between the new clusters and the old clusters: `AssertionError: 2468 out of 2522 geometries are not equal.` I.e. most of the geometries changed since the previous run. Again, this is not comparing to `dev` so I'm not sure what the changes are here that cause this (e.g. is this branch any commits behind `dev`?) - generally we should only runs with `--save` from `dev` going forwards. I think we need to re-run from `dev` then compare the changes from this branch and see if we are still getting a lot of different geometries.

The one thing I wanted to flag is that I had to run cluster.py with `--save` because i changed the tech clusters from a geojson to a parquet to fix the 0/1's issue. 

I merged dev into this branch a couple of days ago. I also didn't change cluster.py, only the file type. So unsure why the geometries are not equal.

Did you compare geometries or the cluster IDs?

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -133,6 +133,8 @@ def extend_df_contextual_features(
         dummy_contextual_feat_df, how="left", on="cluster_id"
     )
 
+    print(uprns_df[["UPRN", "in_listed_building"]])
```
crispy-wonton, line 111: note: legacy print statement

crispy-wonton:
> > On results: I checked the outputs and I think they look ok. I double checked that all the unknown anchor load radius properties are coming from the DESNZ heat network zones so that's all good. However, I noticed there are 32 unknown values in the other
> 
> But this is expected right? We know there are Unknowns across the other variables. But for anchor load this should only be happening for the desnz heat network zones.

I must have edited my original comment and then accidentally cut off the remaining text. I don't think this is expected. There are 4 columns which have exactly 32 clusters with "unknown" values for the following 4 variables: 

```
in_hn_zone
in_city_centre
within_1500m_coastline
in_protected_area
```

For all these variables, I don't think it's expected for them to have unknown values. Because for all of them, they either receive `True` or `False` for the original UPRN - there shouldn't be null values for any UPRNs. I checked their values in `s3://asf-local-heat-planning-tool/outputs/data/plymouth/plymouth_with_features.parquet` (output of `add_features.py`) and there are no null values in any of these columns.

> The one thing I wanted to flag is that I had to run cluster.py with `--save` because i changed the tech clusters from a geojson to a parquet to fix the 0/1's issue.

Cool - I guess we should do those things locally until merging. Once we have the test subset set up for local runs, this will make it easier.

> Did you compare geometries or the cluster IDs?

I compared geometries. As I recall, there were the same number of clusters but you may want to double check if you haven't checked already.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -760,6 +761,58 @@ def append_gdf_heat_network_zone_layer(
...
+        uprns_df (pl.DataFrame): UPRNs with building ID column required.
+        buildings_gdf (gpd.GeoDataFrame): buildings in area of interest with building ID column.
+        clusters_gdf (gpd.GeoDataFrame): clusters with `cluster_id` column required.
+        building_id (str): name of building ID column in both `uprns_df` and `buildings_gdf`.
+
+    Returns:
+        pl.DataFrame: UPRNs mapped to their clusters. Expect some UPRNs to be duplicated if the `clusters_gdf` contains DESNZ Heat Network zones.
+    """
+    # Split clusters into DESNZ heat network zones and clusters
+    desnz_hn_zones_gdf = clusters_gdf[clusters_gdf["cluster_id"].str.contains("DESNZ")]
```
sofiapinto, line 783: Something to action later: we can move the append of the DESNZ heat network zones as the last thing we do in compute_contextual_features.py to avoid this

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -844,14 +896,9 @@ def parse_arguments() -> argparse.Namespace:
...
     if args.save:
-        # Simplify geometry for file size using tolerance
-        clusters_gdf["geometry"] = clusters_gdf["geometry"].simplify(
-            tolerance=tolerance_m
-        )
         save_utils.save_to_s3(
             clusters_gdf,
             config["output"]["dataset"]["tech_clusters"].format(
                 local_authorities=local_authority_dict["url_slug"],
-                tolerance_m=tolerance_m,
```
sofiapinto, line 849: I'm guessing we need to change the name of the tech_clusters file now that we only use tolerance later?

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -760,6 +761,58 @@ def append_gdf_heat_network_zone_layer(
...
+        building_desnz_mapping = building_desnz_mapping.set_index(building_id)[
+            "cluster_id"
+        ].to_dict()
+        desnz_uprns_df = uprns_df.with_columns(
+            pl.col("ID")
+            .replace_strict(building_desnz_mapping, default=None)
+            .alias("cluster_id")
+        ).drop_nulls(subset="cluster_id")
+
+        return pl.concat([uprns_df, desnz_uprns_df])
```
sofiapinto, line 813: Again I think most of this can be avoided by moving the append of the DESNZ zones to compute_contextual_features.py

sofiapinto: Also, we should discuss having the DESNZ zones as a separate layer in the geosjon! for a next pr :)

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -135,7 +135,7 @@ output:
     # Residential buildings with their most suitable low carbon heat technology group
     buildings_most_suitable_tech: "s3://asf-local-heat-planning-tool/outputs/data/{local_authorities}/{local_authorities}_buildings_most_suitable_tech.parquet"
     # Buildings containing domestic UPRNs aggregated into groups of neighbouring clusters of the same tech type
-    tech_clusters: "s3://asf-local-heat-planning-tool/outputs/data/{local_authorities}/{local_authorities}_clustered_tech_polygons_simplified_{tolerance_m}m.parquet"
+    tech_clusters: "s3://asf-local-heat-planning-tool/outputs/data/{local_authorities}/{local_authorities}_clustered_tech_polygons_simplified.parquet"
```
sofiapinto: i think we can also remove the word simplified?

crispy-wonton (dismissed):

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -274,6 +274,11 @@ def extend_df_contextual_features(
         uprns_df=uprns_df,
     )
 
+    # TODO identify source of empty clusters and fix - temporary fix to remove empty clusters
+    clusters_with_contextual_features_df = clusters_with_contextual_features_df.filter(
+        pl.col("n_UPRNs").is_not_null()
+    )
```
sofiapinto, line 280: will these actually be null or 0?

crispy-wonton: these are nulls!

crispy-wonton (approved):

PR #351, opened 2026-05-19
Remove grid squares from config

### LINKED ISSUE #331: Remove requirement for using grid squares

simran-dave:
---

# Description

Previously, we had a list of possible local authority args you could pass to the terminal in the config, written with dashes instead of spaces if the name has multiple words (e.g. south-lanarkshire). These had a corresponding official ONS name (South Lanarkshire) and grid squares associated with them in the config. The outputs would be saved to a folder/filename with the same name as the arg (so with the dashes). You could check to see if the LA you were using was valid from the config, and it would always save outputs in a URL-safe way because you had to input the LA name like that.

I have added some functions to a new file `resolve_las` (I can move them to a different location if you think they could live somewhere else). Now, you can input a LA name (e.g. South Lanarkshire). This is case insensitive. The function `resolve_las.resolve_la_names()` then checks against the list of possible LAs to see if it is valid, and suggests similar ones if not. The function `resolve_las.make_slug()` , which I took from Seth's PR, then converts this into a URL-safe string  for saving files. The function `resolve_las.get_la_grid_squares()` then finds the corresponding grid squares for this LA, based on its boundary. These three functions are now called at the start of all the pipeline scripts. 

Fixes #331 

# Instructions for Reviewer

Try running some of the pipeline scripts. I have run `cluster.py` for plymouth and it ran successfully but I may have missed references to the config grid squares in other scripts. 

I have been finding it hard to get this to work with a list of local authorities as the input. I think all the functions should be able to take a list of LAs or grid squares as the input, but I'm not sure of the best way to pass this as an argument via the terminal. Previously, if you wanted to run for several LAs you could add these to the config under one overall name. I think ideally you should be able to do something like this --local_authorities ["south lanarkshire", "midlothian"]. 

I think also the heat network zone function could need editing as this relied on the LA / group of LA names specified in the config

Note: I have left the grid squares for `sampling_areas` in the config for now as I think these were used to train the decision tree. I've only updated the pipeline scripts in this PR.


asf_heat_pump_suitability/config/base.yaml
```diff
@@ -55,6 +57,8 @@ data:
     inspire_file_names: "s3://asf-local-heat-planning-tool/inputs/processed/inspire/inspire_file_bounds_EWS.geojson"
     # Manually labelled blocks of flats
     manually_labelled_block_of_flats: "s3://asf-local-heat-planning-tool/inputs/processed/manually_labelled_block_of_flats.parquet"
+    # Valid local authority names
+    valid_la_names: "s3://asf-local-heat-planning-tool/inputs/processed/valid_la_names.csv"
```
crispy-wonton, line 81: question: is there a script to create this somewhere? If it's not complicated to create (e.g. just take the LA names from the LA 2023 boundaries file) then you can just add an annotation here about how to create it so it's replicable if we ever needed to make it again.

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -232,14 +232,14 @@ def load_gdf_os_openmap_layer(
         gpd.GeoDataFrame: OS OpenMap Local geometries for specified layer
     """
     if not grid_squares:  # Load all of GB
-        if layer == "greenspace_site":
-            raise ValueError(
-                "Greenspace site not implemented for GB yet. Please select a grid square."
-            )
+        if layer == "GreenspaceSite":
```
crispy-wonton: issue: this needs to be formatted as `greenspace_site`. Otherwise the string formatting in the `else` block below doesn't work and will create an incorrect file path.

asf_heat_pump_suitability/getters/resolve_las.py
```diff
@@ -0,0 +1,125 @@
...
+
+def resolve_la_names(la_names: str | list[str]) -> list:
+    """
+    Check input local authority names are valid (not case sensitive), and suggest a similar LA name if not.
+
+    Args:
+        la_names: local authority names
+
+    Returns:
+        list of valid local authority names, in the format found in the local authority boundary data.
```
crispy-wonton: documentation: missing explicit args and returns types.

asf_heat_pump_suitability/getters/resolve_las.py
```diff
@@ -0,0 +1,125 @@
...
+import pandas as pd
+import polars as pl
+import difflib
+import regex as re
+from asf_heat_pump_suitability.getters import load_geodata, load_boundaries
+from asf_heat_pump_suitability.utils import geo_utils
+from asf_heat_pump_suitability import config
+
+
+def resolve_la_names(la_names: str | list[str]) -> list:
```
crispy-wonton:
```suggestion
def resolve_list_la_names(la_names: str | list[str]) -> list:
```
nitpick: per the style guide :)

asf_heat_pump_suitability/getters/resolve_las.py
```diff
@@ -0,0 +1,125 @@
...
+
+                if not matches:
+                    print(f"No match found for: '{la_name}'")
+                    resolved_list.append(None)
+                else:
+                    suggestions = [possible_la_names_map[m] for m in matches]
+                    print(
+                        f"{la_name} not found. Did you mean: {', '.join(suggestions)}?"
+                    )
+                    resolved_list.append(None)
```
crispy-wonton: suggestion: I think it will be better to raise an error here if either of these happen. I think we wouldn't want the script to run if the input LA names are not found.

asf_heat_pump_suitability/getters/resolve_las.py
```diff
@@ -0,0 +1,125 @@
...
+
+    return resolved_list
+
+
+def make_slug(name: str | list[str]) -> str:
```
crispy-wonton:
```suggestion
def make_str_slug(name: str | list[str]) -> str:
```

asf_heat_pump_suitability/getters/resolve_las.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    Convert an official ONS place name to a URL-safe hyphenated slug for saving data.
```
crispy-wonton: documentation: technically the function will just work on any string so this description and the arg can be generalised. Although good to keep the example of king's lynn.

also, args and returns statements missing types.

asf_heat_pump_suitability/getters/resolve_las.py
```diff
@@ -0,0 +1,125 @@
...
+
+    Rules applied in order:
+    1. Lowercase the name.
+    2. Strip apostrophes and other single-quote characters.
```
crispy-wonton: question: what does 'single-quote' characters mean?

asf_heat_pump_suitability/getters/resolve_las.py
```diff
@@ -0,0 +1,125 @@
...
+    3. Replace any run of non-alphanumeric characters with a single hyphen.
+    4. Strip leading/trailing hyphens.
+
+    Args:
+        name: Official ONS place name (e.g. "King's Lynn and West Norfolk") or a list of names (e.g. ["Glasgow", "Midlothian"]).
+
+    Returns:
+        URL-safe slug (e.g. "kings-lynn-and-west-norfolk" or "glasgow-midlothian")
```
crispy-wonton: question: is the expected behaviour to make a slug from a list? e.g. "glasgow-midlothian"

asf_heat_pump_suitability/getters/resolve_las.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    # Convert a single string to a 1-element list so the rest of the logic is identical
+    names = [name] if isinstance(name, str) else name
+
+    slug_parts = []
+    for item in names:
+        item = str(item).lower()
+        item = re.sub(r"['\u2018\u2019\u02bc]", "", item)
```
crispy-wonton: question: what is the purpose of this regex?

asf_heat_pump_suitability/getters/resolve_las.py
```diff
@@ -0,0 +1,125 @@
...
+        item = re.sub(r"[^a-z0-9]+", "-", item)
+        cleaned = item.strip("-")
+        if cleaned:  # Avoid adding empty strings to the final slug
+            slug_parts.append(cleaned)
+
+    return "-".join(slug_parts)
+
+
+def get_la_grid_squares(local_authorities: list = None, buffer_m: float = 1000) -> list:
```
crispy-wonton:
```suggestion
def get_list_la_grid_squares(local_authorities: list = None, buffer_m: float = 1000) -> list:
```

asf_heat_pump_suitability/getters/resolve_las.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    Return grid squares corresponding to a local authority or list of local authorities. Defaults None to return whole of GB.
```
crispy-wonton: documentation: for clarity, I would move the information about the default to the `local_authorities` argument section below. Can you add here or below what the buffer distance is for.

asf_heat_pump_suitability/getters/resolve_las.py
```diff
@@ -0,0 +1,125 @@
...
+
+    Args:
+        local_authorities: Official ONS place name (e.g. "King's Lynn and West Norfolk") or a list of names (e.g. ["Glasgow", "Midlothian"]).
+        buffer_m: buffer distance around local authority boundary (default = 1000m).
+
+    Returns:
+        list of OS BNG grid squares corresponding to the input local authorities.
```
crispy-wonton: documentation: missing args and return types

asf_heat_pump_suitability/getters/resolve_las.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+
+    # For whole of GB, return None
+    if local_authorities is None:
+        grid_squares = None
```
crispy-wonton: issue: I'm not sure this should return none. I think it should be removed. This is because it doesn't really make sense to return `None` as a standalone function. I'm guessing this was added for compatibility in the rest of the pipeline.

asf_heat_pump_suitability/getters/resolve_las.py
```diff
@@ -0,0 +1,125 @@
...
+
+    else:
+        # Get all BNG grid squares
+        grid_gdf = load_geodata.load_gdf_bng_grid_squares()
+
+        # Total boundary of all local authorities of interest, with a buffer of buffer_m
+        boundary_gdf = load_boundaries.load_gdf_local_authority_boundaries(
+            local_authorities
+        )
+        boundary_gdf = boundary_gdf.geometry.buffer(buffer_m).union_all()
```
crispy-wonton: question: what happens here if you use local authorities which are not neighbouring? Do you get a multipolygon?

asf_heat_pump_suitability/getters/resolve_las.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # clip grid square gdf to local authority boundaries and return grid squares
+        grid_gdf = grid_gdf.clip(boundary_gdf)
+        grid_squares = list(grid_gdf["bng_ref"].drop_duplicates())
```
crispy-wonton: suggestion: you could return a set instead and then you don't need to drop duplicates
```suggestion
        grid_squares = set(grid_gdf["bng_ref"])
```

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -633,21 +633,23 @@ def parse_arguments() -> argparse.Namespace:
     args = parse_arguments()
     local_authorities = args.local_authorities
     tolerance_m = config["constant"]["clustering"]["tolerance_m"]
-    list_las = config["constant"][local_authorities]["la_names"]
+
+    list_las = resolve_las.resolve_la_names(local_authorities)
+    url_slug = resolve_las.make_slug(local_authorities)
+    grid_squares = resolve_las.get_la_grid_squares(list_las)
```
crispy-wonton: suggestion: as you're using this series of functions across multiple scripts, I'd suggest creating a single function that takes in a list[str] or str of local authority values, and outputs a dict where keys are the url_slug and values are the list/set of grid squares or something similar.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -633,21 +633,23 @@ def parse_arguments() -> argparse.Namespace:
     args = parse_arguments()
     local_authorities = args.local_authorities
     tolerance_m = config["constant"]["clustering"]["tolerance_m"]
-    list_las = config["constant"][local_authorities]["la_names"]
+
+    list_las = resolve_las.resolve_la_names(local_authorities)
```
```suggestion
    std_local_authorities = resolve_las.resolve_la_names(local_authorities)
```
suggestion: this is an issue with the original name of `list_las` but for some reason I always find it confusing to read. Could we rename to `valid_local_authorities` or something similar? Please could you implement across scripts if you do implement.

asf_heat_pump_suitability/getters/resolve_las.py
crispy-wonton, line 1: suggestion: could we move this to `/pipeline/transform/` and also rename the file to `local_authority.py`. For some reason I always find the lowercase use of `las` slightly confusing.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -673,6 +675,7 @@ def parse_arguments() -> argparse.Namespace:
     )
 
     # Add heat network zones to clusters_gdf, if they exist
+    # TODO: check this still works now LAs have been removed from config
```
crispy-wonton: note: this should still work because it is calling the `heat_network_zones` key which is separate from the LA keys which you removed.

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -160,9 +159,7 @@ def parse_arguments() -> argparse.Namespace:
     # ADD CITY CENTRE AND HEAT NETWORK ZONE BOOLEAN FLAGS
 
     # Load planned heat network zone polygons (if available)
-    hn_zones_gdf = load_geodata.load_gdf_heat_network_zones(
-        local_authority=local_authorities
-    )
+    hn_zones_gdf = load_geodata.load_gdf_heat_network_zones(local_authority=list_las)
```
crispy-wonton: issue: this function cannot handle a list input.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -633,21 +633,23 @@ def parse_arguments() -> argparse.Namespace:
     args = parse_arguments()
     local_authorities = args.local_authorities
     tolerance_m = config["constant"]["clustering"]["tolerance_m"]
-    list_las = config["constant"][local_authorities]["la_names"]
+
+    list_las = resolve_las.resolve_la_names(local_authorities)
+    url_slug = resolve_las.make_slug(local_authorities)
```
```suggestion
    url_slug = resolve_las.make_slug(list_las)
```
question: Shouldn't this function take in the `list_las` output from above? If that's the case, it needs to be updated across the pipeline scripts.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -205,23 +205,27 @@ def extend_df_contextual_features(
     from asf_heat_pump_suitability.pipeline.transform import uprns
     from asf_heat_pump_suitability import config
     from asf_heat_pump_suitability.utils import save_utils
+    from asf_heat_pump_suitability.getters import resolve_las
 
     args = parse_arguments()
     local_authorities = args.local_authorities
     tolerance_m = config["constant"]["clustering"]["tolerance_m"]
 
+    list_las = resolve_las.resolve_la_names(local_authorities)
```
crispy-wonton: note: this isn't used

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -480,7 +485,7 @@ def identify_gdf_tuple_most_suitable_tech_uprn_and_building(
     )
 
     identify_gdf_tuple_most_suitable_tech_uprn_and_building(
```
crispy-wonton, line 480: documentation: please can you update the docstring of the `local_authorities` arg for this function to say that this is used for the filename to save decision tree outputs to only.

crispy-wonton (changes requested):
Thanks for this work @simran-dave ! The code will be much cleaner to use without having to manually input the grid squares into the config every time.

I have identified some minor issues, but most comments are clarification questions or documentation requests.

Some things I suggest doing:
- Running all the scripts in the pipeline without the `--save` flag for a single LA - probably Plymouth. You can run `add_features.py` with the `--detail simplified` flag to speed it up. This will help identify if there's any errors that appear with this.
- Running the `uprns.py` for a list of LAs just to see if it can run all the way through. E.g. you could use Glasgow and South Lanarkshire.

I think you're right to leave the sampling areas in the config as they are used in the labelling.

asf_heat_pump_suitability/getters/resolve_las.py
simran-dave: single quotation marks like ' or `

asf_heat_pump_suitability/getters/resolve_las.py
simran-dave: yeah so previously, if you were running for a list of LAs you would add these to the config under one overall name (e.g. 'greater_manchester_las') and then this is what all the outputs would be saved as. But since we're not adding LA names to the config anymore we need a way of specifying what to save the files as. I guess could add functionality to specify a string to save outputs as?

asf_heat_pump_suitability/getters/resolve_las.py
simran-dave: this is where the single quotes are removed

asf_heat_pump_suitability/getters/resolve_las.py
simran-dave: yeah I added it because whenever grid squares are used as an arg, if grid squares is None it runs for whole of GB. But makes sense to remove!

asf_heat_pump_suitability/getters/resolve_las.py
simran-dave: Yep, just tested with Glasgow and Midlothian and I got a multipolygon

asf_heat_pump_suitability/pipeline/cluster/cluster.py
simran-dave: that makes sense! have added this function in

asf_heat_pump_suitability/pipeline/cluster/cluster.py
simran-dave: yep, have updated this in the new function I've added

asf_heat_pump_suitability/pipeline/run/add_features.py
simran-dave: ah ok, I've written this line to now load the hn zone for each LA in the list, and then concatenate the result. Let me know if you think it would be better to move that to the actual hn zone function.

simran-dave:
hey @crispy-wonton:

- I have run all the pipeline scripts for plymouth and they ran successfully! :) 
- I have updated the scripts so they can take in multiple LAs from the command line. E.g. you can run with `--local_authorities "midlothian" "glasgow city"` and it runs for both of these, producing one file as the output. I've gone through the pipeline scripts to check that all the functions can handle list inputs, but let me know if I have missed any. 
- I've moved the new functions I've written to `local_authority.py` and added a new function `get_dict_la_data` which uses the other functions to produce a dictionary with the url-slug, grid squares and valid LA names. Now just this one function is called at the start of each script.

asf_heat_pump_suitability/getters/resolve_las.py
crispy-wonton: please could you add a description to explain that's what's happening as it's not very readable :D

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -618,7 +619,8 @@ def parse_arguments() -> argparse.Namespace:
         "--local_authorities",
         help="Local authority or authorities. See base.yaml's `constant` section for options e.g. `plymouth`, `plymouth_similar_cities`, `sampling_areas`, `greater_manchester_las`.",
```
crispy-wonton: nitpick: just realised we need to update the docstring as I think the strings a user could enter here has changed.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -634,33 +636,35 @@ def parse_arguments() -> argparse.Namespace:
     local_authorities = args.local_authorities
     tolerance_m = config["constant"]["clustering"]["tolerance_m"]
 
-    list_las = resolve_las.resolve_la_names(local_authorities)
-    url_slug = resolve_las.make_slug(local_authorities)
-    grid_squares = resolve_las.get_la_grid_squares(list_las)
+    local_authority_dict = local_authority.get_dict_la_data(local_authorities)
```
crispy-wonton, line 639: praise: clean! love it!

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -37,6 +37,7 @@ def parse_arguments() -> argparse.Namespace:
         "--local_authorities",
         help="Local authority or authorities. See base.yaml's `constant` section for options e.g. `plymouth`, `plymouth_similar_cities`, `sampling_areas`, `greater_manchester_las`.",
```
crispy-wonton: same here as before re docstring

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -33,6 +33,7 @@ def parse_arguments() -> argparse.Namespace:
         "--local_authorities",
         help="Local authority or authorities. See base.yaml's `constant` section for options e.g. `plymouth`, `plymouth_similar_cities`, `sampling_areas`, `greater_manchester_las`.",
```
crispy-wonton: same here re docstring

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -52,6 +52,7 @@ def parse_arguments() -> argparse.Namespace:
         "--local_authorities",
         help="Local authority or authorities. See base.yaml's `constant` section for options e.g. `plymouth`, `plymouth_similar_cities`, `sampling_areas`, `greater_manchester_las`.",
```
crispy-wonton: same here re docstring

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -364,7 +365,7 @@ def identify_gdf_tuple_most_suitable_tech_uprn_and_building(
     Saves outputs to S3 if specified.
 
     Args:
-        local_authorities (str): Local authority or authorities.
+        local_authorities (str): filename to save decision tree outputs to.
```
crispy-wonton: nitpick: I'd specify that this is the local authority slug within the filename to save decision tree outputs to. can just see this causing confusion

asf_heat_pump_suitability/pipeline/transform/local_authority.py
```diff
@@ -0,0 +1,153 @@
...
+    Check input local authority names are valid (not case sensitive), and suggest a similar LA name if not.
+
+    Args:
+        la_names (str | list[str]): local authority names
+
+    Returns:
+        list: list of valid local authority names, in the format found in the local authority boundary data.
+
+    Raises:
+        ValueError: If any provided local authority name is not found in the valid list.
```
crispy-wonton, line 23: praise: love this in the docstring ty!

asf_heat_pump_suitability/pipeline/transform/local_authority.py
```diff
@@ -0,0 +1,153 @@
...
+            slug_parts.append(cleaned)
+
+    return "-".join(slug_parts)
+
+
+def get_list_la_grid_squares(
+    local_authorities: list = None, buffer_m: float = 1000
+) -> list:
+    """
+    Return grid squares corresponding to a local authority or list of local authorities. The buffer (m) ensures that buildings straddling the LA boundary are captured.
```
crispy-wonton: > The buffer (m) ensures that buildings straddling the LA boundary are captured.

nitpick: I would suggest rewording to the below. It is slightly confusing to refer to buildings here as the function isn't directly dealing with building footprints but I get what you mean.

> The buffer (m) ensures that geographical features (e.g. buildings) straddling the LA boundary are captured if they fall into neighbouring grid squares.

asf_heat_pump_suitability/pipeline/transform/local_authority.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    Takes a local authority or list of local authority names and returns a dictionary
+    with a single combined URL-slug as the key, and the resolved LAs
+    and grid squares as the values.
+
+    Args:
+        la_names (str | list[str]): List of local authority names.
+
+    Returns:
+        dict: Format -> { "url-slug": { "valid_local_authorities": [...], "grid_squares": [...] } }
```
crispy-wonton: praise: great docstring - love it! I think it slightly mismatches the format of the return statement :)

asf_heat_pump_suitability/pipeline/transform/local_authority.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    resolved_las = resolve_list_la_names(la_names)
+    # whole of GB
+    if resolved_las is None:
+        slug = "gb"
+        grid_squares_set = get_list_la_grid_squares(None)
```
crispy-wonton: suggestion: it's probably faster here to load all the grid squares directly rather than going via the function which also has to load the LA boundaries and union them.

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -328,6 +330,7 @@ def parse_arguments() -> argparse.Namespace:
         "--local_authorities",
         help="Local authority or authorities. See base.yaml's `constant` section for options e.g. `plymouth`, `plymouth_similar_cities`, `sampling_areas`, `greater_manchester_las`.",
```
crispy-wonton: same as above re docstring

asf_heat_pump_suitability/pipeline/run/add_features.py
crispy-wonton: I think it's fine to leave as is :)

crispy-wonton (commented):
hey @simran-dave thanks for making the updates and addressing the comments :)

I have left some final comments mainly on updating some now-outdated docstrings.

I am currently running the `uprns.py` script for Glasgow and South Lanarkshire together to test it out. Let me know if there's anything else I should run or any outputs you want me to review. Will post an update here when it's finished running.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -617,7 +617,7 @@ def parse_arguments() -> argparse.Namespace:
 
     parser.add_argument(
         "--local_authorities",
-        help="Local authority or authorities. See base.yaml's `constant` section for options e.g. `plymouth`, `plymouth_similar_cities`, `sampling_areas`, `greater_manchester_las`.",
+        help="Local authority or authorities (case insensitive) e.g. -- 'plymouth' to run for Plymouth or --'glasgow city' 'south lanarkshire' to run for both Glasgow City and South Lanarkshire.",
```
crispy-wonton, line 620: nice!

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -10,12 +10,6 @@
 
 python asf_heat_pump_suitability/pipeline/transform/decision_tree.py --local_authorities LOCAL_AUTHORITIES
 
-LOCAL AUTHORITIES can be one of the following, as defined in the `constant` section of base.yaml:
-- plymouth
-- plymouth_similar_cities
-- sampling_areas
-- greater_manchester_las
-
```
crispy-wonton, line 18: thanks for updating these as well!

asf_heat_pump_suitability/pipeline/transform/local_authority.py
```diff
@@ -132,13 +132,14 @@ def get_dict_la_data(la_names: str | list[str]) -> dict:
...
-        dict: Format -> { "url-slug": { "valid_local_authorities": [...], "grid_squares": [...] } }
+        dict: Format -> { "url-slug": str, "valid_local_authorities": [...], "grid_squares": [...] }
     """
     resolved_las = resolve_list_la_names(la_names)
     # whole of GB
     if resolved_las is None:
         slug = "gb"
-        grid_squares_set = get_list_la_grid_squares(None)
+        grid_squares = load_geodata.load_gdf_bng_grid_squares()
+        grid_squares_set = set(grid_squares["bng_ref"])
```
crispy-wonton, line 142: nice!

crispy-wonton (approved):
Great work @simran-dave ! Thank you! And thanks for noticing all the script docstrings to modify as well - great catch!

PR #360, opened 2026-05-21
357 fix overlapping polygons

### LINKED ISSUE #357: Fix overlapping polygons
Some polygons are overlapping others with different tech type assignments in the final clustering outputs.

crispy-wonton:
Fixes #357 

# Description

The main changes are in the `cluster.py` file. The issue causing overlapping polygons was this:

I have voronoi cells, I overlay the barrier polygons to cut them out of the cells, then dissolve cells back to their corresponding building footprints. The dissolving step was creating the overlapping polygons. This was due to microscopic bleeds of the cell fragments created by possibly rounding errors in the overlay step, causing a building to be dissolved with its own cells AND erroneously a neighbouring cell. E.g. in some cases, building A was dissolved back with cell A AND cell B because of tiny overlaps.

This has now been corrected by specifying that cell fragments are only dissolved back to the building that they have the largest overlap with.

I have checked that this fixes the problem by running `cluster.py` and using the code snippet below to test the before and after of the results.

```
import geopandas as gpd
from asf_heat_pump_suitability.utils import geo_utils
current_gdf = gpd.read_file(f"s3://asf-local-heat-planning-tool/outputs/data/{local_authorities}/{local_authorities}_clustered_tech_polygons_simplified_5m.geojson").to_crs(epsg=27700)
# Currently on dev
geo_utils.find_gdf_overlapping_geometries(current_gdf, return_geometries=False, id_col="cluster_id")
# Currently on 357_fix_overlapping_polygons
geo_utils.find_gdf_overlapping_geometries(clusters_gdf, return_geometries=False, id_col="cluster_id")
```

Additionally, I found some edge cases during debugging caused by a different problem. Some buildings did not receive a Voronoi polygon for clustering, retaining only their building geometry instead. This was due to matching building polygons to their Voronoi cells using the 'contains' predicate. Due to the way we create the Voronoi cells using interpolating points, I believe this sometimes causes small edges of a building to not be fully contained within the cell, thus they are not paired with a cell. 

I fixed that using the same logic as that required for fixing the overlapping clusters problem.

Although the buildings this affected were still included in the map and tool at the end, they were probably not getting effectively clustered, so this should fix that. It will mean that some clusters will be slightly bigger than before as there will now be more connectivity between the edge case buildings and the rest of the map.

Modified files:
- `asf_heat_pump_suitability/pipeline/cluster/cluster.py` - fix fragmented cell handling AGAIN to prevent overlapping polygons. Add warnings to indicate if there are potential issues with the clustering. These warnings will be converted to proper unit tests once the unit testing infrastructure is present.
- `asf_heat_pump_suitability/pipeline/transform/decision_tree.py` - UPRNs with no building footprint were not being handled correctly by this script which resulted in the creation of a row in the final tech type per building footprint geodataframe with no building ID and an empty geometry. This then creates downstream problems in the clustering, giving the illusion that there is an issue with the clustering. I have used the existing mapping function to map UPRNs to either the building footprint they intersect with or the nearest that they are within 3m of. This handles valid EPC UPRNs which fall outside footprints. The remaining unmatched UPRNs are dropped from the analysis. Until a new `dev` run is completed with the changes here merged, this issue will raise the new warnings added in `cluster.py`.
- `asf_heat_pump_suitability/utils/geo_utils.py` - add new function to identify geometries within the same geodataframe which are overlapping.
- `asf_heat_pump_suitability/pipeline/transform/uprns.py` - removed a redundant docstring that I noticed.

# Instructions for Reviewer

run the `cluster.py` script and then the test snippet above on the results. Run the script using:
`python asf_heat_pump_suitability/pipeline/cluster/cluster.py --local_authorities LOCAL_AUTHORITY`

I ran this on Plymouth and Midlothian already so it would be great if you could test it on a different one, e.g. Vale of Glamorgan.

- the new logic to handle cell fragments
- the new logic in `overlay_gdf_physical_barriers` in `cluster.py` intended to retain more Voronoi cells
- the changes to `decision_tree.py`. Are there any potential unintended consequences of these changes that you can see?

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/pipeline/cluster/cluster.py
+++ b/asf_heat_pump_suitability/pipeline/cluster/cluster.py
@@ -352,14 +352,36 @@ def overlay_gdf_physical_barriers(
     Returns:
         gpd.GeoDataFrame: domestic building cells with overlapping physical barriers removed
     """
-    # Filter to domestic building Voronois only
-    voronoi_gdf = voronoi_gdf.sjoin(
-        tech_gdf[["assigned_tech", "geometry"]], how="inner", predicate="contains"
-    ).drop(columns="index_right")
+    # # Filter to domestic building Voronois only
+    # Add a temporary ID for each Voronoi cell
+    cell_id_col = "_internal_cell_fragment_id"
+
+    voronoi_gdf = voronoi_gdf.assign(**{cell_id_col: np.arange(len(voronoi_gdf))})
+
+    # Get the largest intersecting Voronoi cell for each domestic building.
+    # This method means that buildings with a Voronoi cell that does not completely contain them (e.g. missing a tiny corner)
+    # still retain a Voronoi cell
+    intersection_gdf = sjoin_gdf_max_intersection(
+        container_gdf=voronoi_gdf,
+        within_gdf=tech_gdf,
+        container_id=cell_id_col,
+        within_cols=["ID", "assigned_tech", "geometry"],
+    )
+    # Map each building to its corresponding Voronoi ID
+    cell_to_building_mapping = intersection_gdf.set_index(cell_id_col)["ID"].to_dict()
+
+    # Use the mapping to label the original Voronoi cells with the correct building ID
+    voronoi_gdf["select_id"] = voronoi_gdf[cell_id_col].replace(
+        cell_to_building_mapping
+    )
+    # Filter to the rows where the building ID matches (i.e. only domestic buildings are retained here)
+    domestic_voronoi_gdf = voronoi_gdf[
+        voronoi_gdf["ID"] == voronoi_gdf["select_id"]
+    ].drop(columns="select_id")
 
     # Remove areas covered by polygons and lines
     cells_gdf = (
-        voronoi_gdf.overlay(polygon_overlay_gdf, how="difference")
+        domestic_voronoi_gdf.overlay(polygon_overlay_gdf, how="difference")
         .overlay(line_overlay_gdf, how="difference")
         .explode()
     )
--- a/asf_heat_pump_suitability/pipeline/cluster/cluster.py
+++ b/asf_heat_pump_suitability/pipeline/cluster/cluster.py
@@ -471,10 +471,19 @@ def sjoin_gdf_max_intersection(
     building_cols: list[str] = None,
 ) -> gpd.GeoDataFrame:
     """
-    Match polygons from two different geodataframes which have the maximum intersecting overlap area.
+    Match cells to the buildings they have the greatest intersecting area with.
 
     Args:
-        cell_gdf (gpd.GeoDataFrame): polygons
+        cell_gdf (gpd.GeoDataFrame): polygons of cells which contain buildings
+        building_gdf (gpd.GeoDataFrame): polygons of building footprints
+        cell_id (str): name of column containing unique cell ID
+        cell_cols (list[str]): list of columns to retain from `cell_gdf`. Default `None` to retain only columns required for
+        the operation (`cell_id`, `geometry`).
+        building_cols (list[str]): list of columns to retain from `building_gdf`. Default `None` to retain only columns required for
+        the operation (`geometry`).
+
+    Returns:
+        gpd.GeoDataFrame: cells with the building that has the greatest intersecting area with them. One row per unique cell.
     """
     # Ensure required columns are retained
     if cell_cols:
@@ -484,7 +493,7 @@ def sjoin_gdf_max_intersection(
         cell_cols = list(cell_cols)
     else:
         cell_cols = [cell_id, "geometry"]
-    if building_gdf:
+    if building_cols:
         building_cols = set(building_cols)
         building_cols.add("geometry")
         building_cols = list(building_cols)
@@ -500,6 +509,7 @@ def sjoin_gdf_max_intersection(
     intersections_gdf["max_intersection"] = intersections_gdf.groupby(cell_id)[
         "area"
     ].transform("max")
+
     return (
         intersections_gdf[
             intersections_gdf["area"] == intersections_gdf["max_intersection"]
--- a/asf_heat_pump_suitability/pipeline/cluster/cluster.py
+++ b/asf_heat_pump_suitability/pipeline/cluster/cluster.py
@@ -337,25 +337,51 @@ def _handle_gdf_fragmented_cells(
     Returns:
         gpd.GeoDataFrame: domestic building cells with overlapping physical barriers removed and cell fragments handled
     """
-    # Add a temporary ID for each building
-    id_col = "_internal_building_id"
-    tech_gdf[id_col] = np.arange(len(tech_gdf))
-    cols = [id_col, "geometry"]
+    # Add a temporary ID for each building and cell fragment
+    building_id_col = "_internal_building_id"
+    cell_id_col = "_internal_cell_fragment_id"
 
-    # Keep only cell fragments that intersect with a building and label with the ID of the building
-    cells_gdf = cells_gdf.sjoin(
-        tech_gdf[cols], how="inner", predicate="intersects"
-    ).drop(columns=["index_right"])
+    tech_gdf = tech_gdf.assign(**{building_id_col: np.arange(len(tech_gdf))})
+    cells_gdf = cells_gdf.assign(**{cell_id_col: np.arange(len(cells_gdf))})
+
+    # Keep only cell fragments that intersect with a building and label with the ID of the building.
+    # We use intersection overlay and get the intersection area between each fragment and building, retaining only the
+    # pairing with the largest intersection per cell fragment. This handles cases where one fragment joins to multiple
+    # buildings to prevent the final set of cells from containing any overlapping geometries.
+    intersections_gdf = gpd.overlay(
+        cells_gdf[[cell_id_col, "geometry"]],
+        tech_gdf[[building_id_col, "geometry"]],
+        how="intersection",
+    )
+    intersections_gdf["area"] = intersections_gdf.geometry.area
+    intersections_gdf["max_intersection"] = intersections_gdf.groupby(cell_id_col)[
+        "area"
+    ].transform("max")
+    intersections_gdf = intersections_gdf[
+        intersections_gdf["area"] == intersections_gdf["max_intersection"]
+    ].copy()
+
+    # Map building IDs to best intersecting cell fragments
+    cells_gdf = cells_gdf.merge(
+        intersections_gdf[[cell_id_col, building_id_col]],
+        on=cell_id_col,
+        how="inner",
+    )
+
+    print(cells_gdf[cell_id_col].duplicated().sum())
+    print(cells_gdf["geometry"].duplicated().sum())
+    print(tech_gdf["geometry"].duplicated().sum())
 
     # Dissolve building footprints and their cell fragments together
+    cols = [building_id_col, "geometry"]
     union_gdf = (
-        pd.concat([cells_gdf[cols], tech_gdf[cols]])
-        .dissolve(by=id_col)
+        pd.concat([cells_gdf[cols], tech_gdf[cols]], ignore_index=True)
+        .dissolve(by=building_id_col)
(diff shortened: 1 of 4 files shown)


asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -144,9 +157,49 @@ def generate_gdf_clusters(
...
 
+def sjoin_gdf_buildings_to_clusters(
+    tech_gdf: gpd.GeoDataFrame, clusters_gdf: gpd.GeoDataFrame
+):
+    """
+    Join buildings to their corresponding cluster.
+
+    Args:
+        tech_gdf (gpd.GeoDataFrame): domestic building footprints with assigned tech types.
+        buildings_gdf (gpd.GeoDataFrame): clusters of building footprints with the same assigned technology, one row per cluster.
```
sofiapinto: parameter in docstring does not match actual parameter
```suggestion
        clusters_gdf (gpd.GeoDataFrame): clusters of building footprints with the same assigned technology, one row per cluster.
```

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -144,9 +157,49 @@ def generate_gdf_clusters(
...
+            f"There is a problem with the clustering. {n_missing} buildings have not been assigned to a cluster.",
+            UserWarning,
+        )
+
     return clusters_gdf
 
 
+def sjoin_gdf_buildings_to_clusters(
+    tech_gdf: gpd.GeoDataFrame, clusters_gdf: gpd.GeoDataFrame
+):
```
sofiapinto: return type mising
```suggestion
) -> gpd.GeoDataFrame:
```

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -144,9 +157,49 @@ def generate_gdf_clusters(
...
+    """
+    Join buildings to their corresponding cluster.
```
sofiapinto: Could you please add a mention of the buffering?

asf_heat_pump_suitability/utils/geo_utils.py
```diff
@@ -92,3 +92,39 @@ def verify_gdf_crs(
         return gdf.to_crs(target_crs)
     else:
         return gdf
+
+
+def find_overlapping_geometries(
```
sofiapinto: Function name missing the return type, as you usually like to add :)

asf_heat_pump_suitability/utils/geo_utils.py
```diff
@@ -92,3 +92,39 @@ def verify_gdf_crs(
...
+    gdf: gpd.GeoDataFrame, return_geometries: bool = True, id_col=None
+) -> gpd.GeoDataFrame | set:
+    """
+    Find geometries in the same GeoDataFrame which overlap with any other geometry in the same GeoDataFrame.
```
sofiapinto: Easier to read?
```suggestion
    Find overlapping geometries in the same GeoDataFrame.
```

asf_heat_pump_suitability/utils/geo_utils.py
```diff
@@ -92,3 +92,39 @@ def verify_gdf_crs(
...
+    Returns:
+        gpd.GeoDataFrame | set: geometries or IDs of geometries which overlap with any other geometry within the `GeoDataFrame`
+    """
+    buffer = gdf.copy()
+    # Buffer added to prevent touching neighbours being identified as overlapping
+    buffer["geometry"] = buffer["geometry"].buffer(-0.0001)
+    intersecting_gdf = buffer.sjoin(buffer, predicate="intersects")
+    intersecting_gdf = intersecting_gdf.loc[
+        intersecting_gdf.index != intersecting_gdf.index_right
+    ]
```
sofiapinto, line 119: I'm guessing this is just to avoid checking a geometry against each other?

asf_heat_pump_suitability/utils/geo_utils.py
```diff
@@ -92,3 +92,39 @@ def verify_gdf_crs(
...
+    if return_geometries:
+        return intersecting_gdf
+    else:
+        if id_col:
+            overlapping_ids = set(intersecting_gdf[f"{id_col}_left"])
+        else:
+            overlapping_ids = set(intersecting_gdf.index)
+            id_col = "index"
```
sofiapinto, line 127: will this yield an error if `intersecting_gdf` has length of 0?

asf_heat_pump_suitability/pipeline/transform/uprns.py
sofiapinto, line 296: shouldn't we be setting predicate=predicate here? hence why we need the predicate above?

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -337,37 +412,110 @@ def _handle_gdf_fragmented_cells(
...
+        # Drop building geometries before merging as we already have them included in the dissolve above
+        tech_gdf.drop(columns="geometry")
+        .merge(union_gdf, how="left", on=building_id_col)
+        .set_geometry("geometry", crs=cells_gdf.crs)
+        .drop(columns=[building_id_col])
     )
 
-    # Join unionised cells back to original buildings
-    cells_gdf = tech_gdf.merge(union_gdf, how="left", on=id_col)
+    return cells_gdf[~cells_gdf["geometry"].is_empty]
```
sofiapinto, line 463: How can we have empty geometries here? And do we also need to check for whether geometries are valid at this point?

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -374,8 +351,19 @@ def identify_gdf_tuple_most_suitable_tech_uprn_and_building(
...
             - GeoDataFrame with building footprint-level most suitable tech.
     """
-    uprns_gdf = extend_gdf_building_footprints(
-        gdf=uprns_gdf, buildings_gdf=buildings_gdf, id_col=id_col
+    # Join UPRNs with buildings they intersect with or are up to 3m away.
+    # This allows valid EPC UPRNs which fall just outside building footprints to be matched to a building.
+    uprn_to_building_id = uprns.map_dict_uprns_to_building_id(
+        uprns_gdf=uprns_gdf, buildings_gdf=buildings_gdf, id_col=id_col, max_distance=3
+    )
+    uprns_gdf[id_col] = uprns_gdf["UPRN"].replace(uprn_to_building_id)
```
sofiapinto: Do we need to drop any UPRNs after this mapping?

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -374,8 +351,19 @@ def identify_gdf_tuple_most_suitable_tech_uprn_and_building(
...
+
+    # Join corresponding building geometry to each UPRN
+    buildings_gdf["building_geometry"] = buildings_gdf["geometry"]
+    uprns_gdf = uprns_gdf.merge(
+        buildings_gdf[[id_col, "building_geometry"]],
+        how="left",
+        on=id_col,
     )
```
sofiapinto, line 360: related to my comment above, what happens to UPRNs without a building match?

sofiapinto (commented):
Thanks for debugging this issue @crispy-wonton so quickly and thoroughly! I still need to run the code & do a few extra checks, but have left some questions and feedback in the meantime.

asf_heat_pump_suitability/utils/geo_utils.py
crispy-wonton: yes to avoid checking it against itself

asf_heat_pump_suitability/utils/geo_utils.py
crispy-wonton: no

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
crispy-wonton: They will not be included in the analysis. This was happening previously as well, so the changes here include more UPRNs than previously. Per @simran-dave 's analysis on this, I believe there were some EPC UPRNs which were up to 50m away from the nearest building footprint so there are some which are probably not appropriate to join to a building.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
crispy-wonton: Good question. Will need to check.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: This addresses an issue caused by retaining UPRNs with no building footprint up to this point in the pipeline. These UPRNs had no geometry and so you end up with no Voronoi cell and no building footprint, but you still retain the row as an empty geometry.

This should be redundant now because of the changes in `decision_tree.py` to drop rows with no geometry. However, since it makes the code more robust, I'd like to keep it here.

asf_heat_pump_suitability/utils/geo_utils.py
crispy-wonton: Had deliberately left it out here as there are two different return types but have added `gdf` now as it's the default.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: I think that makes sense. Are we dropping them somewhere?

crispy-wonton: Now we are dropping them in the `decision_tree.py` script with the changes in this PR.

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: I actually think this creates more room for error so I'd rather hard code it until we have a use case for allowing different predicates.

I have checked the usages throughout the repo and they use the default anyway so this should not be a breaking change.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
crispy-wonton: I ended up removing this because it was redundant. The building ID for each row is added in the `add_features.py` script so UPRNs have already been mapped to buildings in the input data.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: Sorry! I had previously misinterpreted this. Yes, I can see this is what's happening when we do the 

`.drop_nulls(subset=id_col)` above!

And this part of the code isn't actually new. You just moved it from an existing function into this function directly. Gotcha.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -144,9 +157,49 @@ def generate_gdf_clusters(
...
+    ):
+        warnings.warn(
+            "Sum of all cluster areas is greater than the area of the cluster union. "
+            "This indicates cluster polygons are overlapping.",
+            UserWarning,
+        )
+
+    # TODO move to testing when sample set available
+    joined_gdf = sjoin_gdf_buildings_to_clusters(
+        tech_gdf=tech_gdf, clusters_gdf=clusters_gdf
```
sofiapinto: This now yields an error because you changed the parameter name in the function definition, but not here.
```suggestion
        buildings_gdf=tech_gdf, clusters_gdf=clusters_gdf
```

sofiapinto (commented):
Found a little bug! Easy to fix.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: great spot thank you!

sofiapinto (approved):

PR #364, opened 2026-05-27
359 create manifest.json

### LINKED ISSUE #359: Create manifest.json

sofiapinto:
---

# Description

This PR adds code to:
- Save outputs to front end S3 bucket
- Create a `manifest.json` file in the front end S3 bucket

Fixes #359 

# Instructions for Reviewer

A few things for you to reflect on @crispy-wonton :
- I currently have a .env file with the front end info. Would you prefer to have this info stored somewhere else? e.g. in our private DS S3 bucket?
- Instead of creating `manifest.json` from scratch each time, I could just add info for new LAs? I avoided that at the moment as file names can change in the next few months. We could go with this for now and update in the future, once the scaling development has stabilised?


asf_heat_pump_suitability/pipeline/run/create_manifest.py
```diff
@@ -0,0 +1,118 @@
...
+You can then run this script with `python asf_heat_pump_suitability/pipeline/run/create_manifest.py`
+"""
+
+from dotenv import load_dotenv
+import boto3
+from typing import List
+import os
+
+
+def fetch_file_paths_from_s3_folder(
```
crispy-wonton:
```suggestion
def fetch_list_file_paths_from_s3_folder(
```

asf_heat_pump_suitability/pipeline/run/create_manifest.py
```diff
@@ -0,0 +1,118 @@
...
+    s3_client: boto3.client, s3_bucket: str, path_folder: str, file_type: str = None
+) -> List[str]:
+    """
+    Fetches file paths from a specified S3 folder.
+
+    Args:
+        s3_client (boto3.client): An initialized boto3 S3 client.
```
crispy-wonton: question: asking because I don't know: is there a reason to take the `boto3.client` as an arg rather than instantiating within the function itself?

asf_heat_pump_suitability/pipeline/run/create_manifest.py
```diff
@@ -0,0 +1,118 @@
...
+if __name__ == "__main__":
+    from asf_heat_pump_suitability.utils import save_utils
+
+    # Load environment variables from .env file
+    load_dotenv()
+
+    # Initialize S3 client
+    s3_client = boto3.client("s3")
+
+    # Load file names in s3 path
```
crispy-wonton: note: I think this annotation doesn't make sense?

asf_heat_pump_suitability/pipeline/run/create_manifest.py
```diff
@@ -0,0 +1,118 @@
...
+        path_folder=front_end_staging_s3_path,
+        file_type=".geojson",
+    )
+
+    geojson_s3_url_prefix = f"https://{front_end_s3_bucket}.s3.eu-west-2.amazonaws.com/{front_end_staging_s3_path}"
+
+    geojson_dict_list = []
+    for file_path in geojson_file_paths_list:
+        file_name = file_path.split("/")[-1]
+        local_authority = file_name.split("_clusters")[0]
```
crispy-wonton: suggestion: i'd avoid hard-coding the `_clusters` part of the filename if possible as it could change in the future.

asf_heat_pump_suitability/pipeline/run/create_manifest.py
```diff
@@ -0,0 +1,118 @@
...
+
+        # if file name does not follow expected format, raise error
+        if len(local_authority) == len(file_name):
```
crispy-wonton: question: why are you only checking for equality here? couldn't there be other indications that the `file_name` is not in the expected format or that LA has not been extracted correctly?

crispy-wonton (commented):
Hey @sofiapinto  :)

Have left a couple of small suggestions and some questions but generally I can't see any issues in the code. I still need to do a test run and set up the env variables later when bucket permissions are updated - please let me know when it's ready to test as well as the instructions for adding the environment variables.

In response to your questions:

> I currently have a .env file with the front end info. Would you prefer to have this info stored somewhere else? e.g. in our private DS S3 bucket?

I think how you have it now is the safest :)

> Instead of creating manifest.json from scratch each time, I could just add info for new LAs? I avoided that at the moment as file names can change in the next few months. We could go with this for now and update in the future, once the scaling development has stabilised?

Yes I agree, let's create from scratch (assuming it's a quick process) but then update this so that it's built into the contextual features script at some point when we expect less frequent changes in the names.

asf_heat_pump_suitability/pipeline/run/create_manifest.py
sofiapinto: Cool! I'll apply this.

One thing I wanted to ask you @crispy-wonton is where you'd like me to move this function. This is a more general util, which might become useful more widely.

asf_heat_pump_suitability/pipeline/run/create_manifest.py
sofiapinto: I believe it's better practice. If we have multiple functions that require initialising boto3 and we do that inside each function, we would be doing it multiple times with no need. Same with loading data within/outisde of functions. If we pass the data as a value to a specific param, we avoid loading it multiple times.

asf_heat_pump_suitability/pipeline/run/create_manifest.py
sofiapinto: Need to change it. I was previously loading the file names, but I changed the approach and forgot to update the annotation.

asf_heat_pump_suitability/pipeline/run/create_manifest.py
sofiapinto: I agree. Initially I was loading each file to get the local authority name from the metadata, but then thought that loading 300 files from S3 just to get the local authority name might not be ideal. Hence, used this as a  quicker way to get the local authority name. Ideally we would add the file name to `base.yaml`. But because you have the full paths on `base.yaml` rather than the file name, I decided against it.

What is you preferred option?

asf_heat_pump_suitability/pipeline/run/create_manifest.py
sofiapinto: Equality would mean that the split() above didn't work, and hence flagging that the filename might have changed.

asf_heat_pump_suitability/pipeline/run/create_manifest.py
crispy-wonton: maybe to `/utils/file_utils.py` or `/utils/s3_utils.py`?

asf_heat_pump_suitability/pipeline/run/create_manifest.py
crispy-wonton: Sorry I'm not following - why would adding the file name to `base.yaml` rather than full path change this code?

asf_heat_pump_suitability/pipeline/run/create_manifest.py
```diff
@@ -15,80 +15,49 @@
...
 
+    # Determine the suffix of the file names for local authority datasets based on the config
+    file_name_suffix = (
+        config["output"]["dataset"]["clusters_tech_contextual_info"]
+        .format(
+            local_authorities="{local_authorities}",
+            tolerance_m=config["constant"]["clustering"]["tolerance_m"],
+        )
+        .split("{local_authorities}")[-1]
+    )
```
crispy-wonton, line 54: nice!

crispy-wonton (dismissed):
Approving the code but with the caveat that I haven't run it given ongoing permissions issues. Once the issues are resolved, then I suggest Aidan tests out running the code and we merge :)

crispy-wonton (approved):
Have tested the script and it works, I can see it saves the json to the appropriate location in S3.

PR #365, opened 2026-05-28
Add data QA checks for uprns.py

### LINKED ISSUE #363: Add data QA checks to uprns.py

simran-dave:
---

# Description

I have added a new file `schemas/uprns_schema.py` which contains the data QA checks from the spreadsheet, and these are then implemented in `uprns.py`.  I have tested to see if `uprns.py` runs and passes the checks, which it does. I have also tried purposefully making it fail the checks (e.g. wrong dtype and wrong expected number of rows) and the appropriate checks fail. 

Fixes #363 

# Instructions for Reviewer

In order to test the code in this PR you need to run `uprns.py`. 

I know that you can use pandera with a polars backend which would be ideal, however when I tried this the check on the number of UPRNs in the final dataset kept failing. For some reason only 2 rows were being passed to pandera each time. The full dataframe was still being created, it's just the dataframe that was being passed to pandera that had 2 rows. When I switched to using the pandas backend (and then converting the df back to polars before saving), this problem stopped and it all worked as expected. I don't know why. 

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/schemas/uprns_schema.py
```diff
@@ -0,0 +1,69 @@
+import pandera as pa
+
+# Validate EPC UPRNs are max 12 digit positive integers
+EPC_UPRN_Schema = pa.DataFrameSchema(
+    {"UPRN": pa.Column("Int64", pa.Check.in_range(1, 999999999999), nullable=True)},
```
crispy-wonton: nitpick: `nullable` could be set to `False` here I think as the validation occurs after dropping null values.

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -115,6 +117,10 @@ def load_set_valid_epc_uprns(epc_type: str) -> set:
         .alias("UPRN")
     ).drop_nulls()  # TODO: Scotland commercial EPC data has a lot (37 %) of null UPRNs.
 
+    # Validate the EPC UPRNs
+    df_validated = uprns_schema.EPC_UPRN_Schema.validate(df.to_pandas(), lazy=True)
+    df = pl.from_pandas(df_validated)
```
crispy-wonton: docs: it would be good to add details of the error this will raise into the docstring if the validation checks fail.

thought: this polars -> pandas -> polars conversion is something we should try to fix if possible. I think it might be as simple as updating the import statement in `uprns_schema.py` to `import pandera.polars as pa` (per [Aidan's script](https://github.com/nestauk/asf_heat_pump_suitability/blob/v1.0.0/asf_heat_pump_suitability/pipeline/data_qc/pandera_eval_hps_outputs.py) I shared with you). There's also example usage with `polars` on this [Data Validation with Polars ](https://pandera.readthedocs.io/en/latest/polars.html)page in the pandera docs. Otherwise, here's some other suggestions for what to look into to help identify the cause:
- version of `pandera` you are using
- implementation of the schema or validation - are there any keyword arguments missing or is there a different way to call the polars validation?
- if stuck, you could try debugging with gemini

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -430,6 +436,44 @@ def parse_arguments() -> argparse.Namespace:
         ]
     )
 
+    # --- PANDERA VALIDATION ---
+
+    # rough expected UPRN range for whole of GB
+    if local_authorities == "gb":
+        min_uprns = 10000000
+        max_uprns = 50000000
```
crispy-wonton: issue: I'm seeing ~29 million households in GB from a quick search online. I would suggest we get the latest total count from the census and then set min_uprns to 5% below and then set max to like 25% above or something. I think we shouldn't expect our pipeline to fall so far below the true total of UPRNs - this is more problematic than retaining more non-domestic UPRNs. We should also print the expected and actual number and the % of true UPRNs that have been identified (the % would be based on counts only so we won't actually know if we are identifying the true UPRN IDs but I think it's sufficient here and we have our evaluation separately).

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -430,6 +436,44 @@ def parse_arguments() -> argparse.Namespace:
...
+
+    else:
+        # TODO: this will need updating with updated grid squares work
+        la_names_list = [config["constant"][local_authorities]["la_names"]]
+        la_names_lower = [name.lower() for name in la_names_list]
```
crispy-wonton: note: you should be able to implement the update now that the grid squares work is merged into `dev`. Also, do we still need to lower the names or is this already handled by your new functions?

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -430,6 +436,44 @@ def parse_arguments() -> argparse.Namespace:
...
+
+        # data on number of households in each LA in GB
+        census_df = pd.read_csv(config["data"]["gb_household_census_data"])
```
crispy-wonton: nitpick: could we use `polars` for consistency?

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -430,6 +436,44 @@ def parse_arguments() -> argparse.Namespace:
...
+
+        if la_census_data.empty:
+            raise ValueError(f"Could not find census data for LAs: {la_names_list}")
+
+        # sum number of households in each LA for list of input LAs
+        census_households = la_census_data["Observation"].sum()
+
+        # +/- 40% buffer on total number of households
+        min_uprns = int(census_households * 0.6)
+        max_uprns = int(census_households * 1.4)
```
crispy-wonton: suggestion: would be great to pack this into a single function

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -430,6 +436,44 @@ def parse_arguments() -> argparse.Namespace:
...
+        la_names_lower = [name.lower() for name in la_names_list]
+
+        # data on number of households in each LA in GB
+        census_df = pd.read_csv(config["data"]["gb_household_census_data"])
+        la_census_data = census_df[
+            census_df["Lower Tier Local Authorities"].str.lower().isin(la_names_lower)
+        ]
+
+        if la_census_data.empty:
+            raise ValueError(f"Could not find census data for LAs: {la_names_list}")
```
crispy-wonton: suggestion: this value error could be quite disruptive if it triggers a lot as it will not allow the `uprns.py` script to complete. I would raise a warning here for now and then use a default fall back option of min and max values of UPRNs.

We should also add a check here to ensure that the len of the la_census_data df matches the len of la_names_list and then, again, add default numbers for each missing local authority if it doesn't match.

Separately, we can map the local authorities to their LAD23CD codes so that it's consistent across files and we don't have to do string matching. Let's do this in a separate PR.

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: issue: I think the min UPRNs will be too low with this. I would set it to 95% for the same reasons discussed above. And then I would definitely print out the actual min and max and the census counts somewhere so we can see how far off it is.

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -61,6 +61,9 @@ data:
       # plymouth council tax data
       plymouth: "s3://asf-local-heat-planning-tool/inputs/geodata/council_tax/PLYMOUTH_CTBANDS_ONSUD_202512.csv"
 
+  # Number of households in each LA recorded in 2021 census
+  gb_household_census_data: "s3://asf-local-heat-planning-tool/inputs/testing/v2021_census_households_GB.csv"
```
crispy-wonton: question: Have you checked that this contains data for all GB? Generally there are 2 different census datasets. Scotland has it's own census, separate to the one for England and Wales. You may need to add another census dataset here for Scotland.

asf_heat_pump_suitability/schemas/uprns_schema.py
```diff
@@ -0,0 +1,69 @@
...
+    strict=False,
+)
+
+
+def create_domestic_uprn_schema(
+    min_expected_rows: int, max_expected_rows: int
+) -> pa.DataFrameSchema:
+    """
+    Validate all columns of final filtered domestic UPRN dataframe prior to saving.
```
crispy-wonton:
```suggestion
    Validate all columns of final filtered domestic UPRN dataframe.
```
nitpick: functions should generally be written and documented as generically as possible to make them more flexible.

asf_heat_pump_suitability/schemas/uprns_schema.py
```diff
@@ -0,0 +1,69 @@
...
+
+    Performs the following checks on the columns:
+    UPRN: integer, unique, maximum 12 digit positive number, exists for all entries
+    X_COORDINATE: float, within expected geographical bounds (BNG), exists for all entries
+    Y_COORDINATE: float, within expected geographical bounds (BNG), exists for all entries
+    LATITUDE: float, within expected geographical bounds
+    LONGITUDE: float, within expected geographical bounds
+    LAD23CD: string
+    LAD23NM: string
```
crispy-wonton:
```suggestion
    Performs the following checks on the columns:
    UPRN: is integer, unique, maximum 12 digit positive number, exists for all entries
    X_COORDINATE: is float, within expected geographical bounds (BNG), exists for all entries
    Y_COORDINATE: is float, within expected geographical bounds (BNG), exists for all entries
    LATITUDE: is float, within expected geographical bounds
    LONGITUDE: is float, within expected geographical bounds
    LAD23CD: is string
    LAD23NM: is string
```

asf_heat_pump_suitability/schemas/uprns_schema.py
```diff
@@ -0,0 +1,69 @@
...
+                float, checks=pa.Check.in_range(0.0, 700000.0), nullable=False
+            ),
+            "Y_COORDINATE": pa.Column(
+                float, checks=pa.Check.in_range(0.0, 1300000.0), nullable=False
+            ),
+            "LATITUDE": pa.Column(
+                float, checks=pa.Check.in_range(49.0, 61.0), nullable=True
+            ),
+            "LONGITUDE": pa.Column(
+                float, checks=pa.Check.in_range(-9.0, 2.0), nullable=True
```
crispy-wonton: note: not sure if it will make a material difference but I'm seeing 2.01E and 61.01N on [epsg.org](https://epsg.org/crs_27700/OSGB36-British-National-Grid.html?sessionkey=uf93d1cg4b) if you click on the link where it says 'extent'.
```suggestion
            "LATITUDE": pa.Column(
                float, checks=pa.Check.in_range(49.0, 61.01), nullable=True
            ),
            "LONGITUDE": pa.Column(
                float, checks=pa.Check.in_range(-9.0, 2.01), nullable=True
```

asf_heat_pump_suitability/schemas/uprns_schema.py
```diff
@@ -0,0 +1,69 @@
...
+                unique=True,
+                checks=pa.Check.in_range(1, 999999999999),
+                nullable=False,
+            ),
+            "X_COORDINATE": pa.Column(
+                float, checks=pa.Check.in_range(0.0, 700000.0), nullable=False
+            ),
+            "Y_COORDINATE": pa.Column(
+                float, checks=pa.Check.in_range(0.0, 1300000.0), nullable=False
+            ),
```
crispy-wonton, line 50: issue: what's the source of these numbers? I don't think the ranges are quite correct. I loaded the local authority boundaries and checked the `total_bounds` for all the LAs in the dataframe and I got the following values for the minx, miny, maxx, maxy values:

```
from asf_heat_pump_suitability.getters import load_boundaries
gdf = load_boundaries.load_gdf_local_authority_boundaries()
list(gdf.total_bounds)
[-116.19280000030994, 5333.8100000005215, 655992.75, 1220309.8807999995]
```

As you can see, minx is <0. I struggled to find a source online that gives the bounds of the BNG system. Maybe you can find one or ask Dan about it.

asf_heat_pump_suitability/schemas/uprns_schema.py
```diff
@@ -0,0 +1,69 @@
...
+            "LATITUDE": pa.Column(
+                float, checks=pa.Check.in_range(49.0, 61.0), nullable=True
+            ),
+            "LONGITUDE": pa.Column(
+                float, checks=pa.Check.in_range(-9.0, 2.0), nullable=True
+            ),
+            "LAD23CD": pa.Column(str, nullable=True),
+            "LAD23NM": pa.Column(str, nullable=True),
```
crispy-wonton: question: do we want these to be nullable? Aren't these values coming from joining the LA boundary to UPRNs so we don't expect them to be null? Or have you seen different behaviour?

asf_heat_pump_suitability/schemas/uprns_schema.py
```diff
@@ -0,0 +1,69 @@
...
+        },
+        checks=[
+            pa.Check(
+                lambda df: (min_expected_rows <= len(df) <= max_expected_rows),
+                name="uprn_census_household_comparison",
+                ignore_na=False,
+            )
```
crispy-wonton, line 69: suggestion: please can we add a custom error message here with a warning.

crispy-wonton (changes requested):
hey @simran-dave !

Great work on this! I like the way you've organised the code here and you've written it clearly and readably as usual!

I have left some thoughts and suggestions. I think some key ones are:
- I think you can update the way you import pandera to allow using pandera on polars dataframes directly.
- I believe you might need to add census data for Scotland.
- I think we should be more restrictive on the minimum number of UPRNs.
- I think the bounds for your checks on X and Y coordinates might need to be updated.

 Check my comments for more info on all of these plus some other suggestions / questions.

asf_heat_pump_suitability/schemas/uprns_schema.py
simran-dave: I think the origin should be (0, 0) based on [this article ](https://www.ordnancesurvey.co.uk/insights/the-british-national-grid-now-in-python-and-r) and this [wikipedia page](https://en.wikipedia.org/wiki/Ordnance_Survey_National_Grid). And the max I got from [here](https://ukmapcentre.com/articles/ordnance-survey-national-grid/?srsltid=AfmBOorADZdjZE5H0xnlGt5l22XmWSyXpUOQeqtFSB_1-xGBMVDWM27m) if you scroll down there is a numbered map. I will ask Dan as I assumed this meant we wouldn't get minus numbers so that's weird. I also struggled to find a good reference for this

asf_heat_pump_suitability/pipeline/transform/uprns.py
simran-dave: yeah makes sense! Have updated to check each LA in the list and see if it is in the census data. If it is, the code finds the census households and adds to a total number. If not, a default min/ max is applied based on the smallest and largest LAs in GB.

asf_heat_pump_suitability/config/base.yaml
simran-dave: Didn't realise this! found scotland census data [here](https://www.scotlandscensus.gov.uk/2022-reports/) under demography and migration and have added :)

asf_heat_pump_suitability/pipeline/transform/uprns.py
simran-dave: Ok have set range to 27 million - 40 million (~ 5% below and 40% above). The above could be too high but for some LAs I was seeing ~ 35% more UPRNs so have used this just to be safe

asf_heat_pump_suitability/schemas/uprns_schema.py
simran-dave: Yeah they shouldn't be null. I guess since we don't really use them in the pipeline (that could be wrong but I have not seen them?) I didn't want to prevent things running if they are null. But have changed to `False.`

asf_heat_pump_suitability/pipeline/transform/uprns.py
simran-dave: So I think the issue was using `len(df)` instead of `pl.len()`. When using `len(df)` pandera was passing a 2-line DataFrame as a dry-run so was always getting a DataFrame of length 2. But `pl.len` is a native polars expression. Apparently this is all to do with pandera using polars LazyFrames (which is like instructions on how to execute operations on data) instead of DataFrames. I don't fully follow what's happening with a LazyFrame so might need to look into it more.

asf_heat_pump_suitability/schemas/uprns_schema.py
crispy-wonton: Yeh, odd. Not sure what the cause is - whether the actual boundaries fall slightly outside of the BNG grid, e.g. for a small island perhaps? Because what I checked was the local authority boundaries implying that this is actually a land boundary, not just the grid square extending to a buffer zone in the sea.

You could also try converting the lat/lon coordinates you're using to BNG and see what that gives you.

asf_heat_pump_suitability/schemas/uprns_schema.py
crispy-wonton: I think if they are null then that implies they didn't get joined to a local authority and therefore would be excluded from analysis which would be incorrect.

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -321,6 +312,91 @@ def map_dict_uprns_to_building_id(
     return uprns_building_dict
 
 
+def get_census_uprn_range(local_authorities: list[str]):
```
crispy-wonton: docs: missing return type in function name and type hinting for return type

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -321,6 +312,91 @@ def map_dict_uprns_to_building_id(
...
+    """
+    Finds the number of households in a Local Authority from the 2021 census, and returns a min and max expected number of UPRNs at 0.95 and 1.4 times the sum of census counts.
```
crispy-wonton: > from the 2021 (England and Wales) and 2022 (Scotland) census

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -321,6 +312,91 @@ def map_dict_uprns_to_building_id(
...
+        local_authorities (list[str]): list of Local Authorities to find census counts for.
+
+    Returns:
+        dict: dictionary with minimum and maximum expected UPRNs based on total census counts. Format: {"min": min_uprns, "max": max_uprns}.
+
+    """
+    # rough expected UPRN range for whole of GB
+    if local_authorities == "gb":
+        min_uprns = 27000000
+        max_uprns = 40000000
```
crispy-wonton: docs: would be good to add a source here and a reference value for all of GB

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -321,6 +312,91 @@ def map_dict_uprns_to_building_id(
...
+
+    else:
+        la_lower = [la.lower() for la in local_authorities]
+
+        # data on number of households in each LA in England and Wales
+        ew_census_df = pl.read_csv(config["data"]["EW_household_census_data"])
+        # data on number of households in each LA in Scotland
+        s_census_df = pl.read_csv(
+            config["data"]["S_household_census_data"], skip_rows=3
```
crispy-wonton, line 339: docs: can we add a note of why we're skipping the first three rows

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -321,6 +312,91 @@ def map_dict_uprns_to_building_id(
...
+        # Identify missing LAs
+        ew_found = (
+            ew_matches["Lower Tier Local Authorities"].str.to_lowercase().to_list()
+        )
+        scot_found = scot_matches["Area name"].str.to_lowercase().to_list()
+        all_found = set(ew_found + scot_found)
+
+        # set default min/ max UPRN range for each missing LA
+        default_min_uprns = 850
+        default_max_uprns = 593000
```
crispy-wonton, line 371: docs: would be great to add a note saying that these are the min/ max values for all of GB already multiplied by 0.95 and 1.4

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -321,6 +312,91 @@ def map_dict_uprns_to_building_id(
...
+
+        missing_las = [la for la in la_lower if la not in all_found]
+        if missing_las:
+            warnings.warn(
+                f"Could not find census data for LAs: {missing_las}. "
+                f"Adding default range (Min: {default_min_uprns}, Max: {default_max_uprns}) for each."
+            )
```
crispy-wonton, line 378: praise: great warning!

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: praise: have left a couple of comments but overall this is a really nice function!

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -438,46 +519,25 @@ def parse_arguments() -> argparse.Namespace:
...
     # perform validation checks on df
     schema = uprns_schema.create_domestic_uprn_schema(
-        min_expected_rows=min_uprns, max_expected_rows=max_uprns
+        min_expected_rows=uprn_bounds["min"], max_expected_rows=uprn_bounds["max"]
     )
 
-    df_validated = schema.validate(df.to_pandas(), lazy=True)
-    df = pl.from_pandas(df_validated)
+    print(f"Min Expected: {uprn_bounds["min"]}")
+    print(f"Max Expected: {uprn_bounds["max"]}")
```
crispy-wonton: suggestion: maybe add to the print statements that the min and max come from 0.95/ 1.4 * the numbers found in census data.

crispy-wonton (commented):
hey @simran-dave !

very nice work, looks great well done.

Mainly some minor documentation requests but I think this looks good. The other final thing to do is confirm the appropriate BNG ranges.

Is there a particular LA you want me to test this on before merging?

asf_heat_pump_suitability/schemas/uprns_schema.py
simran-dave: The ONS LA boundary file contains all LA boundaries in the UK, including N Ireland. The BNG is not used in NI so this was incorrectly giving a minus number for min x of the boundaries. Since we do not use the NI data anywhere in the pipeline, this has been left for now. Dan has confirmed that the BNG boundaries for GB should be (0, 0) up to (700000, 1300000).

crispy-wonton (approved):
Looks good. I also ran it for Dudley and all the checks passed :)

PR #379, opened 2026-06-04
375 bugfix load hn zones

### LINKED ISSUE #375: Bugs in `load_geodata.py`
`load_gdf_heat_network_zones()` using keys previously deleted from `base.yaml`

crispy-wonton:
Fixes #375 

# Description

Update `load_geodata.load_gdf_heat_network_zones` function to load heat networks in a cleaner and more user-friendly way.

Modified files:
- asf_heat_pump_suitability/config/base.yaml - remove all references to specific heat network zone files and replace with single file
- asf_heat_pump_suitability/getters/load_geodata.py - update heat network zone getter to be cleaner. Also had to fix a bug that was caused by a generator in the function to load grid squares which meant the function could originally only be used once per session.
- asf_heat_pump_suitability/pipeline/cluster/cluster.py - updated to work with new hn zone getter. I tested this script for Dudley and the part with the HN Zones ran without errors. It errored at the known non-existent layer issue.
- asf_heat_pump_suitability/pipeline/run/add_features.py - updated to work with new hn zone getter. This also creates the file of concatenated heat network zones.
- asf_heat_pump_suitability/research/exploratory/city_centre_identification/20251205_identify_city_centre_spatial_signatures.py - updated to work with new hn zone getter.
- asf_heat_pump_suitability/research/exploratory/decision_tree_categorisation/applying_decision_tree.py - updated to work with new hn zone getter.
- asf_heat_pump_suitability/research/exploratory/decision_tree_categorisation/manchester_applying_decision_tree.py - updated to work with new hn zone getter.
- asf_heat_pump_suitability/utils/geo_utils.py - add new utils functions to map many geospatial files to their boundaries or concatenate them into one file.

# Instructions for Reviewer

- I have tested the new HN zone getter function works for the following:
- plymouth boundary polygon
- plymouth boundary gdf
- plymouth str
- South Lanarkshire str
- dudley boundary gdf

You can test it as you see fit by adapting the below code:
```
from asf_heat_pump_suitability.getters import load_boundaries, load_geodata
boundary_gdf = load_boundaries.load_gdf_local_authority_boundaries("plymouth")
hnz_gdf = load_geodata.load_gdf_heat_network_zones(local_authority="plymouth")
hnz_gdf = load_geodata.load_gdf_heat_network_zones(boundary=boundary_gdf)
```

- Please could you check that I have updated all the appropriate references to this function correctly? I searched the repository for references and think I've updated all of them. Would be good to double check this is done correctly.

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -5,13 +5,17 @@
...
 
+from tenacity import retry, stop_after_attempt
 from osbng import grids
 
 from asf_heat_pump_suitability import config
-from asf_heat_pump_suitability.getters import base_getters
+from asf_heat_pump_suitability.getters import base_getters, load_boundaries
+from asf_heat_pump_suitability.pipeline.transform import local_authority as la
+
+_BNG_GRID_100KM_FEATURES = None
```
sofiapinto, line 21: A doc above might be helpful :)

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -44,80 +48,69 @@ def load_gdf_bng_grid_squares() -> gpd.GeoDataFrame:
...
-    return gpd.GeoDataFrame.from_features(grids.bng_grid_100km, crs=27700)
+    global _BNG_GRID_100KM_FEATURES
+
+    # Materialize the iterator into a reusable list only once
+    # This is required to make multiple calls to this function in the same session
+    # Otherwise it will raise errors due to iterator exhaustion
+    if _BNG_GRID_100KM_FEATURES is None:
+        _BNG_GRID_100KM_FEATURES = list(grids.bng_grid_100km)
+
+    return gpd.GeoDataFrame.from_features(_BNG_GRID_100KM_FEATURES, crs=27700)
```
sofiapinto, line 62: wow! didn't even think about this issue.

One way to avoid using `if _BNG_GRID_100KM_FEATURES is None:` would be to use the  `functools.lru_cache` decorator (you can do `from functools import lru_cache` and then use the decorator `@lru_cache` just above a function) to handle this without managing global variables. You would cache the whole `gpd.GeoDataFrame.from_features(grids.bng_grid_100km, crs=27700)`.

However, using this would mean that you can't modify the DataFrame in-place later in the scripts without creating a copy.

Just flagging, but happy with either solution :)

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -29,29 +29,8 @@ data:
...
-      nottingham: "s3://asf-local-heat-planning-tool/inputs/geodata/heat_network_desnz_data/heat-network-zone-map-Nottingham.gpkg"
-      peterborough: "s3://asf-local-heat-planning-tool/inputs/geodata/heat_network_desnz_data/heat-network-zone-map-Peterborough.gpkg"
-      sheffield: "s3://asf-local-heat-planning-tool/inputs/geodata/heat_network_desnz_data/heat-network-zone-map-Sheffield.gpkg"
-      southampton: "s3://asf-local-heat-planning-tool/inputs/geodata/heat_network_desnz_data/heat-network-zone-map-Southampton.gpkg"
-      southwark: "s3://asf-local-heat-planning-tool/inputs/geodata/heat_network_desnz_data/heat-network-zone-map-Southwark.gpkg"
-      stoke_on_trent: "s3://asf-local-heat-planning-tool/inputs/geodata/heat_network_desnz_data/heat-network-zone-map-Stoke-on-Trent.gpkg"
-      sunderland: "s3://asf-local-heat-planning-tool/inputs/geodata/heat_network_desnz_data/heat-network-zone-map-Sunderland.gpkg"
+      desnz_files: "s3://asf-local-heat-planning-tool/inputs/geodata/heat_network_desnz_data/"
+      desnz_polygons: "s3://asf-local-heat-planning-tool/inputs/geodata/heat_network_desnz_data/heat_network_zones_all.parquet"
```
sofiapinto, line 33: nice!

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -44,80 +48,69 @@ def load_gdf_bng_grid_squares() -> gpd.GeoDataFrame:
...
+        boundary (shapely.Polygon | shapely.MultiPolygon | gpd.GeoDataFrame): Optional. Boundary to load heat network
+        zone polygons for or geodataframe of multiple boundary polygons.
+        local_authority (str): Optional. Local Authority to load heat network zone polygons for. This is slower than using
+        the boundary directly.
 
     Returns:
-        gpd.GeoDataFrame: polygons of heat network zones in given Local Authority.
+        gpd.GeoDataFrame: polygons of heat network zones in given Local Authority or boundary.
     """
+    hn_gdf = gpd.read_parquet(
```
sofiapinto, line 86: Or similar
```suggestion
    # Load gdf with DESNZ heat network zones for all areas in England
    hn_gdf = gpd.read_parquet(
```

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -44,80 +48,69 @@ def load_gdf_bng_grid_squares() -> gpd.GeoDataFrame:
...
+        path=config["data"]["geodata"]["heat_network_zones"]["desnz_polygons"]
+    ).drop(columns="index_right")
+    # Assume first column with `ID` substring is the zone ID column
+    # Note original ID column retained in case of erroneous ID assignment
+    hn_gdf = _extend_gdf_hn_zone_id(hn_gdf)
+
+    if boundary is not None:
+        if isinstance(boundary, gpd.GeoDataFrame):
```
sofiapinto, line 100: I'm guessing that, at this point, both geometries are in the same crs?

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -44,80 +48,69 @@ def load_gdf_bng_grid_squares() -> gpd.GeoDataFrame:
...
-        **kwargs for `gpd.read_file()`
+        boundary (shapely.Polygon | shapely.MultiPolygon | gpd.GeoDataFrame): Optional. Boundary to load heat network
+        zone polygons for or geodataframe of multiple boundary polygons.
+        local_authority (str): Optional. Local Authority to load heat network zone polygons for. This is slower than using
+        the boundary directly.
 
     Returns:
-        gpd.GeoDataFrame: polygons of heat network zones in given Local Authority.
+        gpd.GeoDataFrame: polygons of heat network zones in given Local Authority or boundary.
     """
```
sofiapinto, line 84: I believe we need something like this to prevent both arguments being None.


```suggestion
    """
    
    if boundary is None and local_authority is None:
        raise ValueError("At least one of 'boundary' or 'local_authority' must be provided.)
```

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -44,80 +48,69 @@ def load_gdf_bng_grid_squares() -> gpd.GeoDataFrame:
...
-
-    gdf = gpd.GeoDataFrame()
-
-    # Local authority (e.g. `plymouth`) or group of local authorities (e.g. `greater_manchester_las`)
-    local_authority = local_authority.lower()
+    else:
+        local_authority_dict = la.get_dict_la_data(local_authority)
+        boundary_gdf = load_boundaries.load_gdf_local_authority_boundaries(
+            select_las=local_authority_dict["valid_local_authorities"]
+        )
```
sofiapinto, line 116: Adding this suggestion as I find it easier to read the code
```suggestion
        valid_local_authorities_list = la.get_dict_la_data(local_authority)["valid_local_authorities"]
        boundary_gdf = load_boundaries.load_gdf_local_authority_boundaries(
            select_las=valid_local_authorities_list
        )
```

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -44,80 +48,69 @@ def load_gdf_bng_grid_squares() -> gpd.GeoDataFrame:
...
+        if isinstance(boundary, gpd.GeoDataFrame):
+            hn_gdf = hn_gdf.sjoin(
+                boundary[["geometry"]], how="inner", predicate="intersects"
+            )
+        else:
+            hn_gdf = hn_gdf[hn_gdf["geometry"].intersects(boundary)]
+        if hn_gdf.empty:
+            print("No heat network zone geodata found within given boundary.")
+        else:
+            return hn_gdf
```
sofiapinto: In some cases I believe we end up with a "index_right" column. Just flagging as you're removing it in the case below, but not here.

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -44,80 +48,69 @@ def load_gdf_bng_grid_squares() -> gpd.GeoDataFrame:
...
     Args:
-        local_authority (str): Local Authority or Local Authorities to load Heat Network zone polygons for.
-        e.g. `plymouth` for Plymouth Local Authority; `greater_manchester_las` for Greater Manchester Combined Authority (all 10 LAs in Greater Manchester).
-        See config/base.yaml under the `constant` key for options.
-
-        **kwargs for `gpd.read_file()`
+        boundary (shapely.Polygon | shapely.MultiPolygon | gpd.GeoDataFrame): Optional. Boundary to load heat network
+        zone polygons for or geodataframe of multiple boundary polygons.
+        local_authority (str): Optional. Local Authority to load heat network zone polygons for. This is slower than using
+        the boundary directly.
```
sofiapinto: Worth pointing out somewhere that if both arguments are passed, then boundary will take precedence. Or an error in case people pass both.

And maybe you could have some prints inside the if's to make that obvious for those using the function?

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -152,16 +153,21 @@ def parse_arguments() -> argparse.Namespace:
...
-    # Load planned heat network zone polygons (if available) for each LA in the list, then concatenate the gdfs
-    hn_zones_gdf_list = [
-        load_geodata.load_gdf_heat_network_zones(local_authority=la)
-        for la in local_authority_dict["valid_local_authorities"]
-    ]
-    hn_zones_gdf = pd.concat(hn_zones_gdf_list, ignore_index=True)
+    geo_utils.concat_gdfs(
+        dir_path=config["data"]["geodata"]["heat_network_zones"]["desnz_files"],
+        save_as=config["data"]["geodata"]["heat_network_zones"]["desnz_polygons"],
+    )
```
sofiapinto, line 159: I'm guessing you left it here as something to be ran each time because it doesn't actually take long? Thinking about how we run these scripts long term, might be worth adding a TODO to move this to a separate file to be ran once before the whole pipeline is ran for all LAs (once we have a shell script, or similar,  that deals with running the whole pipeline for all LAs in GB)

asf_heat_pump_suitability/utils/geo_utils.py
```diff
@@ -128,3 +132,102 @@ def find_gdf_overlapping_geometries(
...
+        dict: mapping of file names to boundary geometries
+    """
+    mapping = dict()
+    geo_files = list_geo_files(dir_path)
+    print(f"Found {len(geo_files)} files to map. Beginning mapping...")
+    for file in geo_files:
+        print(f"\nLoading: {file}")
+        gdf = gpd.read_file(f"s3://{file}").to_crs(epsg=27700)
+        gdf["geometry"] = gdf["geometry"].make_valid()
+        mapping[file] = gdf.dissolve()["geometry"].iloc[0]
```
sofiapinto, line 156:
```suggestion
        # aggregating all the geometries in the gdf into one single combined shape
        mapping[file] = gdf.dissolve()["geometry"].iloc[0]
```

asf_heat_pump_suitability/utils/geo_utils.py
```diff
@@ -128,3 +132,102 @@ def find_gdf_overlapping_geometries(
 
         print(f"ID used: {id_col}. Overlapping ID count: {len(overlapping_ids)}")
         return overlapping_ids
+
+
+def map_dict_files_to_boundaries(dir_path: str, save_as: str = None) -> dict:
```
sofiapinto, line 137: would it ever be helpful to have a `file_extensions: list = None` so that people can select the file extensions if needed and None defaulting to all possible geo file extensions (or similar)?

asf_heat_pump_suitability/utils/geo_utils.py
```diff
@@ -128,3 +132,102 @@ def find_gdf_overlapping_geometries(
...
+            {"file_path": mapping.keys(), "geometry": mapping.values()},
+            geometry="geometry",
+            crs=27700,
+        )
+        save_utils.save_to_s3(gdf, save_as)
+
+    return mapping
+
+
+def list_geo_files(dir_path: str) -> list:
```
sofiapinto, line 169: A few tweaks proposed below
```
def list_geo_files(dir_path: str, extensions:list = None) -> list:
    if extensions is None:
        extensions = ["geojson", "gpkg", "shp"]

    fs = s3fs.S3FileSystem()
    
    dir_path = dir_path.rstrip("/")
    all_files = fs.glob(f"{dir_path}/*")
    valid_extensions = {f".{ext.lower().lstrip('.')}" for ext in extensions}
    
    # filtering happens locally 
    return [f for f in all_files if any(f.lower().endswith(ext) for ext in valid_extensions)]
```

asf_heat_pump_suitability/utils/geo_utils.py
```diff
@@ -128,3 +132,102 @@ def find_gdf_overlapping_geometries(
...
+        gdfs (List[pl.DataFrame | gpd.GeoDataFrame]): list of (geo)dataframes to concatenate. Optional.
+        crs (str): CRS of final geodataframe. Default 27700 (BNG).
+        save_as (str): Optional. Save the mapping to a geospatial file type. Default None which does not save the output.
+
+    Returns:
+        gpd.GeoDataFrame: concatenated geodataframe
+    """
+    if not any([dir_path, file_paths, gdfs]):
+        raise ValueError("One of `dir_path`, `file_paths`, or `gdfs` required.")
+    if sum([arg is not None for arg in [dir_path, file_paths, gdfs]]) != 1:
```
sofiapinto, line 208: this should hopefully work & (i believe) more memory efficient
```suggestion
    if sum(arg is not None for arg in (dir_path, file_paths, gdfs)) != 1:
```

asf_heat_pump_suitability/utils/geo_utils.py
```diff
@@ -128,3 +132,102 @@ def find_gdf_overlapping_geometries(
...
+        raise ValueError(
+            "Please select only one of `dir_path`, `file_paths`, or `gdfs`."
+        )
+
+    if any([dir_path, file_paths]):
```
sofiapinto, line 213: I believe this is faster/more efficient
```suggestion
    if dir_path or file_paths:
```

asf_heat_pump_suitability/utils/geo_utils.py
```diff
@@ -128,3 +132,102 @@ def find_gdf_overlapping_geometries(
...
+        if dir_path:
+            file_paths = list_geo_files(dir_path)
+            print(f"Found {len(file_paths)} files to concatenate.")
+
+        gdfs = []
```
sofiapinto, line 218: I don't mind really, but comment here as per your usual naming conventions.
```suggestion
        gdfs_list = []
```

asf_heat_pump_suitability/utils/geo_utils.py
```diff
@@ -128,3 +132,102 @@ def find_gdf_overlapping_geometries(
...
+def concat_gdfs(
+    dir_path: Optional[str] = None,
+    file_paths: Optional[List[str]] = None,
+    gdfs: Optional[List[gpd.GeoDataFrame]] = None,
+    crs: int = 27700,
+    save_as: Optional[str] = None,
+) -> gpd.GeoDataFrame:
+    """
+    Concatenate list of (geo)dataframes from a single given source (e.g. directory, list of file paths, list of
+    (geo)dataframes) into a single one.
```
sofiapinto: In reality this is only suitable for gdfs, so I believe you don't need the geo inside the parenthesis? I believe a lot of this code would error if applied to dfs (because of the handling of crs, geometries, etc).

sofiapinto (commented):
Hello @crispy-wonton - excellent work!

Have only found one minor edge case that we need to sort that leads to duplicate lines (more details below). Have left suggestions below and I'll leave it to your discretion which of the remaining suggestions you want to take forward.

I checked that you have updated all the appropriate references to this function. Haven't found any other references!

I ran this piece of code both for plymouth and dudley (to test for LAs with and without HNZ):
```
from asf_heat_pump_suitability.getters import load_boundaries, load_geodata
boundary_gdf = load_boundaries.load_gdf_local_authority_boundaries("plymouth")
hnz_gdf = load_geodata.load_gdf_heat_network_zones(local_authority="plymouth")
hnz_gdf_2 = load_geodata.load_gdf_heat_network_zones(boundary=boundary_gdf)
```

Also ran the following which did not work exactly as expected:
```
las = ["Bolton", "Bury", "Manchester", "Oldham", "Rochdale", "Salford", "Stockport", "Tameside", "Trafford", "Wigan"]
boundary_gdf = load_boundaries.load_gdf_local_authority_boundaries(las)
hnz_gdf = load_geodata.load_gdf_heat_network_zones(local_authority=las)
hnz_gdf_2 = load_geodata.load_gdf_heat_network_zones(boundary=boundary_gdf)
```

This leads to duplicate lines in both gdfs, which I believe stem from zones that cross LA boundaries. I believe this means we need to drop duplicates before returning the hnz_gdf in `load_gdf_heat_network_zones()`.

And as expected, the only difference between the two gdfs is the `index_right`, as per one of my comments below.

In addition to testing the function above, I have also ran the `cluster.py` script (without saving) for `plymouth`, `midtlothian` and `vale of glamorgan`.

I had to add the `.replace("-", "_")` in the code below (for Vale). This is a separate issue, and I believe this was being dealt with in a separate PR - is that correct?
```
tech_gdf = (
        gpd.read_parquet(
            config["output"]["dataset"]["buildings_most_suitable_tech"].format(
                local_authorities=local_authority_dict["url_slug"].replace("-", "_")
            )
        )
        .set_geometry("geometry")
        .to_crs(config["constant"]["target_crs"])
    )
```

One minor (likely) separate thing: when I run `cluster.py` I always get two equal print messages one after the other. I was wondering if we multiple prints within the same function or, if we might be loading the boundaries twice somehow? I did a quick skim and didn't find the reason for this, so wanted to leave it here for you to check in your own time. Example below for Plymouth:
```
Loading Local Authority boundaries for ['Plymouth']...
Loading Local Authority boundaries for ['Plymouth']...
...
```

Also ran `add_features.py` (with simplified signatures and without saving) for `plymouth` and it ran as expected. One thing I noticed is that the OS OpenMap Building file layer is loaded twice in this pipeline (again nothing to do with this PR, but thought I'd flag).

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: Interesting. I read a bit about caching when trying to solve this issue but opted for using global variable as I thought it was slightly cleaner and more explicit.

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: good point!

asf_heat_pump_suitability/pipeline/run/add_features.py
crispy-wonton: Yes it is very quick to run and therefore I kept it running each time because it saves us from having to remember to run a separate script if we update the DESNZ polygons.

asf_heat_pump_suitability/utils/geo_utils.py
crispy-wonton: I considered implementing it here although so far we don't need that so I didn't implement it. We can implement it if it becomes useful.

asf_heat_pump_suitability/utils/geo_utils.py
crispy-wonton: what's the reasoning for these proposed changes?

asf_heat_pump_suitability/utils/geo_utils.py
crispy-wonton: Yes! I think I wrote the doc string early on and forgot to update!

crispy-wonton:
Thank you for checking all the different cases and finding the edge cases! Great thinking!

> ```
> las = ["Bolton", "Bury", "Manchester", "Oldham", "Rochdale", "Salford", "Stockport", "Tameside", "Trafford", "Wigan"]
> boundary_gdf = load_boundaries.load_gdf_local_authority_boundaries(las)
> hnz_gdf = load_geodata.load_gdf_heat_network_zones(local_authority=las)
> hnz_gdf_2 = load_geodata.load_gdf_heat_network_zones(boundary=boundary_gdf)
> ```
> 
> This leads to duplicate lines in both gdfs, which I believe stem from zones that cross LA boundaries. I believe this means we need to drop duplicates before returning the hnz_gdf in `load_gdf_heat_network_zones()`.

This should now be fixed.

> And as expected, the only difference between the two gdfs is the `index_right`, as per one of my comments below.

This should now be fixed.

> In addition to testing the function above, I have also ran the `cluster.py` script (without saving) for `plymouth`, `midtlothian` and `vale of glamorgan`.
> 
> I had to add the `.replace("-", "_")` in the code below (for Vale). This is a separate issue, and I believe this was being dealt with in a separate PR - is that correct?

yes it was handled in a separate PR which has been merged to dev now. It should be fixed so I will try merging changes from dev again and see if that works.

> One minor (likely) separate thing: when I run `cluster.py` I always get two equal print messages one after the other. I was wondering if we multiple prints within the same function or, if we might be loading the boundaries twice somehow? I did a quick skim and didn't find the reason for this, so wanted to leave it here for you to check in your own time. Example below for Plymouth:
> 
> ```
> Loading Local Authority boundaries for ['Plymouth']...
> Loading Local Authority boundaries for ['Plymouth']...
> ...
> ```

This is because the boundaries are loaded in L845 and L857. We need to load the boundaries within the `get_dict_la_data` function in order to load the correct grid squares.

> Also ran `add_features.py` (with simplified signatures and without saving) for `plymouth` and it ran as expected. One thing I noticed is that the OS OpenMap Building file layer is loaded twice in this pipeline (again nothing to do with this PR, but thought I'd flag).

Ok this is something we can address separately with proper testing to ensure the duplicate line can be deleted.

asf_heat_pump_suitability/utils/geo_utils.py
sofiapinto: I am not attached to it. Happy for you to disregard this suggestion.

The proposed changes:
- extensions can be defined by the user
- It calls the S3 bucket exactly once using fs.glob(f"{dir_path}/*") to list the contents. (it would only not be okay with larges amounts of files)
- It safely catches .geojson, and .GEOJSON because of .lower().

I guess a con would be that it pulls the entire list of filenames into local memory before filtering.

Your version:
- Shorter and more readable code.
- It calls fs.glob() once per extension. This is fine because it's just 3 extensions.
- wouldn't catch .GEOJSON

sofiapinto (approved):
Hey @crispy-wonton - thank you for actioning the changes so quickly!  Changes look good. **I haven't rerun the scripts though - can you do that please?** Thanks a lot!

I have answered your question about some of the [tweaks I proposed above](https://github.com/nestauk/asf_heat_pump_suitability/pull/379#discussion_r3412924290). 

Separately, here's my answer to this:

> > One minor (likely) separate thing: when I run `cluster.py` I always get two equal print messages one after the other. I was wondering if we multiple prints within the same function or, if we might be loading the boundaries twice somehow? I did a quick skim and didn't find the reason for this, so wanted to leave it here for you to check in your own time. Example below for Plymouth:
> > ```
> > Loading Local Authority boundaries for ['Plymouth']...
> > Loading Local Authority boundaries for ['Plymouth']...
> > ...
> > ```
> 
> This is because the boundaries are loaded in L845 and L857. We need to load the boundaries within the `get_dict_la_data` function in order to load the correct grid squares.

I see. Looking at `get_list_la_grid_squares()` might be worth opening an issue to add `grid_squares` and `boundary_gdf` as function parameters rather than loading data inside the function.

asf_heat_pump_suitability/utils/geo_utils.py
crispy-wonton: Great thanks for clarifying!

crispy-wonton:
@sofiapinto I've rerun `cluster.py` and `add_features.py` with no issues. Will merge :)

PR #383, opened 2026-06-09
381 add first unit tests

### LINKED ISSUE #381: Add first unit tests

crispy-wonton:
Fixes #381 

# Description

Add first unit tests to test the functionality of the `map_dict_uprns_to_building_id` function in `uprns.py`. All tests pass.

New files:
- `asf_heat_pump_suitability/pipeline/transform/tests/test_uprns.py` - first tests and their accompanying fixtures to test `map_dict_uprns_to_building_id` across different cases and edge cases.

Modified files:
- `asf_heat_pump_suitability/pipeline/transform/uprns.py` - update mapping function to reproducibly map UPRNs in cases where they fall on a shared edge.
- `requirements.txt` - add pytest to requirements

# Instructions for Reviewer

```
pip install pytest
pytest
```

All tests should pass.

For @simran-dave :
- All of the different cases I have tested. These should cover the full spectrum of interactions that our UPRN point geometries can have with the building geometries. Are there any cases missing? 
- Any behaviour of the cases that you think should be different? E.g. is there any unexpected results from the tests even if they are passing?
- Anything that could be improved with the layout /structure of the code to make it easier to follow?

For @caldwellst : 
- Anything that could be improved with the layout /structure of the code to make it easier to follow?
- Anything that could be better aligned with best practice? 
- Are there ways I could be using `pytest` better?

# Checklist:

- [x] I have tested the code


asf_heat_pump_suitability/pipeline/transform/tests/test_uprns.py
```diff
@@ -0,0 +1,266 @@
...
+            max_distance=5,
+        )
+
+        expected = {"P2": "B1"}
+        assert results == expected
+
+    def test_uprn_joined_to_single_building(
+        self, buildings_gdf, gdf_uprn_on_shared_edge
+    ):
+        """Test UPRN on shared edge maps to only one building."""
```
simran-dave, line 297: How do we decide which building the UPRN gets joined to when it is mapped to more than one?

asf_heat_pump_suitability/pipeline/transform/tests/test_uprns.py
```diff
@@ -0,0 +1,266 @@
...
+            max_distance=5,
+        )
+
+        expected = {"P4": "B2"}
+        assert results == expected
+
+    def test_uprn_no_building_match(
+        self, buildings_gdf, gdf_one_uprn_far_outside_building
+    ):
+        """Test one UPRN does not map to any building due to distance from nearest building (>5m)."""
```
simran-dave, line 253: should we test a case where the UPRN is exactly 5m away or is that overkill

asf_heat_pump_suitability/pipeline/transform/tests/test_uprns.py
```diff
@@ -0,0 +1,266 @@
...
+            buildings_gdf=buildings_gdf,
+            uprns_gdf=gdf_uprn_intersecting_boundary,
+            id_col="building_id",
+            max_distance=5,
+        )
+
+        expected = {"P2": "B1"}
+        assert results == expected
+
+    def test_uprn_joined_to_single_building(
```
simran-dave, line 294: we could also test a case where a UPRN is outside but equidistant from two buildings

simran-dave (commented):
hey @crispy-wonton ! I have run `pytest` and all tests passed. The layout of the code makes sense to me. I think all the main test cases are covered, I added some ides for additional ones but I don't think they are super necessary. 

The case where a UPRN gets mapped to two buildings is interesting- do we know how often this happens? And from my understanding, right now we just choose the building that comes first in building ID?

asf_heat_pump_suitability/pipeline/transform/tests/test_uprns.py
crispy-wonton: Currently, it's just the building with the first ID after they are sorted into order by ID. However, this is not optimal as it's arbitrary and not replicable across different iterations of the building dataset because the IDs are not updated. I didn't have a better solution at this time though.

asf_heat_pump_suitability/pipeline/transform/tests/test_uprns.py
crispy-wonton: good idea - I just used an existing fixture and reduced the max distance to the distance of the UPRN from the building (3m).

crispy-wonton:
> hey @crispy-wonton ! I have run `pytest` and all tests passed. The layout of the code makes sense to me. I think all the main test cases are covered, I added some ides for additional ones but I don't think they are super necessary.

Awesome thank you! I have implemented your ideas :) 

> The case where a UPRN gets mapped to two buildings is interesting- do we know how often this happens? 

Good question, I checked this for Plymouth and it's 0% of all UPRNs, which is surprising. I expected at least a low count but perhaps I wrote the code incorrectly, have pasted it below - do you mind please checking it?

Either way, there may still be a few edge cases where this happens somewhere in GB and there may be cases as you've flagged where a UPRN is outside the building footprints but equidistant to two buildings where this is relevant.

```

uprns_df = load_geodata.load_df_osopen_uprn()
uprns_gdf = uprns.generate_gdf_uprn_coords(uprns_df)
buildings_gdf = load_geodata.load_gdf_os_openmap_layer(layer="building", grid_squares="SX")

joined_gdf = uprns_gdf.sjoin(buildings_gdf, how="inner", predicate="intersects")
count_uprns = joined_gdf.groupby("UPRN").size().reset_index(name="count")
count_uprns[count_uprns["count"] != 1]
```

> And from my understanding, right now we just choose the building that comes first in building ID?

Correct - see full response here: https://github.com/nestauk/asf_heat_pump_suitability/pull/383#discussion_r3552517352

simran-dave (commented):
@crispy-wonton yeah kind of surprising that there are 0 cases of a UPRN being mapped to two buildings. I can't see anything wrong with the function you wrote. I also tested it using the examples here of a UPRN on the shared edge of two buildings and the function correctly identified that it was being sjoined to two buildings.

I feel like since this seems to be a very rare edge case it is ok that we are using an arbitrary mapping. I agree it's not optimal which is why I was wondering how often this happens.

simran-dave (dismissed):

simran-dave (approved):

PR #389, opened 2026-06-17
384 test clustering

### LINKED ISSUE #384: Add tests for `cluster.generate_gdf_clusters`

### LINKED ISSUE #391: Add tests for `cluster.extend_edges_gdf`

crispy-wonton:
Fix #384 

# Description

Add initial tests for `cluster.py`.

The PR was getting very big so I haven't tested all the functions in `cluster.py`. These can be added in a separate PR. However, this should cover some crucial tests. Note, the clustering with polygon and line overlays needs to be tested separately, as does the anchor property reassignment.

New files:
- asf_heat_pump_suitability/pipeline/cluster/tests/test_cluster.py - add fixtures and tests to test two functions: `generate_gdf_clusters` and `extend_edges_gdf`.
- asf_heat_pump_suitability/pipeline/cluster/tests/utils.py - some utils to assist with the creation of fixtures

Modified files:
- asf_heat_pump_suitability/config/base.yaml - add colour mapping to config for tech types. This was useful when visualising clusters/buildings as I was creating the test sets
- asf_heat_pump_suitability/pipeline/cluster/cluster.py - removed some hard-coding highlighted by the tests.
- asf_heat_pump_suitability/utils/mapping_utils.py - new mapping util for some additional map plotting. Again this was useful for visualisation in development but is not called anywhere in the pipeline / tests.

# Instructions for Reviewer

run the following code from terminal:
```
pip install pytest
pytest
```
There are 7 tests, all of which should pass.
Note there are some warnings which I have looked into, none of which are problematic.
The `keep_geom_type=True` warning is removing linestrings or points which get generated during the clustering which is an expected behaviour.

- Do the test fixtures cover enough edge cases that we want to test on?
- For the functions that have been tested, are we missing any key tests?

# Checklist:

- [x] I have tested the code


asf_heat_pump_suitability/pipeline/cluster/tests/test_cluster.py
sofiapinto, line 1: script level docstring missing :)

asf_heat_pump_suitability/utils/mapping_utils.py
sofiapinto, line 1: script level docstring missing :)

asf_heat_pump_suitability/pipeline/cluster/tests/utils.py
```diff
@@ -0,0 +1,11 @@
+def assign_bool_domestic_status(b_id: str, non_domestic: list):
```
sofiapinto: return type missing

asf_heat_pump_suitability/pipeline/cluster/tests/utils.py
```diff
@@ -0,0 +1,11 @@
...
+    """Returns False for specified non-domestic codes."""
+    # Extract prefix (e.g. 'B03') cleanly regardless of lowercase user input
+    prefix = b_id.split("_")[0].upper()
+    return prefix not in non_domestic
+
+
+def assign_str_tech_type(b_id: str, tech_mapping: dict):
```
sofiapinto: return type missing

asf_heat_pump_suitability/pipeline/cluster/tests/test_cluster.py
```diff
@@ -0,0 +1,576 @@
...
+from asf_heat_pump_suitability.pipeline.cluster.tests import utils
+
+
+@pytest.fixture(scope="module")
+def gdf_mixed_buildings():
+    """
+    Generate a geodataframe containing a selection of test building footprints to test clustering across different scenarios:
+    1. A horseshoe-shaped building that wraps around another smaller building on three sides.
+    2. A central cluster of buildings surrounded on all sides by a selection of buildings of different shapes.
+    3. Buildings which are very close to the neighbouring buildings.
```
sofiapinto: Worth adding some info about the CRS of the geometries below? I can imagine this is British National Grid / EPSG:27700.

asf_heat_pump_suitability/utils/mapping_utils.py
```diff
@@ -7,6 +7,56 @@
 from pathlib import Path
 import os
 
+from asf_heat_pump_suitability import config
+
+TECH_TYPES = config["constant"]["tech_types"]
+COLOUR_MAPPING = config["constant"]["tech_colour_mapping"]
+
+
+def plot_folium_cluster_map(
```
sofiapinto, line 20: This is very useful, thanks for adding. It's not user anywhere above though, but I believe that wasn't the idea anyway.

asf_heat_pump_suitability/utils/mapping_utils.py
```diff
@@ -7,6 +7,56 @@
...
+    )
+
+    # Create map
+    m = folium.Map(
+        location=[centre_map[0][1], centre_map[0][0]],
+        zoom_start=15,
+        tiles="esri_worldimagery",
+    )
+
+    colour_mapping = dict(zip(TECH_TYPES.values(), COLOUR_MAPPING.values()))
```
sofiapinto, line 51: I believe this only works properly if the order of the categories is the same in both.

asf_heat_pump_suitability/utils/mapping_utils.py
```diff
@@ -7,6 +7,56 @@
...
+    popup_col: Optional[str] = None,
+    save_as: Optional[str] = None,
+) -> folium.Map:
+    """
+    Plot cluster polygons on a satellite folium map.
+
+    Args:
+        clusters_gdf (gpd.GeoDataFrame): geodataframe containing (Multi)Polygons to plot.
+        popup_col (str): name of column containing information for popup. Optional.
+        save_as (str): file name to save as. Saves a local html file copy to /outputs/maps/.
```
sofiapinto:
```suggestion
        save_as (str): file name to save as without file extension. Saves a local html file copy to /outputs/maps/. E.g. `cluster_map_plymouth`
```

asf_heat_pump_suitability/pipeline/cluster/cluster.py
sofiapinto, line 1: Looks great, thank you!

Since we're applying all of these changes about using building_id from base.yaml rather than hardcoding, can we please fix the fact that there are two building IDs in base.yaml? Maybe that's for another PR but just flagging that we have two :)

asf_heat_pump_suitability/pipeline/cluster/tests/utils.py
```diff
@@ -0,0 +1,11 @@
+def assign_bool_domestic_status(b_id: str, non_domestic: list):
+    """Returns False for specified non-domestic codes."""
+    # Extract prefix (e.g. 'B03') cleanly regardless of lowercase user input
```
sofiapinto: Needs more context i believe. Extracts B03 from what originally? Add to the doc what the original b_id would be.

asf_heat_pump_suitability/pipeline/cluster/tests/test_cluster.py
sofiapinto, line 24: Not for now, but in the future we might want to consider creating a sample from the actual data.

asf_heat_pump_suitability/pipeline/cluster/tests/test_cluster.py
```diff
@@ -0,0 +1,576 @@
...
+    """
+    Generate a boundary geodataframe with a 12m buffer.
+    """
+    combined_footprints = gdf_mixed_buildings.union_all()
+    site_boundary_geom = combined_footprints.buffer(12).convex_hull
+
+    return gpd.GeoDataFrame(
+        {"name": ["SITE_BOUNDARY"], "geometry": [site_boundary_geom]},
+        crs="EPSG:27700",
+    )
```
sofiapinto, line 222: Is this just for the purpose of the testing or does the 12m have a specific meaning?

sofiapinto: I believe this is just here because the code below requires passing something to `boundary_gdf`. But just checking.

asf_heat_pump_suitability/pipeline/cluster/tests/test_cluster.py
```diff
@@ -0,0 +1,576 @@
...
+        )
+
+        results = clusters_gdf[["geometry"]].sjoin(
+            gdf_mixed_buildings, how="inner", predicate="contains"
+        )["building_id"]
+        expected = gdf_mixed_buildings["building_id"]
+        missing = set(expected).difference(set(results))
+        assert (
+            not missing
+        ), f"Some buildings not contained by a cluster. Building IDs: {missing}"
```
sofiapinto, line 298: One to add: I guess we also want to check that each building is only contained within one cluster too.

asf_heat_pump_suitability/pipeline/cluster/tests/test_cluster.py
```diff
@@ -0,0 +1,576 @@
...
+                (400100.0, 400030.0),
+                (399978.0, 400030.0),
+            ]
+        )
+
+    def test_voronoi_contain_single_polygons(
+        self, gdf_mixed_buildings, gdf_enclosing_boundary
+    ):
+        """Test one Voronoi polygon contains one building footprint."""
+        boundary = gdf_enclosing_boundary.geometry.iloc[0]
```
sofiapinto, line 510: Why are we accessing iloc[0] here?

asf_heat_pump_suitability/pipeline/cluster/tests/test_cluster.py
```diff
@@ -0,0 +1,576 @@
...
+        cells_gdf = extend_edges_gdf(
+            gdf=gdf_polygons_across_boundary, boundary=geometry_boundary_crossed
+        )
+        results = cells_gdf["building_id"]
+        expected = gdf_polygons_across_boundary[
+            gdf_polygons_across_boundary["within_boundary"]
+        ]["building_id"]
+        assert set(results) == set(
+            expected
+        ), "Polygons outside or crossing boundaries are not handled correctly"
```
sofiapinto, line 548: this is one we'll need to review/change once we deal with the buildings at the boundaries. perhaps we could add a TODO

asf_heat_pump_suitability/pipeline/cluster/tests/test_cluster.py
```diff
@@ -0,0 +1,576 @@
...
+
+    def test_clip_voronoi_to_buffer(
+        self, gdf_far_apart_polygons, geometry_far_apart_boundary
+    ):
+        """Test Voronoi polygons are clipped to the desired buffer."""
+        cells_gdf = extend_edges_gdf(
+            gdf=gdf_far_apart_polygons, boundary=geometry_far_apart_boundary, buffer=20
+        )
+        results = cells_gdf.area.sum()
+        expected = gdf_far_apart_polygons.buffer(20, join_style=2).area.sum()
```
sofiapinto, line 596: maybe add a note on what join_style=2 means

asf_heat_pump_suitability/pipeline/cluster/tests/test_cluster.py
crispy-wonton: I did consider this but in the end I decided that it would be much harder to locate the different cases I wanted to test.

asf_heat_pump_suitability/pipeline/cluster/tests/test_cluster.py
crispy-wonton: We only need the shapely polygon of the boundary in the function in the line below, not a gdf. This line extracts it.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: I'm confused - I don't think I'm accessing the base config anywhere here. What do you mean?

asf_heat_pump_suitability/utils/mapping_utils.py
crispy-wonton: I have used these functions separately in a notebook to generate different maps, including to generate the test geodataframes.

asf_heat_pump_suitability/utils/mapping_utils.py
crispy-wonton: yes it does. The mappings are in the same order in the config but I have sorted them in this script to make it more robust.

asf_heat_pump_suitability/pipeline/cluster/tests/test_cluster.py
crispy-wonton: Just was slightly above the 10m minimum that I wanted so that no buildings were touching the edge of the map.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
sofiapinto: Not here no, but because I noticed that in cluster.py were accesing the building IDs I wanted to make a note that the base.yaml has the building ID coded in two different ways and that that should be sorted in a separate PR.

asf_heat_pump_suitability/pipeline/cluster/tests/test_cluster.py
sofiapinto: Very fair - my suggestion was to have this sample as an extra, as it's based on real data. So in addition to the specific cases you have here. This is fine for now though!

asf_heat_pump_suitability/pipeline/cluster/tests/test_cluster.py
sofiapinto: perhaps add a comment with this?

asf_heat_pump_suitability/pipeline/cluster/tests/test_cluster.py
sofiapinto: maybe you could add a line of doc?

asf_heat_pump_suitability/pipeline/cluster/tests/test_cluster.py
sofiapinto: just noticed you did :)

asf_heat_pump_suitability/pipeline/cluster/tests/test_cluster.py
sofiapinto: just noticed you did :)

sofiapinto (approved):
Have reviewed latest changes, looks great, thank you!

PR #396, opened 2026-06-22
395 add GitHub actions

### LINKED ISSUE #395: Add github actions for tests

crispy-wonton:
Fixes #395 

# Description

Add github actions workflow to run tests.

New files:
- `.github/workflows/run_tests.py` - add workflow to run github actions

Modified files:
- `environment.yaml` - upgrade from python 3.10 to 3.11. This is required to allow the tests to pass as version 3.11 is required to load `osbng` package.

Note this is currently failing the `Run tests` check (see below) due to the absence of tests on this branch. However, we can see from the report that tests are being searched for and coverage is reported.


.github/workflows/run_tests.yaml
```diff
@@ -0,0 +1,25 @@
...
+          environment-name: asf_heat_pump_suitability
+          cache-environment: true
+          create-args: python=3.12
+      - name: Install package and dev dependencies
+        run: |
+          pip install -r requirements.txt -r requirements_dev.txt
+          pip install -e .
+        shell: micromamba-shell {0}
+      - name: Run tests
+        run: pytest --cov=asf_heat_pump_suitability --cov-report=term-missing
```
sofiapinto: I think we either need to specify the folders where to look for tests, or we need to add something to update `pyproject.toml` to say where not to look for tests. otherwise I believe `--cov-report=term-missing` will look for tests in folders where tests aren't needed. 

This is one of the reasons why it's so handy to have all tests in the same path: you can just easily say where to look for tests.

requirements_dev.txt
sofiapinto, line 1: don't know which one you wanted to merge first, but if the uv PR is merged first, we'll need to add these changes to pyproject.toml

.github/workflows/run_tests.yaml
```diff
@@ -0,0 +1,25 @@
...
+        uses: mamba-org/setup-micromamba@v2
+        with:
+          environment-file: environment.yaml
+          environment-name: asf_heat_pump_suitability
+          cache-environment: true
+          create-args: python=3.12
+      - name: Install package and dev dependencies
+        run: |
+          pip install -r requirements.txt -r requirements_dev.txt
+          pip install -e .
```
sofiapinto: Jusr flagging that a lot of this will need to change now that we're moving to uv.

You can take a look at Seth's PR `.github/workflows/ci.yml` for the changes.

sofiapinto (commented):
Hey @crispy-wonton - this is really exciting, thank you for doing this work! 

Have left a few initial comments. I think we should probably wait for one of the tests PRs is merge and potentially also the uv PR before we merge this.

.github/workflows/run_tests.yaml
crispy-wonton: Yep! On my radar :)

requirements_dev.txt
crispy-wonton: I don't mind which is merged first. If this is merged first, we just need to add an issue to update the yaml to be compatible with `uv` :)

.github/workflows/run_tests.yaml
crispy-wonton: I don't think this is an issue (?) - unless I am missing something about speed gains in terms of searching the repository. As I understand, `pytest` simply searches for all files labelled with `test_` prefix or `_test` suffix in the repository and runs them.

In the coverage report, the coverage is reported for `/getters` `/pipeline` and `/utils` only which is the desired folders I think.

Nothing seems problematic to me in the set up but let me know if I'm missing something! :)

sofiapinto: > In the coverage report, the coverage is reported for /getters /pipeline and /utils only which is the desired folders I think.

Ah nice, in that case i think it's fine. Where is this defined?

crispy-wonton: Not explicitly defined. The reason these ones are tracked is because they have `__init__.py` files in the subdirectories and contain python modules.

To be more explicit, I have now added coverage to the `pyproject.toml`.

.github/workflows/run_tests.yaml
sofiapinto, line 1: This looks good!

The only thing potentially missing is what Seth added at the end of this file with dummy AWS credentials (note that they should effectively be dummy): .`github/workflows/ci.yml` in https://github.com/nestauk/asf_heat_pump_suitability/pull/271/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR1-R10

```
 env:
      AWS_DEFAULT_REGION: eu-west-2
      AWS_ACCESS_KEY_ID: testing
      AWS_SECRET_ACCESS_KEY: testing
      LOCAL_DEV: "true"
```

this is something we can also add in the future too - if any of the tests require connecting to AWS, so that the tests don't crash due to the lack of credentials.

sofiapinto (approved):

.github/workflows/run_tests.yaml
crispy-wonton: Yes - will leave out until required.

PR #400, opened 2026-06-23
215 move to uv

### LINKED ISSUE #213: Makefile conda env remove bug fix
`conda env remove` returns an error when running `make install ` for the first time https://github.com/nestauk/asf_heat_pump_suitability/blob/99fe472c42a93b35ac7de9cc413cac8b459f569b/Makefile#L65C1-L70C16

### LINKED ISSUE #215: Move from conda to uv for environment set up

### LINKED ISSUE #399: Upgrade to python 3.12

crispy-wonton:
Fixes #215 #213 #399

# Description

Move to `uv` instead of `conda` for environment setup. Transition to python 3.12.

Note, I've prompted claude to check through for any breaking changes or silent fails caused by transitioning to 3.12 and there were no concerns. The versions of the packages we use all support 3.12 and there are no unexpected behaviours that were flagged.

I tried running a script with the `uv` environment installed and it worked fine.

New files:
- `.python-version` - add python version
- `pyproject.toml` - add requirements and dependencies etc.
- `uv.lock` - package details for replicability. No need for review - autogenerated by `uv`.

Modified files:
- `.envrc` - activate `uv` environment instead of `conda`
- `.gitignore` - ignore uv.lock and .venv
-  `Makefile` - update to use `uv` instead of `conda`
- `README.md` - update setup instructions

Deleted files as content covered by `toml`:
- `environment.yaml`
- `requirements.txt`
- `requirements_dev.txt`
- `setup.cfg`
- `setup.py`

# Instructions for Reviewer

Follow the set up instructions in the README and see if your set up works.

- Is there anything that should have been transitioned to `uv` that hasn't been? I've compared with the original `conda` set up code and I think it's all covered but please double check there aren't key components missing.

# Checklist:

- [ ] I have tested the code


.envrc
```diff
@@ -3,7 +3,7 @@
 
 # export S3_INPUT_PATH="s3://nesta-ds-projects/<MISSION OR PRACTICE>/<PROJECT NAME>"
 # For Windows users, use bash to run scripts
-export OMP_NUM_THREADS=$(sysctl -n hw.physicalcpu)
-# Activate the conda environment
-eval "$(conda shell.bash activate "asf_heat_pump_suitability")"
-[ "$CONDA_DEFAULT_ENV" = "asf_heat_pump_suitability" ] || echo "Have you run \`make install\`?"
+export OMP_NUM_THREADS=$(nproc 2>/dev/null || sysctl -n hw.physicalcpu 2>/dev/null || echo 1)
```
sofiapinto, line 6: Are we sure we want to change this?

.envrc
```diff
@@ -3,7 +3,7 @@
...
+# Activate the virtual environment
+. .venv/bin/activate
+[ -n "$VIRTUAL_ENV" ] || echo "Have you run \`make install\`?"
```
sofiapinto, line 9: I don't think we should be activating the env inside envrc?

Makefile
sofiapinto, line 1: I believe that we no longer need a MakeFile if we're moving to uv? worth double checking with Seth.

pyproject.toml
```diff
@@ -0,0 +1,107 @@
+[build-system]
+requires = ["setuptools>=61", "wheel"]
+build-backend = "setuptools.build_meta"
```
sofiapinto: In our ds cookiecutter we have the following instead:

```
requires = ["hatchling"]
build-backend = "hatchling.build"
```

I believe it's a more modern alternative.

pyproject.toml
```diff
@@ -0,0 +1,107 @@
...
+    "flake8",
+    "flake8-black",
+    "flake8-import-order",
+    "flake8-bandit",
+    "flake8-bugbear",
+    "black",
+    "Sphinx",
+    "sphinxcontrib-napoleon",
+    "sphinx-rtd-theme",
+    "pytest",
```
sofiapinto, line 63: You can also already add `pytest-cov` if you wish

pyproject.toml
```diff
@@ -0,0 +1,107 @@
...
+    "pre-commit",
+    "pre-commit-hooks",
+    "flake8-pyproject",
+]
+
+[tool.setuptools.packages.find]
+exclude = ["docs*"]
+
+[tool.setuptools.package-data]
+"*" = ["*.txt", "*.yaml"]
```
sofiapinto:
```suggestion
[tool.hatch.build.targets.wheel]
packages = ["asf_heat_pump_suitability"]
```

pyproject.toml
```diff
@@ -0,0 +1,107 @@
...
+application-import-names = "asf_heat_pump_suitability"
+import-order-style = "smarkets"
+docstring-convention = "google"
+exclude = ["*/notebooks/"]
+per-file-ignores = {"tests/*" = ["S"]}
+
+[tool.darglint]
+# Don't pester about parameters for a one-line docstring
+strictness = "short"
+docstring_style = "google"
```
sofiapinto, line 105: I know adding ruff at this point would be a real pain, but it effectively replaces flake8 and darglint. I believe flake8 might actually ignore pyproject.toml and completely ignore this config. Worth checking with someone else who's more familiar with pyproject.toml what we should have here if we truly don't want ruff.

README.md
```diff
@@ -23,15 +23,16 @@ Source code for v1.0.0 is available under [Releases](https://github.com/nestauk/
...
+  - [`direnv`](https://direnv.net/) (environment variable management)
 - Clone the repo and navigate to your local repo folder
 - Run `direnv allow`
 - Run `make install` to configure the development environment:
-  - Setup the conda environment
+  - Create the `.venv` virtual environment with all dependencies
   - Configure `pre-commit`
-  - Install requirements
-- Run `conda activate asf_heat_pump_suitability`
+- The virtual environment activates automatically when you enter the repo directory (via direnv)
```
sofiapinto, line 34: nothing else we need to run on the command line?

sofiapinto (commented):
Hello @crispy-wonton , this looks great!

I have left a couple of notes/questions/comments for your consideration below. I haven't tested that uv actually works. Have you done that and tried to run any of our scripts?

Also, I don't mind if it happens in a separate PR, but should we update the script level docstrings with the new way of running the scripts `uv run my_script.py`?

.envrc
crispy-wonton: sorry if it's a stupid question but why not? I thought this will mean the environment is activated automatically on set up rather than having a separate step to manually activate?

.envrc
crispy-wonton: This was changed because I specified that the repo should have cross-platform support. Previously it supported maxOS only but this adds support for Linux equivalent. 

If we only want to support macOS, I can revert to the original . What do you think? I don't feel strongly but saw an opportunity for more compatability.

Makefile
crispy-wonton: Yes a `Makefile` is no longer required and I specifically prompted claude about why this was retained. The reasoning was that the `Makefile` still have non-environment targets e.g. docs, and therefore the recommendation was that the file is still valuable although it could be refactored out with further changes. Will check with Seth when I'm back.

pyproject.toml
crispy-wonton: The `Makefile` runs `uv sync --extra dev` which installs `flake8-pyproject` (which is specified here in the dev dependencies) alongside `flake8` which will pick up the config from `pyproject.toml`. This means that flake8 should be configured according to the settings here.

Is that the only reason you were considering `ruff` here or do you prefer to implement `ruff` for other reasons as well?

README.md
crispy-wonton: Don't think so. I tested this out and running `direnv allow` and then `make install` worked for me to install and activate the environment.

crispy-wonton:
> Hello @crispy-wonton , this looks great!
> 
> I have left a couple of notes/questions/comments for your consideration below. I haven't tested that uv actually works. Have you done that and tried to run any of our scripts?

Yes I have tested the installation of the `uv` environment and tested running on the clustering script and it worked for me for Plymouth. I also tested running the UPRNs script.

> Also, I don't mind if it happens in a separate PR, but should we update the script level docstrings with the new way of running the scripts `uv run my_script.py`?

I don't think this is essential - you can either run scripts like that or you can use the original `python module.py` format as long as the environment is active.

pyproject.toml
sofiapinto: no, that was it! :)

sofiapinto:
Hey @crispy-wonton - thank you for the changes!

I ran

```
direnv allow
make install
```

Resulting in the following warnings
```
[WARNING] hook id `end-of-file-fixer` uses deprecated stage names (commit, push) which will be removed in a future version.  run: `pre-commit migrate-config` to automatically fix this.
[WARNING] hook id `trailing-whitespace` uses deprecated stage names (commit, push) which will be removed in a future version.  run: `pre-commit migrate-config` to automatically fix this.
pre-commit installed at .git/hooks/pre-commit
[WARNING] hook id `end-of-file-fixer` uses deprecated stage names (commit, push) which will be removed in a future version.  run: `pre-commit migrate-config` to automatically fix this.
[WARNING] hook id `trailing-whitespace` uses deprecated stage names (commit, push) which will be removed in a future version.  run: `pre-commit migrate-config` to automatically fix this.
```

I also run upnrs.py with uv run and it worked.


> > Also, I don't mind if it happens in a separate PR, but should we update the script level docstrings with the new way of running the scripts uv run my_script.py?
> 
> 
> I don't think this is essential - you can either run scripts like that or you can use the original python module.py format as long as the environment is active.
> 
you're right, not essential, but there are a few differences between the two that might make it better to use uv run:
- uv auto detects .venv
- uv handles inline dependencies (not that we're using them at the moment
- uv ensures environment matches uv.lock (this means that if you added a dependency recently, uv run updates the environment on the fly before executing)

It might be useful to at least update the channel about this change so that other DS' are aware of it and so that they run direnv allow, make install and that they are aware they can run scripts with `uv run script.py` as an alternative to `python script.py`

sofiapinto (dismissed):

sofiapinto (dismissed):

sofiapinto (approved):

PR #402, opened 2026-06-23
Bugfix 401 aggregate within radius

### LINKED ISSUE #401: Within anchor radius flag aggregated incorrectly

crispy-wonton:
Fixes #401 

# Description

Fix bug causing all values in anchor property flag to show True for all tech types and all clusters except networked heat pump.

Previously, there was a dissolve aggregation step that was dissolving all cells of the same tech type into one multipolygon with the max `within_50m_from_anchor_load` value across the whole tech type. This would be true for all tech types except networked HP which only remains assigned as the tech if the cell is NOT within the anchor load radius. I've fixed this.

Fyi, a small number of clusters will show NAN values at the moment for the boolean flag. This should be fixed by merging the changes in #389 

# Instructions for Reviewer

run the following line in terminal without saving and check the results of `cluster_gdf`
`python asf_heat_pump_suitability/pipeline/cluster/cluster.py --local_authorities plymouth`

# Checklist:

- [ ] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/pipeline/cluster/cluster.py
+++ b/asf_heat_pump_suitability/pipeline/cluster/cluster.py
@@ -133,11 +133,10 @@ def generate_gdf_clusters(
     )
 
     # Create cluster geometries
-    cluster_id = "_internal_cluster_id"
     clusters_gdf = (
         cells_gdf.dissolve(by="assigned_tech")
         .explode()
-        .reset_index(names=cluster_id)[["assigned_tech", "geometry"]]
+        .reset_index()[["assigned_tech", "geometry"]]
     )
 
     # Create an ID for each geometry that starts with the tech code and ends with a unique number
@@ -155,9 +154,9 @@ def generate_gdf_clusters(
         reassigned_gdf[[f"within_{radius}m_from_anchor_load", "geometry"]],
         how="left",
         predicate="contains",
-    )
+    ).drop(columns="index_right")
 
-    clusters_gdf = clusters_gdf.dissolve(by="cluster_id", aggfunc="max")
+    clusters_gdf = clusters_gdf.dissolve(by="cluster_id", aggfunc="max").reset_index()
 
     # TODO move to testing when sample set available
     if round(clusters_gdf["geometry"].area.sum(), 3) > round(
--- a/asf_heat_pump_suitability/pipeline/cluster/cluster.py
+++ b/asf_heat_pump_suitability/pipeline/cluster/cluster.py
@@ -56,7 +56,7 @@ TECH_CODES = {
 NETWORKED = TECH_TYPES["networked"]
 COMMUNAL = TECH_TYPES["communal"]
 
-ANCHOR_REASSINGMENT_MAPPING = {NETWORKED: COMMUNAL}
+ANCHOR_REASSIGNMENT_MAPPING = {NETWORKED: COMMUNAL}
 
 
 def generate_gdf_clusters(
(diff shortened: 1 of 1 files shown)


asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -56,7 +56,7 @@
 NETWORKED = TECH_TYPES["networked"]
 COMMUNAL = TECH_TYPES["communal"]
 
-TECH_MAPPING = {NETWORKED: COMMUNAL}
+ANCHOR_REASSIGNMENT_MAPPING = {NETWORKED: COMMUNAL}
```
sofiapinto, line 59: nice - this is clearer!

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -690,7 +683,7 @@ def load_transform_anchor_property_gdfs(
     combined_anchor_gdf = pd.concat(
         [anchors_with_footprint, important_building_gdf], join="inner"
     )
-    combined_anchor_gdf["geometry"] = combined_anchor_gdf.normalize()
```
sofiapinto, line 693: interesting how this didn't throw an error before.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -165,6 +149,15 @@ def generate_gdf_clusters(
...
     )
 
+    # Join boolean flag for each building contained in the cluster back to the cluster to aggregate
+    clusters_gdf = clusters_gdf.sjoin(
+        reassigned_gdf[[f"within_{radius}m_from_anchor_load", "geometry"]],
+        how="left",
+        predicate="contains",
+    ).drop(columns="index_right")
+
+    clusters_gdf = clusters_gdf.dissolve(by="cluster_id", aggfunc="max").reset_index()
```
sofiapinto, line 162: Can you add a doc above this line to say something like: touching polygons will "dissolve" into one larger polygon, etc etc

sofiapinto: Question: at this point, what other variables are present in clusters_gdf? is "max" always the correct aggfunc to use?

I'm guessing for "within_{radius}m_from_anchor_load" we use max to get the True's? Can we add a note about that too please?

sofiapinto (commented):
Hey @crispy-wonton - thank you for this work. I've left some questions for you!

I'm not sure if I did something wrong but when I run

`python -i asf_heat_pump_suitability/pipeline/cluster/cluster.py --local_authorities plymouth`

and then inspect `clusters_gdf` by doing

```
clusters_gdf[~pd.isnull(clusters_gdf["within_50m_from_anchor_load"]) & (clusters_gdf["within_50m_from_anchor_load"])]["assigned_tech"].value_counts()
```

I still get techs that aren't just communal.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Yeh, not sure! Maybe it does actually work but at least from the `geopandas` documentation I couldn't see that this was a valid implementation of the `normalize` method.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: At this point, we only have cluster ID (the groupby), geometry (which is being dissolved), within_radius_from_anchor_load (which we want the max of as max will be `True`), and assigned tech (which will be the same across cluster ID).

crispy-wonton:
> Hey @crispy-wonton - thank you for this work. I've left some questions for you!
> 
> I'm not sure if I did something wrong but when I run
> 
> `python -i asf_heat_pump_suitability/pipeline/cluster/cluster.py --local_authorities plymouth`
> 
> and then inspect `clusters_gdf` by doing
> 
> ```
> clusters_gdf[~pd.isnull(clusters_gdf["within_50m_from_anchor_load"]) & (clusters_gdf["within_50m_from_anchor_load"])]["assigned_tech"].value_counts()
> ```
> 
> I still get techs that aren't just communal.

This is expected. The clusters for different technology types can still be within 50m of an anchor property. Whether they are reassigned to communal depends on if they were originally labelled as a networked GSHP. Individual solutions and district heat network clusters can still be within 50m of an anchor property but will not be reassigned to communal based on our logic.

The only thing that shouldn't happen is we should never get a Networked GSHP cluster that is within 50m of an anchor property.

sofiapinto:
> This is expected. The clusters for different technology types can still be within 50m of an anchor property. Whether they are reassigned to communal depends on if they were originally labelled as a networked GSHP. Individual solutions and district heat network clusters can still be within 50m of an anchor property but will not be reassigned to communal based on our logic.
> 
> The only thing that shouldn't happen is we should never get a Networked GSHP cluster that is within 50m of an anchor property.

correct!

sofiapinto (approved):

PR #403, opened 2026-06-24
382 initial land parcel exploration

crispy-wonton:
---

# Description

I have added in voronoi cell area and separately added the 5 nearest neihbour garden sizes and distances. I am also now clipping the data so any property with a garden size >300m2 gets assigned a garden size of 300m2 to deal with the outliers. I think this has improved performance significantly. I am also now looking at new evaluation metrics: median absolute percentage error and root mean squared log error. 

Fixes # (issue)

# Instructions for Reviewer

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/research/exploratory/garden_size/missing_land_parcel_exploration.py
```diff
@@ -0,0 +1,248 @@
...
+    Returns:
+        list: list of OS BNG grid squares corresponding to the input local authorities.
+    """
+    boundary = boundary_gdf.geometry.buffer(buffer_m).union_all()
+
+    # clip grid square gdf to local authority boundaries and return grid squares
+    grid_gdf = grid_gdf.clip(boundary)
+    grid_squares = list(grid_gdf["bng_ref"])
+
+    return grid_squares
```
crispy-wonton, line 30: question: how come you've created this function instead of using the existing one? was it not working properly? asking because I've had issues with the existing function previously.

asf_heat_pump_suitability/research/exploratory/garden_size/missing_land_parcel_exploration.py
```diff
@@ -0,0 +1,248 @@
...
+    ].unique()
+
+    land_parcels_gdf = pd.concat(
+        [
+            outdoor_space.load_transform_gdf_land_parcels(f"s3://{file}")
+            for file in inspire_file_names
+        ],
+        ignore_index=False,
+    )
+    land_parcels_gdf["geometry"] = land_parcels_gdf.normalize()
```
crispy-wonton, line 47:
```suggestion
    land_parcels_gdf["geometry"] = land_parcels_gdf["geometry"].normalize()
```
Without this you are normalising the numerical values in the geodataframe but not the geometry :)

asf_heat_pump_suitability/research/exploratory/garden_size/missing_land_parcel_exploration.py
```diff
@@ -0,0 +1,248 @@
...
+    missing_gdf = missing_gdf.explode(index_parts=False).reset_index(drop=True)
+
+    # Calculate the area for each missing gap
+    missing_gdf["area"] = missing_gdf.geometry.area
+
+    return missing_gdf
+
+
+# %%
+def get_gdf_missing_garden_data(parcels_gdf, grid_squares, boundary_gdf):
```
crispy-wonton, line 73: note: the name is slightly confusing here without docstring as I understand it is getting **building footprints** with missing outdoor space data.

asf_heat_pump_suitability/research/exploratory/garden_size/missing_land_parcel_exploration.py
```diff
@@ -0,0 +1,248 @@
...
+    parcels_gdf = get_gdf_land_parcels(boundary_gdf=boundary_gdf)
+    missing_gdf = get_gdf_missing_land_parcels(
+        parcels_gdf=parcels_gdf, boundary_gdf=boundary_gdf
+    )
+    missing_gardens = get_gdf_missing_garden_data(
+        parcels_gdf=parcels_gdf, grid_squares=grid_squares, boundary_gdf=boundary_gdf
+    )
+
+    print(f"--- Missing Area Statistics for {la} ---")
+    print(missing_gdf["area"].describe())
```
crispy-wonton, line 167: suggestion: it would be more useful to join this with building footprints first and see what the missing area looks like for only domestic buildings missing an outdoor space.

asf_heat_pump_suitability/research/exploratory/garden_size/missing_land_parcel_exploration.py
```diff
@@ -0,0 +1,248 @@
...
+    # Get outdoor space
+    outdoor_space_gdf = outdoor_space.generate_gdf_outdoor_space(
+        building_intersections_gdf=intersection_gdf, land_parcels_gdf=parcels_gdf
+    )
+
+    garden_size = buildings_gdf.sjoin(
+        outdoor_space_gdf, how="left", predicate="intersects"
+    )
+
+    missing_gardens = garden_size[garden_size["total_outdoor_space_area_m2"].isna()]
```
crispy-wonton, line 95: suggestion: it would be useful to join this with domestic UPRNs so that we are only looking at buildings containing domestic UPRNs. You could also add roads here to separate the missing land parcels from bigger chunks of public land like parks etc.

asf_heat_pump_suitability/research/exploratory/garden_size/missing_land_parcel_exploration.py
```diff
@@ -0,0 +1,248 @@
...
+        parcels_gdf=parcels_gdf, grid_squares=grid_squares, boundary_gdf=boundary_gdf
+    )
+
+    print(f"--- Missing Area Statistics for {la} ---")
+    print(missing_gdf["area"].describe())
+
+    total_missing = missing_gdf["area"].sum()
+    total_parcel = parcels_gdf.geometry.area.sum()
+    print(f"\nTotal Missing Area: {total_missing:.2f}")
+    print(f"Total Parcel Area: {total_parcel:.2f}")
```
crispy-wonton, line 172: suggestion: as above, I'd be interested in these numbers just related to buildings/ building segments containing domestic UPRNs. I.e. % of UPRNs missing outdoor space per LA; difference overlay polygons (i.e. missing land parcels) that intersect with building footprints containing domestic UPRNs); average of missing area that intersects domestic UPRNs.

asf_heat_pump_suitability/research/exploratory/garden_size/missing_land_parcel_exploration.py
```diff
@@ -0,0 +1,248 @@
...
+def plot_garden_size_by_signature(
+    df,
+    la_name,
+    target_col="max_contiguous_outdoor_space_area_m2",
+    sig_col="spatial_signature_types",
+):
+    plt.figure(figsize=(12, 6))
+
+    # Drop NaNs just for the plot
+    plot_data = df.dropna(subset=[target_col, sig_col])
```
crispy-wonton, line 196: suggestion: as we have really microscopic slivers of outdoor space here, I'd probably drop anything below like 50cm2 here as well.

asf_heat_pump_suitability/research/exploratory/garden_size/missing_land_parcel_exploration.py
```diff
@@ -0,0 +1,248 @@
...
+]
+
+# %%
+for la in local_authority_slugs:
+    uprn_with_features = pd.read_parquet(
+        f"s3://asf-local-heat-planning-tool/outputs/data/{la}/{la}_with_features.parquet"
+    )
+    uprn_with_features["spatial_signature_types"] = uprn_with_features[
+        "spatial_signature_types"
+    ].apply(lambda x: x[0] if isinstance(x, (np.ndarray, list)) and len(x) > 0 else x)
```
crispy-wonton, line 239: question: what's the sort of proportion of UPRNs that have multiple values? If it's a very tiny proportion then this is ok but if it's larger, I think we should do this more carefully. E.g. creating a hierarchy of signature types and taking the densest one for example.

asf_heat_pump_suitability/research/exploratory/garden_size/missing_land_parcel_exploration.py
```diff
@@ -0,0 +1,248 @@
...
+
+    # Drop NaNs just for the plot
+    plot_data = df.dropna(subset=[target_col, sig_col])
+
+    sns.boxplot(
+        data=plot_data,
+        x=sig_col,
+        y=target_col,
+        palette="viridis",  # Uses a nice color gradient
+    )
```
crispy-wonton, line 203: suggestion: on the boxes, it would also be good to add counts for each group so we get a sense of the scale of where these numbers come from.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+
+    known_gardens = gdf.dropna(subset=["max_contiguous_outdoor_space_area_m2"])
+    known_centroids = np.column_stack(
+        (known_gardens.geometry.centroid.x, known_gardens.geometry.centroid.y)
+    )
+    known_sizes = known_gardens["max_contiguous_outdoor_space_area_m2"].values
+    known_ids = known_gardens[unique_id_col].values
+
+    # We query 10 neighbours to ensure we have 5 left after dropping the self-match
+    nn = NearestNeighbors(n_neighbors=10, algorithm="kd_tree")
```
crispy-wonton: question: wouldn't we just need to query 6 neighbours to ensure we have 5 after the self match? It would probably increase the run time to reduce the number of neighbours here.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+# %%
+def _get_int_garden_area_nearest_neighbours(gdf, unique_id_col):
+
+    all_centroids = np.column_stack((gdf.geometry.centroid.x, gdf.geometry.centroid.y))
+    all_ids = gdf[unique_id_col].values
+
+    known_gardens = gdf.dropna(subset=["max_contiguous_outdoor_space_area_m2"])
+    known_centroids = np.column_stack(
+        (known_gardens.geometry.centroid.x, known_gardens.geometry.centroid.y)
+    )
```
crispy-wonton, line 63: question: are you accounting for duplicate land parcels matched to different UPRNs here? Not sure what `gdf` contains, I'm assuming building footprints due to having a centroid, but it would be important to get unique neighbour outdoor space sizes. I think this is already happening but want to double check.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+        # Apply the mask to get the sizes of only the valid neighbours
+        valid_sizes = known_sizes[row_indices[valid_mask]]
+
+        # Take the mean of the first 5 valid neighbours
+        if len(valid_sizes) >= 5:
+            mean_5nn.append(np.mean(valid_sizes[:5]))
+        elif len(valid_sizes) > 0:
+            mean_5nn.append(np.mean(valid_sizes))
+        else:
+            mean_5nn.append(np.nan)
```
crispy-wonton: question: are there cases where this happens? I'm assuming no because there is no radius set?

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+        {
+            "target_id": gdf[unique_id_col],
+            "uprn_weight": gdf["uprn_weight"],
+            "geometry": gdf.geometry.centroid,
+        },
+        crs=gdf.crs,
+    )
+
+    # find target points inside buffer radius
+    joined = gpd.sjoin(buffers, points, how="left", predicate="contains")
```
crispy-wonton, line 172: suggestion: for UPRN level, you could try looking at counts of UPRNs outside the building that the target UPRN is in.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+    # remove temporary weight column we added to the original GeoDataFrame
+    gdf.drop(columns=["uprn_weight"], inplace=True)
+
+    return uprn_sums
+
+
+# %%
+def _calculate_plot_ratio_proxy(buildings_gdf, buffer_radius=100):
+    # calculate the area of the buffer circle
+    buffer_area = 3.14159 * (buffer_radius**2)
```
crispy-wonton: note: you can use `math.pi`

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+
+    # Create the buffers
+    buffers = buildings_gdf.geometry.centroid.buffer(buffer_radius)
+    buffers_gdf = gpd.GeoDataFrame(geometry=buffers, index=buildings_gdf.index)
```
crispy-wonton: suggestion: I'd suggest using building `ID` here as it's probably more robust than using the index.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+
+    # get building polygons that are within the buffer circles
+    # the geometry will be of the full building (not just the bit inside the circle)
+    # explicitly give them the exact same index as the 'joined' dataframe.
+    right_geoms = gpd.GeoSeries(
+        buildings_gdf.loc[joined["index_right"], "geometry"].values, index=joined.index
+    )
+
+    # now clip building geometries to just what is within the buffer circle areas
+    exact_intersections = left_geoms.intersection(right_geoms)
```
crispy-wonton: question: I might be missing something, but why are we using this methodology instead of just using `overlay` with `intersection` and finding the area of the intersections?

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+        # Sum the coordinates for all pieces of the multi-polygon
+        return sum(len(poly.exterior.coords) for poly in geom.geoms)
+    return 0
+
+
+# %%
+aggregation_rules = {
+    "spatial_signature_types": "first",
+    "UPRN": "count",
+    "max_contiguous_outdoor_space_area_m2": "first",
```
crispy-wonton, line 306: question: I'm not sure if this is correct. E.g. if there are multiple gardens for one building, it will only capture the first garden. However, you can't do the sum either because it will give you duplicates. You would need to sum the unique outdoor space area values for a building.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+    "ATTACHMENT": "first",
```
crispy-wonton, line 307: thought: I'm also not sure if this is correct. There will be different terrace attachment types within the same building (end-terrace, mid-terrace) etc. The other thing is that there could be erroneous data points as this comes from the EPC. Suggestion: aggregate all types of terrace into one group, and then check the distribution of how many unique attachment types buildings that have >1 UPRN have. If they generally have 1 attachment type, fine to agg with first as you have here, otherwise you might want to take the most frequent attachment type.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+
+    # Divide the actual building area by the convex hull area
+    buildings_gdf["building_convexity"] = (
+        buildings_gdf.geometry.area / convex_hull_areas
+    )
+
+    # Fill any 0-division errors etc with 1.0
+    buildings_gdf["building_convexity"] = buildings_gdf["building_convexity"].fillna(
+        1.0
+    )
```
crispy-wonton, line 348: thought: I'm not sure if 1 is the best number to fill this with? If you are getting a zero division error it implies all the points from the building are in a perfect line or are all in the same coordinate. A value of 1 indicates the convexity is 0 to me, whereas if there is 0m2 of convex hull area, this implies very high convexity to me. So perhaps you should replace any zero values in `convex_hull_areas` with a small number close to zero (order of magnitude dependent on the rest of the convex_hull_area values).

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+    gdf_with_features["perimeter_to_area_ratio"] = (
+        gdf_with_features["building_perimeter_m"]
+        / gdf_with_features["building_area_m2"]
+    )
+
+    # convert spatial signature types to be one column per spatial signature type, with 1 being True and 0 being False
+    gdf_with_features["spatial_signature_types"] = (
+        gdf_with_features["spatial_signature_types"]
+        .astype(str)
+        .str.replace(r"[\[\]\'\"]", "", regex=True)
```
crispy-wonton, line 417: docs: it would be good to add an annotation here to explain the intended targets of this regex, e.g. 'remove square brackets and apostrophes'.

Secondly, do you know why are these characters in the column? I can't see them in the column when I first load the data.

asf_heat_pump_suitability/research/exploratory/garden_size/missing_land_parcel_exploration.py
```diff
@@ -0,0 +1,248 @@
...
+        local_authority=la,
+    )
+
+    plot_statistical_distribution_land_parcels(
+        missing_gdf=missing_gdf, local_authority=la
+    )
+
+
+# %%
+def plot_garden_size_by_signature(
```
crispy-wonton, line 187: comment: I think we would see more correlation here if we filtered to domestic UPRNs only. I can see how including all buildings might skew the distributions as large commercial plots can appear in most signatures.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+
+# %%
+def prepare_test_train_datasets(df, feature_cols):
+    df = df.fillna(
+        {
+            "mean_5nn_garden_size": df["mean_5nn_garden_size"].median(),
+            "uprns_within_100m": 0,
+            "area_per_uprn": df["area_per_uprn"].median(),
+        }
+    )
```
crispy-wonton: question: are there a lot of NA values here? It might be better to drop them?

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+from asf_heat_pump_suitability.pipeline.transform import local_authority
+
+
+# %%
+def _get_int_garden_area_nearest_neighbours(gdf, unique_id_col):
+
+    all_centroids = np.column_stack((gdf.geometry.centroid.x, gdf.geometry.centroid.y))
+    all_ids = gdf[unique_id_col].values
+
+    known_gardens = gdf.dropna(subset=["max_contiguous_outdoor_space_area_m2"])
```
crispy-wonton, line 48: suggestion: it might be worth accounting for the unknown garden sizes somehow as well. e.g. proportion of unknowns within Xm radius, and the N rankings of the selected 5 nearest neighbours below (e.g. 11, 12, 13, 14, 15 if the 10 nearest neighbours had null values for example). Otherwise, you could end up using neighbours from a variety of rankings to train the model without making that information available for the model to learn from.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+
+    # Prediction data: Buildings missing their garden size
+    predict_df = df[df["max_contiguous_outdoor_space_area_m2"].isna()].copy()
+
+    coords = np.column_stack(
+        (known_df.geometry.centroid.x, known_df.geometry.centroid.y)
+    )
+    spatial_clusters = KMeans(n_clusters=5, random_state=42, n_init=10).fit_predict(
+        coords
+    )
```
crispy-wonton, line 475: note: there are geospatial clustering algorithms that exist for geospatial data so you could also look into those in case there's a more suitable option.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+            "uprns_within_100m": 0,
+            "area_per_uprn": df["area_per_uprn"].median(),
+        }
+    )
+    print("Splitting datasets...")
+    # Training data: Buildings where we already know the garden size
+    known_df = df[df["max_contiguous_outdoor_space_area_m2"].notna()].copy()
+
+    # Prediction data: Buildings missing their garden size
+    predict_df = df[df["max_contiguous_outdoor_space_area_m2"].isna()].copy()
```
crispy-wonton: note: you can exclude the prediction data from any model training functions and just introduce it when you have a trained model.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+        (known_df.geometry.centroid.x, known_df.geometry.centroid.y)
+    )
+    spatial_clusters = KMeans(n_clusters=5, random_state=42, n_init=10).fit_predict(
+        coords
+    )
+    known_df["spatial_cluster"] = spatial_clusters
+
+    features = [col for col in known_df.columns if col in feature_cols]
+    X_known = known_df[features]
+    y_known = known_df["max_contiguous_outdoor_space_area_m2"]
```
crispy-wonton:
```suggestion
    X = known_df[features]
    y = known_df["max_contiguous_outdoor_space_area_m2"]
```
I would just standardise to `X` and `y` here as it's easier to read

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+    "plot_ratio_proxy",
+    "perimeter_to_area_ratio",
+    "dist_to_nearest_building_centroid_m",
+    "building_convexity",
+    "building_vertex_count",
+]
+
+# %%
+feature_cols = [
+    "uprns_in_building",
```
crispy-wonton: suggestion: could we rename this variable?
```suggestion
    "n_uprns_in_building",
```

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+# %%
+rf = RandomForestRegressor()
+
+param_distributions = {
+    "n_estimators": [100, 250, 500],
+    "max_depth": [10, 20, 30, None],
+    "min_samples_split": [2, 5, 10],
+    "min_samples_leaf": [1, 2, 4],
+    "max_features": ["sqrt", "log2", 1.0],
+}
```
crispy-wonton, line 638: suggestion: for further refinement of the model: it looks like some of your parameters are at the limit of the distributions you tested. I recommend expanding your parameter distributions to refine the hyperparameters to their optimal values, e.g. `max_depth` is currently at 30 so you could create much higher limits in your hyperparameter space with more frequent intervals e.g. range 25-100 with interval 5. `n_estimators` is at the lower limit but you have quite large intervals so again, might be worth exploring 10-200 with interval 20 for example. You can start with large intervals if you prefer and then keep narrowing it down.

With `min_samples_split` i would also test an interval of 1 within your current range of 2-10.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+
+# %%
+rf_random = RandomizedSearchCV(
+    estimator=rf,
+    param_distributions=param_distributions,
+    n_iter=15,
+    cv=test_train_dict["gkf"],  # Uses spatial GroupKFold
+    n_jobs=-1,
+    scoring="neg_mean_absolute_error",
```
crispy-wonton: question: for my understanding, why have you chosen to use negative MAE here?

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+param_distributions = {
+    "n_estimators": [100, 250, 500],
+    "max_depth": [10, 20, 30, None],
+    "min_samples_split": [2, 5, 10],
+    "min_samples_leaf": [1, 2, 4],
+    "max_features": ["sqrt", "log2", 1.0],
+}
+
+# %%
+rf_random = RandomizedSearchCV(
```
crispy-wonton, line 643: suggestion: when you increase the hyperparameter space according to the suggestions above, I recommend testing out either `HalvingGridSearchCV` or `HalvingRandomSearchCV` (if grid search is taking a long time). This will check the different hyperparameters more thoroughly.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+    cv=test_train_dict["gkf"],  # Uses spatial GroupKFold
+    n_jobs=-1,
+    scoring="neg_mean_absolute_error",
+)
+
+# %%
+rf_random.fit(
+    test_train_dict["X_train"],
+    test_train_dict["y_train"],
+    groups=test_train_dict["groups_train"],
```
crispy-wonton: question: which object is the `groups` parameter being passed to?

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+
+# %%
+bdt = GradientBoostingRegressor()
+
+bdt_param_distributions = {
+    "n_estimators": [100, 250, 500],
+    "learning_rate": [0.01, 0.05, 0.1, 0.2],
+    "max_depth": [3, 5, 7, 10],
+    "subsample": [0.8, 0.9, 1.0],
+}
```
crispy-wonton, line 672: suggestion: here as well, your best parameters are at the limits of your search space so you can expand it.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+    groups=test_train_dict["groups_train"],
+)
+
+# %%
+bdt_best = bdt_random.best_estimator_
+bdt_predictions = bdt_best.predict(test_train_dict["X_test"])
+
+# %%
+
+print("\n--- Side-by-Side Model Comparison ---")
```
crispy-wonton, line 696: suggestion: I had another look at metrics to use here and I came across a couple that I think would be really useful to test out and refine, although they both have caveats. The benefit of both is that they look at relative/ percentage error rather than absolute error. E.g. being off by 5m2 in an actual garden of 10m2 is treated the same as being 1000m2 off a 2000m2 estate. It means the outliers won't affect the metrics so much:
1. Mean Absolute Percentage Error (MAPE). Downsides, it can't handle zero values (it will yield an error) and can be unstable with near-zero values (e.g. the error can become very large).
2. Root Mean Squared Log Error (RMSLE). This can handle zero and non-zero values fine because it adds 1 to the values before calculating the metrics. Downsides, it penalises underestimation much more heavily than overestimation. I think this is problematic because over or underestimation around our threshold values for the decision tree will affect the outcome equally. I think, if anything, overestimation is worse because it recommends a tech type that won't actually fit in the space (i.e. individual solution in a garden that's too small).

Both also would require any negative predictions to be corrected first.

We would need to resolve these issues before implementing the metrics - please could you look into ways we could implement either of these without running into the problems, or look into alternative metrics that handle relative error instead.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+# %%
+rf_uprn = RandomForestRegressor()
+
+param_distributions = {
+    "n_estimators": [100, 250, 500],
+    "max_depth": [10, 20, 30, None],
+    "min_samples_split": [2, 5, 10],
+    "min_samples_leaf": [1, 2, 4],
+    "max_features": ["sqrt", "log2", 1.0],
+}
```
crispy-wonton: as above, re hyperparameter space

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -0,0 +1,720 @@
...
+    uprn_test_train_dict["y_train"],
+    groups=uprn_test_train_dict["groups_train"],
+)
+
+# %%
+best_model_uprn = rf_random_uprn.best_estimator_
+print(f"\nBest Parameters: {rf_random_uprn.best_params_}")
+
+# To get a realistic performance metric, we predict on the training data using cross-val
+# (In a real strictly rigorous pipeline, you'd hold out a completely separate spatial test set)
```
crispy-wonton: agree with this comment that we should be implementing a separate hold-out test set to run the model on after cross validation.

crispy-wonton (commented):
hi @simran-dave 

Thanks for all the great work on this! This is a lot of great exploration and analysis.

I have left some comments and ideas for how you can take this work forward to improve the model. My main suggestions are:

- Conduct your exploration on outdoor space missing from building footprints containing domestic properties only. You may end up seeing more patterns after removing buildings which our pipeline is identifying as non-domestic, because I expect commercial buildings will generally have larger plots which could be skewing the data and masking any correlations.
- Train and apply the model to domestic UPRNs only.
- I have suggested some refinements related to the nearest neighbours feature like accounting for neighbour ranking and unknown values. It would also be good to include distances as discussed.
- Expand the hyperparameter space for tuning as the best parameters are often at the limits of the space for the models you've trained.
- Use a different hyperparameter search - e.g. `HalvingRandomSearchCV` for more thorough searching.
- We should implement a different metric for evaluating the models, I don't think the MAE is fit for purpose due to massive outliers. I have suggested looking into MAPE and RMSLE to start with, although both have their limitations.

asf_heat_pump_suitability/research/exploratory/garden_size/missing_land_parcel_exploration.py
simran-dave: Yeah I had trouble with running `load_geodata.load_gdf_bng_grid_squares()` more than once in the same notebook/ script so I moved it outside the function in this notebook. I'm not sure why but this is the error I get when I try run it more than once: `ValueError: Assigning CRS to a GeoDataFrame without a geometry column is not supported.`

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: yeah so this should be a gdf with building footprints, and garden size info. I am getting the centroid of all the building footprints for the nearest neighbours estimation and dropping any which do not have a garden size associated.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: yeah I suppose this shouldn't happen and 5 neighbours should always be found

asf_heat_pump_suitability/research/exploratory/garden_size/missing_land_parcel_exploration.py
crispy-wonton: Hmm that's the same error I was getting before but I thought I had addressed it with [this fix](https://github.com/nestauk/asf_heat_pump_suitability/commit/9c2152d9bbea72ca58c285c5a8aa667083f55762). Clearly it's not working for imports as I can see you have this change on your branch.

The reason for the error is that the object created is an iterator, so once you access it once, you exhaust the iterator and so there is no geometry the second time you access it.

Please can you open an issue with this?

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -4,66 +4,123 @@
...
 
 # %%
-def _get_int_garden_area_nearest_neighbours(gdf, unique_id_col):
+def _get_gdf_5nn_spatial_features(gdf, unique_id_col):
+    """
+    Takes a GeoDataFrame of buildings, finds the 5 nearest neighbors that have
+    known garden sizes, and extracts their individual sizes and distances.
+
+    Args:
+        gdf (gpd.GeoDataFrame): GeoDataFrame of buildings footprints or UPRN point coordinates
```
crispy-wonton, line 37: question: does this actually work for UPRNs or just buildings? Thinking about this because you are generating a centroid from a point geometry and not sure if /how that works?

Ultimately, we would want to conduct this at UPRN level.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
crispy-wonton: If using this for UPRN level, I think it's possible you will still end up with duplicate gardens. E.g. if the 5 nearest neighbours all share the same garden, then you will get 5x the same garden size. I'm genuinely not sure if this is a problem - perhaps it is actually the behaviour that we want, although I think not as overall it would mean the model is provided with much less information for those UPRNs. Alternatively, you may want to use the 5 nearest outdoor spaces to each UPRN.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -177,6 +231,69 @@ def _count_vertices(geom):
     return 0
 
 
+# %%
+# I have not made this work with the building level data yet but we aren't training on that anyway
```
crispy-wonton, line 240: I wouldn't worry about building level from this point forward - let's focus on UPRNs.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -177,6 +231,69 @@ def _count_vertices(geom):
...
+        "Joining Voronois to original building footprints and dissolving per footprint..."
+    )
+    # Join the original building points with IDs to the Voronoi cells and dissolve to get one polygon per internal building ID
+    voronoi_gdf = (
+        voronoi_gdf.sjoin(gdf, how="inner", predicate="contains")
+        .dissolve(by=id_col)
+        .reset_index()
+    ).clip(boundary["geometry"])
+
+    voronoi_gdf = voronoi_gdf[[id_col, "geometry"]].reset_index(drop=True)
```
crispy-wonton, line 276: Since you are creating Voronoi cells from UPRNs which are point coordinates, I think this code is overly complex. The intention of these lines is to convert a building footprint into its component point geometries and then generate Voronoi around each point. Instead, you should just be able to generate the Voronoi directly around each UPRN with something as simple as: 

```
# Drop UPRNs located at the same coordinates
voronoi_gdf = gdf.copy().drop_duplicates(subset="geometry")

# Create polygons around UPRNs
voronoi_gdf = voronoi_gdf["geometry"].voronoi_polygons(extend_to=boundary)

# Join UPRNs back to their corresponding Voronoi cells
voronoi_gdf = gdf.sjoin(voronoi_gdf, how="left", predicate="within")
```

As above, you should probably deduplicate the UPRNs which are located at the exact same coordinates first, and then after you have created the voronoi polygons, `sjoin` the UPRNs back to them so that you have one for each UPRN.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -177,6 +231,69 @@ def _count_vertices(geom):
...
+    linestring_barriers = cluster.load_tranform_gdf_linestring_barriers(
+        grid_squares=grid_squares
+    )[["geometry"]].reset_index(drop=True)
+
+    voronoi_gdf = (
+        voronoi_gdf.overlay(polygon_barriers, how="difference")
+        .overlay(linestring_barriers, how="difference")
+        .explode()
+    )
+
```
crispy-wonton, line 290: Before calculating the area, you should join the Voronoi cells back to the UPRNs. This will discard any parts of the cell that no longer contain the UPRN and give you a more accurate representation of area. When you do this, you need to account for some edge cases where a UPRN may no longer intersect with the cell (due to a barrier completely overlapping with a UPRN point). You would need to think about how to best to handle these cases, e.g. setting area to 0, setting a minimum area, setting the area of the original voronoi cell, setting a negative value. Perhaps something to bring to the sprint review for discussion.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -186,7 +303,17 @@ def _count_vertices(geom):
 }
 
 
-def engineer_gdf_features(grid_squares, local_authorities, return_uprn_level=False):
+def engineer_gdf_features(local_authorities, return_uprn_level=False):
+    """
+    Engineer set of features for model training
+    Args:
+        local_authorities (str | list[str]): Local Authority or Authorities to engineer features for
+        return_uprn_level: whether to return the features at the building or UPRN level
```
crispy-wonton, line 316: Note: generally when refactoring throughout this script, I would drop the building level and just have it operate at UPRN level only.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -275,6 +401,10 @@ def engineer_gdf_features(grid_squares, local_authorities, return_uprn_level=Fal
         / gdf_with_features["building_area_m2"]
     )
 
+    gdf_with_features = _compute_voronoi_area(
+        gdf_with_features, grid_squares, boundary_gdf
```
crispy-wonton: suggestion: add kwargs to avoid silent fails.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -305,41 +432,60 @@ def engineer_gdf_features(grid_squares, local_authorities, return_uprn_level=Fal
...
+    gdf = gdf.fillna(
         {
-            "mean_5nn_garden_size": df["mean_5nn_garden_size"].median(),
+            #'mean_5nn_garden_size': df['mean_5nn_garden_size'].median(),
             "uprns_within_100m": 0,
-            "area_per_uprn": df["area_per_uprn"].median(),
+            "area_per_uprn": gdf["area_per_uprn"].median(),
         }
     )
+
```
crispy-wonton, line 461: question: how come you've chosen to fill NA values instead of dropping rows?

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -604,51 +750,56 @@ def prepare_test_train_datasets(df, feature_cols):
 # %%
 uprn_test_train_dict = prepare_test_train_datasets(uprn_df, feature_cols=feature_cols)
 
+# %%
+uprn_test_train_dict["y_train"]
+
+# %%
+y_train_clipped = np.clip(uprn_test_train_dict["y_train"], a_min=None, a_max=300.0)
```
crispy-wonton, line 767: question: how did you select 300? was it an educated guess or based on a distribution or a percentile? E.g. we could use the 90th or 75th percentile if either are sensible (if you haven't already).

crispy-wonton (commented):
Nice work and great additions!

The original metrics show the error is much much better with the clipping and new features so that's great. It looks like the new metrics may need some extra work to make them suitable for use here as the MAPE and RMSLE are both quite large.

My main comments are:
- I believe the function to generate the Voronoi polygons can be significantly simplified and have left a suggestion in a comment.
- As discussed and as Dan suggested, you may want to look into geographically weighted regression (starting with `pysal` implementation) as an alternative model which already accounts for geospatial variation and would remove the need to manually engineer features that account for this variation (e.g. your engineering of the nearest neighbour features). I think one benefit of this is from the sounds of it, it should be able to innately handle sparse data caused by missingness - the current model would need that information additionally engineered as a new feature.
- Generally as you go forward and refactor, I think we can drop the compatibility with building level of the functions and stick to UPRN level only.
- The hyperparameters fit within the space better now but you have quite sparse distributions that you're searching in. I would iterate on what you have to further optimise the hyperparameters by narrowing the space and searching across more values around the currently identified best params.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: yeah just a rough first guess. The main thinking was that we don't really care about the difference between gardens >70m2 but I wanted to choose a number far away enough that gardens on the boundary were dealt with properly. I'll have a look at optimising this when I refactor- using the percentiles makes sense.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: I think otherwise `overlay` would include the overlaps between the buffer circles

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: I'm leaving this for now since I am now only looking at the URPN level, so no aggregation rules needed

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: same as above

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: yeah a value of 0 means that the points are all in a line/ in the same place. i think this means that the area of the building itself will also be 0. So I think 1 makes sense? as a small number close to zero would indicate a really complex shape

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: I think I'm not fully understanding how a geospatial clustering algorithm would be different

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: I've tried a few different metrics here and it didn't make much of a difference to overall performance. Negative because the scorer always assumes a greater score is better.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: I'll implement this when refactoring

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: I'll implement this when refactoring

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: This line does the hyperparemeter search and then the full model training in one. During the hyperparameter search, the data is split into a test and validation set. This is then iterated on, changing which data is put in the test/validation sets each time using group k fold. The groups parameter tells the algorithm to keep the data in the same spatial group (here given by groups_train) within the same test/validation set.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: Again, will do this when refactoring

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: Have looked into some alternative metrics, as well as the performance at categorising into >30, 30-70 and 70 m2 buckets

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: also for refactoring

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: Mainly for the prediction set, but could exclude any rows with NA values from the test/train set and just fill values for those we need to predict a garden size for

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: i think that with shapely, the centroid should be the same as the point geometry. so I wrote it like this to handle both buildings and UPRNs. Though now we are not using buildings this isn't necessary

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: ah yeah, can simplify now I am only using UPRNs

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: Oh weird, I had a bunch of square brackets when I loaded this data

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: So I've tried dropping duplicate coordinates from the known gardens. So now the nearest neighbour search is done with unique gardens which I think helps with this

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: Will leave this for refactoring

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -44,11 +46,23 @@ def _get_gdf_5nn_spatial_features(gdf, unique_id_col):
 
     # get just the known garden sizes and the centroids of them
     known_gardens = gdf.dropna(subset=["max_contiguous_outdoor_space_area_m2"])
+
+    known_gardens["coord_round"] = (
+        known_gardens.geometry.centroid.x.round(4).astype(str)
+        + "_"
+        + known_gardens.geometry.centroid.y.round(4).astype(str)
+    )
```
crispy-wonton, line 54: question: do you know if you will end up with duplicates from the rounding that are actually unique points?

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
```diff
@@ -107,6 +121,9 @@ def _get_gdf_5nn_spatial_features(gdf, unique_id_col):
     gdf["nn4_distance_m"] = nn4_dist
     gdf["nn5_distance_m"] = nn5_dist
 
+    if "coord_sig" in gdf.columns:
```
crispy-wonton: question: should this say `coord_round`?

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
crispy-wonton: I think you should be able to use something like this:
`intersections_gdf = gpd.overlay(buffers_gdf, buildings_gdf, how="intersection")` and it will give you the parts of the buildings that intersect with the buffer. Then you can flatten the intersections by buffer ID and get the area of the intersections / buffer_area to get your ratio. It's simpler than the above.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
crispy-wonton: I'm imagining that if points are in a line or same place, the building could either be a terrace, or block of flats, respectively. The area of the building footprint would not be 0m2 but the convex area would be 0.

Conversely, a convexity of 1 means that building area == convex area, which implies the UPRNs are at each corner of the building. That's why I think 1 is inappropriate for the case where convex area is 0.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
crispy-wonton: Sorry yes I didn't explain this very well at all.

I was more thinking that you might want to group points based on their spatial position AND other features. Basically grouping similar properties that are in the same area together. You can do this with geospatial clustering algorithms like we discussed previously when talking about geospatially weighted regression. I think another relevant differences is that geospatial clustering account better for density / sparsity of points whereas k-means forces you to select the number of clusters ahead of time and so doesn't account for this as well. 

Anyway, I think this is a low priority on the list of improvements but I just wanted to note it as a thought here.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
crispy-wonton: Ah true! Makes sense

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
crispy-wonton: Thank you that's clearer - I still can't figure out which underlying object the parameter is passed to? Just trying to look at the docs.

crispy-wonton (commented):
Responded to some of your comments. Am just re-running the notebook to see the changes with the 3 classes you created.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
crispy-wonton: Note to self - this is calculating BUILDING footprint convexity, not convexity of UPRN points within a building. So if the points are all in a line / in the same space, that means the building footprint is erroneous.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: No I've not checked. Although, here I am finding the centroid of each garden and dropping any duplicates (rounded to 4 dp) so if two different gardens became the same with the rounding that would imply that the centres of the two gardens are within 0.00005 m of each other. Which I think wouldn't really happen unless there was an error in the garden geometries.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: I think it gets passed to the `cv` parameter that is defined when you define `RandomizedSearchCV` here:
```
rf_random_uprn = RandomizedSearchCV(
    estimator=rf_uprn,
    param_distributions=param_distributions,
    n_iter=15,
    cv=uprn_test_train_dict[
        "gkf"
    ],  # telling the rf to use Group K fold when splitting the training set into cross-validation sets
    n_jobs=-1,
    scoring="neg_mean_absolute_error",  # I tried a couple different metrics at this step and the results didn't change too much
)
```

So above, we are telling it to use group K fold to split the data, but we haven't told it what the groups are. And then in the code you've commented on we are passing the info on what the actual groups are to the splitter. That's my understanding of what's happening.

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
simran-dave: ok I think I have it working with the method you've given now !

asf_heat_pump_suitability/research/exploratory/garden_size/garden_size_model_training.py
crispy-wonton: Yes i think thats a fair assumption

PR #405, opened 2026-07-02
404 Temp fix for missing outdoor space

### LINKED ISSUE #404: Temporary mixed tech output for properties with missing outdoor space

sofiapinto:
---

# Description

This PR introduces changes to `decision_tree.py` as a temporary fix for the missing outdoor space, where areas are displayed as "Individual or Networked Heat Pump" / "Individual or Heat Network Zone" if outdoor space is unknwon

Fixes #404 

# Instructions for Reviewer

Can you please check if you agree to the changes in the logic?

### DIFF
--- a/asf_heat_pump_suitability/pipeline/transform/decision_tree.py
+++ b/asf_heat_pump_suitability/pipeline/transform/decision_tree.py
@@ -278,38 +278,42 @@ def assign_df_unique_solution(solutions_per_footprint_df: pl.DataFrame) -> pl.Da
                 & pl.col("assigned_tech").list.contains(TECH_TYPES["networked"])
             )
             .then(pl.lit(TECH_TYPES["communal"]))
+            # heat network takes precedence
             .when(pl.col("assigned_tech").list.contains(TECH_TYPES["heat_networks"]))
             .then(pl.lit(TECH_TYPES["heat_networks"]))
-            # networked heat pump
+            # networked heat pump takes precedence
             .when(pl.col("assigned_tech").list.contains(TECH_TYPES["networked"]))
             .then(pl.lit(TECH_TYPES["networked"]))
-            # communal solution
+            # communal solution takes precedence
             .when(pl.col("assigned_tech").list.contains(TECH_TYPES["communal"]))
             .then(pl.lit(TECH_TYPES["communal"]))
-            # Could either be individual or networked HP: outdoor space is unknown for some properties in the building
+            # When outdoor space is unknown for some properties in the building
+            # At this point, some properties in the building might be assigned "individual", while others are assigned "individual_or_networked" if the outdoor space is missing
             .when(
                 pl.col("assigned_tech").list.contains(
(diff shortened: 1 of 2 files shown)


asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -280,41 +280,37 @@ def assign_df_unique_solution(solutions_per_footprint_df: pl.DataFrame) -> pl.Da
...
             .then(
-                pl.when(
-                    (pl.col("perc_properties_available_outdoor_space_data") >= 50)
-                    & (
-                        pl.col("median_contiguous_outdoor_space_area_m2")
-                        > OUTDOOR_SPACE_THRESHOLD_M2.get("outside_hn_zone")
-                    )  # Using your column name from previous step
-                )
+                # Properties with known outdoor space are guaranteed to exceed the threshold.
+                # Therefore, we only need to check that the area is not null/unknown.
```
crispy-wonton: question: for my understanding - isn't it the case that all properties going through this block will not have outdoor space because they can only be labelled 'individual_or_networked' if there's no outdoor space?

crispy-wonton (commented):
I don't think it's wrong, but I also don't fully understand the reasoning in your logic when you check for null outdoor space size - is that just for robustness or is it actually required?

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: I've just noticed that the docstring in 296/297 hasn't been updated so it's confusing. i've updated it now.


>  isn't it the case that all properties going through this block will not have outdoor space because they can only be labelled 'individual_or_networked' if there's no outdoor space?

No - at this point you might have a mixed of properties assigned 'individual' and others assigned "individual_or_networked". So what we're doing her is basically checking for whether they're all assigned "individual_or_networked" (in which case the outdoor space would be null) or if some are assigned "individual" (median would not be null)

sofiapinto:
> I don't think it's wrong, but I also don't fully understand the reasoning in your logic when you check for null outdoor space size - is that just for robustness or is it actually required?

It is actually required! I've added some info above, but let me bring it to life here.

Imagine a building with 10 properties: 
- 8 with known outdoor space and assigned individual solution at UPRN level
- 2 with unknown outdoor space

We check for whether the median outdoor space is unknown. Because it's not, we assign the building as individual.

Another example, a building with 10 properties but all of them with unknown outdoor space. We keep it as "individual_or_networked_hp"

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
crispy-wonton: Ah that makes sense. I didn't think through that there would be individual solutions at this stage but of course that makes sense! Thank you!

crispy-wonton (approved):
Thanks - this makes sense and the annotations are really useful!

PR #408, opened 2026-07-02
406 and 407: fix missing error from except block and enhance cluster id with LA slug

### LINKED ISSUE #406: Bug fix: wrong error captured whe loading OS open map layer
We need to capture `pyogrio.errors.DataSourceError` in the try/except bloc of`getters.load_geodata.load_gdf_os_openmap_layer()`

### LINKED ISSUE #407: Add prefix with area info to cluster ID

sofiapinto:
---

# Description

This PR:
- adds a prefix with LA slug to cluster ID. This is required by the front end, to avoid confusion between clusters across LAs (that previously had the same ID e.g. IND_101 in LA1 and LA2 should now be, respectively, IND_102_la1_slug and IND_102_la2_slug)
- Fixes a bug when loading OS open map data layers

Fixes #406 #407

# Instructions for Reviewer

You can check the code using the snippet below:

```

grid_square = 'SP' # you can check for 'SO' too for example
for layer in ['building',
        'car_charging_point',
        'electricity_transmission_line',
        'foreshore',
        'functional_site',
        'glasshouse',
        'greenspace_site',
        'important_building',
        'motorway_junction',
        'named_place',
        'railway_station',
        'railway_track',
        'railway_tunnel',
        'road',
        'road_tunnel',
        'roundabout',
        'surface_water_area',
        'surface_water_line',
        'tidal_boundary',
        'tidal_water',
        'woodland']:
    # checking the code runs
    load_geodata.load_gdf_os_openmap_layer(layer = layer, grid_square = grid_square)
```

No need to run anything to test the cluster_id, as I've run the pipeline and checked it's generating the correct output.

### DIFF
--- a/asf_heat_pump_suitability/getters/load_geodata.py
+++ b/asf_heat_pump_suitability/getters/load_geodata.py
@@ -4,6 +4,7 @@ import os
 import pandas as pd
 from typing import Optional, List
 import boto3
+import pyogrio
 import s3fs
 import shapely
 import logging
@@ -273,7 +274,7 @@ def load_gdf_os_openmap_layer(
             print(f"\nLoading OS OpenMap layer - {layer.title()} file: {file}")
             try:
                 gdfs.append(gpd.read_file(file, **kwargs))
-            except FileNotFoundError as e:
+            except (FileNotFoundError, pyogrio.errors.DataSourceError) as e:
                 # dealing with non-existing layers (e.g. no bodies of water in the grid square so no surface water area layer)
                 print(
                     f"Error loading OS OpenMap layer - {layer.title()} file {file}: {e}"
--- a/asf_heat_pump_suitability/pipeline/cluster/cluster.py
+++ b/asf_heat_pump_suitability/pipeline/cluster/cluster.py
@@ -67,6 +67,7 @@ def generate_gdf_clusters(
     polygon_overlay_gdf: gpd.GeoDataFrame,
     combined_anchor_gdf: gpd.GeoDataFrame,
     radius: float,
+    local_authorities_slug: str,
     id_col: str = "ID",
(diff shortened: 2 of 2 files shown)


asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -273,7 +274,7 @@ def load_gdf_os_openmap_layer(
             print(f"\nLoading OS OpenMap layer - {layer.title()} file: {file}")
             try:
                 gdfs.append(gpd.read_file(file, **kwargs))
-            except FileNotFoundError as e:
+            except (FileNotFoundError, pyogrio.errors.DataSourceError) as e:
```
helloaidank, line 277: This has overlap that's my doing, I hit this same bug whilst testing the GB grid square loading in #424 and fixed it before this was flagged (apologies, should've checked open PRs first). 

So we have two versions of the same except block, whichever merges second will conflict. Proposal for resolving this is to let #424 carry the fix, and this PR drops it. A couple reasons for this are that #424 version allows auth failures/corrupt files to crash loudly rather than being skipped and it also has some unit tests which pin this behaviour. 

Happy to do the reverse too (I drop mine and merge) if you'd rather keep #406 here, in that case I'd suggest adding the "does not exist" guard to this version.

helloaidank:
Proposed a resolution to the two separate solutions to same issue from this PR and #424 , the LA slug stuff looks good to merge!

asf_heat_pump_suitability/getters/load_geodata.py
sofiapinto: hey @helloaidank - as we discussed this morning, i'll merge this now and then you'll bring in your changes when you merge your PR next week.

helloaidank (approved):
Good to merge

PR #410, opened 2026-07-03
409 create temporary run_pipeline.sh

### LINKED ISSUE #409: Temp bash script to run pipelines

sofiapinto:
---

# Description

This is a temporary script to run all pipelines in sequence for a list of local authorities. We will use orbit future to automate this in a way that scales.

Fixes #409 

# Instructions for Reviewer

Can you please check there's enough documentation and that all scripts have been added and in the right order?

You can then narrow down the list of LAs to 2 and test the script works for you by doing:
```
chmod +x asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
./asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
```

### DIFF
--- a/asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
+++ b/asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+
+local_authorities=(
+    "plymouth"
+    "dudley"
+    "vale of glamorgan"
+    "midlothian"
+    "glasgow city"
+    "south lanarkshire"
+    "east lothian"
+)
+
+for la in "${local_authorities[@]}"; do
+    echo "=================================================="
+    echo "Starting pipeline for: ${la^^}"
+    echo "=================================================="
+
+    # Step 1: Domestic UPRNs
+    echo "--> Running: uprns.py"
+    python asf_heat_pump_suitability/pipeline/transform/uprns.py --local_authorities "$la" --save
+    if [ $? -ne 0 ]; then echo "Error in uprns.py for $la. Skipping..."; continue; fi
+
+    # Step 2: Add features
+    echo "--> Running: add_features.py"
+    python asf_heat_pump_suitability/pipeline/run/add_features.py --local_authorities "$la" --save
+    if [ $? -ne 0 ]; then echo "Error in add_features.py for $la. Skipping..."; continue; fi
+
+    # Step 3: Decision tree
+    echo "--> Running: decision_tree.py"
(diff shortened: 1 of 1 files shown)


asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+# This script runs the entire pipeline for a list of local authorities. It executes each step in sequence and checks for errors after each step.
+# To make the script executable, run: chmod +x asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
+# You can then run the script with: ./asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
```
helloaidank, line 4: All the `python asf_heat_pump_suitability/...` calls below are relative to the repo root, so would be nice if we could run the script from any other directory (e.g. from inside `pipeline/run/`) `. Making the script `cd` to the repo root first means it works from anywhere:

```suggestion
# You can then run the script with: ./asf_heat_pump_suitability/pipeline/run/run_pipeline.sh

# Run from the repo root regardless of where the script is invoked from
cd "$(dirname "$0")/../../.." || exit 1
```

asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
```diff
@@ -0,0 +1,60 @@
...
+    # Step 5: Compute contextual features
+    echo "--> Running: compute_contextual_features.py"
+    python asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py --local_authorities "$la" --save
+    if [ $? -ne 0 ]; then echo "Error in compute_contextual_features.py for $la. Skipping..."; continue; fi
+
+    echo "Successfully finished pipeline for: $la"
+    echo ""
+done
+
+echo "All local authorities data generated!"
```
helloaidank: This banner prints unconditionally, so it announces success even when every LA failed and was skipped via `continue` (which is exactly what happened on a bad run today — all 7 LAs errored, then "All local authorities data generated!" printed). Counting successes keeps the summary honest (`${succeeded:-0}` defaults to 0 so no initialisation is needed before the loop):

```suggestion
    echo "Successfully finished pipeline for: $la"
    succeeded=$((succeeded+1))
    echo ""
done

echo "Pipeline completed for ${succeeded:-0} of ${#local_authorities[@]} local authorities."
```

helloaidank:
Just a couple of suggestions on some small improvements, I was able to get the script to run and it seemed to generate the output no problem. Happy to merge otherwise!

helloaidank (dismissed):
Good to merge!

helloaidank (approved):

PR #413, opened 2026-07-07
Version output datasets by release date (#412)

### LINKED ISSUE #412: Version output datasets with {LA_slug}/{YYYYMMDD} release directories
Pipeline outputs on S3 (six per-LA datasets under `outputs/data/`) are currently written to fixed paths and silently overwritten on every `--save` run, so releases can't be pinned, compared, or rolled back.

Decided approach:

- Outputs are written under dated release directories laid out as `{LA_slug}/{YYYYMMDD}`, e.g. `vale_of_glamorgan/20260708/`.
- All six datasets for an LA share one release directory (stages read/write the same date).
- Release date defaults to the run date, with an explicit `--release_date` override so multi-day runs land in one release.
- The current undated paths keep being written as "latest" until downstream dependencies (frontend geojson copy) are confirmed.
- No S3 bucket versioning.

Full briefing: `specs/1216293268467175_versionOutputDatasets.md` (added on the implementation branch).

helloaidank:
## Description

Every `--save` pipeline run wrote to fixed, undated per-LA output paths, so each run silently overwrote the previous outputs. There was no way to pin a release, compare two runs, or roll back to an earlier one.

This PR versions the six S3 output datasets into dated release directories, keyed on a release date that is resolved once per run and threaded through every stage.

## Related Issue(s)

- Closes #412

## Changes Made

- Output path templates in `config/base.yaml` gain a `{release_date}` (YYYYMMDD) segment:

      s3://asf-local-heat-planning-tool/outputs/data/{local_authority}/{release_date}/{local_authority}_domestic_uprns.parquet

- A `--release_date` CLI arg is threaded through the five pipeline stages (`uprns.py`, `add_features.py`, `decision_tree.py`, `cluster.py`, `compute_contextual_features.py`) for both inter-stage reads and writes, so a multi-day run pins to a single release directory. The date is also stamped into the final geojson metadata.
- The existing undated files on S3 are deliberately left untouched. We can pin a dated release.

### What to Focus On

- **Known impact:** four research scripts still format the now dated templates without `release_date` and will raise `KeyError` if run:
  - `research/exploratory/domestic_filtering/domestic_filtering.py:47`
  - `research/exploratory/domestic_filtering/council_tax_domestic_filtering.py:29`
  - `research/exploratory/city_centre_identification/20251205_identify_city_centre_spatial_signatures.py:44`
  - `research/exploratory/decision_tree_categorisation/applying_decision_tree.py:122,160`

  Deliberately not fixed here, they read the old undated Plymouth outputs, which are now frozen, so repointing them at dated paths is a separate decision.
- **Question:** `decision_tree.py` saves via raw `gdf.to_parquet()` while every other stage routes through `save_utils.save_to_s3` (which has handled GeoDataFrames since 2026-03-05, before these calls were written...parallel work, not a capability gap). Should it switch to `save_to_s3` for the logging chokepoint?

### How to Test

- Run the five stages in order for a small LA, pinning the date, e.g.:

      python asf_heat_pump_suitability/pipeline/transform/uprns.py --local_authorities "east lothian" --save --release_date 20260708

  (then `add_features.py`, `decision_tree.py`, `cluster.py`, `compute_contextual_features.py` with the same flags). Omitting `--release_date` defaults to today's date.
- Verify all six datasets land under one directory: `aws s3 ls s3://asf-local-heat-planning-tool/outputs/data/east_lothian/20260708/`
- Verify the undated files one level up are untouched (unchanged timestamps).
- Verify the guard: run `add_features.py` with `--release_date` for a date with no release...it should fail immediately with a `FileNotFoundError` suggesting `--release_date` for an existing release.

## Checklist

- [x] Code has been tested locally (East Lothian end-to-end run; guard tested with a non-existent release date)
- [x] No local file dependencies (all inputs/outputs on S3)
- [x] Changes are reasonably scoped (one issue: output dataset versioning)

## Known follow-ups (deliberately out of scope)

- #426 — five call sites in research/exploratory format the output templates directly and now raise KeyError; fix depends on the release-lifecycle discussion, so deferred.
- Existing pre-#413 outputs live at the old un-dated S3 paths; a one-off backfill sync into a dated directory is needed on merge day so downstream stages can read them.


asf_heat_pump_suitability/utils/tests/test_save_utils.py
```diff
@@ -0,0 +1,49 @@
+"""
+Tests for asf_heat_pump_suitability.utils.save_utils and the dated release
```
helloaidank, line 2: wondering what our current approach to testing is?

asf_heat_pump_suitability/config/base.yaml
sofiapinto, line 1: Looks great, thank you!

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -12,6 +12,9 @@
 Defaults to `GB` (all of Great Britain), but this is not yet implemented.
 
 Set --save to save the outputs to S3. By default, outputs are not saved.
+
+Set --release_date to specify the YYYYMMDD dated release directory to save outputs to.
+Defaults to today's date.
```
sofiapinto: This is the first step in our pipeline. Why did you add this as a param, rather than using today's date, i.e. the date of running the pipeline?

For all the remaining scripts, we could just get the latest version of the output from the previous pipeline? I.e. have a function that checks for the latest version on s3.


I guess my question is: was there any reason for adding as a param, rather than automating it in the code?

asf_heat_pump_suitability/utils/save_utils.py
```diff
@@ -4,9 +4,71 @@
...
+    try:
+        parsed_date = datetime.strptime(release_date, date_format)
+    except ValueError:
+        raise ValueError(
+            f"release_date must be a valid date in YYYYMMDD format, got '{release_date}'."
+        )
+    return parsed_date.strftime(date_format)
+
+
+def get_str_output_path(
```
sofiapinto, line 48: Nice!!

docs/specs/412_versionOutputDatasets.md
sofiapinto, line 1: i'm guessing this is something we would delete before merging?

asf_heat_pump_suitability/utils/tests/test_save_utils.py
sofiapinto: Something to discuss with @crispy-wonton. She has decided to create subfolders for tests rather than an overall tests folder. I typically prefer the overall tests folder, but please check in with her on which approach we're following.

tests/test_save_utils.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def test_get_str_release_date_returns_valid_date_unchanged():
+    """A valid YYYYMMDD date string is returned as-is."""
+    assert save_utils.get_str_release_date("20260708") == "20260708"
+
+
+@pytest.mark.parametrize(
+    "invalid_release_date",
+    ["2026-07-08", "8 July 2026", "202607", "20261332", ""],
```
sofiapinto: you might want to add a few like: "08072026", "080726"

asf_heat_pump_suitability/utils/tests/test_save_utils.py
```diff
@@ -0,0 +1,49 @@
...
+        save_utils.get_str_release_date(invalid_release_date)
+
+
+def test_output_dataset_templates_format_to_dated_release_paths():
+    """Every output dataset path template resolves to a {la_slug}/{YYYYMMDD}/ release directory."""
+    release_date = "20260708"
+    for name, template in config["output"]["dataset"].items():
+        path = template.format(
+            local_authority="plymouth",
+            local_authorities="plymouth",
```
sofiapinto, line 45: Would be good to open a separate issue to sort out having local_authority vs local_authorities in different scripts.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -9,6 +9,9 @@
 Required args:
 --local_authorities to specify which local authority / authorities to run the script for
 --save - Set to save output GeoDataFrame to S3.
+
+Set --release_date to specify the YYYYMMDD dated release directory to read inputs from and
+save outputs to. Defaults to today's date.
```
sofiapinto: In all scrips:

```suggestion
save outputs to. Defaults to running the pipeline using today's date.
```

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -305,6 +315,7 @@ def create_json_contextual_features_metadata(
     )
     metadata = {
         "Data file date of creation": datetime.now().strftime("%Y-%m-%d"),
+        "Release date": release_date,
```
sofiapinto: These are the same, so you can remove the one above I had oririginally created.

sofiapinto (commented):
Hey @helloaidank , 

Thanks for getting this ready so quickly! I haven't yet run the scripts, but i'll leave you with some thoughts in the meantime and will then post a separate review when the scripts finish running.

My main question for you is why set release_date as a param, rather than always using today's date for uprns.py (the first pipeline) and then for all other pipelines, having a function that automatically retrieves the latest version of the outputs required for that pipeline to run?

asf_heat_pump_suitability/utils/tests/test_save_utils.py
helloaidank: Ok sounds good! Will put it as an agenda item for a stand-up.

asf_heat_pump_suitability/pipeline/transform/uprns.py
helloaidank: Some thoughts around this: 

So the design is that we resolve the date once and thread it through the pipeline, so for example if it starts Monday and finishes on a Wednesday, we still assemble one coherent release under 20260708. Omitting —release_date defaults to today so a single sitting run is possible. But we could imagine this might crash if for whatever reason we run the pipeline overnight or over multiple dates.   

In terms of auto-latest, I think the risk with it is in regards to failed runs as s3 has no indicator around if the data has been written out correctly or just partially. We could make this safe with a complete marker e.g. some file to indicate a successful run or a per release manifest written after the final stage (this would indicate which ‘latest’ to filter on). Just means some additional code to implement this. The other risk is around picking up someone elses run, so might involve an ID.

docs/specs/412_versionOutputDatasets.md
helloaidank: I think this might be another agenda item to discuss in a stand-up, this is the spec alongside some of the decision logs (interaction between myself and agent). We might want to keep a record of this for transparency and context.

asf_heat_pump_suitability/pipeline/transform/uprns.py
helloaidank: Thinking about this at a higher level, I think there might be some tension between having a dev outputs and outputs we're planning on putting into production. When we are making a methodological change and doing development, we might want to have these separate to the outputs meant for production.

tests/test_save_utils.py
helloaidank: Makes sense, will add a couple more

asf_heat_pump_suitability/utils/tests/test_save_utils.py
helloaidank: Agreed to follow the subfolder layout — moved to `asf_heat_pump_suitability/utils/tests/test_save_utils.py` in 7b84de9, matching e.g. `pipeline/cluster/tests/`.

tests/test_save_utils.py
helloaidank: Added both in 42bb4e3 and good call: `"080726"` caught a real bug. `strptime` leniently parsed it as year 807 / Feb 6 and the validator silently normalised it to `"08070206"`, i.e. a wrong release directory. Now rejected via a strict round-trip check (`parsed.strftime() == input`).

asf_heat_pump_suitability/pipeline/cluster/cluster.py
helloaidank: Applied to the module docstrings of all five pipeline scripts in a15f263.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
helloaidank: Removed in 08a6b36 — `Release date` is now the single date field in the geojson metadata.

asf_heat_pump_suitability/utils/tests/test_save_utils.py
helloaidank: Opened #425 for this.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
helloaidank: One follow-up before we resolve: this changes the metadata contract of the geojson the front end consumes...the key `Data file date of creation` (format `2026-07-14`) is gone, replaced by `Release date` (format `20260714`). Can you confirm the front end doesn't read the old key anywhere (and is happy with the undashed date format if it displays it)? If it does, we should coordinate the front-end change with this PR merging.

helloaidank:
Thanks for the review @sofiapinto !

Might be more changes with regards to the auto-detection of the date (e.g. create a separate `manifest.json` for the intermediary and final outputs) but this might be worth tackling in a separate issue as this PR is already pretty meaty. 

Summary of some of the changes I've made.

Responses to review / hardening `release_date` handling:
- `save_utils` path builder now requires an explicit `release_date` instead of defaulting internally ([`9987824`](https://github.com/nestauk/asf_heat_pump_suitability/commit/9987824)). Why do this? If each helper defaults to "today", every call resolves today's date independently and a run that crosses midnight would silently split one release across two dated directories. 
- Test you suggested failed, so made the criteria stricter in terms of formatting ([`42bb4e3`](https://github.com/nestauk/asf_heat_pump_suitability/commit/42bb4e3)).
- Made some clarification to some doc strings so it's clearer that multi-day runs should pin to one date (subject to change depending on your response to the comment) ([`a15f263`](https://github.com/nestauk/asf_heat_pump_suitability/commit/a15f263), [`fa1d787`](https://github.com/nestauk/asf_heat_pump_suitability/commit/fa1d787)).

Merging with dev:
- merged some conflicts in relation to `compute_contextual_features.py` and `create_manifest.py` ([`99556dc`](https://github.com/nestauk/asf_heat_pump_suitability/commit/99556dc), [`ea79d7d`](https://github.com/nestauk/asf_heat_pump_suitability/commit/ea79d7d))

Small tidy-ups
- dropped the creation date from geojson metadata, the release directory now carries that information, keeping a second independently stamped date doesn't make sense ([`08a6b36`](https://github.com/nestauk/asf_heat_pump_suitability/commit/08a6b36)).
- moved `save_utils` test into the sub-folder ([`7b84de9`](https://github.com/nestauk/asf_heat_pump_suitability/commit/7b84de9))

Spec/docs housekeeping

- Changed filename convention to be more interpretable and link to the github issue, with initial section now linking to the issue, PR and the Asana task ([`48bde4f`](https://github.com/nestauk/asf_heat_pump_suitability/commit/48bde4f)).
- moved specs under docs/specs, docs/decisions/ is for permanent decision records (will explain in separate slack message) ([`d8a2c25`](https://github.com/nestauk/asf_heat_pump_suitability/commit/d8a2c25)).

asf_heat_pump_suitability/pipeline/transform/uprns.py
sofiapinto: > Thinking about this at a higher level, I think there might be some tension between having a dev outputs and outputs we're planning on putting into production. When we are making a methodological change and doing development, we might want to have these separate to the outputs meant for production.

Could you please elaborate on this? 

Did you mean that we should separate **methodological changes** from other **development changes**? Or did you mean that we should separate dev outputs from production outputs? 

Or something different?

The production outputs live in a separate S3 bucket and folder, the production outputs are not replaced when running these pipelines.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
sofiapinto: That's fine. That's something Simon currently prefers as a written update. I summarise all changes ahead of sharing new data with him. We don't regenerate data after each branch is merged to dev.

asf_heat_pump_suitability/utils/save_utils.py
sofiapinto: It says it returns a release dade in YYYYMMDD format, but that will actually depend on `config["constant"]["release_date_format"]`

docs/specs/412_versionOutputDatasets.md
sofiapinto: We've decided to keep this for now, but just to say that I can't imagine myself going back through these. As we've discussed in one of the past sprint reviews, having one file generated from automatically scanning closed PRs might prove more useful in future. No action for now.

sofiapinto (commented):
Hey @helloaidank ,

Thank you for the changes!

I have left replies to your original questions and some additional questions throughout.

> save_utils path builder now requires an explicit release_date instead of defaulting internally ([9987824](https://github.com/nestauk/asf_heat_pump_suitability/commit/9987824)). Why do this? If each helper defaults to "today", every call resolves today's date independently and a run that crosses midnight would silently split one release across two dated directories.

One clarification on this. If now we run each of the scripts individually (uprns.py, add_features.py, etc) without passing a date, it will still default to using today's date am I right? So if we run add_features.py with today's date, but there's no file generated from uprns.py with today's date, then add_features.py will error correct?)

---

Separately, we should probably open an issue to add release date as a param to the bash file that runs all pipelines in order.


Otherwise happy for this to be merged.

asf_heat_pump_suitability/pipeline/transform/uprns.py
helloaidank: Yes, I just didn't have the clarity around production vs dev pipelines that we just discussed in the stand-up here. I think potentially it could be nice to have a --prod flag (maybe) that writes out to the separate S3 bucket and folder!

asf_heat_pump_suitability/utils/save_utils.py
helloaidank: Yep, that's right! fixed in 78b4258. The docstring now says `config["constant"]["release_date_format"]`, and the `ValueError` message renders a live example date from the config (e.g. `'20260730'`) instead of hard-coding YYYYMMDD.

helloaidank:
> If now we run each of the scripts individually (uprns.py, add_features.py, etc) without passing a date, it will still default to using today's date am I right? So if we run add_features.py with today's date, but there's no file generated from uprns.py with today's date, then add_features.py will error correct?

Yep, that's right. Each script resolves its own date when `--release_date` is omitted, so `add_features.py` run on a later day than `uprns.py` looks under today's release directory, finds nothing, and fails immediately with a `FileNotFoundError`. So any run over multiple days (or crossing midnight) must pin an explicit `--release_date` on every stage.

> Separately, we should probably open an issue to add release date as a param to the bash file that runs all pipelines in order.

Agreed, opened #460: `run_pipeline.sh` will resolve the date once at the top (optional arg, defaulting to today) and pass the same `--release_date` to every stage.

asf_heat_pump_suitability/pipeline/transform/uprns.py
sofiapinto: For now, we want to keep this last step as manual after checks. Once development stabilises we can move to something more automated.

sofiapinto (approved):

PR #418, opened 2026-07-09
417 simplify polygons while preserving topology

### LINKED ISSUE #417: Reduce number of coordinates in cluster polygons

sofiapinto:
---

# Description

This PR includes adds 5m simplification to the clusters while preserving their typology, ahead of saving the final output.

Fixes #417 

# Instructions for Reviewer

Can you please check the logic holds?

### DIFF
--- a/asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
+++ b/asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
@@ -284,7 +284,7 @@ def create_gdf_contextual_features(
 
     return gpd.GeoDataFrame(
         clusters_with_contextual_features_gdf, geometry="geometry", crs="EPSG:27700"
-    ).to_crs(epsg=4326)
+    )
 
 
 def create_json_contextual_features_metadata(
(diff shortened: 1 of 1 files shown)


asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -378,6 +378,18 @@ def create_json_contextual_features_metadata(
         uprns_df=uprns_df, clusters_gdf=clusters_gdf
     )
 
+    print("Simplifying geometries using tolerance_m...")
+    clusters_with_contextual_features_gdf["geometry"] = (
+        clusters_with_contextual_features_gdf["geometry"].simplify(
+            tolerance=tolerance_m, preserve_topology=True
```
crispy-wonton, line 384: fyi: `simplify` defaults to preserving topology so this is the same as the simplification we used before :) just flagging to say we don't expect this to change results compared to previous simplification.

crispy-wonton (approved):

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
sofiapinto: agreed! but that code somehow just disappeared.

PR #422, opened 2026-07-09
415 postcode analysis for individual solutions

### LINKED ISSUE #415: UPRN to postcode mapping analysis for a set of LAs

sofiapinto:
---

# Description

This PR adds an analysis script to:
- **create a list of postcodes "less suitable" for individual solutions for specific LAs**: the assumption being that any clusters identified as networked HP, communal or DHN are less suitable for individual solutions.
- **identify all postcodes in each area including a summary of information for each**:
   - number of addresses in the postcode (both domestic and overall)
   - the average outdoor space on each address
   - distribution of tenure (we're showing owner occupiers and unknown info)
   - a tag of whether the postcode overall is most suitable/less suitable for individual solutions (if at least one property within a postcode is suitable for individual then the postcode is overall suitable of individual - we'd rather have false positives, than false negatives)

Fixes #415 

# Instructions for Reviewer

hey @crispy-wonton ,

You can check if the code runs for you with 
```
python -i asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py --local_authority plymouth
```

Can you please check the logic holds and that I made no mistake coding it?

### DIFF
--- a/asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
+++ b/asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
@@ -41,7 +41,6 @@ def parse_arguments() -> argparse.Namespace:
     return parser.parse_args()
 
 
-# Lowercase keys to avoid case-sensitivity bugs with multi-word inputs
 la_code_mapping = {
     "babergh": "E07000200",
     "wiltshire": "E06000054",
@@ -76,11 +75,14 @@ if __name__ == "__main__":
         )
     )
 
+    # Load buildings geodataframe for the local authority
+    print(f"Loading {local_authority} buildings geodataframe...")
     buildings_gdf = load_geodata.load_gdf_os_openmap_layer(
         layer="building", grid_squares=local_authority_dict["grid_squares"]
     )
 
-    print("Loading clusters...")
+    # Load clusters geodataframe for the local authority
+    print(f"Loading {local_authority} clusters ...")
     clusters_gdf = gpd.read_parquet(
         config["output"]["dataset"]["tech_clusters"].format(
             local_authorities=local_authority_dict["url_slug"],
@@ -88,17 +90,16 @@ if __name__ == "__main__":
         ),
     ).to_crs(epsg=27700)
 
+    # Map UPRNs to clusters
+    print(f"Mapping {local_authority} UPRNs to clusters...")
     uprns_df = cluster.map_df_uprns_to_clusters(
         uprns_df=uprns_df, buildings_gdf=buildings_gdf, clusters_gdf=clusters_gdf
     )
 
-    print(len(uprns_df))
-    # Drop cluster_ids starting with DESNZ_HNZ
+    # Dropping UPRNs in HN zone
     uprns_df = uprns_df.filter(~pl.col("cluster_id").str.starts_with("DESNZ_HNZ"))
-    print(len(uprns_df))
 
-    # Map each UPRN to its postcode
-    print("Mapping UPRNs to postcodes...")
+    print("Loading UPRN to postcode mapping from S3...")
     s3_client = boto3.client("s3")
 
     path = config["data"]["geodata"]["gb_uprn_country_mapping"]
@@ -121,7 +122,7 @@ if __name__ == "__main__":
         pl.col("lad25cd") == la_code_mapping[local_authority]
     )
 
-    # Merge the UPRN to country mapping with the UPRNs DataFrame
+    # Extend uprns_df with postcode information
     uprns_df = uprns_df.join(
         uprn_to_postcode_df.select(["UPRN", "postcode"]),
         on="UPRN",
@@ -133,12 +134,12 @@ if __name__ == "__main__":
         pl.col("UPRN").count().over("postcode").alias("n_domestic_uprns_in_postcode")
     )
 
-    # Number of UPRNs per postcode overall
+    # Aggregate UPRN counts per postcode
     uprns_per_postcode_df = uprn_to_postcode_df.group_by("postcode").agg(
         pl.col("UPRN").count().alias("n_uprns_in_postcode")
     )
 
-    # Merge overall UPRN counts
+    # Extend uprns_df with overall UPRN counts per postcode
     uprns_df = uprns_df.join(
         uprns_per_postcode_df.select(["postcode", "n_uprns_in_postcode"]),
         on="postcode",
@@ -146,12 +147,14 @@ if __name__ == "__main__":
     )
 
     # Identify suitability flags
+    # Suitable for individual solutions if cluster_id starts with "IND"
     uprns_df = uprns_df.with_columns(
         pl.col("cluster_id")
         .str.starts_with("IND")
         .alias("suitable_for_individual_solutions")
     )
 
+    # Less suitable for individual solutions if cluster_id starts with "NHP", "COM", or "DESNZ"
     uprns_df = uprns_df.with_columns(
         (
             pl.col("cluster_id").str.starts_with("NHP")
--- a/asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
+++ b/asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
@@ -187,7 +187,7 @@ if __name__ == "__main__":
         )
     )
 
-    # Group and generate summary DataFrame
+    # Group and generate summary df
     summary_per_postcode_df = (
         uprns_df.group_by("postcode")
         .agg(
@@ -211,7 +211,7 @@ if __name__ == "__main__":
         )
         .with_columns(
             [
-                # Percent UPRNs in owner-occupied tenure (with safety for division-by-zero)
+                # Percent UPRNs in owner-occupied tenure
                 (
                     pl.col("n_owner-occupied")
                     / pl.col("n_domestic_uprns_in_postcode")
@@ -223,7 +223,7 @@ if __name__ == "__main__":
                     / pl.col("n_domestic_uprns_in_postcode")
                     * 100
                 ).alias("percent_unknown_tenure"),
-                # Properly aliased boolean suitability column
+                # Postcode suitability for individual solutions: True if at least one UPRN is suitable for individual solutions, False otherwise
                 (
                     pl.col("one_uprn_or_more_suitable_for_individual_solutions")
                     & ~pl.col("one_uprn_or_more_less_suitable_for_individual_solutions")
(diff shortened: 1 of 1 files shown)


asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
```diff
@@ -0,0 +1,248 @@
...
+    "east suffolk": "E07000244",
+    "ipswich": "E07000202",
+    "mid suffolk": "E07000203",
+    "west suffolk": "E07000245",
+    "breckland": "E07000143",
+    "west berkshire": "E06000037",
+    "somerset": "E06000066",
+    "dorset": "E06000059",
+    "plymouth": "E06000026",
+}
```
crispy-wonton, line 56: note: I haven't manually checked this mapping is correct. Let me know if you want me to do that.

asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
```diff
@@ -0,0 +1,248 @@
...
+    buildings_gdf = load_geodata.load_gdf_os_openmap_layer(
+        layer="building", grid_squares=local_authority_dict["grid_squares"]
+    )
+
+    # Load clusters geodataframe for the local authority
+    print(f"Loading {local_authority} clusters ...")
+    clusters_gdf = gpd.read_parquet(
+        config["output"]["dataset"]["tech_clusters"].format(
+            local_authorities=local_authority_dict["url_slug"],
+            tolerance_m=tolerance_m,
```
crispy-wonton: question: I don't think these have a tolerance? They shouldn't be simplified at this point.

asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
```diff
@@ -0,0 +1,248 @@
...
+
+    # Extend uprns_df with overall UPRN counts per postcode
+    uprns_df = uprns_df.join(
+        uprns_per_postcode_df.select(["postcode", "n_uprns_in_postcode"]),
+        on="postcode",
+        how="left",
+    )
+
+    # Identify suitability flags at UPRN level based on cluster_id prefixes
+    # 1) Suitable for individual solutions if cluster_id starts with "IND"
```
crispy-wonton, line 149: question: what's the value if the outdoor space is unknown and it's assigned individual or (heat) networked?

Also, by the logic below, it doesn't look like these properties are included - do they just remain as null values?

asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
```diff
@@ -0,0 +1,248 @@
...
+
+    # Group and generate summary df
+    summary_per_postcode_df = (
+        uprns_df.group_by("postcode")
+        .agg(
+            [
+                pl.col("n_domestic_uprns_in_postcode")
+                .first()
+                .alias("n_domestic_uprns_in_postcode"),
+                pl.col("n_uprns_in_postcode").first().alias("n_uprns_in_postcode"),
```
crispy-wonton: i think you possibly don't need the `alias()` here as it will default to the original name but not 100% sure. Anyway it doesnt matter but just for future reference.

asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
```diff
@@ -0,0 +1,248 @@
...
+                pl.col("suitable_for_individual_solutions")
+                .any()
+                .alias("one_uprn_or_more_suitable_for_individual_solutions"),
+                pl.col("less_suitable_for_individual_solutions")
+                .any()
+                .alias("one_uprn_or_more_less_suitable_for_individual_solutions"),
```
crispy-wonton, line 201: suggestion: it might be useful here to add the count and proportion of UPRNs less suitable for individual solutions here as well

asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
```diff
@@ -0,0 +1,248 @@
...
+            ]
+        )
+        .with_columns(
+            [
+                # Percent UPRNs in owner-occupied tenure
+                (
+                    pl.col("n_owner-occupied")
+                    / pl.col("n_domestic_uprns_in_postcode")
+                    * 100
+                ).alias("percent_owner_occupied"),
```
crispy-wonton:
```suggestion
                ).alias("percent_domestic_owner_occupied"),
```

asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
```diff
@@ -0,0 +1,248 @@
...
+                # Percent UPRNs with unknown tenure
+                (
+                    pl.col("n_unknown_tenure")
+                    / pl.col("n_domestic_uprns_in_postcode")
+                    * 100
+                ).alias("percent_unknown_tenure"),
```
crispy-wonton, line 240:
```suggestion
                ).alias("percent_domestic_unknown_tenure"),
```

asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
```diff
@@ -0,0 +1,248 @@
...
+                # Postcode suitability for individual solutions: True if at least one UPRN is suitable for individual solutions, False otherwise
+                (
+                    pl.col("one_uprn_or_more_suitable_for_individual_solutions")
+                    & ~pl.col("one_uprn_or_more_less_suitable_for_individual_solutions")
```
crispy-wonton: bug: i think this is the wrong way round

crispy-wonton (changes requested):
I think there is one bug at the end of the script. Otherwise, I didn't really find issues with the logic.

One thing that I think it slightly confusing, is including the count of all UPRNs as well as domestic ones. It makes the other numbers more confusing to interpret as it's then unclear if they are referring to all UPRNs or just domestic ones. So I would remove that unless it's been requested.

I'm currently running the script for Plymouth so will leave a comment about how that went separately.

asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
sofiapinto: yes, please :)

crispy-wonton: checked - all correct.

asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
sofiapinto: Ah! fair, I forgot about these. Will add those two new categories in. We should count the grey areas/areas of unknown outdoor space as individual.

asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
sofiapinto: you're right!

asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
sofiapinto: Good catch!!!

asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
sofiapinto: Actually I didn't forget about them because the grey areas cluster IDs also start with IND. They're either: IND_or_NHP or IND_or_DHNZ

crispy-wonton: So for this analysis, you want them to be considered as individual solutions?

crispy-wonton (dismissed):

crispy-wonton:
Approved - also, the script ran through for Plymouth successfully.

asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
```diff
@@ -265,10 +267,34 @@ def parse_arguments() -> argparse.Namespace:
...
-    assert postcodes_less_suitable_individual_df.frame_equal(
-        postcodes_less_suitable_individual_df_check
-    ), "Postcodes in postcodes_less_suitable_individual_df do not match postcodes in summary_per_postcode_df where postcode_less_suitable_for_individual_solutions is True"
+    postcodes_less_suitable_individual_set_check = set(
+        summary_per_postcode_df.filter(
+            pl.col("postcode_less_suitable_for_individual_solutions")
+        )
+        .select("postcode")
+        .to_series()
+    )
```
crispy-wonton, line 274: It's not problematic and I wouldn't change it but here and below you should just be able to call the series like this:
```
set(
        summary_per_postcode_df.filter(
            pl.col("postcode_less_suitable_for_individual_solutions")
        )["postcode"]
    )
```

crispy-wonton (dismissed):
Logic all looks correct.

asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
sofiapinto: I think we discussed this yesterday, but just to have written record: yes, because there's still potential that they are suitable for individual. And we'd rather have false positives than false negatives.

asf_heat_pump_suitability/research/analysis/postcodes_individual_solutions/postcodes_individual_solutions.py
sofiapinto: good shout!

i've replaced with:

```
set(
    summary_per_postcode_df
    .filter(pl.col("postcode_less_suitable_for_individual_solutions"))
    .get_column("postcode")
)
```
which is more idiomatic and has performance benefits.

crispy-wonton (approved):

Issue #423, opened 2026-07-09
Grid-square getters crash when a layer is missing from ALL of an LA's squares (e.g. tidal_water for inland LAs)


helloaidank:
## Problem

The missing-file skip added in #416 handles squares that lack a layer, but if **every** square an LA touches lacks it, `gdfs` stays empty and `pd.concat([])` raises `ValueError: No objects to concatenate` (`getters/load_geodata.py`, `load_gdf_os_openmap_layer` and `load_gdf_os_openroad`).

This is common, not exotic: `cluster.py` unconditionally requests `tidal_water` per LA, and OS only ships a TidalWater file for squares containing tidal water. Verified: **Oxford, Milton Keynes and West Northamptonshire are each entirely within square SP, which has no TidalWater/Foreshore files** — clustering any of them crashes. (Leicester survives only because its second square, SK, clips the Humber estuary.)

The same opaque error also masks a misconfigured `base.yaml` path (every read 404s → all skipped → same ValueError).

## Proposed fix (not implemented in #416's PR)

In both getters, guard the empty case and return a truthful empty result:

```python
if not gdfs:
    logging.warning(f"No {layer} files found for grid squares {grid_squares}. Returning empty GeoDataFrame.")
    return gpd.GeoDataFrame(geometry=[], crs=27700)
```

- Semantically correct: an inland LA has zero tidal-water barriers; `cluster.py` concatenates barrier layers, so an empty frame contributes nothing.
- Also avoids a secondary KeyError: the empty frame would otherwise hit `drop_duplicates(subset=[id_col, ...])` on a column it doesn't have.
- Keep the per-file skip messages as `logging.warning` so a typo'd config path still produces a loud wall of warnings rather than silence.

While in there, consider extracting the duplicated skip logic (two divergent except shapes in the two getters) into one shared helper, and hardening the `"does not exist"` GDAL-message substring with an `s3fs.exists()` check on the failed path only.

Issue #421, opened 2026-07-09
Domestic UPRN filtering over-counts in dense mixed-use LAs (City of London fails census validation)


helloaidank:
## Finding

Running `pipeline/transform/uprns.py` for City of London + Orkney Islands (now possible with GB-wide grid squares, #416) fails the pandera census validation in `schemas/uprns_schema.py`:

- Expected (0.95–1.4 × census households): 14,738–21,719
- Actual: 28,958 (~1.87×), failure examples all City of London (E09000001)

A Plymouth control run on the same new data passes comfortably (147,526 vs range 108,830–160,381), so this is not a data or config issue — the residential-filtering heuristics (UPRN inside a building footprint not classified non-residential, plus EPC registers) appear to over-count homes in dense mixed-use areas like the Square Mile, which none of the 6 sampling LAs resemble.

## Suggested next steps

- Split the run per LA to confirm Orkney alone passes.
- Inspect a sample of the excess City of London UPRNs against the non-domestic EPC register / POI categories to see which filter is letting offices through.
- Consider whether the census row-count check should run per LA rather than on the combined dataframe (a passing LA can currently mask a failing one, and vice versa).

helloaidank:
Per-LA split run confirms the attribution:

- **Orkney Islands alone: passes** — 14,192 UPRNs vs expected 10,070–14,839 (1.34×, near the ceiling but inside).
- **City of London (implied from combined run): ~14,766 UPRNs vs ~4,915 census households (~3×)** — accounts for the entire combined-run violation.

So the over-counting is isolated to City of London / dense mixed-use territory; remote-sparse LAs validate fine. Also note Orkney sitting at 1.34× suggests LAs with high second-home rates may brush the 1.4 ceiling — worth keeping in mind if the per-LA check is implemented.

Issue #420, opened 2026-07-09
Clean up superseded OS grid-square input prefixes on S3


helloaidank:
## Context

#416 repointed the grid-square input config at new dated prefixes covering all GB squares. The superseded prefixes remain on S3 as rollback targets and should be removed once the new paths have been in use for a while:

- `s3://asf-local-heat-planning-tool/inputs/geodata/v202510_OSOpenMapLocal_geometries_selected/` (15 squares)
- `s3://asf-local-heat-planning-tool/inputs/geodata/v202510_OSOpenMapGreenspace_geometries_selected/` (15 squares)
- `s3://asf-local-heat-planning-tool/inputs/geodata/oproad_essh_gb/20260507/` (old flat road files)
- `s3://asf-local-heat-planning-tool/inputs/geodata/opgrsp_essh_gb/20260708/` (GB-wide greenspace shapefile, superseded by per-square `20260709/`)

## Before deleting

Two research scripts hard-code the `v202510_OSOpenMapLocal_geometries_selected/SX/` path and need repointing first:

- `research/exploratory/block_of_flats_classification_model/sample_buildings_for_labelling.py`
- `research/exploratory/block_of_flats_classification_model/build_classification_model_block_of_flats.py`

Issue #419, opened 2026-07-09
Add script to stream OS open data grid squares from the OS Downloads API to S3


helloaidank:
## Motivation

The per-grid-square OS inputs (OpenMap Local, Open Roads, Open Greenspace) are currently refreshed by downloading tiles from the OS portal and uploading via the S3 console. This is error-prone: the last manual refresh silently dropped 4 files, and the greenspace upload was initially missing 5 island tiles (HP/HT/HU/HY/HZ).

The public OS Downloads API (`https://api.os.uk/downloads/v1/products/{product}/downloads`, no auth for OpenData) lists every per-area file with download URLs, so the refresh can be one reproducible command.

## Proposal

New script `pipeline/run/stream_os_open_data.py` (precedent: `pipeline/run/stream_inspire_files.py`, which streams external zips to S3):

- CLI: `--products OpenMapLocal OpenRoads OpenGreenspace --save`; dry-run (list tiles + sizes) unless `--save` is set.
- Name the dated S3 prefix from the API's product `version` field (e.g. `2026-04`) rather than the download date.
- Normalise the internal layout across products at upload time (the three current prefixes have three different layouts because they mirror hand-unzipped OS downloads); update the `base.yaml` path templates once, at the next refresh.
- After upload, reconcile S3 squares against the API's offered areas and fail loudly on mismatch.
- API base URL in `base.yaml`, not hard-coded.

helloaidank:
Kickoff interview outcomes (2026-07-23) — two deviations from the proposal above, plus confirmations:

**Deviations**
1. **No layout normalisation.** Each product keeps its current on-S3 internal layout (OpenMap Local: `data/{square}/{square}_{layer}.shp`; Open Roads: `data/{square}_{layer}.shp`; Greenspace: `{square}/data/{square}_{layer}.shp`). Only the dated segment of each `base.yaml` template changes at repoint. Rationale: match what's already on S3; less churn for consumers and reviewers.
2. **Test rehearsal before production run.** The script first runs against a clearly-named test prefix inside `asf-local-heat-planning-tool`, is verified end-to-end (getters load from it), and only then runs against the real product prefixes with the `base.yaml` repoint as the final reviewed commit — all on the same branch.

**Confirmations**
- Dated prefixes named from the API `version` field verbatim (e.g. `2026-04`), not download date.
- Post-upload reconciliation against the API's offered areas lives in this script (complementary to the #434 preflight, not a duplicate).
[...]

PR #424, opened 2026-07-09
416 update pipeline to GB wide OS grid square data

### LINKED ISSUE #416: Upload per-grid-square OS data for all GB squares to S3
The pipeline loads OS datasets per 100km OS National Grid square, but S3 only holds squares for the development sampling areas (NS NT NX SD SE SJ SK SN SO SP SS ST SU SX SZ), so the pipeline can only run for LAs inside them.

Upload per-grid-square files for **all GB squares** for the three grid-square datasets:

- **OS OpenMap Local** (`grid_square_os_openmap_local`)
- **OS Open Roads** (`grid_square_os_openroad`)
- **OS Open Greenspace** (`grid_square_os_openmap_greenspace`)

keeping the existing `{square}/{square}_{layer}.shp` layout so the getters (`load_gdf_os_openmap_layer`, `load_gdf_os_openroad`) work unchanged for any LA.

**Versioning:** implement versioning for these grid-square inputs as part of the upload — dataset version/release info carried in the S3 prefix (consistent with the input naming convention `{pub_date}_v{version}_...` and with the output-dataset versioning introduced in #412), with `config/base.yaml` paths and `config/README.md` citations updated to match.

Acceptance: a pipeline stage (e.g. `pipeline/transform/uprns.py`) runs for an LA outside the current 15 squares without a missing building layer, plus a spot-check of one remote LA (Scilly/Northern Isles squares often lack several layers — the getter's skip-missing behaviour should handle this).

helloaidank:
## Description

The pipeline could previously only run for LAs falling inside the 15 development grid squares. Per-grid-square OS files now exist on S3 for **all of GB** under dated prefixes (uploaded outside this PR). This PR wires the pipeline up to that GB wide data so any LA can be processed.

Closes #416.

## Changes

- **Repoint grid-square config** — the four grid-square keys in `base.yaml` now point at the GB-wide dated prefixes (`opmplc_essh_gb/20260708`, `oproad_essh_gb/20260708`, `opgrsp_essh_gb/20260709`). 
- **Tolerate legitimately missing files** — the two grid-square getters now skip files that are genuinely absent from S3 (sparse/sea squares lack some layers; `pyogrio` raises `DataSourceError` for these), with unit tests covering both the skip and the re-raise guard.
- **Citations** — `config/README.md` updated to the April 2026 OS release and given the previously missing Open Roads row.

## For Reviewers

**What to focus on:** the skip vs fail logic in the grid-square getters...a missing *optional* layer should be skipped, but a genuine error should still surface.

**How to test:** `pytest tests/test_load_geodata.py` (4/4). Optionally run `pipeline/transform/uprns.py` for a non-dev grid square LA (e.g. Orkney Islands).

## Verification

- Unit tests: `tests/test_load_geodata.py` 4/4 pass (skip + re-raise guard).
- End-to-end `uprns.py` on new squares: Plymouth (control) passes pandera census validation; Orkney Islands passes (14,192 UPRNs vs expected 10,070–14,839); clustering barrier loaders verified for Orkney incl. missing-file skips.
- City of London fails census validation — a pre-existing method issue *exposed* (not caused) by GB coverage, tracked in #421.

## Known limitations

- If **all** of an LA's squares lack a layer (e.g. `tidal_water` for LAs entirely within square SP: Oxford, Milton Keynes), `pd.concat` raises `ValueError`. Fix proposed in #423, TODO noted in docstrings.
- A missing *core* layer file is currently skipped with a warning rather than failing loudly. We should implement a fix so that we distinguish between layers that are everywhere (e.g. building/road vs tidal water).  
## Follow-ups

Related issues: #419 (scripted OS Downloads API refresh), #420 (S3 cleanup of superseded prefixes), #421, #423.

## Checklist:

- [x] I have tested the code


asf_heat_pump_suitability/config/base.yaml
sofiapinto, line 1: Hey @helloaidank - I'll make a similar point here to the one I made in the versioning of outputs PR. An alternative to hardcoding the input date here would be pass {date} and then change our getters so that we always get the latest version of the outputs when running the pipeline. This would avoid having to change base.yaml every time we get an new dataset.

I guess the downside is that it's less explicit which versions of the datasets we're using - I don't think this should prevent us from going in this direction though. We just need to make that explicit somewhere else e.g. a README.md with the versions of each dataset.

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -278,6 +284,13 @@ def load_gdf_os_openmap_layer(
                 print(
                     f"Error loading OS OpenMap layer - {layer.title()} file {file}: {e}"
                 )
+            except DataSourceError as e:
+                # pyogrio raises DataSourceError for files missing from S3; skip those only
+                if "does not exist" not in str(e):
+                    raise
+                logging.warning(
+                    f"Skipping missing OS OpenMap layer - {layer.title()} file {file}: {e}"
+                )
```
sofiapinto, line 293: Just to flag that this was already being tackled in another PR: https://github.com/nestauk/asf_heat_pump_suitability/pull/408

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -318,7 +335,13 @@ def load_gdf_os_openroad(
...
         for file in files:
             print(f"\nLoading OS OpenRoad file: {file}")
-            gdfs.append(gpd.read_file(file))
+            try:
+                gdfs.append(gpd.read_file(file))
+            except (FileNotFoundError, DataSourceError) as e:
+                # skip grid squares with no road data (e.g. sea-only squares)
+                if isinstance(e, DataSourceError) and "does not exist" not in str(e):
+                    raise
+                logging.warning(f"Skipping missing OS OpenRoad file {file}: {e}")
```
sofiapinto, line 344: We would never use sea only squares though right? and if that was the case we would actually want it to error because it would mean wrong assignment of grid squares?

sofiapinto (commented):
Hey @helloaidank,

@crispy-wonton will be reviewing this and I haven't reviewed the tests for this PR, but have looked at everything else as this is connected to some of the work I've been doing and also some of the comments I left on the outputs versioning PR.

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: Agree with this - let's avoid hard-coding dates where possible so that it's easier to update the input files.

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -4,16 +4,16 @@ data:
     uk_osopen_uprn: "s3://asf-local-heat-planning-tool/inputs/geodata/osopenuprn_202510_csv.zip"
     # OpenMap Local Geopackage
     gb_os_openmap_local: "s3://asf-local-heat-planning-tool/inputs/geodata/opmplc_gb.gpkg"
-    # OpenMap Local layers for certain grid squares
-    grid_square_os_openmap_local: "s3://asf-local-heat-planning-tool/inputs/geodata/v202510_OSOpenMapLocal_geometries_selected/{square}/{square}_{layer}.shp"
+    # OpenMap Local layers per 100km grid square, all GB (April 2026 release)
```
crispy-wonton, line 7:
```suggestion
    # OpenMap Local layers per 100km grid square, all GB
```
Here and below I would remove references to the date otherwise we have to update them when we update the inputs.

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -202,6 +203,11 @@ def load_gdf_os_openmap_layer(
     Load specified OS OpenMap Local or Greenspace layer for Great Britain or optionally for a specific grid square.
     CRS British National Grid (27700).
 
+    In grid square mode, files missing from S3 are skipped with a warning (OS only ships a layer file for squares
+    containing that feature, e.g. no tidal water file for inland squares), so the result may cover fewer squares
+    than requested. TODO: if all requested squares lack the layer, pd.concat raises ValueError; return an empty
+    GeoDataFrame instead.
```
crispy-wonton, line 209: question: I'm confused about the TODO note here. E.g. what if only one square is requested and it lacks the layer, e.g. tidal water - wouldnt that be a valid case of a missing layer?

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton, line 293: question: why are we only raising a warning if 'does not exist' is not in the string? Shouldn't we be raising an error because it's failing for another reason and then raising a warning when it actually doesnt exist? Lmk if I misunderstand.

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: Agree this should not happen so this warning should never get triggered unless there is an error in grid square assignment. But I'm happy to keep this `try` `except` block as potentially this would be required when running for all of GB (not sure yet).

tests/test_load_geodata.py
crispy-wonton, line 1: can we move this file to the `getters` subdir into a folder called `tests/` thank you :)

You can also take a look at this PR to see the structure of the test scripts I'd like us to use: https://github.com/nestauk/asf_heat_pump_suitability/pull/383

You can also see slides 38 and 39 in the Sprint 2 slides (will share on slack) for the conventions I'm using.

tests/test_load_geodata.py
```diff
@@ -0,0 +1,65 @@
+"""
+Tests for grid-square getters in getters/load_geodata.py.
+
+Missing grid-square files on S3 (e.g. sea-only squares with no road layer) raise
```
crispy-wonton:
```suggestion
Genuine missing grid-square files on S3 (due to true absence of data e.g. sea-only squares with no road layer) raise
```

tests/test_load_geodata.py
```diff
@@ -0,0 +1,65 @@
...
+    return fake_read_file
+
+
+def test_load_gdf_os_openmap_layer_skips_missing_square(monkeypatch):
+    """Squares whose layer file is missing from S3 are skipped; the rest are returned."""
+    monkeypatch.setattr(gpd, "read_file", _fake_read_file_missing_square("HX"))
+    gdf = load_geodata.load_gdf_os_openmap_layer(
+        "important_building", grid_squares=["ND", "HX"]
+    )
+    assert len(gdf) == 1
```
crispy-wonton:
```suggestion
def test_load_gdf_os_openmap_layer_skips_missing_square(mocker):
    """Squares whose layer file is missing from S3 are skipped; the rest are returned."""

    def read_file_side_effect(path, **kwargs):
        if "HX" in str(path):
            raise DataSourceError(
                f"'{path}' does not exist in the file system, and is not recognised as a supported dataset name."
            )
        return SAMPLE_GDF.copy()

    mocker.patch("geopandas.read_file", side_effect=read_file_side_effect)
    gdf = load_geodata.load_gdf_os_openmap_layer(
        "important_building", grid_squares=["ND", "HX"]
    )
    assert len(gdf) == 1
```

I think this is cleaner to use a mocker with `pytest-mock` (note if implementing, you would need to add this package to `pyproject.toml`).

tests/test_load_geodata.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def test_load_gdf_os_openroad_skips_missing_square(monkeypatch):
+    """Squares whose road file is missing from S3 are skipped; the rest are returned."""
+    monkeypatch.setattr(gpd, "read_file", _fake_read_file_missing_square("HX"))
```
crispy-wonton: same as above.

tests/test_load_geodata.py
```diff
@@ -0,0 +1,65 @@
...
+
+@pytest.mark.parametrize(
+    "getter",
+    [
+        lambda: load_geodata.load_gdf_os_openmap_layer(
+            "important_building", grid_squares=["ND"]
+        ),
+        lambda: load_geodata.load_gdf_os_openroad(grid_squares=["ND"]),
+    ],
+)
```
crispy-wonton: nice use of parametrize - forgot this exists!

tests/test_load_geodata.py
```diff
@@ -0,0 +1,65 @@
...
+def test_getters_reraise_other_data_source_errors(monkeypatch, getter):
```
crispy-wonton: question - just double checking the definition - does 'reraise' means you intend to trigger the error inside a try/except block?

tests/test_load_geodata.py
```diff
@@ -0,0 +1,65 @@
...
+    """DataSourceErrors unrelated to missing files (e.g. credentials) must not be skipped."""
+
+    def fake_read_file(path, **kwargs):
+        raise DataSourceError("AWS error: access denied")
+
+    monkeypatch.setattr(gpd, "read_file", fake_read_file)
+    with pytest.raises(DataSourceError):
+        getter()
```
crispy-wonton: suggestion, again I think you should be able to use `mocker` here.

crispy-wonton (changes requested):
All tests passed when I ran and I couldn't find errors in the logic. There are some suggestions I'd like to implement, like using `mocker` if possible for cleaner mocking of errors and also following the same conventions that I've been using in terms of formatting the tests.

Lastly, I'd like to remove the hard-coding of the file dates in the input data BUT i'm happy for this PR to be merged sooner and have that addressed in a separate PR as it's not crucial to move forward and I think it might take a fair bit of new code.

asf_heat_pump_suitability/getters/load_geodata.py
helloaidank: I hit the same bug on the GB wide loading and fixed it here before spotting #408  had already tackled it. I've proposed a way to untangle the overlap on the [#408 thread](https://github.com/nestauk/asf_heat_pump_suitability/pull/408).

asf_heat_pump_suitability/config/base.yaml
helloaidank: I guess the couple of outstanding issues (they might not be worth the inconvenience of changing the base.yaml) would be around incomplete data downloads (e.g. maybe having a success marker after some test has passed to make sure all relevant data has been downloaded) or when the data is being actively downloaded and we try to run the pipeline.

helloaidank: Happy to tackle this in a separate PR! Opened #429 to track it — includes the incomplete-download/success-marker and concurrent-upload considerations from above.

asf_heat_pump_suitability/getters/load_geodata.py
helloaidank: Will resolve merge conflict

asf_heat_pump_suitability/getters/load_geodata.py
helloaidank: We compute the grid squares from the local authority boundaries, potentially some local councils would own a tiny roadless island out at sea, and this would be a legitimate case? In practice, this might not happen.

asf_heat_pump_suitability/getters/load_geodata.py
helloaidank: Think this was phrased maybe not in the best way. We should essentially return an empty geodataframe if all requested squares lack say a tidal water layer but how it is currently implemented means it crashes. Opened an issue with regards to this: https://github.com/nestauk/asf_heat_pump_suitability/issues/423

asf_heat_pump_suitability/getters/load_geodata.py
helloaidank: I tested this. I ran `get_list_la_grid_squares` to derive the squares from the LA boundaries, for all GB LAS and compared the output against the OpenRoad files we have under `oproad_essh_gb/20260708/data/` on S3.

Three squares get requested but have no road file, so this happens on a valid grid square assignment:

- HW: intersected by Na h-Eileanan Siar (it contains North Rona and Sula Sgeir, which have no roads)
- HX: intersected by Orkney Islands (Sule Skerry / Sule Stack)
- OV: sea-only, but North Yorkshire's boundary plus the 1km buffer clips its corner

asf_heat_pump_suitability/getters/load_geodata.py
helloaidank:
```
if "does not exist" not in str(e):
    raise
logging.warning('skip file').
```

So the raise happens when file is corrupt/auth failure etc.

Warning happens when the file does not exist.

Does that make sense?

tests/test_load_geodata.py
helloaidank: Yep, reraise means we want to catch the DataSourceError and for anything other than a "does not exist", calls the bare raise.

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: Yes, I had misunderstood the line and thought you were raising the warning if the string did not contain 'does not exist'. Have checked the documentation and note to self: the `raise` here raises the original exception to catch any unexpected instances of this error.

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: Great - thanks for checking @helloaidank !

crispy-wonton (approved):
Changes look great thank you!

All tests passed as well.

PR #432, opened 2026-07-22
423 fix error when a layer is missing from ALL of an LA's squares (e.g. tidal_water for inland LAs) in `load_gdf_os_open_map`

### LINKED ISSUE #423: Grid-square getters crash when a layer is missing from ALL of an LA's squares (e.g. tidal_water for inland LAs)

sofiapinto:
---

# Description

Quick fix to `load_gdf_os_open_map()` function of `getters.load_geodata.py` to account for the possibility of a layer missing from ALL of an LA's squares

Fixes #423

# Instructions for Reviewer

Hi @helloaidank , this is me applying the quick fix you proposed.


asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -292,6 +291,12 @@ def load_gdf_os_openmap_layer(
                     f"Skipping missing OS OpenMap layer - {layer.title()} file {file}: {e}"
                 )
 
+        if not gdfs:
+            logging.warning(
+                f"No {layer} files found for grid squares {grid_squares}. Returning empty GeoDataFrame."
+            )
+            return gpd.GeoDataFrame(geometry=[], crs=27700)
```
helloaidank: changed to `config["constant"]["target_crs"]` and this module already imports `config`.

```suggestion
            return gpd.GeoDataFrame(geometry=[], crs=config["constant"]["target_crs"])
```

helloaidank (dismissed):
Thanks Sofia for implementing the fix. I've updated the branch to include latest changes from dev, it's easily mergeable as the code has no overlaps. However I did put a docstring with this as a TODO in #424, so would make the fix without editing the TODO (would be confusing).  

I've also taken the liberty to add a unit test for the all squares missing path for this PR. Tested that all the getter tests pass locally. One inline suggestion below on CRS.

asf_heat_pump_suitability/getters/tests/test_load_geodata.py
sofiapinto: i'm guessing we should also import the CRS from the config in this case, or was there any reason not to in this case? @helloaidank

asf_heat_pump_suitability/getters/tests/test_load_geodata.py
```diff
@@ -49,6 +49,21 @@ def test_reraises_other_data_source_errors(self, mocker):
...
+            side_effect=DataSourceError(
+                "'path' does not exist in the file system, and is not recognized as a supported dataset name."
+            ),
+        )
+        gdf = load_geodata.load_gdf_os_openmap_layer(
+            "important_building", grid_squares=["ND", "HX"]
+        )
+        assert gdf.empty
+        assert "geometry" in gdf.columns
+        assert gdf.crs.to_epsg() == 27700
```
sofiapinto, line 68: i'm guessing we should also import the CRS from the config in this case, or was there any reason not to in this case? @helloaidank

asf_heat_pump_suitability/getters/tests/test_load_geodata.py
```diff
@@ -49,6 +49,21 @@ def test_reraises_other_data_source_errors(self, mocker):
...
 
+    def test_returns_empty_gdf_when_all_squares_missing(self, mocker):
+        """When the layer file is missing from every requested square, an empty GeoDataFrame is returned."""
+        mocker.patch(
+            "geopandas.read_file",
+            side_effect=DataSourceError(
+                "'path' does not exist in the file system, and is not recognized as a supported dataset name."
+            ),
+        )
+        gdf = load_geodata.load_gdf_os_openmap_layer(
```
sofiapinto, line 63: I haven't used mocker before btw, very cool.

helloaidank: It's @crispy-wonton suggestion!

asf_heat_pump_suitability/getters/tests/test_load_geodata.py
helloaidank: getter returns crs = config["constant"]["target_crs"], if the test then asserted gdf.crs.to_epsg = config["constant"]["target_crs"] then it would be comparing the config to itself.  I guess it's useful to flag that the coordinate system has changed, and fail the test to flag to the user.

asf_heat_pump_suitability/getters/tests/test_load_geodata.py
helloaidank: Implemented this! Thanks for the suggestion

helloaidank (approved):

PR #448, opened 2026-07-23
434 Add S3 input preflight check before pipeline runs

### LINKED ISSUE #434: Add input preflight check before pipeline runs
### Current behaviour

Nothing verifies S3 input paths exist before a run starts. A missing or misconfigured path
only surfaces deep inside a stage once a lazy S3 read attempts it — as happened when an
input dataset was reorganised into new dated prefixes and the pipeline crashed mid-run on a
single missing file, well after other stages had already completed.

### Relevant code

`pipeline/run/run_pipeline.sh` is the live per-LA orchestrator (merged, #410) — its first
step is where this check plugs in. It walks the S3 paths already declared under
`config["data"]` in `config/base.yaml`.

### Desired behaviour

A preflight step walks every configured S3 input path, checks each one exists (existence
check only, not a full read), and reports **all** missing paths in one pass rather than
failing on the first one encountered deep in a stage. The pipeline orchestrator runs it
first and stops before processing any local authority if paths are missing.

### Acceptance criteria

- [ ] Checks every configured S3 input path, resolving templated paths to their prefix
- [ ] Reports all missing paths in one pass and exits non-zero
- [ ] Wired into `run_pipeline.sh` as its first step
- [ ] Unit tests cover at least one missing-path case and one all-present case

### Out of scope

- Per-square existence checks (would need the LA→squares mapping; prefix-level only for v1)
- Auto-detecting the latest dated prefix (tracked separately in #429)

---

Part of the pipeline validation checks initiative (see `docs/specs/pipeline_validation_checks.md` and `docs/specs/pipeline_validation_checks_issues.md`). Independent of the run-manifest/comparison-script stack — can be built in parallel.

helloaidank:
---

# Description

Adds a preflight step that verifies every S3 input path configured under `config["data"]` exists before a pipeline run starts. Previously a missing or misconfigured path only surfaced deep inside a stage once a lazy S3 read attempted it.

- **`pipeline/validate/check_inputs.py`** (new subpackage): recursively collects every `s3://` leaf in `config["data"]`, truncates templated paths (e.g. `.../{square}/{square}_{layer}.shp`) to the prefix before the first `{`, checks each exists, reports **all** missing paths in one pass, and exits non-zero if any are missing.
- **`utils/s3_utils.py`**: new `get_bool_s3_path_exists` helper using `list_objects_v2` with `MaxKeys=1`, which works for both folder prefixes and exact file keys.
- **`run_pipeline.sh`**: calls the check as its first step and aborts before the local-authority loop if it fails.
- **`config/base.yaml`**: removed the stale `geodata.council_tax_data` entry (should we do this?) the acceptance run surfaced that the file was never uploaded to S3 (empty folder) and only research/exploratory scripts referenced it. For the config, should we just use this config for 'production', what the actual pipeline uses. 

Spec with full decision record: `docs/specs/434_inputPreflightCheck.md`.

Fixes #434

# Instructions for Reviewer

In order to test the code in this PR you need to run:

```
uv run python asf_heat_pump_suitability/pipeline/validate/check_inputs.py
```

Expected: `INFO: All 26 configured S3 input paths exist.` and exit 0. To see the failure mode, temporarily break a path in `base.yaml` and re-run — every broken path is listed before the non-zero exit. Unit tests: `uv run pytest asf_heat_pump_suitability/pipeline/validate/tests/`.

Please pay special attention to the removal of the `council_tax_data` config entry (rationale in the spec's Verification section) and to the convention it establishes for what belongs under `config["data"]`.

# Checklist:

- [x] I have tested the code


asf_heat_pump_suitability/config/base.yaml
```diff
@@ -1,3 +1,8 @@
+# Production pipeline inputs only. Every s3 path under `data` is treated as a
+# required input: the preflight (pipeline/validate/check_inputs.py) checks each
+# one exists before any pipeline run and aborts if any are missing. Datasets
+# used only by research/exploratory scripts must not live here — keep their
+# paths in the script or a research-local config.
```
crispy-wonton: issue: I am fine with this being the standard in theory and only using this `config/base.yaml` for the production pipeline eventually. However, currently we can't delete files referenced here that are used in the research scripts so that we break the research scripts. I suggest restoring the council tax data reference here for now and then opening a new issue and separate PR (low priority) to clean up the config to files and constants used in the pipeline only and in that PR create a separate config specific to the research folder. 

I also think this should happen much further down the line when we have a stable pipeline with stable inputs, otherwise, we risk duplicating stuff in the research config and base config as we move stuff from research into the main pipeline.

asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
```diff
@@ -6,6 +6,14 @@
 # Run from the repo root regardless of where the script is invoked from
 cd "$(dirname "$0")/../../.." || exit 1
 
+# Preflight: verify every configured S3 input path exists before starting
+echo "--> Checking S3 input paths exist: check_inputs.py"
+python asf_heat_pump_suitability/pipeline/validate/check_inputs.py
+if [ $? -ne 0 ]; then
+    echo "Error running check_inputs.py: missing S3 input paths. Aborting."
+    exit 1
+fi
```
crispy-wonton: wth is this syntax!! lol (this is not aimed at you, i just looked it up and learned it for the first time and it's crazy. generally this is a very nice addition thank u)

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
```diff
@@ -0,0 +1,101 @@
+"""
+Preflight check that all S3 input paths configured under `config["data"]` exist.
+
+Recursively collects every s3:// path in the nested `config["data"]` mapping,
+truncates templated paths (e.g. ".../{square}/{square}_{layer}.shp") to the
+prefix before the first "{" token, and checks each one exists in S3. All
+missing paths are reported in one pass and the script exits non-zero if any
+are missing, so run_pipeline.sh can abort before the local-authority loop
+starts. Templated paths are checked at prefix level only, not resolved per
+grid square or local authority.
```
crispy-wonton: suggestion: I think it might be worth checking for each grid square. Otherwise, I can see us running into the same issues you've run into already, when one particular grid square is missing. What do you think?

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
```diff
@@ -0,0 +1,101 @@
...
+    paths = []
+    for value in config_section.values():
+        if isinstance(value, dict):
+            paths.extend(get_list_s3_paths(value))
+        elif isinstance(value, list):
+            paths.extend(
+                item
+                for item in value
+                if isinstance(item, str) and item.startswith("s3://")
+            )
```
crispy-wonton: question: do we have lists? looking at the config I don't think we have them for data sources but just for my understanding, can you let me know if there are any lists under `config["data"]`

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
```diff
@@ -0,0 +1,101 @@
...
+        elif isinstance(value, str) and value.startswith("s3://"):
+            paths.append(value)
+    return paths
+
+
+def get_str_checkable_prefix(path: str) -> str:
```
crispy-wonton:
```suggestion
def get_str_common_prefix(path: str) -> str:
```
just a suggestion to make it more generic and clear imo

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
```diff
@@ -0,0 +1,101 @@
...
+            (truncated) prefix.
+    """
+    missing_paths = []
+    for path in paths:
+        bucket, _, key = (
+            get_str_checkable_prefix(path).removeprefix("s3://").partition("/")
+        )
+        if not s3_utils.get_bool_s3_path_exists(s3_client, bucket, key):
+            missing_paths.append(path)
+    return missing_paths
```
crispy-wonton, line 135: suggestion: this could return `None` if there are no paths instead of an empty list.

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
```diff
@@ -0,0 +1,101 @@
...
+        s3_client: an initialized boto3 S3 client.
+
+    Returns:
+        list: configured paths with no S3 object at or under their
+            (truncated) prefix.
+    """
+    missing_paths = []
+    for path in paths:
+        bucket, _, key = (
+            get_str_checkable_prefix(path).removeprefix("s3://").partition("/")
```
crispy-wonton: praise: nice use of builtins with `removeprefix` and `partition` here!

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
```diff
@@ -0,0 +1,101 @@
...
+        paths: configured s3:// paths, possibly templated.
+        s3_client: an initialized boto3 S3 client.
+
+    Returns:
+        list: configured paths with no S3 object at or under their
+            (truncated) prefix.
+    """
+    missing_paths = []
+    for path in paths:
+        bucket, _, key = (
```
crispy-wonton:
```suggestion
        bucket, sep, key = (
```
just a suggestion to make the code easier to understand

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
```diff
@@ -0,0 +1,101 @@
...
+    Args:
+        path: S3 path, optionally containing template tokens such as "{square}".
+
+    Returns:
+        str: the path unchanged, or its prefix before the first "{".
+    """
+    return path.split("{", 1)[0]
+
+
+def get_list_missing_s3_paths(paths: list, s3_client: boto3.client) -> list:
```
crispy-wonton, line 116: praise: really nicely written function!

asf_heat_pump_suitability/pipeline/validate/tests/test_check_inputs.py
```diff
@@ -0,0 +1,113 @@
...
+    get_str_checkable_prefix,
+)
+
+
+class FakeS3Client:
+    """Minimal stand-in for a boto3 S3 client's `list_objects_v2` method."""
+
+    def __init__(self, existing_paths: list):
+        """Store full s3:// paths of the objects that "exist" in the fake S3."""
+        self.existing_paths = existing_paths
```
crispy-wonton, line 18: question: where is the `existing_paths` variable coming from here?

asf_heat_pump_suitability/pipeline/validate/tests/test_check_inputs.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def list_objects_v2(self, **kwargs) -> dict:
+        """Return KeyCount 1 if any existing path sits under Bucket/Prefix."""
```
crispy-wonton: question: this won't _necessarily_ return 1 right? it could be any int value for the count of keys under the bucket/prefix combination, including 0.

asf_heat_pump_suitability/pipeline/validate/tests/test_check_inputs.py
```diff
@@ -0,0 +1,113 @@
...
+    def test_collects_s3_paths_inside_lists(self):
+        """s3:// strings inside list values are collected, not silently skipped."""
+        config_section = {
+            "regions": ["s3://bucket/inputs/north.csv", "s3://bucket/inputs/south.csv"],
+            "mixed": ["not a path", 42],
+        }
+        assert get_list_s3_paths(config_section) == [
+            "s3://bucket/inputs/north.csv",
+            "s3://bucket/inputs/south.csv",
+        ]
```
crispy-wonton:
```suggestion
            "mixed": ["not a path", 42],
            "path_and_not": ["s3://bucket/inputs/east.csv", "https://github.com/nestauk"]
        }
        assert get_list_s3_paths(config_section) == [
            "s3://bucket/inputs/north.csv",
            "s3://bucket/inputs/south.csv",
            "s3://bucket/inputs/east.csv",
        ]
```

asf_heat_pump_suitability/pipeline/validate/tests/test_check_inputs.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+class TestGetStrCheckablePrefix:
+    """Tests for `get_str_checkable_prefix`."""
+
+    def test_truncates_templated_path_at_first_brace(self):
+        """A templated path is truncated to the prefix before the first '{'."""
+        path = "s3://bucket/inputs/opmplc/data/{square}/{square}_{layer}.shp"
+        assert get_str_checkable_prefix(path) == "s3://bucket/inputs/opmplc/data/"
```
crispy-wonton: question: can the `get_str_checkable_prefix` function handle a path like this: "s3://bucket/inputs/opmplc/data/grid_square_{square}/{square}_{layer}.shp"

would be good to add to the tests as well to ensure expected behaviour with path formats as such

crispy-wonton (commented):
Hi @helloaidank !

Looks good and the tests all passed. I couldn't find any bugs. I mainly left some questions and a couple of suggestions. Great work!

asf_heat_pump_suitability/pipeline/validate/tests/test_check_inputs.py
crispy-wonton, line 1: issue (non-blocking): as in previous discussion on PR #449 , the assert statements should print an informative error message to console if they fail  :)

asf_heat_pump_suitability/config/base.yaml
helloaidank: Happy to restore the council tax data reference here for now, and open up an issue around separating out the production config and the research config. However if we restore it naively under config["data"], the preflight treats every "s3://" path under config["data"] as a required input. Also worth saying that the "research/exploratory/domestic_filtering/council_tax_domestic_filtering.py" script is already broken regardless of the config key as the relevant s3 bucket doesn't have the input, i.e. `PLYMOUTH_CTBANDS_ONSUD_202512.csv` is not in the s3 bucket. We can write a specific exclusion for the moment but not the cleanest. Will do this change for the moment but [opened issue](https://github.com/nestauk/asf_heat_pump_suitability/issues/469#issue-5133239809) related to config split.

asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
helloaidank: Shell script can get a bit crazy. `$0` is the path to this script, `dirname` takes the folder the script lives in, and `/../../..` climbs three folders up from there, which lands at the repo root. The $(...)` wrapper just means "run this and paste the result in place". So the whole line says "cd to the repo root, no matter where the script was called from". The `|| exit 1` means "and if that cd fails for any reason, stop immediately" rather than running the pipeline from the wrong folder.

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
helloaidank: Good shout, agreed. Implemented in the latest commit: paths with a `{square}` token are now expanded over`config["constant"]["sampling_areas"["grid_squares"]` before checking, so a missing square gets reported individually (e.g. `.../data/SD/SD_{layer}.shp`). The remaining `{layer}` token is still truncated to a prefix, so the check requires at least one layer file per square rather than every layer, since some layers are legitimately absent in some squares and the getter deliberately skips those.

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
helloaidank: No lists under config["data"] at the moment, every leaf is a string or a nested dict. The list branch was there defensively in case we add one later, but agreed it's simpler to apply YAGNI here, so I've removed the branch (and its test).

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
helloaidank: Done!

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
helloaidank: I would probably just keep the empty list here? I guess the empty list already means "nothing is missing", and we need to loop over it and count it, so returning None would require an extra guard.

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
helloaidank: Done!

asf_heat_pump_suitability/pipeline/validate/tests/test_check_inputs.py
helloaidank: It's the constructor argument of `FakeS3Client`: `__init__`!

asf_heat_pump_suitability/pipeline/validate/tests/test_check_inputs.py
helloaidank: For the real function we use in the pipeline it can only ever be 0 or 1 here (see `get_bool_s3_path_exists` in s3_utils.py,  I've made that explicit within the doc string now.

asf_heat_pump_suitability/pipeline/validate/tests/test_check_inputs.py
helloaidank: Implemented!

asf_heat_pump_suitability/pipeline/validate/tests/test_check_inputs.py
helloaidank: Yes it can, and applied this test case, thanks!

asf_heat_pump_suitability/pipeline/validate/tests/test_check_inputs.py
helloaidank: Done, added failure messages to all asserts.

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: Interesting. Didn't realise it was broken separately 🤔 Thanks for the temp solution though!

asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
crispy-wonton: I specifically am in awe of the ridiculous 
```
if
...
fi
``` 
syntax LOL! that's the weirdest thing I ever seen

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
crispy-wonton: Thanks - re layers I think that makes sense!

Re the sampling areas - I would probably just use the full set of grid squares rather than a sample :)

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
crispy-wonton: no probs, happy to keep

asf_heat_pump_suitability/pipeline/validate/tests/test_check_inputs.py
crispy-wonton: must have been tired when i first read this. makes sense now in the light of day, lol

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
```diff
@@ -25,6 +28,13 @@
 from asf_heat_pump_suitability import config
 from asf_heat_pump_suitability.utils import s3_utils
 
+# Research-only inputs configured under config["data"] but not read by the
```
crispy-wonton: suggestion: can we label this with TODO so its discoverable upon searching the repo

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
```diff
@@ -40,18 +50,35 @@ def get_list_s3_paths(config_section: dict) -> list:
...
+
+    Returns:
+        list: paths with "{square}" substituted per square; paths without the
+            token are passed through unchanged.
+    """
+    expanded_paths = []
+    for path in paths:
+        if "{square}" in path:
+            expanded_paths.extend(
+                path.replace("{square}", square) for square in squares
```
crispy-wonton: suggestion: you should also be able to used `str.format()` here which might be more readable.

E.g.
```
path.format(square=square) for square in squares
```

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
```diff
@@ -88,7 +115,19 @@ def get_list_missing_s3_paths(paths: list, s3_client: boto3.client) -> list:
 
 if __name__ == "__main__":
     logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
-    input_paths = get_list_s3_paths(config["data"])
+    input_paths = [
+        path
+        for path in get_list_s3_paths(config["data"])
+        if path not in RESEARCH_ONLY_PATHS
+    ]
```
crispy-wonton, line 165: oh lol I see what you mean about this being jank now. Thanks for adding this in though! we can change streamline later.

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
```diff
@@ -88,7 +115,19 @@ def get_list_missing_s3_paths(paths: list, s3_client: boto3.client) -> list:
...
+    if not input_paths:
+        logging.error(
+            "No S3 input paths found under config['data'] - check config parsing."
+        )
+        sys.exit(1)
+    input_paths = generate_list_expanded_square_paths(
+        input_paths, config["constant"]["sampling_areas"]["grid_squares"]
```
crispy-wonton: suggestion: as suggested in other comment, I'd use the full list of grid squares rather than just the sampling areas. You could also use the grid squares generated in the `local_authorities_dict` if you just wanted to check for the grid squares that are actually required for a particular run.

crispy-wonton (commented):
Great changes, thank you!

Main suggestion is to use either full list of grid squares, or grid squares required by the current pipeline run for the checks rather than just the sampling areas.

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
helloaidank: Done — labelled as `TODO(#469)` in the latest commit so it shows up when grepping the repo for TODOs, and ties straight to the config-split issue.

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
helloaidank: Good idea in principle, but `str.format()` has a catch: it needs every placeholder in the string filled in at once. Two of our templated paths contain both `{square}` and `{layer}` (e.g. `.../{square}/{square}_{layer}.shp`), so `path.format(square="SX")` would crash with `KeyError: 'layer'`.

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
helloaidank: Ok going to talk through the thinking on this, be useful to just talk through it in a stand-up:

- `check_inputs.py` takes `--local_authorities` as an argument. The named LAs expand `{square}` over the union of their grid squares using `get_list_la_grid_squares`. So the preflight checks exactly the squares the run will request (i.e. your local_authorities_dict suggestion).
- calling it without a flag runs it over the whole of GB. 

The whole GB run reported 18 missing paths, all in 7 grid squares (HW, HX, NQ, NV, OV, SA, SB) and this means the pre-flight fails. These are genuine OS gaps, not upload gaps. There is zero road/greenspace features in these seven squares (meaning the at least one layer for this square doesn't quite work as`geodata/oproad_essh_gb/20260708/data/HW_RoadLink.dbf` won't exist). I've documented these missing layers in `config["constant"]["os_data_absent_grid_squares"]` locally for the moment, but would be good to discuss first.

helloaidank: Following up post stand-up: discussed and agreed, now pushed in c5e1655. Final shape, with two refinements over my comment above:

- **The exclusions are per product, not a flat square list**: `config["constant"]["os_data_absent_grid_squares"]` maps product folder → absent squares, and the expansion skips a square only for the product it's listed against. That matters because OpenMap Local *does* ship HW/HX/OV (St Kilda has buildings, no roads), so those checks are kept.
- **Verified these are OS gaps, not upload gaps**: cross-checked against the separately-downloaded GB-wide GeoPackage copies already in the bucket (`opmplc_gb.gpkg`/`opgrsp_gb.gpkg`/`oproad_gb.gpkg`, single-file downloads from a different download).

Acceptance: bare GB run = 59 derived squares, 182 paths, exit 0; `run_pipeline.sh` now passes its LA list to the preflight; the failure mode was demonstrated en route (pre-exclusion GB run listed all 18 absences in one pass and exited 1). Unit tests cover the per-product skip.

Follow-ups opened: #479 (`get_dict_la_data(None)` returns the raw unclipped grid - the two "GB" code paths disagree, and the GB derivation currently includes NI boundaries).

helloaidank:
Also included a one-line fix in `uprns.py` (c87beaa) that surfaced while building the preflight's CLI: its `--local_authorities` used `default="GB"` — a bare string. With `nargs="+"`, argparse turns *typed* values into a list but passes the *default* through as-is, so running `python uprns.py` without the flag iterated the string character-by-character (`["g", "b"]`) and crashed LA name resolution with a confusing "No match found for: g". Never hit in practice because `run_pipeline.sh` always passes the flag. Now `default=["GB"]`, matching `cluster.py` — the same form the new `check_inputs.py` uses. Fixes #480.

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -71,6 +71,10 @@ constant:
     outdoor_space_threshold_m2:
       within_hn_zone: 70
       outside_hn_zone: 30
+  os_data_absent_grid_squares: # 100km squares specific OS products publish no data for (remote islets and NI-overlap squares on the BNG; the tool is GB-only). Keyed by the product folder name in the S3 path. The input preflight check skips these square/product combinations. Verified 2026-08-18 against the GB-wide gpkg copies (opmplc_gb/opgrsp_gb/oproad_gb.gpkg): zero features per listed square.
+    opmplc_essh_gb: ["NQ", "NV", "SA", "SB"]
+    opgrsp_essh_gb: ["HW", "HX", "NQ", "NV", "OV", "SA", "SB"]
```
crispy-wonton, line 85: question: what's this dataset?

asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
```diff
@@ -23,6 +16,13 @@ local_authorities=(
     "east lothian"
 )
 
+echo "--> Checking S3 input paths exist: check_inputs.py"
+python asf_heat_pump_suitability/pipeline/validate/check_inputs.py --local_authorities "${local_authorities[@]}"
+if [ $? -ne 0 ]; then
+    echo "Error running check_inputs.py: missing S3 input paths. Aborting."
+    exit 1
+fi
```
crispy-wonton, line 61: question: how come this is hard coded to check only the testing local authorities here?

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
```diff
@@ -55,23 +68,32 @@ def get_list_s3_paths(config_section: dict) -> list:
...
 
     Returns:
         list: paths with "{square}" substituted per square; paths without the
             token are passed through unchanged.
     """
+    absent_squares = absent_squares or {}
     expanded_paths = []
     for path in paths:
         if "{square}" in path:
+            skip = next((set(v) for k, v in absent_squares.items() if k in path), set())
```
crispy-wonton: question on syntax here: what is `next()` doing exactly? why is it just getting the next set of grid square values? Why not just collect all the absent squares for a given path into a single set? Sorry, i'm clearly missing something!

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
```diff
@@ -113,8 +135,29 @@ def get_list_missing_s3_paths(paths: list, s3_client: boto3.client) -> list:
...
+    return parser.parse_args()
+
+
 if __name__ == "__main__":
     logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
+    args = parse_arguments()
+    resolved_las = local_authority.resolve_list_la_names(args.local_authorities)
+    grid_squares = sorted(local_authority.get_list_la_grid_squares(resolved_las))
+    logging.info(
+        f"Checking inputs for {'whole of GB' if resolved_las is None else ', '.join(resolved_las)} "
```
crispy-wonton, line 180: praise: this is a cool use of fstring which makes perfect sense but I haven't thought to do before - nice!

crispy-wonton (commented):
Thanks for the fixes!

Have run the tests again and they all passed and reviewed them to ensure the logic makes sense.

Couldn't find any issues, but have left some final clarifying questions where things confused me!

crispy-wonton:
Separately, merge conflicts are awaiting resolution before merge :) good time to address those

asf_heat_pump_suitability/config/base.yaml
helloaidank: referring to `opgrsp_essh_gb`? It's the green space one I believe. 

If 'opmplc_essh_gb', it's the OS OpenMap – Local dataset.

asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
helloaidank: I think we agreed that's what was useful for the moment in the run script? We can expand to **GREAT BRITAIN** if required

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
helloaidank: so my understanding is that next() checks key 1, key 2, key 3... and the moment one of them matches, hands it over. We go through this iteration of the absent square config paths. We could collect all the absent squares for a given path into a single set, that's an equally valid solution :).

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: Thanks - it was the green space one! :)

asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
crispy-wonton: Cool no worries - just wanted to double check this was desired. Let's leave expansion to a separate PR as there are other scaling actions todo as well.

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
crispy-wonton: Ok thank you - I understand the intent here. Is this possible to rewrite in a more readable way? It's quite complex to understand.

crispy-wonton (changes requested):
Have just requested a rewrite of one line to make it more readable. 

Additionally, there are some merge conflicts to be resolved with dev so good time to address those.

Thank you!

asf_heat_pump_suitability/pipeline/validate/check_inputs.py
helloaidank: I have separated it out into a much more readable function! Lmk if that looks good.

asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
helloaidank: Maybe having a python runner script might make more sense long term for these sort of complexities

crispy-wonton (approved):

PR #449, opened 2026-07-23
440 Add run manifest for output lineage

### LINKED ISSUE #440: Add a run manifest
**Stacked on #413** (`412_versionOutputDatasets`, currently open) — touches the same five
entrypoints #413 already edits for dated output paths. Branch off `412_versionOutputDatasets`
rather than `dev` until #413 merges.

### Current behaviour

Nothing records which input versions, git commit, or parameters produced a given output.
Only the final geojson carries any metadata (creation date + licence — no lineage). Note:
the already-merged `create_manifest.py` is unrelated — it's a front-end dataset listing
(local authority + geojson URL), not run lineage; this manifest needs a different name to
avoid the collision.

### Desired behaviour

Each pipeline entrypoint writes a small JSON manifest alongside its output, recording:
stage, local authority, run timestamp, git commit, resolved input dataset versions, output
row count, key run parameters.

### Acceptance criteria

- [ ] A manifest JSON is written next to each of the five entrypoints' outputs (uprns,
      add_features, decision_tree, cluster, compute_contextual_features)
- [ ] Manifest includes: stage, local_authority, run_at, git_commit,
      input_versions, row_count, params
- [ ] Filename avoids collision with the existing `create_manifest.py` output
- [ ] Unit test covers manifest content for at least one entrypoint

### Out of scope

- Auto-detecting input versions from S3 (tracked separately in #429); until it lands,
  `input_versions` is read from the dated prefixes already in `base.yaml`
- Any reporting or comparison logic on manifest contents (that's the cross-version
  comparison script's job — first follow-on issue in this stack)

---

Part of the pipeline validation checks initiative (see `docs/specs/pipeline_validation_checks.md` and `docs/specs/pipeline_validation_checks_issues.md`). First issue in the comparison-script stack — the cross-version comparison script (base checks, then cluster geometry, then distribution, then thresholding) all depend on this landing first.

helloaidank:
## Description

Right now nothing records **how** a pipeline output was produced. If you open a
`plymouth_domestic_uprns.parquet` sitting in S3, there's no way to tell which
input dataset versions, which git commit, or which command line arguments
created it. The only metadata we keep is on the final geojson (a creation date
and a licence), and that carries no lineage.

This PR makes each of the five pipeline entrypoints write a small companion
**run manifest** (a JSON file) next to every output it saves. The manifest
records the lineage of that output so a later comparison script can diff two
runs and see exactly what changed.

This is the **first issue in the cross version comparison stack** — the
comparison script (and everything after it) can't be built until runs start
leaving this lineage behind.

## Related Issue(s)

- Closes #440

## Stacked PR — read this first

This PR is **stacked on top of #413** (`412_versionOutputDatasets`). Its base
branch is `412_versionOutputDatasets`, **not `dev`**, so the diff here shows
only the manifest work and none of #413's changes.

## Changes Made

- **New module `asf_heat_pump_suitability/utils/manifest_utils.py`** (the core of
  the PR). It provides:
  - `generate_dict_run_manifest(...)` — builds the manifest dict with keys
    `stage`, `local_authority`, `run_at`, `git_commit`, `input_versions`,
    `row_count`, `params`. This is a **pure function** (no file I/O), so it's
    easy to unit-test on its return value.
  - `save_manifest_to_s3(...)` — writes the manifest as JSON next to the
    output. Named `{output_basename}.manifest.json` (e.g.
    `plymouth_domestic_uprns.parquet` -> `plymouth_domestic_uprns.manifest.json`)
    so it can never collide with the unrelated `manifest.json` that
    `create_manifest.py` writes to the front-end bucket.
  - `generate_dict_input_versions(...)` + `STAGE_INPUT_KEYS` — a **curated
    per-stage list** of the `config["data"]` datasets each stage actually
    reads, defined in `base.yaml` under `run_manifest.stage_input_keys`.
    Each stage's manifest records only its own inputs (as raw resolved
    path strings), so a diff points at inputs that could have changed the
    output rather than 27 identical keys. 
- **Edits to the five entrypoints** (`transform/uprns.py`,
  `run/add_features.py`, `transform/decision_tree.py`, `cluster/cluster.py`,
  `run/compute_contextual_features.py`): each now writes its manifest inside
  the same `if args.save:` block as its output write, so a manifest only
  appears when the output does. `decision_tree` has two outputs, so it writes
  two manifests (six manifests total across the five stages).
- **Unit tests** in `utils/tests/test_manifest_utils.py` (17 tests).
- **Spec** `docs/specs/440_addRunManifest.md` records the design and every
  decision made during kickoff, implementation, and review.

### What to Focus On

- **`run_manifest.stage_input_keys` in `base.yaml`** — this is the part most
  likely to drift over time. Each entry is a `config["data"]` key. Please
  sanity check that the datasets listed for a stage match what that stage
  actually reads. A test asserts every listed key
  resolves in config, but it can't tell you if a real input is *missing* from
  a list.
- **The degrade don't fail choices**: `save_manifest_to_s3` swallows and logs
  any write exception, and `get_str_git_commit` falls back to `"unknown"`.
  These are deliberate and worth looking at to see if you agree. 
- **`decision_tree`'s save path** — resolved at review: saves now route through `save_utils.save_to_s3` at the entrypoint, deriving each output path once for both the write and its manifest (d9a2fd5).

### How to Test

Run the unit tests:

```
uv run pytest asf_heat_pump_suitability/utils/tests/test_manifest_utils.py
```

Expected: 17 passing tests, covering the manifest dict contents and exact key
set, curated input version resolution (correct selection; loud `KeyError` on
an unknown key or a key that resolves to a subtree; every `STAGE_INPUT_KEYS`
entry resolves), the swallowed and logged write failure, the `git_commit`
fallback, and the manifest path derivation.

To see it end-to-end , run any entrypoint with
`--save` and check that a `*.manifest.json` lands next to the output in S3, e.g.:

```
python asf_heat_pump_suitability/pipeline/transform/uprns.py --local_authorities east_lothian --save
```

## Checklist

- [x] Code has been tested locally (17 unit tests pass)
- [x] No local file dependencies (unit tests run with no AWS/S3 access)
- [x] Docstrings added on every new function; module docstring explains the
      naming choice
- [x] Changes are scoped to one task (the run manifest) — `decision_tree`'s
      save-path refactor was initially left out, then folded in at review
      (agreed with @crispy-wonton it was quick, and it removes a double
      path derivation that risked the manifest pointing at the wrong path)

---

Closes #440. Stacked on #413.


helloaidank:
**Open question — where should `stage_input_keys` live?**

The curated per stage input lists currently sit in `base.yaml` under `run_manifest.stage_input_keys` (~45 lines in an already 390-line file). They're read by exactly one module (`utils/manifest_utils.py`), which raises the question of whether single consumer config belongs in the shared file at all.

A decision to be made whether we:

1. **Keep in `base.yaml`** (as implemented). One config file, matches the `features:` / `tech_types:` idiom of curated lists.
2. **Multi-file config loader** (merge a `config/` directory into the global `config` object). This is likely the general fix for config growth, but touches the import path every module uses, so it would be its own issue/PR.

Parking this rather than extending this PR further, happy to hear what y'all think / preferences. None of the options change the manifest JSON contract.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -480,11 +480,33 @@ def identify_gdf_tuple_most_suitable_tech_uprn_and_building(
...
 
-    identify_gdf_tuple_most_suitable_tech_uprn_and_building(
-        local_authorities=local_authority_dict["url_slug"],
-        buildings_gdf=buildings_gdf,
-        id_col="ID",
-        uprns_gdf=uprns_with_features_gdf,
-        save=args.save,
-        release_date=release_date,
+    uprns_tech_gdf, buildings_tech_gdf = (
+        identify_gdf_tuple_most_suitable_tech_uprn_and_building(
```
helloaidank, line 459: We have an inconsistency here around how we're saving the files, the save happens within the transform function itself. So for this script we have to derive the paths which isn't the cleanest. Maybe worth thinking about whether we want consistency in saving across all the pipeline scripts? Would be a quick change.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -905,11 +905,19 @@ def parse_arguments() -> argparse.Namespace:
...
-                release_date=release_date,
-                local_authorities=local_authority_dict["url_slug"],
-            ),
+        output_path = save_utils.get_str_output_path(
+            "tech_clusters",
+            release_date=release_date,
+            local_authorities=local_authority_dict["url_slug"],
+        )
+        save_utils.save_to_s3(clusters_gdf, output_path)
+        manifest_utils.save_run_manifest_to_s3(
```
helloaidank: made a conscious decision to put this save in a domain specific file rather than save_utils.

asf_heat_pump_suitability/config/base.yaml
```diff
@@ -343,3 +339,49 @@ tech_types:
   - "SGL_N"
   - "HN_S"
   - "HN_N"
+
+run_manifest:
+  # Remember to update stage_input_keys below when a stage starts reading a new input data source.
```
crispy-wonton: question: is there a way to automate this or even better just use the original keys? I'd prefer to update in one place only.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: makes sense

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -351,11 +351,20 @@ def parse_arguments() -> argparse.Namespace:
...
+        save_utils.save_to_s3(features_df, path=output_path)
+        manifest_utils.save_run_manifest_to_s3(
+            output_path,
+            stage="add_features",
+            local_authority=local_authority_dict["url_slug"],
+            row_count=len(features_df),
+            params={
+                "local_authorities": args.local_authorities,
+                "release_date": release_date,
+                "detail": args.detail,
```
crispy-wonton, line 368: praise: nice addition of the detail level!

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -454,3 +426,17 @@ def create_json_contextual_features_metadata(
                 "s3://", front_end_s3_bucket, front_end_staging_s3_path, file_name
             ),
         )
+
+        # Only the dated data-science copy gets a run manifest; the undated
+        # front-end copy above is overwritten every run, so there is no
+        # version history to attach lineage to.
```
crispy-wonton, line 460: praise: good annotation, thanks!

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
crispy-wonton: Agree, we should separate the saving out as in the other scripts. Feel free to do it here if it's quick but otherwise, happy for this to go into a separate issue.

asf_heat_pump_suitability/utils/manifest_utils.py
```diff
@@ -0,0 +1,185 @@
...
+import json
+import logging
+import subprocess
+from datetime import datetime, timezone
+
+import fsspec
+
+from asf_heat_pump_suitability import PROJECT_DIR, config
+
+MANIFEST_SUFFIX = ".manifest.json"
```
crispy-wonton, line 21: question: I haven't seen this style of file naming before, how come it is .manifest.json? is that 2 file extensions stacked?

asf_heat_pump_suitability/utils/manifest_utils.py
```diff
@@ -0,0 +1,185 @@
...
+
+STAGE_INPUT_KEYS = config["run_manifest"]["stage_input_keys"]
+
+
+def get_str_git_commit() -> str:
+    """
+    Get the git commit hash of the checked-out code.
+
+    Runs in the project directory so the hash reflects the imported package,
+    not the caller's working directory.
```
crispy-wonton: question: why would the imported package be different from the caller's working dir?

asf_heat_pump_suitability/utils/manifest_utils.py
```diff
@@ -0,0 +1,185 @@
...
+
+    Returns:
+        str: 40-character commit hash, or "unknown" if git is unavailable
+    """
+    try:
+        return subprocess.run(
+            ["git", "rev-parse", "HEAD"],
+            cwd=PROJECT_DIR,
+            capture_output=True,
```
crispy-wonton, line 47: question: have read the docs for `capture_output` but still don't really understand what it's doing - why have you set to True here? Is it to capture errors or more than that?

asf_heat_pump_suitability/utils/manifest_utils.py
```diff
@@ -0,0 +1,185 @@
...
+            "Could not read git commit hash; recording '%s' in run manifest.",
+            UNKNOWN_GIT_COMMIT,
+        )
+        return UNKNOWN_GIT_COMMIT
+
+
+def generate_dict_input_versions(input_keys: list[str]) -> dict:
+    """
+    Resolve config["data"] key paths to their dataset path strings, whose
+    dated prefixes carry the input versions.
```
crispy-wonton: issue: i'm finding these docstrings quite jargony and hard to understand - is it possible to make them a bit more reader friendly?

asf_heat_pump_suitability/utils/manifest_utils.py
```diff
@@ -0,0 +1,185 @@
...
+
+    Args:
+        input_keys (list[str]): dot-separated key paths into `config["data"]`,
+            typically one of the `STAGE_INPUT_KEYS` lists
```
crispy-wonton:
```suggestion
        input_keys (list[str]): dot-separated key paths in `config["data"]`,
            typically one of the `STAGE_INPUT_KEYS` lists
```
question: is this meant to say 'in'?

asf_heat_pump_suitability/utils/manifest_utils.py
```diff
@@ -0,0 +1,185 @@
...
+                raise KeyError(
+                    f"Run manifest input key '{dot_key}' not found in config['data']"
+                ) from error
+        if isinstance(value, dict):
+            raise KeyError(
+                f"Run manifest input key '{dot_key}' resolves to a config['data'] "
+                "subtree, not a dataset path"
+            )
+        input_versions[dot_key] = value
+    return input_versions
```
crispy-wonton, line 102: suggestion: this could benefit from some annotation

asf_heat_pump_suitability/utils/manifest_utils.py
```diff
@@ -0,0 +1,185 @@
...
+    input_versions = {}
+    for dot_key in input_keys:
+        value = config["data"]
+        for part in dot_key.split("."):
+            try:
+                value = value[part]
+            except (KeyError, TypeError) as error:
+                raise KeyError(
+                    f"Run manifest input key '{dot_key}' not found in config['data']"
+                ) from error
```
crispy-wonton, line 93: question: what is the `raise KeyError() from error` doing? as in, why do we need the `from error` at the end?

docs/specs/440_addRunManifest.md
crispy-wonton, line 1: Have not read. Let me know if I should.

asf_heat_pump_suitability/utils/manifest_utils.py
```diff
@@ -0,0 +1,185 @@
...
+    local_authority: str,
+    row_count: int,
+    params: dict,
+    input_keys: list[str] | None = None,
+) -> dict:
+    """
+    Generate run manifest dict recording the lineage of one pipeline output.
+
+    Args:
+        stage (str): pipeline entrypoint that produced the output, e.g. "uprns"
```
crispy-wonton:
```suggestion
        stage (str): pipeline script that produced the output, e.g. "uprns"
```
question: would it be incorrect to just refer to this as script? what will happen if you run the whole pipeline end-to-end. Will all the outputs be labelled as 'uprns' stage (the first script) or will they all have their own script as the stage?

Same where you have this arg in other functions.

asf_heat_pump_suitability/utils/manifest_utils.py
```diff
@@ -0,0 +1,185 @@
...
+        local_authority (str): local authority slug the output was generated for
+        row_count (int): number of rows (or geojson features) in the output
+        params (dict): CLI arguments the entrypoint was run with
```
crispy-wonton: suggestion: could add format of dict to the documentation e.g. `arg name: arg value`. Also, isn't this specifically **additional** arguments as you are retrieving the local authority name above in a separate kwarg, and that's the main CLI of most of the scripts.

Same where you have this arg in other functions.

asf_heat_pump_suitability/utils/manifest_utils.py
```diff
@@ -0,0 +1,185 @@
...
+
+    Returns:
+        dict: run manifest
+    """
+    if input_keys is None:
+        input_keys = STAGE_INPUT_KEYS[stage]
+    return {
+        "stage": stage,
+        "local_authority": local_authority,
+        "run_at": datetime.now(timezone.utc).isoformat(timespec="seconds"),
```
crispy-wonton, line 138: question: does the timezone match the timezone of the S3 bucket? otherwise I can see this causing some confusion.

asf_heat_pump_suitability/utils/manifest_utils.py
```diff
@@ -0,0 +1,185 @@
...
+    """
+    Get the S3 path of the run manifest for an output file.
+
+    Args:
+        output_path (str): S3 path of the output file the manifest describes
+
+    Returns:
+        str: co-located path ending `.manifest.json`
+    """
+    return output_path.rsplit(".", 1)[0] + MANIFEST_SUFFIX
```
crispy-wonton: suggestion: better to use `pathlib` here. More robust and readable.

`from pathlib import Path`
```suggestion
    return Path(output_path).stem + MANIFEST_SUFFIX
```

asf_heat_pump_suitability/utils/manifest_utils.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def save_manifest_to_s3(manifest: dict, output_path: str) -> None:
+    """
+    Save run manifest as JSON next to the output file it describes.
+
+    Write failures are logged and swallowed: lineage degrades rather than
+    aborting a pipeline run.
```
crispy-wonton: question: what does 'swallowed' mean here? and what does `lineage degrades` mean?

asf_heat_pump_suitability/utils/manifest_utils.py
```diff
@@ -0,0 +1,185 @@
...
+    manifest_path = get_str_manifest_path(output_path)
+    logging.info(f"Saving run manifest to {manifest_path}")
+    try:
+        with fsspec.open(manifest_path, "w") as f:
+            json.dump(manifest, f, ensure_ascii=False, indent=2)
+    except Exception:
+        logging.warning(
+            "Failed to write run manifest to %s; continuing without it.",
+            manifest_path,
+            exc_info=True,
```
crispy-wonton, line 212: question: can you link me to the docs for this `kwarg`?

asf_heat_pump_suitability/utils/tests/test_manifest_utils.py
```diff
@@ -0,0 +1,209 @@
...
+import pytest
+
+from asf_heat_pump_suitability import config
+from asf_heat_pump_suitability.utils import manifest_utils
+
+
+class TestGetStrGitCommit:
+    """Tests for `get_str_git_commit`."""
+
+    def test_returns_current_commit_hash(self):
```
crispy-wonton: issue: this name is slightly misleading as it doesn't check that it returns the current commit hash, only that **a** commit hash is returned. I think this is a good test and would keep it. I also think it would be great if there's a way to test that the most recent hash is returned. Is there a way to mock a hash here?

asf_heat_pump_suitability/utils/tests/test_manifest_utils.py
crispy-wonton, line 1: issue: general note - please could you add info docstrings that are printed to console if the assert statements fail?

asf_heat_pump_suitability/utils/tests/test_manifest_utils.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def test_unknown_key_raises_keyerror(self):
+        """A key path missing from config["data"] fails loudly at run time rather
+        than silently omitting lineage."""
+        with pytest.raises(KeyError, match="epc.nonexistent"):
+            manifest_utils.generate_dict_input_versions(["epc.nonexistent"])
+
+    def test_subtree_key_raises_keyerror(self):
+        """A key path resolving to a config subtree rather than a dataset path
+        string is a curated-list mistake and fails loudly."""
```
crispy-wonton: question: what is a 'curated-list mistake'?

asf_heat_pump_suitability/utils/tests/test_manifest_utils.py
```diff
@@ -0,0 +1,209 @@
...
+            )
+            == "s3://bucket/outputs/data/plymouth/20260722/plymouth_clusters_contextual_features_5m.manifest.json"
+        )
+
+    def test_filename_never_collides_with_front_end_manifest_json(self):
+        """The derived filename keeps the output basename, so it can never be the
+        bare `manifest.json` owned by pipeline/run/create_manifest.py."""
+        path = manifest_utils.get_str_manifest_path("s3://bucket/dir/output.parquet")
+        assert path.split("/")[-1] != "manifest.json"
+        assert path.endswith(".manifest.json")
```
crispy-wonton: suggestion: again, I'd use `pathlib`'s `Path` here.

asf_heat_pump_suitability/utils/tests/test_manifest_utils.py
```diff
@@ -0,0 +1,209 @@
...
+        hands it to `save_manifest_to_s3` with the output path."""
+        save = mocker.patch.object(manifest_utils, "save_manifest_to_s3")
+        manifest_utils.save_run_manifest_to_s3(
+            "s3://bucket/dir/output.parquet",
+            stage="uprns",
+            local_authority="plymouth",
+            row_count=123,
+            params={"local_authorities": ["plymouth"], "release_date": "20260722"},
+        )
+        manifest, output_path = save.call_args.args
```
crispy-wonton:
```suggestion
        # args passed to 'save_manifest_to_s3' within the 'save_run_manifest_to_s3' function
        manifest, output_path = save.call_args.args
```
This confused me initially as the names are so similar. Might be worth updating the names as well.

crispy-wonton (changes requested):
hey @helloaidank :) 

Great work - the test suite looks comprehensive and the code is clean and effective.

I have left some clarifying questions for you plus some suggestions.

My main suggestions are:
- Update some of the docstrings to be less jargon-heavy. Some of them are a bit difficult to understand as they currently are.
- Is there a way to avoid creating a separate list of 'STAGE_INPUT_KEYS' or automating the creation of it? It just makes the manifest creation very brittle as it will fail unless the keys are updated in two places if we ever make changes. I understand that one of your tests will fail because you check for matches between STAGE INPUT KEYS and config data keys so that's not too bad but still inconvenient and extra work. Also, if we add input keys but forget to add them to the manifest stage input keys, I don't think this will create any errors or test failures and so will silently fail at adding all the input keys into the manifest.
- Suggest using `pathlib` and `Path` when working with paths as it's more robust and fit-for-purpose than dealing with paths as strings.

asf_heat_pump_suitability/config/base.yaml
helloaidank: There's not really a single existing list to point at within this config, I guess we could do this for some of the input sources?

Each stage's inputs aren't always declared somewhere machine readable today. A lot of it is implicit, in terms of which getters a script runs within transform/run code. We could make every getter self-declare its config key at the point of definition but this would require a refactor of the getters (probably out of scope for this PR).

helloaidank: I am quite keen to have this as well tbh, it's annoying having to potentially update in two places.

asf_heat_pump_suitability/utils/manifest_utils.py
helloaidank: it's just a compound suffix, same pattern as say` .tar.gz`, motivation was so it would never collide with the unrelated flat `manifest.json` which `pipeline/run/create_manifest.py`writes to the front-end bucket.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
helloaidank: Done in d9a2fd5 — saves now go through `save_utils.save_to_s3` in `__main__`, deriving each output path once and sharing it with the manifest write, so the manifest can't drift from where the output actually lands. `identify_gdf_tuple_most_suitable_tech_uprn_and_building` is a pure transform now (dropped its save params; no external callers). Also added `pipeline/transform/tests/test_decision_tree.py`, including a test pinning that the transform does no saving.

asf_heat_pump_suitability/utils/manifest_utils.py
helloaidank: Not explained well enough in the documentation, it captures both stdout and stderr into `result.stdout/result.stderr `instead of letting them print to the terminal/logs, it's so .stdout.strip() can grab the commit hash that git rev-parse HEAD prints, without this the hash would just print to the console and the result.stdout would be None.

asf_heat_pump_suitability/utils/manifest_utils.py
helloaidank: Will do a scan and update doc strings so that things are clearer.

asf_heat_pump_suitability/utils/manifest_utils.py
helloaidank: yeah happy to change to script. They will have their own stage label, it's hard-coded within `STAGE_INPUT_KEYS`.

asf_heat_pump_suitability/utils/manifest_utils.py
helloaidank: here you go: https://docs.python.org/3/library/logging.html#logging.Logger.debug

docs/specs/440_addRunManifest.md
helloaidank: Fine not to

asf_heat_pump_suitability/utils/manifest_utils.py
helloaidank: Good question! `raise KeyError(...) from error` chains the two exceptions together. Python shows both tracebacks either way, but the label between them changes. Without `from error` it says *"During handling of the above exception, another exception occurred"* — which reads as if our error handling itself broke by accident. With `from error` it says the original error *"was the direct cause of the following exception"* — i.e. we caught the original error on purpose and re-raised it with a clearer message, and here's the original underneath for debugging. So it's purely about making the traceback honest and readable...no behaviour change.

asf_heat_pump_suitability/utils/manifest_utils.py
helloaidank: Both good suggestions — implemented in 80ab585. The docstring now describes `params` as the **additional** CLI arguments and shows the dict shape: `{argument name: value}, e.g. {"release_date": "20260722"}`. Same wording applied on the wrapper function.

asf_heat_pump_suitability/utils/manifest_utils.py
helloaidank: Fair point, that was jargon, reworded in 80ab585. It now reads: *"If the write fails, the error is logged as a warning and not raised: a failed manifest write should never crash a pipeline run whose output has already saved successfully. The run just goes without its manifest."*

asf_heat_pump_suitability/utils/manifest_utils.py
helloaidank: Good thing to check, but there's no mismatch possible here. S3 buckets don't have a timezone of their own — S3 records every file's timestamp in UTC, whichever region the bucket is in. And `run_at` is also UTC (`datetime.now(timezone.utc)`). The timestamp string even carries its timezone with it — it ends `+00:00` — so whoever reads the manifest doesn't have to guess. The manifest time and S3's own file timestamps will always be directly comparable. Added a short comment in the code saying this (40bffc7).

asf_heat_pump_suitability/utils/manifest_utils.py
helloaidank: Really good instinct — the raw string split here *was* fragile — but `pathlib` specifically can't be used on these paths, because they're S3 URLs rather than real file paths. `Path` doesn't know `s3://` is a protocol; it just applies normal filesystem rules, and two things go wrong:

```python
>>> Path("s3://bucket/dir/output.parquet").stem
'output'
```

`.stem` means "just the filename, minus its extension" — the bucket and all the folders are dropped by design, so the manifest would be written to a bare filename with no location instead of next to its output (two of the path tests fail under this for exactly that reason). `pathlib` also collapses the double slash, turning `s3://bucket` into `s3:/bucket`.

So I've gone with `os.path.splitext` instead (f2641e0) — it's string-based, so it leaves the URL alone, but it only ever splits the extension off the final part of the path, never a folder:

```python
>>> os.path.splitext("s3://bucket/dir/output.parquet")[0] + ".manifest.json"
's3://bucket/dir/output.manifest.json'
```

asf_heat_pump_suitability/utils/tests/test_manifest_utils.py
helloaidank: Same reasoning as on the main thread (https://github.com/nestauk/asf_heat_pump_suitability/pull/449#discussion_r3750819063) — these tests assert on the full S3 URLs as plain strings, and they're exactly what catches the `Path.stem` problem, so keeping them string-based. The implementation now uses `os.path.splitext` (f2641e0) and these tests pass unchanged.

asf_heat_pump_suitability/utils/tests/test_manifest_utils.py
helloaidank: Both done in 23222df: renamed to `test_returns_a_valid_commit_hash_format` so the name matches what it checks, and added the mock test you suggested — `test_returns_exactly_the_hash_git_reports` patches `subprocess.run` to return a known hash (with a trailing newline, as git prints it) and asserts the function returns exactly that hash, newline stripped.

asf_heat_pump_suitability/utils/tests/test_manifest_utils.py
helloaidank: Done in 23222df — every assert in the file now carries a plain-English message that pytest prints on failure, e.g. `"git_commit is neither a 40-character hex hash nor 'unknown'"`. Will carry this pattern forward as the default for new tests.

asf_heat_pump_suitability/utils/tests/test_manifest_utils.py
helloaidank: "curated list" meant the hand-maintained `stage_input_keys` lists in base.yaml, and a "curated-list mistake" is the likely typo when hand-editing them: writing a key that stops at a *group* of datasets (e.g. `"epc"`) instead of a single dataset path (e.g. `"epc.domestic"`). This test checks that typo fails loudly instead of quietly recording the wrong thing. Reworded the docstring in 23222df so it says exactly that.

asf_heat_pump_suitability/utils/tests/test_manifest_utils.py
helloaidank: you're right,`save_manifest_to_s3` vs `save_run_manifest_to_s3` are very similar with nothing saying which one a script should call. Renamed the wrapper to `generate_and_save_run_manifest_to_s3` in 733ef69 (the name now says it builds the manifest *and* writes it, while `save_manifest_to_s3` keeps its narrow write-only meaning). Updated everywhere — the five pipeline scripts, tests, and the annotation comment here.

asf_heat_pump_suitability/utils/manifest_utils.py
helloaidank: Because the pipeline can be launched from anywhere — another folder, a notebook, a scheduled job — and `git rev-parse HEAD` answers for whatever git repo the *current directory* happens to be in. If you ran one of our scripts while sat inside a different repo, we'd record that repo's commit hash by mistake. Pinning `cwd=PROJECT_DIR` makes git always answer for this repo, i.e. the code that's actually running. Reworded the docstring to spell this out (80ab585).

asf_heat_pump_suitability/utils/manifest_utils.py
helloaidank: Yes — typo, thanks. That whole docstring got reworded as part of the plain-English pass (80ab585) and now reads "key paths in `config["data"]`" with a worked example (`"epc.domestic"`).

asf_heat_pump_suitability/utils/manifest_utils.py
helloaidank: Done in 80ab585 — the loop now has worked-example comments (how the dot-key walk descends `config["data"]`, and what hitting a dict means).

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
crispy-wonton: amazing! thank you!

asf_heat_pump_suitability/utils/manifest_utils.py
crispy-wonton: ah great - thanks for looking into that and explaining :D

asf_heat_pump_suitability/utils/tests/test_manifest_utils.py
```diff
@@ -41,17 +55,18 @@ def test_resolves_exactly_the_requested_keys(self):
...
             manifest_utils.generate_dict_input_versions(["epc.nonexistent"])
 
     def test_subtree_key_raises_keyerror(self):
-        """A key path resolving to a config subtree rather than a dataset path
-        string is a curated-list mistake and fails loudly."""
+        """A key that stops at a group of datasets (e.g. "epc") rather than a
+        single dataset path (e.g. "epc.domestic") — the likely typo when
+        hand-editing the stage_input_keys lists in base.yaml — fails loudly."""
         with pytest.raises(KeyError, match="subtree"):
             manifest_utils.generate_dict_input_versions(["epc"])
```
crispy-wonton, line 71: issue (minor; non-blocking): missing statement printed when assertion fails

asf_heat_pump_suitability/utils/tests/test_manifest_utils.py
```diff
@@ -41,17 +55,18 @@ def test_resolves_exactly_the_requested_keys(self):
...
             ]["UK_ons_lad_bounds"],
-        }
+        }, "Resolved input versions did not match the config['data'] paths"
 
     def test_unknown_key_raises_keyerror(self):
         """A key path missing from config["data"] fails loudly at run time rather
-        than silently omitting lineage."""
+        than silently omitting the input from the manifest."""
         with pytest.raises(KeyError, match="epc.nonexistent"):
             manifest_utils.generate_dict_input_versions(["epc.nonexistent"])
```
crispy-wonton, line 64: suggestion: missing error message if test fails

asf_heat_pump_suitability/utils/manifest_utils.py
```diff
@@ -51,22 +58,31 @@ def get_str_git_commit() -> str:
...
 def generate_dict_input_versions(input_keys: list[str]) -> dict:
     """
-    Resolve config["data"] key paths to their dataset path strings, whose
-    dated prefixes carry the input versions.
+    Look up the S3 path of each input dataset, as a record of which version
+    of the data was used.
+
+    Input dataset files in this repo carry a date in their filename or folder
+    (e.g. `202412_v1_...`), so the full path doubles as a version record: if
+    the path is different between two runs, the input data changed.
```
crispy-wonton: docs: I think it's likely we'll change the format of the input file names so I'd remove this reference so that it doesn't later become outdated and confusing.

asf_heat_pump_suitability/utils/manifest_utils.py
```diff
@@ -51,22 +58,31 @@ def get_str_git_commit() -> str:
...
-            in a curated list must fail loudly, not silently omit lineage
+        KeyError: if a key is missing from `config["data"]`, or stops at a
+            group of datasets rather than a single dataset path. A typo in
+            the config lists should stop the run with a clear error, not
+            quietly leave an input out of the manifest.
     """
     input_versions = {}
     for dot_key in input_keys:
+        # Walk down config["data"] one key at a time, e.g. "epc.domestic"
+        # -> config["data"]["epc"] -> config["data"]["epc"]["domestic"]
```
crispy-wonton, line 85: praise: nice! This is helpful :)

asf_heat_pump_suitability/utils/manifest_utils.py
```diff
@@ -75,6 +91,8 @@ def generate_dict_input_versions(input_keys: list[str]) -> dict:
                 raise KeyError(
                     f"Run manifest input key '{dot_key}' not found in config['data']"
                 ) from error
+        # Reaching a dict means the key stopped at a group of datasets
+        # (e.g. "epc") instead of a single dataset path (e.g. "epc.domestic")
```
crispy-wonton, line 95: helpful!

asf_heat_pump_suitability/utils/manifest_utils.py
crispy-wonton: nice addition with the annotation thanks!

asf_heat_pump_suitability/utils/manifest_utils.py
```diff
@@ -158,15 +183,18 @@ def get_str_manifest_path(output_path: str) -> str:
     Returns:
         str: co-located path ending `.manifest.json`
     """
-    return output_path.rsplit(".", 1)[0] + MANIFEST_SUFFIX
+    # os.path.splitext, not pathlib: Path mangles "s3://" URLs and Path.stem
+    # drops the directory. splitext only touches the extension.
```
crispy-wonton, line 187: good addition! I can see myself coming back with the same question later otherwise!

asf_heat_pump_suitability/pipeline/transform/tests/test_decision_tree.py
```diff
@@ -0,0 +1,125 @@
...
+def uprns_gdf():
+    """Three UPRNs with the decision tree's input features: two flats in B1
+    (one in a city centre / heat network zone), one house in B2."""
+    return gpd.GeoDataFrame(
+        {
+            "UPRN": [1, 2, 3],
+            "ID": ["B1", "B1", "B2"],
+            "in_block_of_flats": [True, True, False],
+            "max_contiguous_outdoor_space_area_m2": [0.0, 0.0, 120.0],
+            "in_city_centre_or_hn_zone": [True, True, False],
```
crispy-wonton, line 54: note: because we are removing the city centre / HN zone branch of the decision tree in a separate PR, this test will fail at some point. It can be resolved in whichever PR is merged second - so just remember to merge changes from dev before merging this PR! :)

crispy-wonton, line 54: suggestion: we should also add a separate fixture and test to test that the function can handle buildings containing different combinations of UPRNs which have different attributes (e.g. one has a garden above  and another below the threshold). This can definitely be done in a separate PR.

asf_heat_pump_suitability/pipeline/transform/tests/test_decision_tree.py
```diff
@@ -0,0 +1,125 @@
...
+                buildings_gdf=buildings_gdf,
+                id_col="ID",
+                uprns_gdf=uprns_gdf,
+            )
+        )
+        techs_per_building = uprns_tech_gdf.groupby("ID")["assigned_tech"].nunique()
+        assert (techs_per_building == 1).all(), (
+            f"Buildings with more than one assigned tech: "
+            f"{techs_per_building[techs_per_building > 1].index.tolist()}"
+        )
```
crispy-wonton, line 111: great test addition!

asf_heat_pump_suitability/pipeline/transform/tests/test_decision_tree.py
```diff
@@ -0,0 +1,125 @@
...
+        entrypoint, so no output path is built and nothing is written."""
+        mock_output_path = mocker.patch.object(save_utils, "get_str_output_path")
+        mock_save = mocker.patch.object(save_utils, "save_to_s3")
+        decision_tree.identify_gdf_tuple_most_suitable_tech_uprn_and_building(
+            buildings_gdf=buildings_gdf,
+            id_col="ID",
+            uprns_gdf=uprns_gdf,
+        )
+        mock_output_path.assert_not_called()
+        mock_save.assert_not_called()
```
crispy-wonton: thought: this seems like a weird test to have here. I understand why you added it as it confirms that you've updated the script correctly but I would probs delete this as the function is no longer intending to save data and it doesn't have any save kwarg or anything. Basically, this test seems a bit out of scope now.

asf_heat_pump_suitability/pipeline/transform/tests/test_decision_tree.py
crispy-wonton, line 1: praise: really nice addition with these tests - thank you!

crispy-wonton (commented):
hey @helloaidank 

Great updates and changes!! I have made some minor suggestions but otherwise, looks great.

asf_heat_pump_suitability/pipeline/transform/tests/test_decision_tree.py
helloaidank: Ok cool makes sense!

asf_heat_pump_suitability/pipeline/transform/tests/test_decision_tree.py
helloaidank: yeah, you're right, we should delete it as it's a bit of a legacy test.

asf_heat_pump_suitability/utils/manifest_utils.py
helloaidank: Thank you!

asf_heat_pump_suitability/pipeline/transform/tests/test_decision_tree.py
helloaidank: Done in d9319e3 — deleted the test along with the `save_utils` import, which only existed for it.

asf_heat_pump_suitability/pipeline/transform/tests/test_decision_tree.py
helloaidank: Agreed, put this as an issue #476.

asf_heat_pump_suitability/utils/manifest_utils.py
helloaidank: Done in d9319e3 — dropped the filename format example. The docstring is much more general so should stand the test of time.

asf_heat_pump_suitability/utils/tests/test_manifest_utils.py
helloaidank: No message slot on this one I'm afraid....it's a `pytest.raises` block rather than an `assert`. If the code fails to raise, pytest fails the test with its own built-in message (`Failed: DID NOT RAISE <class 'KeyError'>`), and that text can't be customised (`message=` was removed in pytest 5 because people confused it with `match=`).

asf_heat_pump_suitability/utils/tests/test_manifest_utils.py
helloaidank: Same situation as `test_unknown_key_raises_keyerror` above — this is a `pytest.raises` block, not an assert, so pytest supplies its own failure message (`Failed: DID NOT RAISE <class 'KeyError'>`) and a custom one can't be attached.

asf_heat_pump_suitability/pipeline/transform/tests/test_decision_tree.py
helloaidank: Merged `dev` in — currently already up to date (the city-centre removal hasn't landed on `dev` yet). Will re-merge just before merging this PR and fix the fixture if that lands first.

asf_heat_pump_suitability/utils/tests/test_manifest_utils.py
helloaidank: Could do a try except if strong feelings about this!

crispy-wonton: all good - ty for explaining!

crispy-wonton (approved):

PR #458, opened 2026-07-29
452 change decision tree to assign only individual/networked HP/communal and add layer with areas of DHN potential

### LINKED ISSUE #452: Decision tree update
Remove district heat network from decision tree and add it as separate layer.

sofiapinto:
---

# Description

Changes to the code include: 
- `decision_tree.py`: changes decision tree logic to assign only individual, networked HP or communal. DHN will be shown as a separate layer.
- `add_features.py`: flags `in_city_centre` and `in_hn_zone` no longer required at UPRN level, so code to create these has been removed
- `cluster.py`: remove code to append HN zones. These are added a separate layer later in the pipeline
- `compute_contextual_features.py`: we create a separate layer of data with DHN potential. This includes DESNZ HN zones as layers as well as spatial signature polygons. For this layer, there's a feature called `annotation` to enable distinguishing between the different sources of DHN potential.
- `load_geodata.py`: changes to `load_gdf_spatial_signatures_gb()` to enable allow filtering by boundary/LA and signature type. Changes both to `load_gdf_spatial_signatures_gb()` and `load_gdf_heat_network_zones()` to create a new column called annotation.

Might be useful to review the code in the order above.

Fixes #452 

# Instructions for Reviewer

**1) Please check you agree with the changes to the code & the logic still makes sense.** 

At UPRN level, this simplifies the decision tree quite substantially. 

image

That bottom branch will disappear once outdoor space is known for all UPRNs.

I also had to make changes to how we do aggregations at building level.

**2) Please also check the format of the new output and that you agree with it.** 

To load the geodata within the file you can do:

```
import geopandas as gpd

clusters_gdf = gpd.read_file(
        config["output"]["dataset"]["clusters_tech_contextual_info"].format(
            local_authorities="wiltshire",
            tolerance_m=5,
        ),
    ).to_crs(epsg=27700)

```
This will include both layers (clusters with filters OR the areas of DHN potential). And you can filter for which layer you want by doing 
- `clusters_gdf[clusters_gdf["layer"]=="clusters_with_contextual_features"]`. Note that he column `annotation` will be have missing values only.
- `clusters_gdf[clusters_gdf["layer"]=='areas_of_district_heat_network_potential']`: only the `geometry`, `layer` and `annotation` columns should be non missing.

If you load as a json, you can see how the json is organised:
```

# File path
file_path = config["output"]["dataset"]["clusters_tech_contextual_info"].format(
    local_authorities="wiltshire",
    tolerance_m=5,
)

import json
from smart_open import open 

# Load the raw GeoJSON data from S3
with open(file_path, "r") as f:
    geojson_data = json.load(f)

geojson_data.keys()
```
You'll see there are three keys: `dict_keys(['type', 'metadata', 'features'])`

You can extract a key, like the metadata, as usual:
```
# Extract your custom metadata block
file_metadata = geojson_data.get("metadata", {})
print("Metadata:", file_metadata)
```

**3) You can check the code still runs with (I have done so myself too):**

```
python asf_heat_pump_suitability/pipeline/run/add_features.py --local_authorities wiltshire
python asf_heat_pump_suitability/pipeline/transform/decision_tree.py --local_authorities wiltshire
python asf_heat_pump_suitability/pipeline/cluster/cluster.py --local_authorities wiltshire
python asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py  --local_authorities wiltshire
```


asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -81,56 +82,32 @@ def identify_dict_most_suitable_tech(
...
+    else:
+        if pd.isna(outdoor_space):
             return {
-                "assigned_tech": TECH_TYPES["heat_networks"],
-                "decision_tree_path": "1. blocks of flats, in HNZ/ city centre",
+                "assigned_tech": f"{TECH_TYPES['individual_or_networked']}",
+                "decision_tree_path": "4. Not in block of flats. Unknown outdoor space",
             }
-        else:
+        elif outdoor_space > OUTDOOR_SPACE_THRESHOLD_M2.get("outside_hn_zone"):
```
crispy-wonton, line 101: note: i guess we can delete the 'within_hn_zone' value from `config`.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -259,16 +235,11 @@ def assign_df_unique_solution(solutions_per_footprint_df: pl.DataFrame) -> pl.Da
     Assigns a unique solution for each building footprint based on the combination of solutions in the set
     and the median outdoor space of properties within the building footprint, following the logic below:
 
-    - If the set of solutions contains both "District heat network" and "Networked heat pump", assign "Communal".
-    - Else if the set of solutions contains "District heat network", assign "District heat network".
+    - If the set of solutions contains "Communal", assign "Communal"
     - Else if the set of solutions contains "Networked heat pump", assign "Networked heat pump".
-    - Else if the set of solutions contains "Communal", assign "Communal"
     - Else if the set of solutions contains both "Individual heat pump" and "Networked heat pump", assign:
```
crispy-wonton, line 240:
```suggestion
    - Else if the set of solutions contains "individual_or_networked", assign:
```
question: should it say this? as I understand from the code below, if the set of solutions contains both, it will still be labelled as 'networked'

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -280,23 +251,12 @@ def assign_df_unique_solution(solutions_per_footprint_df: pl.DataFrame) -> pl.Da
...
-            # this happens at the edge of HN zones
-            # where some properties are assigned district heat network and some networked GSHP
-            # due to being just outside the HN zone boundary
-            pl.when(
-                pl.col("assigned_tech").list.contains(TECH_TYPES["heat_networks"])
-                & pl.col("assigned_tech").list.contains(TECH_TYPES["networked"])
-            )
+            # communal solution takes precedence
+            pl.when(pl.col("assigned_tech").list.contains(TECH_TYPES["communal"]))
             .then(pl.lit(TECH_TYPES["communal"]))
```
crispy-wonton, line 256: question: I'm not saying I disagree here but i'm just noticing that the order of precedence has changed now. Previously, networked solutions took precedence over communal but now it's the other way round. Is this desired here?

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -312,20 +272,6 @@ def assign_df_unique_solution(solutions_per_footprint_df: pl.DataFrame) -> pl.Da
                 # In this case, the outdoor space is unknown for all properties in the building, so we assign "individual_or_networked"
                 .otherwise(pl.lit(TECH_TYPES["individual_or_networked"]))
             )
```
crispy-wonton, line 274: question: sorry if i'm missing some context here and also I appreciate this part of the code is unchanged in this PR, but this doesn't look like it checks if 50% of properties in the building have outdoor space measurements. Is that accounted for somehow in this step or are we missing it?

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -312,20 +272,6 @@ def assign_df_unique_solution(solutions_per_footprint_df: pl.DataFrame) -> pl.Da
                 # In this case, the outdoor space is unknown for all properties in the building, so we assign "individual_or_networked"
                 .otherwise(pl.lit(TECH_TYPES["individual_or_networked"]))
```
crispy-wonton, line 273: note: according to the docstring, this `otherwise` should result in `networked`. Either the code needs updating here or the docstring above does.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
crispy-wonton, line 233: thought: I don't think we should change it here and I don't think it's a high priority right now, but I just want to flag that we might want to revisit the aggregation methodology at some point. Looking at it again, I'm not sure that solution precedence is enough - maybe we also need to bring in proportions. E.g. if a terrace has 30 properties, and 29 are individual solutions and 1 is communal, currently this will assign all 30 to communal. Maybe that's desired but it seems maybe too strict, but anyway I just thought of this now looking at it again. 

Perhaps this is causing some of the weird islands of different tech types in terraced areas. Let me know if I'm forgetting something!

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -1,15 +1,13 @@
...
-- boolean flags to indicate whether UPRN is in a block of flats; in a heat network zone; and in a city centre; in a listed building; off-gas; near the coast; and in a protected area
+- boolean flags to indicate whether UPRN is in a block of flats; in a listed building; off-gas; near the coast; and in a protected area
 - estimated max contiguous and total outdoor space (m2)
 - EPC-derived features of tenure; attachment type of property; and current energy rating, solar PV info, estimated current energy consumption.
 
 Run:
 python asf_heat_pump_suitability/pipeline/run/add_features.py --local_authorities LOCAL_AUTHORITIES
 
-Set -- `--detail "simplified"` to use simplified spatial signature polygons to label city centres. The default is "full" which uses the fully detailed spatial signatures framework.
-
```
crispy-wonton, line 12: makes the optimisation a lot easier on this one 👯

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -2,12 +2,14 @@
 Script to compute contextual information for clusters including:
 - Proportion of attachment types, tenure types, EPC ratings of properties within clusters
 - Median outdoor space of properties within clusters
-- Whether any properties within clusters are in HN zones, city centres, protected areas, off-gas, within 1500m of coastline
+- Whether any properties within clusters are in protected areas, off-gas, within 1500m of coastline
```
crispy-wonton, line 5: suggestion: you could add that we are adding HN zone layer here.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -273,14 +266,19 @@ def extend_df_contextual_features(
...
+    spatial_signatures_gdf: gpd.GeoDataFrame,
 ) -> gpd.GeoDataFrame:
     """
     Create geodataframe with cluster_id, geometry and contextual features for each cluster
 
     Args:
         uprns_df (pl.DataFrame): dataframe of UPRNs and UPRN-level features
         clusters_gdf (gpd.GeoDataFrame): geodataframe of clusters with geometry and cluster_id
+        hn_zones_gdf (gpd.GeoDataFrame): geodataframe of heat network zones with geometry and annotation
+        spatial_signatures_gdf (gpd.GeoDataFrame): geodataframe of spatial signatures with geometry and annotation
```
crispy-wonton:
```suggestion
        hn_zones_gdf (gpd.GeoDataFrame): geodataframe of heat network zones with geometry and source annotation
        city_centre_signatures_gdf (gpd.GeoDataFrame): geodataframe of spatial signatures representing city centre areas with geometry and source annotation
```
suggestion: to clarify this is city centre spatial signatures only because this confused me before I looked at the implementation of the code later in this script.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -298,21 +296,32 @@ def create_gdf_contextual_features(
...
-    return gpd.GeoDataFrame(
-        clusters_with_contextual_features_gdf, geometry="geometry", crs="EPSG:27700"
+    # Add in_hn_zone and in_city_centre flags to clusters_gdf
+    clusters_with_contextual_features_gdf["in_hn_zone"] = (
+        clusters_with_contextual_features_gdf.intersects(hn_zones_gdf.union_all())
+    )
+    clusters_with_contextual_features_gdf["in_city_centre"] = (
+        clusters_with_contextual_features_gdf.intersects(
+            spatial_signatures_gdf.union_all()
+        )
```
crispy-wonton, line 316: suggestion: it might be worth adding some CRS checks here using `geo_utils.verify_gdf_crs()` here.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -331,9 +341,17 @@ def create_json_contextual_features_metadata(
...
         clusters_with_contextual_features_gdf.to_json(drop_id=True)
     )
+
+    hn_potential_geojson = json.loads(hn_potential.to_json(drop_id=True))
+
+    for feature in clusters_json["features"]:
+        feature["properties"]["layer"] = "clusters_with_contextual_features"
+    for feature in hn_potential_geojson["features"]:
+        feature["properties"]["layer"] = "areas_of_district_heat_network_potential"
+
```
crispy-wonton, line 354: suggestion: it might be worth adding some CRS checks here using `geo_utils.verify_gdf_crs()` here.

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -93,6 +93,8 @@ def load_gdf_heat_network_zones(
     hn_gdf = _extend_gdf_hn_zone_id(hn_gdf)
     hn_gdf = geo_utils.verify_gdf_crs(gdf=hn_gdf)
 
+    hn_gdf["annotation"] = "DESNZ advanced heat network zoning"
```
crispy-wonton:
```suggestion
    hn_gdf["source_annotation"] = "DESNZ advanced heat network zoning"
```
Makes it more obvious what this is meant to represent.

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -170,9 +178,16 @@ def load_gdf_spatial_signatures_gb(
...
 
+    Allows filtering by:
+    - boundary or local authority: if specified, only polygons that intersect the boundary or local authority are returned. Otherwise, GB-wide polygons are returned.
+    - signature_types: if specified, only polygons of the specified spatial signature types are returned. Otherwise, all types are returned.
+
     Args:
         detail_level (str, optional): Which level of descriptive detail to load.
             Must be either "simplified" or "full". Defaults to "simplified".
+        signature_types (List[str], optional): Optional. List of spatial signature types to load. If None, all types are loaded.
+        boundary (shapely.Polygon | shapely.MultiPolygon | gpd.GeoDataFrame, optional): Optional. Boundary to load spatial signature polygons for or geodataframe of multiple boundary polygons.
```
crispy-wonton:
```suggestion
        boundary (shapely.Polygon | shapely.MultiPolygon | gpd.GeoDataFrame, optional): Optional. Boundary to load spatial signature polygons for, or geodataframe of multiple boundary polygons.
```

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -193,6 +208,48 @@ def load_gdf_spatial_signatures_gb(
         f"Spatial signatures {detail_level} geodataframe successfully loaded with CRS {gdf.crs}."
     )
 
+    if signature_types is not None:
+        gdf = gdf[gdf["type"].isin(signature_types)]
+        if gdf.empty:
+            print(
+                f"No spatial signatures found for the specified signature types: {signature_types}."
+            )
```
crispy-wonton, line 216: suggestion: perhaps this could raise a warning instead of print statement

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -193,6 +208,48 @@ def load_gdf_spatial_signatures_gb(
...
+        if isinstance(boundary, gpd.GeoDataFrame):
+            boundary = geo_utils.verify_gdf_crs(gdf=boundary)
+            gdf = gdf.sjoin(
+                boundary[["geometry"]], how="inner", predicate="intersects"
+            ).drop(columns="index_right")
+        else:
+
+            gdf = gdf[gdf["geometry"].intersects(boundary)]
+        if gdf.empty:
+            print("No spatial signatures found within given boundary.")
```
crispy-wonton, line 232: suggestion: as above re warning

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -193,6 +208,48 @@ def load_gdf_spatial_signatures_gb(
...
+            select_las=local_authorities_list
+        )
+
+        gdf = gdf.sjoin(
+            boundary_gdf[["geometry"]], how="inner", predicate="intersects"
+        ).drop(columns="index_right")
+        if gdf.empty:
+            print(
+                f"No spatial signatures found for Local Authority: {local_authority}."
+            )
```
crispy-wonton, line 247: suggestion: same here re warning

crispy-wonton (commented):
hi @sofiapinto !

Great work implementing this, it cleans the code up quite a lot and will increase the speed of the pipeline already.

I couldn't see issues with the code - I mainly left some questions and suggestions.

> 1) Please check you agree with the changes to the code & the logic still makes sense.

I think the proposed logic in the diagram in your PR description makes sense. I did leave a question on the implementation of the logic as there is a minor change that I'm not sure if it's desired or not.

> 2) Please also check the format of the new output and that you agree with it.

When I load both the geojson and the gdf, I cannot see the 'layer' column. I am getting a keyerror. Note that I had to change the code you left in the description to load the data as I was getting an error due to the updated config with the date in the file path. 

The only thing I changed is the file path to this:
`# File path
file_path = "s3://asf-local-heat-planning-tool/outputs/data/wiltshire/wiltshire_clusters_contextual_features_5m.geojson"`

Let me know if I'm loading the wrong file.

I also think that `geopackage` might be a better file format than `geojson` at this point. It does require some more research in terms of speed, but geopackages are designed to contain separate layers of data. I have checked with Simon if this would be an appropriate format for him and he said he would be able to migrate to using it instead.

> 3) You can check the code still runs with (I have done so myself too):

Per our discussion int he stand up, I did not run these files due to the missing data error caused by the recent changes to config.

Note to @helloaidank: this will affect your PR for the manifest json as now we are not using the spatial signatures or HN zones in some scripts where we were using them before.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: Good shout. I think we can delete both and have one threshold only? The threshold is now independent of whether we're within or outside of the HN zone.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: "individual_or_networked" is an output from this function, it doesn't exist before this.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: I changed it because I think it was wrong, but let me know if you disagree. 

We always say the most communal takes precedence. But then we had networked HP taking precedence over communal solution. Networked HP is a subset of communal, but if some of the properties were assigned communal and others networked HP, then we're not sure which type of communal solution we'll have so it makes more sense to assign communal solution.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: ah! good flag. In a previous PR I removed the 50% check from the code but forgot to update the docstring.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: this should definitely be "individual_or_networked", so it's the dosctring that needs updating. From your comment above and this one, it seems like when I last updated this function (with the grey areas work!) I forgot to update the dosctring.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: I totally agree - we should review this soon and bring it to the team. I've been having this on my mind but for networked HPs in specific. I actually opened an issue last week because I wanted to do a little investigation for buildings being assigned networked HP as in a lot of cases the avg outdoor space is high, so it would be good to have a sense of the prop of properties within the building driving the assignment to networked HP rather than individual.

> E.g. if a terrace has 30 properties, and 29 are individual solutions and 1 is communal, ...

I don't think this example would ever happen though. The only way to have communal at this point is to have a block of flats, so should in theory have a high number of flats in a block. Am I mistaken?

asf_heat_pump_suitability/pipeline/run/add_features.py
sofiapinto: woop woop 💃

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
sofiapinto: Correct! Forgot to add back.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
sofiapinto: Good shout!

asf_heat_pump_suitability/getters/load_geodata.py
sofiapinto: agree!

asf_heat_pump_suitability/getters/load_geodata.py
sofiapinto: agree!

asf_heat_pump_suitability/getters/load_geodata.py
sofiapinto: agree!

asf_heat_pump_suitability/getters/load_geodata.py
sofiapinto: agree!

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: I've now clarified this, let me know if you have any questions.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: I've now clarified this, let me know if you have any questions.

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: have now updated, let me know if it's still unclear :)

sofiapinto:
Thanks for the review @crispy-wonton 🌟  I have made the changes and replied to your comments.


> I also think that `geopackage` might be a better file format than `geojson` at this point. It does require some more research in terms of speed, but geopackages are designed to contain separate layers of data. I have checked with Simon if this would be an appropriate format for him and he said he would be able to migrate to using it instead.

As we discussed with Simon, happy for us to use geopackage. We should do it in a separate PR, and I'll leave it to you and Simon to check the pros and cons.

> Per our discussion in the stand up, I did not run these files due to the missing data error caused by the recent changes to config.

Perhaps the best would be for you to run the pipeline (with --save) for a new local authority, let's say "Southamption", and then try the below.

To load the geodata within the file you can do:


```
import geopandas as gpd

clusters_gdf = gpd.read_file(
        config["output"]["dataset"]["clusters_tech_contextual_info"].format(
            local_authorities=...,
            tolerance_m=5,
            release_date=....
        ),
    ).to_crs(epsg=27700)

```

This will include both layers (clusters with filters OR the areas of DHN potential). And you can filter for which layer you want by doing

```
clusters_gdf[clusters_gdf["layer"]=="clusters_with_contextual_features"]. Note that he column annotation will be have missing values only.
clusters_gdf[clusters_gdf["layer"]=='areas_of_district_heat_network_potential']: only the geometry, layer and annotation columns should be non missing.
If you load as a json, you can see how the json is organised:

# File path
file_path = config["output"]["dataset"]["clusters_tech_contextual_info"].format(
    local_authorities=...,
    tolerance_m=5,
    release_date=....
)

import json
from smart_open import open 

# Load the raw GeoJSON data from S3
with open(file_path, "r") as f:
    geojson_data = json.load(f)

geojson_data.keys()
```

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -306,6 +311,9 @@ def create_gdf_contextual_features(
         crs="EPSG:27700",
     )
 
+    geo_utils.verify_gdf_crs(hn_zones_gdf, target_crs="EPSG:27700")
+    geo_utils.verify_gdf_crs(spatial_signatures_gdf, target_crs="EPSG:27700")
```
crispy-wonton: suggestion: you could set the target CRS once within the function and then call it where required so it is guaranteed to be the same consistently.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -470,8 +482,17 @@ def create_json_contextual_features_metadata(
...
+            hn_zones_gdf[["geometry", "source_annotation"]],
+            # Create a single polygon for all spatial signatures to represent city centres
+            gpd.GeoDataFrame(
+                {
+                    "source_annotation": [
+                        spatial_signatures_gdf["source_annotation"].iloc[0]
+                    ],
+                    "geometry": [spatial_signatures_gdf.geometry.union_all()],
+                },
+                crs=spatial_signatures_gdf.crs,
```
crispy-wonton, line 495: question: how come you are unioning all the polygons here?

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
crispy-wonton: ah ok - in that case - how come we've completely removed this requirement rather than lower it? Is it based on the assumption that even if we have only one outdoor space for the cluster, it is likely that the other properties in the cluster all have a similar outdoor space size?

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
crispy-wonton: much clearer thank you

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
crispy-wonton: note: just noticing, I think we don't need some of the code in this function now. E.g. counting the properties with available space data or the proportion below. It's no longer used right?

crispy-wonton (commented):
A couple of final questions and I think there's now some redundant code in `decision_tree.py`?

Have looked further into using geopackage as the file type and it seems like an appropriate transition to make. Happy for this to happen in a separate PR as low priority.

Also, note there are merge conflicts with dev that need to be resolved.

Will run for a new LA and check the results.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
sofiapinto: Applying union_all() merges individual signature polygons into a unified city centers boundary. Because we don't need to distinguish between signature types on the front end, this cleans up the visual display without altering the total area.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
sofiapinto: Have done this!

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
sofiapinto: I tested both lowering and removing, but it didn't really work to just lower it in areas with high missingness.

> Is it based on the assumption that even if we have only one outdoor space for the cluster, it is likely that the other properties in the cluster all have a similar outdoor space size?

Yes, that's it. And this will effectively be removed altogether once we finish the outdoor space modelling.

sofiapinto:
hey @crispy-wonton - I have:
- taken care of the conflicts with dev
- answered your questions above
- removed the code that was no longer needed

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
crispy-wonton: Great, happy with that! I think it's a very reasonable assumptioN!

crispy-wonton (approved):

PR #462, opened 2026-07-31
453 Layer with anchor loads

### LINKED ISSUE #453: Separate layer with anchor loads

sofiapinto:
---

# Description

This adds a separate layer to each output file with the locations of anchor loads.

Fixes #453 

# Instructions for Reviewer

Can you please check I am using `cluster.load_transform_anchor_property_gdfs()` properly?

As we discussed earlier today we might want to pass the layers into `create_json_contextual_features_metadata()` with `layers = {...}` rather than each separate layer as an argument.


asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -347,10 +349,14 @@ def create_json_contextual_features_metadata(
 
     hn_potential_geojson = json.loads(hn_potential.to_json(drop_id=True))
 
+    anchor_loads_geojson = json.loads(anchor_loads.to_json(drop_id=True))
```
crispy-wonton, line 352: suggestion: as in previous PR, you may want to make use of the `geo_utils` function to verify gdf CRS here.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -475,10 +481,22 @@ def create_json_contextual_features_metadata(
...
+    print("Loading anchor property geodataframes and transforming to EPSG:4326...")
+    combined_anchor_gdf = cluster.load_transform_anchor_property_gdfs(
+        buildings_gdf=buildings_gdf, grid_squares=local_authority_dict["grid_squares"]
+    )[["geometry"]]
+
+    combined_anchor_gdf = gpd.overlay(
+        combined_anchor_gdf,
+        boundary_gdf[["geometry"]],
+        how="intersection",
+    ).to_crs(epsg=4326)
```
crispy-wonton: question: how come you are doing an overlay here? are you just wanting to clip the anchor properties to the boundary of the LA? If so, I think you can use:
`combined_anchor_gdf = combined_anchor_gdf[combined_anchor_gdf["geometry"].intersects(boundary_gdf.union_all())]`

I dont think the overlay is wrong but I think potentially it could be a heavier transformation? Not sure.

crispy-wonton (commented):
> Can you please check I am using cluster.load_transform_anchor_property_gdfs() properly?

Looks like you have :)

> As we discussed earlier today we might want to pass the layers into create_json_contextual_features_metadata() with layers = {...} rather than each separate layer as an argument.

I think this is a good idea - happy for you to implement here if you want but as it's not essential, it can also come as a later improvment.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
sofiapinto: yep, that's exactly what I wanted to do. Will change, not sure why i've used overlay tbh.

sofiapinto:
Thank you for the review @crispy-wonton - I have implemented your suggestions and this PR is now ready for a second review.

crispy-wonton (dismissed):
I suggest squashing the commits for this PR on merging to prevent polluting the history on dev with lots of the same commit message, as most of them are just merging changes from another branch.

crispy-wonton (approved):

PR #466, opened 2026-08-06
459 Adds logic trace to each cluster

sofiapinto:
---

# Description

Adds a trace with the logic behind the tech assigned to each cluster.

Fixes #459 

# Instructions for Reviewer

Please check the logic makes sense. There's a few todo's still, but worth getting a first review in the meantime to check no cases are missing etc.

Update: there were a few bugs and the code is currently a bit messy, but ran out of time to make it nicer. So for now the focus of the review is to make sure it works and the logic makes sense.

PR #467, opened 2026-08-11
465 optimise add features

### LINKED ISSUE #465: Optimise add_features.py script for speed and memory

crispy-wonton:
Fixes #465 

# Description

First pass at optimising `add_features.py` script for speed and memory. The main speed gains come from:
1. Removing the spatial signatures dataset
2. Partitioning the UPRN lookup (NSUL) by grid square and then loading by grid square instead of all GB
3. Removing all the additional steps to fill missing postcode, using NSUL instead

New files:
- `asf_heat_pump_suitability/pipeline/transform/lookups.py` - add function to clean national statistics UPRN lookup table ahead of usage
- `asf_heat_pump_suitability/schemas/nsul_schema.py` - schema for NSUL data. required for consistency of schema when partitioning.

Modified files:
- `asf_heat_pump_suitability/config/base.yaml` - add new path to partitioned GB UPRN lookup
- `asf_heat_pump_suitability/getters/load_data.py` - add getters to load UPRN lookup (NSUL) and to load partitioned domestic EPC
- `asf_heat_pump_suitability/getters/load_geodata.py` - remove `load_transform_dict_uprn_to_country_mapping` as it's no longer required
- `asf_heat_pump_suitability/pipeline/run/add_features.py` - implement new speed gains into the main run script
- `asf_heat_pump_suitability/pipeline/run/generate_inspire_file_bounds.py` - save files as `parquet` rather than `geojson` for faster loading later
- `asf_heat_pump_suitability/pipeline/run/optimise_inputs.py` - update functions to allow partitioning of files that intersect with multiple grid squares and add code to partition NSUL UPRN lookup by grid square
- `asf_heat_pump_suitability/pipeline/transform/off_gas.py` - remove redundant complex code and replace with simple cleaning function
- `asf_heat_pump_suitability/pipeline/transform/protected_areas.py` - edit for compatibility with new 'country' column added to features dataframe
- `asf_heat_pump_suitability/utils/s3_utils.py` - add function to extract bucket name and folder prefix from an s3 path

# Instructions for Reviewer

run the following command from the terminal. For me, it took around 5mins versus 1hr15m on `dev`.
```uv run asf_heat_pump_suitability/pipeline/run/add_features.py --local_authorities plymouth --release_date 20260805```

- are there other obvious improvements to make to improve the speed or memory load of the `add_features.py` script?

# Checklist:

- [ ] I have tested the code

### CODE REVIEW

File: asf_heat_pump_suitability/pipeline/run/add_features.py [hunk 1]
```diff
@@ -164,42 +178,46 @@ def parse_arguments() -> argparse.Namespace:
...
+    # # Load spatial signature polygons and label UPRNs in city centres
+    # spatial_signatures_gdf = load_geodata.load_gdf_spatial_signatures_gb(
+    #     detail_level=detail_level
+    # )
+    # features_df = city_centres.extend_df_city_centre_labels(
+    #     uprns_df=features_df,
+    #     uprns_gdf=uprns_gdf,
+    #     spatial_signatures_gdf=spatial_signatures_gdf,
+    # )
+    # del hn_zones_gdf, spatial_signatures_gdf
```
crispy-wonton: Note: I have commented this out temporarily to be potentially deleted later because PR #458 deletes the usage of these features in the decision tree. However, upon reviewing PR #466 I noticed that these features were still in use. Please could you confirm whether or not these features should therefore be retained in this script - or does it need further discussion?

If we want to retain them, I will definitely need to optimise the spatial signatures dataset for speed.

helloaidank: Just for my understanding, are we keeping these in here or adding them later? think this may have been confirmed in todays stand-up

helloaidank: We are getting rid of this

helloaidank: So for `features["in_city_centre"]` and `features["in_hn_zone"] `in `decision_tree.py` will raise an error if we go ahead with this deletion. Same for `compute_contextual_features.py`.

asf_heat_pump_suitability/pipeline/transform/outdoor_space.py
```diff
@@ -137,12 +137,28 @@ def generate_gdf_outdoor_space(
...
+    )
+
+    # Merge buildings back to land parcels
+    land_and_buildings_gdf = land_parcels_gdf.merge(
+        aggregated_intersections_gdf, how="left", on="NATIONALCADASTRALREFERENCE"
+    )
+
+    # Use a row-wise difference. This is quicker than using overlay with difference due to not needing to create a spatial index.
+    # Difference geometry - equivalent to land minus buildings - MultiPolygons will be created for land parcels which get split into multi-parts
+    diff_geom = land_and_buildings_gdf["geometry"].difference(
```
helloaidank, line 155: This speed up results in a parcel with no buildings having nothing to join to, so its building_geometry is null after the merge. Subtracting null from a shape gives… null. A later explode() step then drops the null-geometry row entirely. Result: a garden-only parcel vanishes from the output, and its properties get no outdoor space at all... when the old code gave them the whole parcel.

asf_heat_pump_suitability/pipeline/run/optimise_inputs.py
```diff
@@ -52,42 +61,140 @@ def assign_df_grid_squares(
...
+        if not isinstance(df, list):
+            raise TypeError(
+                "No list of dataframes detected despite setting `multi-file` to True. Set to False or enter a list of dataframes for processing."
+            )
+
+        existing_gs = set()
+        existing_suffixes = defaultdict(int)
+
+        for chunk_df in df:
+            grid_squares = chunk_df["grid_square"].unique()
```
helloaidank, line 120: we are redefining the grid_squares parameter (input parameter) here, was that intentional?

asf_heat_pump_suitability/pipeline/run/optimise_inputs.py
```diff
@@ -52,42 +61,140 @@ def assign_df_grid_squares(
...
+        # Generate all suffixes for the grid square
+        gs_suffixes = list(range(1, suffixes[grid_square] + 1, 1))
+
+        # Load all files for that grid square and save a final concatenated single file for the grid square
+        fsuffixes = [grid_square + "_" + str(suffix) for suffix in gs_suffixes]
+        suffix_paths = [fpath.format(grid_square=fsuffix) for fsuffix in fsuffixes]
+        dfs.extend([pl.read_parquet(fpath) for fpath in suffix_paths])
+        # Cast to same schema - some dfs read ruc21ind col as Int while others read as string because of formatting
+        # diffs between Scotland and England & Wales
+        dfs = [df.cast(nsul_schema.schema) for df in dfs]
```
helloaidank, line 185: We only do this casting in this flatten step (which means only runs for grid squares that in appeared in >2 CSVs). A square whose data came from a single region skips the cast (and is written with whatever types were guessed). 

E.g. for rural/urban classsification we end up it being a int for E&W but a string for S. If we try to concat these for whatever reason then it results in an error. This doesn't error atm, but would it be possible to do this casting right after reading it?

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -280,14 +267,26 @@ def parse_arguments() -> argparse.Namespace:
...
+    if len(countries) > 1:
+        nation = "GB"
+    else:
+        mapping = {
+            "E": "England",
+            "S": "Scotland",
+            "W": "Wales",
+        }
+        nation = mapping[countries[0]]
+    # TODO this seems to have gotten slower after specifying nation over GB?
```
helloaidank, line 286: **Hypothesis as to why it might be slow:** So the country column came from a left join against NSUL, any property NSUL doesn't know about (e.g. a brand new build) has country that is null, and .unique() includes null as a value. Plymouth then looks like two "countries", so the code falls back to loading all of GB. 

If that is the issue, we could drop the nulls using `features_df["country"].drop_nulls().unique()`

asf_heat_pump_suitability/getters/load_data.py
```diff
@@ -78,3 +84,61 @@ def load_df_off_gas_pcds() -> pl.DataFrame:
...
+    uri = config["data"]["geodata"]["gb_uprn_lookup_partitioned"]
+    if grid_squares:
+        return pl.concat(
+            [
+                pl.read_parquet(uri.format(grid_square=grid_square), **kwargs)
+                for grid_square in grid_squares
+            ]
+        )
+    else:  # Load whole of GB
+        bucket, prefix = s3_utils.extract_tuple_bucket_prefix(uri)
```
helloaidank, line 110: The config value is a template here, not a path. The if grid_squares branch does .format(grid_square="SX") before its used. The GB branch doesn't do the same.

asf_heat_pump_suitability/getters/load_data.py
```diff
@@ -78,3 +84,61 @@ def load_df_off_gas_pcds() -> pl.DataFrame:
...
+
+        fs = boto3.client("s3")
+        files = s3_utils.fetch_list_file_paths_from_s3_folder(
```
helloaidank, line 113: I think there might be a small bug here, when we do pl.read_parquet(file, **kwargs), the file returned is something like "inputs/lookup/NSUL_DEC_2025_partitioned/SX_NSUL.parquet" so it looks for a local file instead of reading in from the s3 bucket.

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -298,24 +297,28 @@ def parse_arguments() -> argparse.Namespace:
...
-    code_point_gdf = load_geodata.load_gdf_code_points()
-
-    features_df = off_gas.extend_df_off_gas(
-        features_df=features_df,
-        uprns_gdf=uprns_gdf,
-        code_point_gdf=code_point_gdf,
-        off_gas_list=off_gas_list,
-        id_col=config["constant"]["id"]["building"],
-        max_distance_m=500,  # to be conservative
+    off_gas_postcodes = off_gas.load_transform_list_off_gas_postcodes()
```
helloaidank, line 302: Just to note here that we've gone from using False to Null for off-gas column.

helloaidank:
Great work! I've tested the code, much speedier. I've left a couple of suggestions, and hypothesised around why the nation loading is so slow...


One potential small win/speed-up below:

The buildings layer is loaded twice, so we should delete one call. load_gdf_os_openmap_layer(layer="building", grid_squares=...) runs at add_features.py:143 (for the block-of-flats model) and again, identically, at line 203 (for outdoor space) and buildings_gdf is never deleted in between, so the second call re-downloads and re-parses the same shapefiles into a variable that already holds the result. 


Thanks so much!

asf_heat_pump_suitability/pipeline/run/add_features.py
crispy-wonton: Yes at the moment this will cause an error. This is dependent on the changes in #458 being merged. This PR removes the `in_hn_zone` and `in_city_centre` features and any reliance on them in the scripts mentioned.

crispy-wonton: And yes, to confirm, we are deleting these. Not adding them later.

asf_heat_pump_suitability/getters/load_data.py
crispy-wonton: Yes, I see the issue. This is a bug but I actually fixed it in a different way when I encountered it myself in PR #483 . Actually, I am just trying to extract the directory that the grid squares are in, rather than the path to a single file which is what happens here. In the correction, I extract just the directory. This also fixes the error mentioned below.

crispy-wonton: I have moved the changes to this branch.

asf_heat_pump_suitability/getters/load_data.py
crispy-wonton: See above thread

asf_heat_pump_suitability/pipeline/run/add_features.py
crispy-wonton: GREAT thinking! I will apply the drop nulls!

asf_heat_pump_suitability/pipeline/run/add_features.py
crispy-wonton: Can you clarify what you mean - doesn't the line below address this and result in True/False depending if postcode is in off gas postcodes? Am I missing something?

asf_heat_pump_suitability/pipeline/run/optimise_inputs.py
crispy-wonton: good point. have updated to have `grid_squares` as an optional param and then set the default to using all in the df.

asf_heat_pump_suitability/pipeline/run/optimise_inputs.py
crispy-wonton: great idea!

asf_heat_pump_suitability/pipeline/transform/outdoor_space.py
crispy-wonton: Yes, I also had thought about that and this was a deliberate design choice because:
1. Our pipeline selectively retains UPRNs which are joined to a building footprint. The only exception is a small number of EPC records whose UPRNs we retain due to them being in the EPC register (i.e. we know they are domestic properties). Some of these EPC UPRNs don't match to a building footprint but we match them to the closest one (within 3m) for further analysis.
2. That small number of EPC UPRNs should be the only ones that could possibly end up with no buildings here. In which case, I think it makes sense that the outdoor space is null because clearly there is some data issue (either the building doesn't exist when it should, or the UPRN is in the wrong location). I thought it would be better to keep null values here (and drop later) than fill with the size of the land parcel. Let me know if you disagree.

In the meantime, I've added annotation to explain this.

helloaidank: Happy to go with that, seems to make sense to me. Just wanted to double check that it was a deliberate choice!

asf_heat_pump_suitability/pipeline/run/add_features.py
helloaidank: Answered below 👍

asf_heat_pump_suitability/pipeline/run/add_features.py
```diff
@@ -298,24 +297,28 @@ def parse_arguments() -> argparse.Namespace:
...
+    features_df = features_df.with_columns(
+        pl.col("postcode").is_in(off_gas_postcodes).alias("off_gas")
```
helloaidank, line 292: `is_in` gives True/False only for non-null postcodes. A null postcode (say UPRN is absent from NSUL because it was created after December 2025 - i.e. new builds) would yield `off_gas = null`. In the previous implementation, we had `extend_df_off_gas` ending with `.otherwise(False)` which guaranteed a boolean!

helloaidank:
Happy to open a separate small issue for: 

> The buildings layer is loaded twice, so we should delete one call. load_gdf_os_openmap_layer(layer="building", grid_squares=...) runs at add_features.py:143 (for the block-of-flats model) and again, identically, at line 203 (for outdoor space) and buildings_gdf is never deleted in between, so the second call re-downloads and re-parses the same shapefiles into a variable that already holds the result.

and then left a clarification for the off gas null comment.  


Otherwise look like some good changes 👯‍♂️ 🚀 

Thanks!

asf_heat_pump_suitability/pipeline/transform/outdoor_space.py
crispy-wonton: Absolutely - good to check and to prompt me adding it to annotation!

asf_heat_pump_suitability/pipeline/run/add_features.py
crispy-wonton: This is probably the more desired behaviour and before it was less accurate. I.e. if we don't know the postcode, we don't know what the off gas label should be and therefore it should remain null.

sofiapinto (dismissed):

sofiapinto (dismissed):
The changes to the getter look good, thank you - have you made a function call to check that it works as expected?

PR #471, opened 2026-08-13
Pin one release date across pipeline stages (#460)

### LINKED ISSUE #460: Add release date param to run_pipeline.sh
## Context

After #413, each pipeline stage resolves `--release_date` independently, defaulting to today's date. `asf_heat_pump_suitability/pipeline/run/run_pipeline.sh` invokes all five stages per local authority without pinning a date, so:

- a run that crosses midnight silently splits one release across two dated S3 directories;
- the first stage invoked after midnight fails with `FileNotFoundError` because the upstream output lives under the previous day's release directory.

## Proposed change

Resolve the release date once at the top of `run_pipeline.sh` (accept it as an optional script argument, defaulting to today) and pass the same `--release_date` to every stage invocation, so the whole run is pinned to a single release directory regardless of how long it takes.

`create_manifest.py` is unaffected — it intentionally leaves the `{release_date}` placeholder unresolved.

Raised by @sofiapinto in the review of #413.

helloaidank:
## Description
Since #413 each stage resolves `--release_date` on its own, defaulting to today. A run that crosses midnight splits one release across two dated S3 folders, then fails when a stage looks for upstream output under the new date. `run_pipeline.sh` now resolves the date once at the top and passes the same value to every stage.

## Related Issue(s)
- Closes #460

## Changes Made
- New optional `--release_date YYYYMMDD` flag on `run_pipeline.sh`, defaulting to today
- The date is validated once via `get_str_release_date` before any stage runs, so the date format stays defined only in config
- The same date goes to all five stage invocations. `create_manifest.py` is untouched on purpose, it keeps its `{release_date}` placeholder
- Fixed `succeeded = 0`, which bash read as a command instead of an assignment
- Spec added under `docs/specs/`

### What to Focus On
- The parsing block and Python one-liner at the top of `run_pipeline.sh`. Bad dates (empty, malformed, impossible) must all exit before any stage runs
- A malformed date prints the right error but wrapped in a Python traceback. Fine as is, or worth a tidier message?
- The less intuitive bash is explained in inline comments on the diff

### How to Test
```bash
bash asf_heat_pump_suitability/pipeline/run/run_pipeline.sh --bogus
bash asf_heat_pump_suitability/pipeline/run/run_pipeline.sh --release_date
bash asf_heat_pump_suitability/pipeline/run/run_pipeline.sh --release_date ""
bash asf_heat_pump_suitability/pipeline/run/run_pipeline.sh --release_date 2026-08-01
```

To see the pinning without running the real pipeline

```bash
sed 's|^\( *\)python asf|\1echo python asf|' asf_heat_pump_suitability/pipeline/run/run_pipeline.sh > /tmp/stub.sh
bash /tmp/stub.sh --release_date 20260801 | grep -o -- '--release_date [0-9]*' | sort | uniq -c
rm /tmp/stub.sh
```

Expect one line, `35 --release_date 20260801`

### Checklist:
- [ ] I have refactored my code out from notebooks/ (n/a)
- [x] I have tested the code
- [x] I have run pre-commit and addressed any issues not automatically fixed
- [x] I have merged any new changes from dev
- [ ] Major functions have docstrings (n/a, no Python changed)
- [ ] Appropriate information has been added to READMEs (n/a)


asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
```diff
@@ -2,10 +2,44 @@
...
+# Usage: ./asf_heat_pump_suitability/pipeline/run/run_pipeline.sh [--release_date YYYYMMDD]
+# The release date defaults to today and is pinned across all stages, so a run
+# crossing midnight still writes to a single dated release directory.
 
 # Run from the repo root regardless of where the script is invoked from
 cd "$(dirname "$0")/../../.." || exit 1
 
+# Print how to call this script, then abort
+usage() {
+    echo "Usage: $0 [--release_date YYYYMMDD]" >&2
```
helloaidank, line 15: `>&2` sends this message to stderr rather than stdout, so it never mixes into piped or logged pipeline output. The `exit 1` below it stops the whole script, not just the function, and reports failure to whatever ran it.

asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
```diff
@@ -2,10 +2,44 @@
...
+    exit 1
+}
+
+# Read the arguments: accept an optional --release_date value, reject anything else
+release_date=""
+while [ $# -gt 0 ]; do
+    case "$1" in
+        --release_date)
+            [ $# -ge 2 ] || usage
+            [ -n "$2" ] || usage
```
helloaidank: Reading these two guards: `$#` is how many words remain on the command line and `$2` is the word after the flag. `[ ... ] || usage` means the check must pass or we quit with the help message. The first guard requires a value after the flag, the second requires it to be non-empty. The quotes around `$2` matter, without them an empty value would vanish before the test sees it and the check would silently pass.

asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
```diff
@@ -2,10 +2,44 @@
...
+            release_date="$2"
+            shift 2
```
helloaidank, line 29: Bash holds the command line words in a numbered queue. The loop only ever looks at the front of the queue, `$1`. `shift 2` deletes the front two entries so everything behind them moves up.

Trace for `./run_pipeline.sh --release_date 20260801`

```text
before shift 2:  $1 = --release_date   $2 = 20260801   $# = 2
after  shift 2:  queue empty                           $# = 0
```

The loop condition is `while [ $# -gt 0 ]`, meaning "while any words are left". After the shift the count is 0, so the loop ends and the script carries on.

Without the shift nothing is ever removed, so on the next lap `$1` is still `--release_date` and the loop repeats the same branch forever.

asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
```diff
@@ -2,10 +2,44 @@
...
+    esac
+done
+
+# Resolve the release date once (defaulting to today) and validate its format
+# via get_str_release_date, failing fast before any stage runs
+release_date=$(python -c "
+import sys
+from asf_heat_pump_suitability.utils.save_utils import get_str_release_date
+print(get_str_release_date(sys.argv[1] if len(sys.argv) > 1 else None))
+" ${release_date:+"$release_date"}) || exit 1
```
helloaidank, line 43: Two bits of shell here worth unpacking. `$( ... )` runs the command inside and stores whatever it prints, so the validated date Python prints becomes the new value of `release_date`. The tail `${release_date:+"$release_date"}` means pass the value along if one was given, otherwise pass nothing at all. In the nothing case the `sys.argv` check hands `None` to `get_str_release_date`, which falls back to today. The `|| exit 1` aborts before any stage runs if validation raised.

asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
```diff
@@ -16,35 +50,35 @@ local_authorities=(
     "east lothian"
 )
 
-succeeded = 0
+succeeded=0
```
helloaidank, line 56: No spaces on purpose. Bash reads `succeeded = 0` as a command called `succeeded` with two arguments, not an assignment. The old line only appeared to work because `$((succeeded+1))` later in the loop treats an unset variable as 0.

asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
crispy-wonton, line 15: question: is `usage` a standard term to use in this situation? this confused me slightly because of too much brevity. happy to keep if standard tho - now I understand

crispy-wonton (commented):
```
bash asf_heat_pump_suitability/pipeline/run/run_pipeline.sh --bogus
bash asf_heat_pump_suitability/pipeline/run/run_pipeline.sh --release_date
bash asf_heat_pump_suitability/pipeline/run/run_pipeline.sh --release_date ""
bash asf_heat_pump_suitability/pipeline/run/run_pipeline.sh --release_date 2026-08-01
```
Everything looks good to me and I ran the test cases plus some additional ones. Just want to check - should the last test case here fail? I suppose if so, we could update it at some point in a separate issue to allow taking in differently formatted date strings and standardising them within the `get_str_release_date` function.

helloaidank:
> ```
> bash asf_heat_pump_suitability/pipeline/run/run_pipeline.sh --bogus
> bash asf_heat_pump_suitability/pipeline/run/run_pipeline.sh --release_date
> bash asf_heat_pump_suitability/pipeline/run/run_pipeline.sh --release_date ""
> bash asf_heat_pump_suitability/pipeline/run/run_pipeline.sh --release_date 2026-08-01
> ```
> 
> Everything looks good to me and I ran the test cases plus some additional ones. Just want to check - should the last test case here fail? I suppose if so, we could update it at some point in a separate issue to allow taking in differently formatted date strings and standardising them within the `get_str_release_date` function.

The last test case is supposed to fail, I think that's good to flag as a follow up issue but guessing not an immediate priority ATM!

asf_heat_pump_suitability/pipeline/run/run_pipeline.sh
helloaidank: Made a couple of changes to make it overall readable and easier to understand why it might crash.

crispy-wonton (approved):
Nice additions!

PR #473, opened 2026-08-14
468 optimise cluster

### LINKED ISSUE #468: Optimise `cluster.py` script for speed and memory

crispy-wonton:
Fixes #468 

# Description

First pass optimisation of `cluster.py` script for speed and memory gains. Cut from 10m4s to 5m2s.

Modified files:
- `asf_heat_pump_suitability/pipeline/cluster/cluster.py` - conduct some vectorisation of geospatial manipulations rather than using heavy processes like `dissolve`

# Instructions for Reviewer

I have tested the code and it works. You can run the following code from terminal to test it:
```uv run python -i asf_heat_pump_suitability/pipeline/cluster/cluster.py --local_authorities plymouth --release_date 20260812```

This will run the code on this branch using the output from the upstream scripts run on `dev`.

I have compared the outputs to the outputs of the `cluster.py` script run on `dev` using the same upstream inputs. I have found some differences which I have outlined in the sprint 6 slide deck. I believe the diffs are negligible. Let me know if there's further comparisons you think I should do :)

Please pay special attention to...
The new logic applied in the clustering to ensure we are achieving the same end goal as before.

# Checklist:

- [x] I have tested the code

### DIFF
--- a/asf_heat_pump_suitability/pipeline/cluster/cluster.py
+++ b/asf_heat_pump_suitability/pipeline/cluster/cluster.py
@@ -98,12 +98,13 @@ def generate_gdf_clusters(
 
     # Create Voronoi polygons and overlay physical barriers for all local authority boundaries
     for boundary in boundary_gdf["geometry"].unique():
+        bounded_tech_gdf = tech_gdf[tech_gdf.within(boundary)]
         voronoi_gdf = extend_edges_gdf(gdf=buildings_gdf, boundary=boundary)
 
         # One cell per building
         cells_gdf = overlay_gdf_physical_barriers(
             voronoi_gdf=voronoi_gdf,
-            tech_gdf=tech_gdf,
+            tech_gdf=bounded_tech_gdf,
             line_overlay_gdf=line_overlay_gdf,
             polygon_overlay_gdf=polygon_overlay_gdf,
             id_col=id_col,
@@ -325,6 +326,8 @@ def extend_edges_gdf(
     arr = voronoi_gdf.sort_values(building_id_col)
     unique_ids, first_idx = np.unique(arr[building_id_col].values, return_index=True)
     # Split the Voronoi geometries into groups. One group == one building ID
+    # first_idx[1:] skips the first element which is always 0. This is because np.split splits the array at the locations of
+    # each index. If we included index 0, it would create an empty group at the start.
     geom_groups = np.split(arr.geometry.values, first_idx[1:])
     voronoi_gdf = gpd.GeoDataFrame(
         {building_id_col: unique_ids},
--- a/asf_heat_pump_suitability/pipeline/cluster/cluster.py
+++ b/asf_heat_pump_suitability/pipeline/cluster/cluster.py
@@ -168,7 +168,19 @@ def generate_gdf_clusters(
     # At this point we have multiple rows of each cluster geometry with one row for every building within the cluster.
     # We need to flatten the cluster geometries to one row per cluster, aggregating the within_anchor_radius boolean flag.
     # Selecting `max` of the boolean will mean any clusters containing one building within the anchor radius will be labelled as within the radius.
-    clusters_gdf = clusters_gdf.dissolve(by="cluster_id", aggfunc="max").reset_index()
+    # clusters_gdf = clusters_gdf.dissolve(by="cluster_id", aggfunc="max").reset_index()
+    clusters_gdf = (
+        clusters_gdf.groupby(by="cluster_id")
+        .agg(
+            {
+                "geometry": "first",
+                "assigned_tech": "first",
+                f"within_{radius}m_from_anchor_load": "max",
+            }
+        )
+        .reset_index()
+        .set_geometry(column="geometry", crs=clusters_gdf.crs)
+    )
 
     # TODO move to testing when sample set available
     if round(clusters_gdf["geometry"].area.sum(), 3) > round(
@@ -285,11 +297,12 @@ def extend_edges_gdf(
     print(
         "Joining Voronois to original building footprints and dissolving per footprint..."
     )
+
     # Join the original building points with IDs to the Voronoi cells and dissolve to get one polygon per internal building ID
     # Intersects (rather than contains) allows for small floating point errors without dropping Voronois
     voronoi_gdf = voronoi_gdf.sjoin(points_gdf, how="inner", predicate="intersects")
     voronoi_gdf.geometry = voronoi_gdf.geometry.make_valid()
-    voronoi_gdf = (voronoi_gdf.dissolve(by=id_col).reset_index()).clip(boundary)
+    voronoi_gdf = voronoi_gdf.clip(boundary).dissolve(by=id_col).reset_index()
 
     # Clip Voronoi cells to a max buffer
     print("Clip Voronoi cells to maximum buffer...")
--- a/asf_heat_pump_suitability/pipeline/cluster/cluster.py
+++ b/asf_heat_pump_suitability/pipeline/cluster/cluster.py
@@ -250,17 +250,18 @@ def extend_edges_gdf(
     gdf = gdf[gdf.within(boundary)]
 
     # Add an internal unique ID to each building
-    id_col = "_internal_building_id"
-    gdf[id_col] = np.arange(len(gdf))
+    building_id_col = "_internal_building_id"
+    gdf[building_id_col] = np.arange(len(gdf))
 
     all_points = []
-    all_ids = []
+    all_building_ids = []
 
     print("Densifying building polygon edges...")
     # Densify polygon edges with additional points to prepare for Voronoi diagram
     for _, row in gdf.iterrows():
         geom = row.geometry
-        id = row[id_col]
+        # Assign building ID
+        building_id = row[building_id_col]
         # Deal with any multipolygon buildings
         polys = geom.geoms if isinstance(geom, MultiPolygon) else [geom]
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: asf_heat_pump_suitability/pipeline/cluster/cluster.py [hunk 1]
```diff
@@ -168,7 +168,19 @@ def generate_gdf_clusters(
...
+        .agg(
+            {
+                "geometry": "first",
+                "assigned_tech": "first",
+                f"within_{radius}m_from_anchor_load": "max",
+            }
+        )
+        .reset_index()
+        .set_geometry(col="geometry", crs=clusters_gdf.crs)
+    )
```
sofiapinto: I think some of the differences potentially come from this change to the code :)

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -168,7 +168,19 @@ def generate_gdf_clusters(
     # At this point we have multiple rows of each cluster geometry with one row for every building within the cluster.
     # We need to flatten the cluster geometries to one row per cluster, aggregating the within_anchor_radius boolean flag.
     # Selecting `max` of the boolean will mean any clusters containing one building within the anchor radius will be labelled as within the radius.
-    clusters_gdf = clusters_gdf.dissolve(by="cluster_id", aggfunc="max").reset_index()
+    # clusters_gdf = clusters_gdf.dissolve(by="cluster_id", aggfunc="max").reset_index()
+    clusters_gdf = (
+        clusters_gdf.groupby(by="cluster_id")
+        .agg(
+            {
+                "geometry": "first",
```
sofiapinto, line 174: Remind me: at this point although each row represents one building, does the geometry represent the geometry of the cluster or the building?

If cluster, then we're fine. If the building, then we would need to set `"geometry": unary_union`.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -266,42 +279,71 @@ def extend_edges_gdf(
             # Add corner vertices of building into list, dropping the last one which is a duplicate of the starting point
             pts.extend(Point(coord) for coord in exterior.coords[:-1])
             all_points.extend(pts)
-            all_ids.extend([id] * len(pts))
+            all_building_ids.extend([building_id] * len(pts))
 
     print(f"Generated {len(all_points)} points.")
-    # Create a gdf of all densified points, where one row is one point
-    points_gdf = gpd.GeoDataFrame({id_col: all_ids}, geometry=all_points, crs=gdf.crs)
+    # Extract a flat (N, 2) float64 array of all point coordinates
```
sofiapinto, line 283: i believe the changes below might be a reason why there are some differences in the outputs, but the code looks fine to me!

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -266,42 +279,71 @@ def extend_edges_gdf(
...
-    # Intersects (rather than contains) allows for small floating point errors without dropping Voronois
-    voronoi_gdf = voronoi_gdf.sjoin(points_gdf, how="inner", predicate="intersects")
+
     voronoi_gdf.geometry = voronoi_gdf.geometry.make_valid()
-    voronoi_gdf = (voronoi_gdf.dissolve(by=id_col).reset_index()).clip(boundary)
+    # Sort building IDs and then get the first index of each unique building ID
+    arr = voronoi_gdf.sort_values(building_id_col)
+    unique_ids, first_idx = np.unique(arr[building_id_col].values, return_index=True)
+    # Split the Voronoi geometries into groups. One group == one building ID
+    geom_groups = np.split(arr.geometry.values, first_idx[1:])
```
sofiapinto, line 328: `first_idx[1:]`: i think this requires some additional documentation, or even just an example (the why this is needed)

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -266,42 +279,71 @@ def extend_edges_gdf(
             # Add corner vertices of building into list, dropping the last one which is a duplicate of the starting point
             pts.extend(Point(coord) for coord in exterior.coords[:-1])
             all_points.extend(pts)
-            all_ids.extend([id] * len(pts))
+            all_building_ids.extend([building_id] * len(pts))
```
sofiapinto, line 280: I believe this `for` loop with `iterrows()` is probably still the slowest part of the script. Could we replace it with something like the below? Haven't checked if results are the same, so you would need to do those checks.

```
# Split MultiPolygons into one row per Polygon, keeping building_id_col aligned
gdf = gdf.explode(index_parts=False)
gdf = gdf[gdf.geometry.geom_type == "Polygon"]

exteriors = shapely.get_exterior_ring(gdf.geometry.values)
# Insert vertices so no edge segment exceeds `spacing`
densified = shapely.segmentize(exteriors, spacing)
# Flatten to a (N, 2) coord array plus the index of the parent ring for each coordinate
coords_arr, ring_idx = shapely.get_coordinates(densified, return_index=True)
all_building_ids = gdf[building_id_col].values[ring_idx]
```

asf_heat_pump_suitability/pipeline/cluster/cluster.py
sofiapinto, line 574: Just looking at the whole file again and wanted to ask a question. What if two buildings tie exactly for the largest intersection area with a cell? Could that happen?

sofiapinto, line 111: This doesn't currently affect things given the way we're running the pipeline (one LA at a time), but if we did decide to run cluster.py for a set of LAs at once, then this change would make it faster.

```suggestion
for boundary in boundary_gdf["geometry"].unique():
    local_tech_gdf = tech_gdf[tech_gdf.within(boundary)]
    voronoi_gdf = extend_edges_gdf(gdf=buildings_gdf, boundary=boundary)

    # One cell per building
    cells_gdf = overlay_gdf_physical_barriers(
         voronoi_gdf=voronoi_gdf,
         tech_gdf=local_tech_gdf,
         line_overlay_gdf=line_overlay_gdf,
         polygon_overlay_gdf=polygon_overlay_gdf,
         id_col=id_col,
    )
# TODO No reassignment based on neighbouring cells - TBC if wanted by user testing
# gdfs.append(reassign_gdf_communal_networked(cells_gdf))
gdfs.append(cells_gdf)
```

sofiapinto: This is probably faster? As it doesn't require two overlays.
```suggestion
barriers_gdf = pd.concat([polygon_overlay_gdf, line_overlay_gdf], ignore_index=True)
cells_gdf = domestic_voronoi_gdf.overlay(barriers_gdf, how="difference").explode()
```

sofiapinto (commented):
Hey @crispy-wonton - great work! 🥳 

I left some questions and some suggestions for other parts of `cluster.py` where we could make some changes to increase speed. I think it's also worth start discussing how the pipeline will be run for all LAs as that might change how we do things within each scirpt.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: yes - 100%, will update!

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: at this point, the geometry is the geometry of the whole cluster, even though each row is a building. So there are lots of duplicate geometries of the same cluster. Thus, we can take the 'first' for each cluster ID.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Looking at the scalene profile, the lines that are the slowest/ most memory intensive are:

- line 332, taking 1m45s: `geometry=[shapely.union_all(g) for g in geom_groups]` -- line to group the voronoi polygons back into one that represents a total voronoi for a single building polygon. This has already been optimised from an even slower `dissolve` that was used previously.
- line 306, 3G of memory (but only 13s): `voronoi_collection = shapely.voronoi_polygons(coords, extend_to=boundary, ordered=True
    )` -- line to create the voronoi polygons from each point.

I don't see this `for` loop looking too problematic in the scalene profile. The time per line is very minimal, except for line 278: `pts = [exterior.interpolate(i * spacing) for i in range(num_pts)]` which takes 30s.

That said, your suggestions could still speed up this loop by replacing the interpolation line with `segmentize`. Will look at it more closely and implement if achieving the same goal.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Yes agree this is the likely cause of the changes - I outlined this in more detail in the related slide deck :)

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Yes! Not sure why I kept these separate in the first place

crispy-wonton: The code doesn't account for this at the moment because I think there is almost no chance of this happening, given that we will have float values for max intersection with many decimal places. I could add extra code to handle this edge case as perhaps there are a handful of edge cases where this would happen - although I don't know what would really cause that.

crispy-wonton: If you think it's worth adding, I think let's do it in another PR. If it did happen, the result would be clusters that merged together too much (the same problem we had before).

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Amazing thank you for reviewing!

Yes, I realised the changes were coming from an earlier version of this code which was taking the minimum space. I outlined this in the sprint slidedeck from a previous sprint. Separately, I then realised that section of the code actually contained a bug so this is the updated version - everything is outlined in the slides :)

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Let me know if still unclear :)

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Have decided to just combine the production of these datasets into a single function and updated the code accordingly.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Given the time constraints of the modelling and other priorities, I'm going to leave this for now as it will require additional work to rerun the script and compare to a previous version of outputs and I want to be careful implementing it. It will be much easier to test the implementation of this once Aidan's cluster comparison work has been completed and merged so I'd leave further major changes til after that.

sofiapinto: Yeah - I don't think the problem is necessarily the time per line, but having to loop through with the for loop, especially for bigger LAs - and then thinking how long this adds on average per LA, given we'll need to run this for all LAs. Happy for this to be actioned/checked separately, if time is still an issue.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
sofiapinto: No need to action if we don't think this is even a problem. Or if we're unsure we can open a separate issue to be actioned later.

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -442,11 +437,9 @@ def overlay_gdf_physical_barriers(
     ].drop(columns="select_id")
 
     # Remove areas covered by polygons and lines
-    cells_gdf = (
-        domestic_voronoi_gdf.overlay(polygon_overlay_gdf, how="difference")
-        .overlay(line_overlay_gdf, how="difference")
-        .explode()
-    )
+    cells_gdf = domestic_voronoi_gdf.overlay(
+        polygon_overlay_gdf, how="difference"
```
sofiapinto, line 441: do we still want to call it polygon overlay, now that it contains lines too? same for the test_ script

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -628,6 +584,11 @@ def load_transform_gdf_polygon_barriers(
     - Tidal boundaries
     - Woodland
 
+    Additionally, load physical barriers with (Multi)LineString geometries - major roads and railways - for the
+    specified grid squares. A buffer is added around each geometry to cover the width of the road / railway.
+
+    # TODO add road types to docstring
```
sofiapinto: is this something we want to do in a separate PR then?

sofiapinto (dismissed):
left a couple of final questions, but otherwise happy for this to be merged!

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Yes - at the point of use it is polygons only. Some of the polygons were created from original linestrings, but at this point in the code, the geodataframe contains polygons only :) it was sort of a misnomer before!

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Added to backlog!

asf_heat_pump_suitability/pipeline/cluster/cluster.py
crispy-wonton: Thanks! This was a legacy TODO statement that was copied over. I've now added the road types to docstring :)

sofiapinto (approved):

Issue #474, opened 2026-08-17
Decompose the comparison script before extending it


helloaidank:
## Current behaviour

The cross-version comparison script is a single module of ~1,330 lines (plus a ~1,640-line test module) holding S3 path resolution and loading, check computations, markdown report rendering, plotting, and the CLI. The functions are small and independent, but the stacked follow-on issues (distribution checks #TBD, thresholding #TBD) will each add compute + render pairs, extrapolating to ~2,000 lines — raising navigation, review, and merge-conflict costs for exactly the "add one more check" extensions the validation stack is designed for.

## Desired behaviour

The comparison tool becomes a small package of focused modules inside the validation subpackage, split along the seams the code already has:

- **loading** — path resolution, version discovery, output loaders, manifest reading
- **checks** — the pure DataFrame-to-summary check functions, including the shared distribution-stats helper
- **report** — the markdown section renderers
- **plots** — the distribution plot functions
- **compare_versions** — CLI and orchestration only, remaining the entrypoint so the run command is unchanged

The refactor is behaviour-preserving: the report output, CLI, and config contract do not change. The test module splits along the same seams with test content unchanged — that is the evidence of behaviour preservation.

## Acceptance criteria

- [ ] Entrypoint module contains only CLI parsing and orchestration; the run command is unchanged
- [ ] Loading, checks, report rendering, and plotting each live in their own module
- [ ] Existing tests pass with content unmodified (relocation and imports aside)
- [ ] No public function's behaviour or signature changes

## Out of scope

- Any new checks or report content (the stacked follow-on issues)
- Consolidating the per-stage config mappings (#470)

PR #475, opened 2026-08-17
472 Filter geometry checks to the clusters layer (3/3)

### LINKED ISSUE #472: Cross-version comparison script — cluster geometry checks
## Current behaviour

The comparison script's tabular checks (#447) can't see geometry-level drift — a clustering change could alter cluster count, total area, or size distribution while every tabular check still passes. The loader drops geometry columns by design.

## Desired behaviour

Extend the report, for the clustering and contextual-features stages, with: cluster count delta; total area delta (EPSG:27700, m² — both stages, with the report noting that contextual-features areas are computed on reprojected simplified geometry); and size-distribution comparisons measured both ways — polygon area, and UPRNs per cluster (the `n_UPRNs` count computed at the contextual-features stage). Distribution statistics (Q1, Q3, min, max, mean) come from one shared column-parameterized helper — introduced here and reused by the distribution-checks follow-on — with target columns listed per stage in `base.yaml` (real columns only; derived area is fed through the same helper in code). Each distribution also gets an old-vs-new overlaid plot saved next to the report and embedded in it.

## Acceptance criteria

- [ ] Cluster count delta reported for both stages
- [ ] Total area delta reported for both stages with CRS/units stated; simplified-geometry caveat in the contextual-features section
- [ ] Cluster-area distribution reported: Q1, Q3, min, max, mean per version
- [ ] UPRNs-per-cluster distribution reported with the same statistics
- [ ] One shared column-parameterized distribution function; target columns per stage in `base.yaml`
- [ ] Overlaid old-vs-new plots for cluster area and UPRNs per cluster, embedded in the report
- [ ] Unit tests cover at least one genuine geometry-drift case and one stable case

## Out of scope

- Categorical mix deltas, numeric drift (stacked follow-on issue 5 in the comparison stack)
- Automatic expected/suspicious flagging (stacked follow-on issue 6)
- PSI or other distribution-drift scalars

helloaidank:
> **Stacked PR, 3 of 3.** Base is `472_addGeometryReportSections` (PR #487, 2/3), not `dev`. Please review only this PR's 5 commits.

# Description

The contextual-features outputs are now multi-layer front-end files: one `layer` column holding `clusters`, `ward_boundaries`, `anchor_loads` and DHN areas. The geometry checks from 1/3 and 2/3 counted every layer, so six whole-county ward polygons swamped the totals...East Lothian showed a +679.5M m² area delta that was entirely wards, not clusters.

- Checks, distributions and plots now cover the clusters layer only, named by `compare_versions.cluster_layer` in `base.yaml`.
- Outputs from before layers existed have no `layer` column, so they count entirely as clusters.

The decomposition follow-on (#474) stacks on this branch.

Closes #472.

# Instructions for Reviewer

## Please pay special attention to

- **The pre-layers fallback** — treating a missing `layer` column as all-clusters. It is the assumption that makes old-vs-new comparisons work across the format change, and it is wrong if any pre-layers output was ever multi-layer.

## How to test

```bash
uv run python -m pytest asf_heat_pump_suitability/pipeline/validate/tests/test_compare_versions.py -v
```

Expect **175 passed**; 14 test functions are new here.

Acceptance run, East Lothian

```bash
uv run python -m asf_heat_pump_suitability.pipeline.validate.compare_versions \
   --stage cluster --local_authority east_lothian

uv run python -m asf_heat_pump_suitability.pipeline.validate.compare_versions \
   --stage compute_contextual_features --local_authority east_lothian
```

# Checklist:

- [x] I have tested the code
- [ ] I have requested a code review — draft while the branches underneath are still open

### DIFF
--- a/asf_heat_pump_suitability/config/base.yaml
+++ b/asf_heat_pump_suitability/config/base.yaml
@@ -405,6 +405,10 @@ compare_versions:
     cluster: "tech_clusters"
     compute_contextual_features: "clusters_tech_contextual_info"
   decision_tree_buildings_dataset: "buildings_most_suitable_tech"
+  # The `layer` value carrying the clusters in the multi-layer front-end
+  # geojson; the cluster checks filter to it. Versions without a `layer`
+  # column (pre-August-2026 outputs) are treated as all-clusters.
+  cluster_layer: "clusters_with_contextual_features"
   # Columns whose distributions the report compares per stage (Q1/Q3, min,
   # max, mean, old vs new). Real output columns only: derived metrics such
   # as cluster area are computed from geometry in code and fed through the
--- a/asf_heat_pump_suitability/config/base.yaml
+++ b/asf_heat_pump_suitability/config/base.yaml
@@ -405,6 +405,14 @@ compare_versions:
     cluster: "tech_clusters"
     compute_contextual_features: "clusters_tech_contextual_info"
   decision_tree_buildings_dataset: "buildings_most_suitable_tech"
+  # Columns whose distributions the report compares per stage (Q1/Q3, min,
+  # max, mean, old vs new). Real output columns only: derived metrics such
+  # as cluster area are computed from geometry in code and fed through the
+  # same shared helper.
+  distribution_columns:
+    cluster: []
+    compute_contextual_features:
+      - "n_UPRNs"
   # Repo paths counted as each stage's module when scoping the commit log
   # between two versions' recorded commits: the stage entrypoint plus the
   # pipeline modules it directly drives. Cross-cutting code (getters/, utils/,
--- a/asf_heat_pump_suitability/pipeline/validate/compare_versions.py
+++ b/asf_heat_pump_suitability/pipeline/validate/compare_versions.py
@@ -47,10 +47,14 @@ TECH_COL = "assigned_tech"
 NULL_TECH_LABEL = "(null)"
 CLUSTER_ID_COL = "cluster_id"
 AREA_COL = "area_m2"
+# Multi-layer front-end outputs (August 2026 onwards) tag every row with
+# the layer it belongs to; earlier outputs have no such column.
+LAYER_COL = "layer"
 # Stages whose outputs carry cluster geometry, and so get the cluster
 # count, area and distribution sections.
 GEOMETRY_STAGES = ("cluster", "compute_contextual_features")
 
+CLUSTER_LAYER = config["compare_versions"]["cluster_layer"]
 STAGE_MODULE_PATHS = config["compare_versions"]["stage_module_paths"]
 STAGE_OUTPUT_DATASETS = config["compare_versions"]["stage_output_datasets"]
 BUILDINGS_DATASET = config["compare_versions"]["decision_tree_buildings_dataset"]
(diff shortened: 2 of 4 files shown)


asf_heat_pump_suitability/config/base.yaml
```diff
@@ -405,6 +405,10 @@ compare_versions:
     cluster: "tech_clusters"
     compute_contextual_features: "clusters_tech_contextual_info"
   decision_tree_buildings_dataset: "buildings_most_suitable_tech"
+  # The `layer` value carrying the clusters in the multi-layer front-end
+  # geojson; the cluster checks filter to it. Versions without a `layer`
+  # column (pre-August-2026 outputs) are treated as all-clusters.
+  cluster_layer: "clusters_with_contextual_features"
```
crispy-wonton, line 411: praise: good idea to add this to the config!

asf_heat_pump_suitability/pipeline/validate/compare_versions.py
```diff
@@ -773,25 +807,30 @@ def load_transform_df_stage_output(path: str) -> pl.DataFrame:
 
 def load_df_cluster_areas(path: str) -> pl.DataFrame:
     """
-    Load a geometry-bearing stage output as per-cluster areas in m².
+    Load a geometry-bearing stage output as per-row areas in m².
```
crispy-wonton, line 810: suggestion: I think here and elsewhere, it still makes sense to refer to `clusters` rather than `rows`. Makes the docstrings more clear and understandable.

asf_heat_pump_suitability/pipeline/validate/tests/test_compare_versions.py
```diff
@@ -1040,9 +1040,49 @@ def test_distribution_sections_report_the_statistics_per_version(
...
+            "| Min | 1,165 | 1,165 |" in report
+        ), "an Int64 and a Float64 version must render the same min"
+        assert (
+            "| Mean | 1,738 | 1,738 |" in report
+        ), "integral float means must render like ints, with no trailing .0"
+        assert (
+            "1,738.0" not in report
+        ), "no integral statistic may keep a trailing .0 in either column"
+
+    def test_non_integral_stats_keep_one_decimal_place(self):
```
crispy-wonton, line 1070: question: does 'integral' mean like 'non_integer' here? If so, I would change to the latter for understandability

asf_heat_pump_suitability/pipeline/validate/tests/test_compare_versions.py
```diff
@@ -1419,6 +1525,86 @@ def test_unreadable_file_type_raises(self):
         with pytest.raises(ValueError, match="csv"):
             compare_versions.load_df_cluster_areas("s3://bucket/dir/output.csv")
 
+    def test_geojson_layer_column_is_carried_alongside_areas(self, mocker):
+        """A multi-layer front-end geojson's areas must keep each feature's
+        layer, so the checks can filter to clusters."""
+        gdf = self.gdf_two_squares("EPSG:4326")
```
crispy-wonton, line 1531: question: what is this line doing?

asf_heat_pump_suitability/pipeline/validate/tests/test_compare_versions.py
```diff
@@ -1419,6 +1525,86 @@ def test_unreadable_file_type_raises(self):
...
+        ), "a frame without a layer column must pass through as the same frame"
+
+    def test_clusters_layer_name_comes_from_config(self):
+        """The clusters layer name is config, not code, and must match the
+        front-end file's real layer value (verified against the 20260806
+        East Lothian output)."""
+        assert (
+            config["compare_versions"]["cluster_layer"]
+            == "clusters_with_contextual_features"
+        ), "base.yaml must name the front-end file's real clusters layer"
```
crispy-wonton, line 1602: issue: this seems circular. I think a better thing would be to just verify that the key value pair exists in config. if we change it later, this test is very fragile.

asf_heat_pump_suitability/pipeline/validate/tests/test_compare_versions.py
```diff
@@ -1419,6 +1525,86 @@ def test_unreadable_file_type_raises(self):
...
+        assert (
+            compare_versions.CLUSTER_LAYER
+            == config["compare_versions"]["cluster_layer"]
+        ), "the module must read the clusters layer name from config"
```
crispy-wonton, line 1606: issue: same here, what's the purpose of this test? we basically just have the same code in two places

asf_heat_pump_suitability/pipeline/validate/tests/test_compare_versions.py
```diff
@@ -1529,6 +1715,46 @@ def test_stage_without_geometry_gets_no_distributions(self, df_old):
...
+            df_tabular,
+            df_tabular,
+            df_areas,
+            df_areas,
+        )
+        assert frames["n_UPRNs"][0]["n_UPRNs"].drop_nulls().to_list() == [
+            5,
+            8,
+        ], "ward rows must not enter the n_UPRNs distribution"
+        assert frames["area_m2"][1]["area_m2"].to_list() == [
```
crispy-wonton, line 1753: question: what are the indexes ([0] and [1]) doing in lines 1749 and 1753? and why are you calling the column name twice? don't understand the syntax here

PR #481, opened 2026-08-24
Bugfix 468 uprns for gb

### LINKED ISSUE #478: Fix running `uprns.py` for GB
Bugs detected so far include:
- reading GB as a list of 'g' and 'b'
- searching for grid squares which have no data because they are sea only squares

crispy-wonton:
Fixes #478 

# Description

Fixes bugs encountered when running `uprns.py` script for all of GB.

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/pipeline/transform/uprns.py
helloaidank: Should this be ["gb"] too?

helloaidank, line 561: These validation ranges may read the wrong source of truth in some edge cases. `get_dict_census_uprn_range` gets the raw CLI list, but `resolve_list_la_names` treats any list containing "gb" as a GB run (i.e. returns None). So `--local_authorities gb plymouth` runs nationally while validating against Plymouth's household count. I also checked, we don't have any guards against putting gb and a separate LA together....

helloaidank: Would suggest inputting "local_authority_dict["valid_local_authorities"]" into this function instead of local_authorities, think that would fix this issue!

helloaidank: Line 476 tests `local_authorities == "gb".` But local_authorities is a list, so it holds `["gb"]`. A list never equals a string, so this test is always false. Every GB run silently takes the else path instead.

Correcting this comparison will lead to an error further down, script would fail with `KeyError: LAD23CD` as we never join the LA boundaries in the if branch. 

If we delete lines 476 to 483 and unindent the rest, the surviving path is already be able to handle GB (guess we were actually doing this above anyway). `valid_local_authorities` is None, the getter is able to  load all boundaries, and `grid_squares` holds the 54 land squares from the config.

helloaidank (commented):
Hey! The actual diffs look good to me, there are just a couple of other bugs that might be worth fixing in this script (one involves deleting lines, and the other is a few line fix I think).

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: Great spot!

Have fixed an made a further edit in `load_geodata.py` so that we just load one file for all of GB instead of looping through all the grid squares.

crispy-wonton: yes! great spot ! I actually encountered this bug when running the `uprns.py` script for the whole of GB. This errored out my script because GB wasn't triggering the right validation and so it was falling back on the per-LA default counts.

crispy-wonton: have looked at the `resolve_list_la_names` and it handles lists containing GB plus other LAs:

```
    if la_names is None or "gb" in [name.lower() for name in la_names]:
        resolved_list = None
```

asf_heat_pump_suitability/getters/load_geodata.py
```diff
@@ -47,8 +47,11 @@ def load_df_osopen_uprn(
                 for sq in grid_squares
             ]
             return pl.read_parquet(paths, **kwargs)
-        path = config["data"]["geodata"]["uk_osopen_uprn_parquet"]
-        return pl.read_parquet(path, **kwargs)
+        if not grid_squares or set(grid_squares) == set(
+            config["constant"]["land_grid_squares"]
```
helloaidank: `set(grid_squares) == set(config["constant"]["land_grid_squares"])`can never be true in practice because we're checking in the above if statement that it is a list right?

helloaidank: Would suggest switching to:

```
if parquet:
    if not grid_squares or set(grid_squares) == set(config["constant"]["land_grid_squares"]):
        path = config["data"]["geodata"]["uk_osopen_uprn_parquet"]
        return pl.read_parquet(path, **kwargs)

    paths = [
        config["data"]["geodata"]["uk_osopen_uprn_partitioned"].format(grid_square=sq)
        for sq in grid_squares
    ]
    return pl.read_parquet(paths, **kwargs)
```

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -473,30 +474,22 @@ def parse_arguments() -> argparse.Namespace:
...
+        select_las=local_authority_dict["valid_local_authorities"]
+    )
+    uprns_df = load_geodata.load_df_osopen_uprn(
+        grid_squares=local_authority_dict["grid_squares"]
+    )
+    uprns_gdf = generate_gdf_uprn_coords(uprns_df)
+    del uprns_df
+
+    # Reduce UPRNs to only those within the LA boundaries and join LA code and name for later use
+    uprns_gdf = uprns_gdf.sjoin(
```
helloaidank, line 488: as an aside, have we ran this for GB? just realise it might be quite a big spatial join

asf_heat_pump_suitability/pipeline/transform/uprns.py
```diff
@@ -473,30 +474,22 @@ def parse_arguments() -> argparse.Namespace:
...
-            la_boundaries_gdf[["LAD23CD", "LAD23NM", "geometry"]],
-            how="inner",
-            predicate="intersects",
-        ).drop(columns="index_right")
+    print(f"Creating residential UPRN dataset for {local_authorities}...")
+    la_boundaries_gdf = load_boundaries.load_gdf_local_authority_boundaries(
+        select_las=local_authority_dict["valid_local_authorities"]
+    )
+    uprns_df = load_geodata.load_df_osopen_uprn(
+        grid_squares=local_authority_dict["grid_squares"]
```
helloaidank, line 482: think it would be worth adding a test for this function covering the three cases of the full GB list, one LA, and None.

helloaidank (commented):
Just one follow up fix but think it looks good to go to me otherwise! Thanks for implementing the fixes 👯

asf_heat_pump_suitability/pipeline/transform/uprns.py
helloaidank: can do this in a follow up issue too.

crispy-wonton: yes agree! we are generally missing tests for this module. Will leave for a separate PR!

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: On your first point about never being true, I'm not understanding why? This is essentially testing if someone has manually input all the land grid squares for GB, in which case - it triggers loading a single whole GB file instead of all the partitioned files separately. Am I missing your point here?

Second point: yeh good point - have made more efficient! Although I opted for explicit `else` statement as I personally find it easier to read :D

asf_heat_pump_suitability/pipeline/transform/uprns.py
crispy-wonton: Yes! Have run the script for all of GB and amazingly it did manage to run this join somehow. I was surprised.

asf_heat_pump_suitability/getters/load_geodata.py
helloaidank: I think my understanding was that the if branch above this condition was checking if it was a list and then it would go off to that branch. And that this below condition was then checking again for a list with the set condition `set(config["constant"]["land_grid_squares"])` but it would never make it there because it would be caught in the previous if statement. 

I think the above suggested change resolves this as an issue anyhow!

helloaidank (approved):
happy to merge!

asf_heat_pump_suitability/getters/load_geodata.py
crispy-wonton: Ahhhh!!! Yes I understand - you're absolutely right - woops! And yes I think the updated code fixes that!

PR #483, opened 2026-08-25
477 sampling script

crispy-wonton:
Fixes #477 

# Description

Add first version of sampling script to create sample of buildings containing flats for manual labelling in block of flats classification model.

### To be aware of: things still TODO:
There are some changes still to be made to the script, but as they don't affect the overall initial sampling methodology, I wanted to put this out for review anyway. Still TODO:
- Filter out buildings which have previously been sampled.
- Replace hard-coded file names with config keys.
- One thing we may want to do that requires more thinking: take the engineered features of all relevant buildings containing flats, compress them into principal components or flatten them in some other way, and then ensure that our sample is representative of the distribution of these features as well.

Note:
- These samples will be affected by the UPRN domestic labelling. Changes to the domestic labelling will change the count of domestic UPRNs in each building and potentially the population of buildings to sample from would change (mostly, improving the domestic identification model should reduce the population of buildings to sample from as fewer get labelled as containing domestic properties).

New files:
- `asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py` - main script to conduct sampling.

Modified files:
- `asf_heat_pump_suitability/config/base.yaml` - new IMD decile datasets and Scotland postcode lookup
- `asf_heat_pump_suitability/getters/load_data.py` - add filtering capacity to UPRN lookup getter
- `asf_heat_pump_suitability/pipeline/impute/property_type.py` - increase flexibility of flat type imputation function
- `asf_heat_pump_suitability/pipeline/run/optimise_inputs.py` - update in line with changes to `s3_utils.py`
- `asf_heat_pump_suitability/pipeline/transform/local_authority.py` - capitalise GB slug for better file saving
- `asf_heat_pump_suitability/utils/s3_utils.py` - fix bucket/prefix extraction function. Required in this PR 

# Instructions for Reviewer

You can test running this on Plymouth only to check that the code works (I haven't actually tested for a single area myself yet):
**PLEASE DO NOT RUN WITH `--save` AS THIS WILL OVERWRITE THE EXISTING FILE** 
```
uv run asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py --local_authorities plymouth --release_date [SELECT_RELEASE_DATE]
```
You can add the additional optional parameters as you wish.

To run for GB (this is what I've been running. Not recommended for test runs as it takes a while.):
```
uv run asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py --release_date 20260820
```

- The grouping of the attributes we are sampling for. We end up with over 1000 combinations of attributes so it would probably be useful to reduce the number of groups for each attribute. Are there any meaningful groupings that come to mind? I can also expand this to the wider team.
- What do you think of the sampling methodology? Do you think it will give us an appropriate sample to start labelling with or do you think we are undersampling from the most populous groups? My expectation is, although there are 1000+ combinations of attributes, I don't expect an equivalent number of archetypes (!!!). My goal with the sampling is just to ensure we are capturing enough variety of flat types to build a good model. My worry is that rarer archetypes maybe over-represented and subsequently the model may underperform on the most common types. However, I think it is likely we will need to conduct additional labelling as we continue to find model blind spots. So, my main question is: **is this a good enough sampling strategy to start labelling from?**

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/config/base.yaml
```diff
@@ -63,6 +63,17 @@ data:
...
+  imd_deciles:
+    # England source: https://www.gov.uk/government/statistics/english-indices-of-deprivation-2025
+    england: "s3://asf-local-heat-planning-tool/inputs/lookup/IMD2025_All_Ranks_Scores_Deciles_Population_Denominators_England.csv"
+    # Scotland source: https://www.gov.scot/publications/scottish-index-of-multiple-deprivation-2020v2-data-zone-look-up/
+    scotland: "s3://asf-local-heat-planning-tool/inputs/lookup/SIMD_2020v2_datazone_lookup_updated_2025.xlsx"
+    # Wales source: https://stats.gov.wales/en-GB/9706edd9-73ad-4902-bb12-7ccd7038626e#data
+    wales: "s3://asf-local-heat-planning-tool/inputs/lookup/welsh-index-of-multiple-deprivation-wimd-2025-index-and-domain-ranks-and-groups-for-lower-layer-super-output-areas-lsoa-v7-en-gb.csv"
+  lookups:
+    # Scotland source: https://www.nrscotland.gov.uk/publications/scottish-statistics-postcode-lookup/
+    scotland: "s3://asf-local-heat-planning-tool/inputs/lookup/Scotland_postcode_lookup.csv"
```
sofiapinto: can we please add these new data sources to the README.md

sofiapinto: I don't mind it at all, but just flagging that the names of these datasets aren't following the style guide.

asf_heat_pump_suitability/getters/load_data.py
```diff
@@ -87,13 +87,16 @@ def load_df_off_gas_pcds() -> pl.DataFrame:
 
 
 def load_df_uprn_lookup(
-    grid_squares: Optional[List[str]] = None, **kwargs
+    grid_squares: Optional[List[str]] = None,
+    uprn_filter: Optional[set] = None,
+    **kwargs,
 ) -> pl.DataFrame:
     """
     Load UPRN national statistics lookup for all of GB or a given list of grid squares if specified.
```
sofiapinto: docstring requires update based on the new param

asf_heat_pump_suitability/getters/load_data.py
```diff
@@ -102,12 +105,13 @@ def load_df_uprn_lookup(
...
-            [
-                pl.read_parquet(uri.format(grid_square=grid_square), **kwargs)
-                for grid_square in grid_squares
-            ]
-        )
+        dfs = [
+            pl.read_parquet(uri.format(grid_square=grid_square), **kwargs)
+            for grid_square in grid_squares
+        ]
+        if uprn_filter is not None:
```
sofiapinto:
```suggestion
        if uprn_filter:
```

asf_heat_pump_suitability/getters/load_data.py
```diff
@@ -118,7 +122,17 @@ def load_df_uprn_lookup(
...
         )
 
-        return pl.concat([pl.read_parquet(file, **kwargs) for file in files])
+        result = []
+        columns = kwargs.get("columns")
+        for file in files:
+            print(f"Loading UPRN national statistics lookup file: {file}...")
+            df = pl.read_parquet(file, **kwargs)
+            if columns is None or "ruc21ind" in columns:
+                df = df.with_columns(pl.col("ruc21ind").cast(pl.String))
```
sofiapinto, line 134: can we add a dosctring above please. why this specific col but no other?

asf_heat_pump_suitability/getters/load_data.py
```diff
@@ -118,7 +122,17 @@ def load_df_uprn_lookup(
...
+            if uprn_filter is not None:
+                df = df.join(uprn_filter, how="semi", on="UPRN")
```
sofiapinto, line 136: so cool, have never used `how=semi` before.

asf_heat_pump_suitability/pipeline/impute/property_type.py
```diff
@@ -1,25 +1,40 @@
...
+        duplicate_geoms = geom_counts[geom_counts > 1].index
 
-    # Filter the GeoDataFrame to only those geometries
-    flats = set(uprns_gdf[uprns_gdf["geometry"].isin(duplicate_geoms)]["UPRN"])
+        # Filter the GeoDataFrame to only those geometries
+        flats = set(uprns_gdf[uprns_gdf["geometry"].isin(duplicate_geoms)]["UPRN"])
+    else:
+        flats = set(
+            uprns_gdf.with_columns(
+                count_geoms=pl.col("UPRN").count().over([x_col, y_col])
```
sofiapinto, line 34: we don't expect multiple entries with the same UPRN right? if so we'd need to use `.n_unique()` here, otherwise we'd be counting the same UPRN twice and acknowledging it as having the same geometry as another property (which happens to be the same)

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
+"""
+Create a disproportional stratified sample of buildings containing flats for manual labelling to use in block of flats
+classifier model training.
+
+Buildings are grouped by area, rurality, construction age band, number of flats, deprivation
+group, and whether the building is predominantly flats (>80%). A target number of samples is
```
sofiapinto, line 6: which ones of these have missing values? is it just construction age band?

sofiapinto: Have you looked through the building archetypes and thought about whether setting predominantly flats (>80%) we could potentially be removing any of the archetypes out of our sample?

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
+"""
+Create a disproportional stratified sample of buildings containing flats for manual labelling to use in block of flats
```
sofiapinto: it's meant to be disproportional for what?

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+
+Options:
+    --local_authorities     One or more local authority names to filter to.
+                            Defaults to all of GB.
+                            e.g. --local_authorities plymouth
+                                 --local_authorities "glasgow city" "south lanarkshire"
+    --release_date          Release date of the input UPRN file in YYYYMMDD format.
+                            Defaults to today's date.
+    --seed                  Random seed for reproducibility. Default: 7.
+    --target_n              Target number of buildings in the sample. Default: 3000.
```
sofiapinto, line 23: Not about the code, more about the sampling we're doing now. How many are you hoping to sample, and is the idea to completely ignore the previously labelled samples? or to use those too?

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
sofiapinto, line 1: really neat script level dosctring, thank you! 🥇

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+    "1983-1991": "8_1983-1991",
+    "1991-1998": "9_1991-1998",
+    "1996-2002": "10_1996-2002",
+    "2003-2007": "11_2003-2007",
+    "2007 onwards": "12_2007 onwards",
+    "unknown": "",
+}
+
+
+def load_df_scotland_postcode_lookup() -> pl.DataFrame:
```
sofiapinto, line 66: could be moved to getters?

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+    Returns:
+        pl.DataFrame: Scotland postcode lookup
+    """
+    df = pl.read_csv(config["data"]["lookups"]["scotland"])
+    return df.with_columns(
+        pl.col("Postcode").str.strip_chars().name.keep(),
+    ).select(["Postcode", "DataZone2011Code"])
+
+
+def load_df_lsoa_imd_decile(nation: str = None) -> pl.DataFrame:
```
sofiapinto, line 79: could be moved to getters?

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+    grid_squares = local_authority_dict["grid_squares"]
+
+    print("Map UPRNs to building footprints...")
+    buildings_gdf = load_geodata.load_gdf_os_openmap_layer(
+        layer="building", grid_squares=grid_squares
+    )
+
+    if args.map_uprns_to_building:
+        print("Generating UPRN to building ID mapping...")
+        all_uprns_df = load_geodata.load_df_osopen_uprn()
```
sofiapinto: We only need to load all UPRNs if dealing with GB right?

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+        all_uprns_gdf = uprns.generate_gdf_uprn_coords(all_uprns_df)
+
+        uprn_building_mapping = uprns.map_dict_uprns_to_building_id(
+            uprns_gdf=all_uprns_gdf, buildings_gdf=buildings_gdf, id_col="ID"
+        )
+
+        del all_uprns_gdf, all_uprns_df
+        # Save mapping
+        fpath = "s3://asf-local-heat-planning-tool/outputs/models/block_of_flats_classifier/gb_uprn_to_building_mapping_non_domestic_and_domestic.parquet"
```
sofiapinto: want to move this to base.yaml?

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+            ],
+        )
+        .with_columns(
+            pl.col("lad25cd").str.starts_with("E09").alias("in_london"),
+        )
+        .rename({"ctry25cd": "country"})
+        .join(domestic_uprns, how="left", on="UPRN")
+        .with_columns(pl.col("is_domestic").fill_null(False))
+    )
+    del domestic_uprns
```
sofiapinto, line 275: I don't remember the uprns_df having all all these columns. Is this a latest addition?

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
sofiapinto, line 6: Remind me: what was the reasoning for using deprivation? i guess it is so we can have social housing blocks? we have deprivation at LSOA level right?

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+                )
+            )
+            .then(pl.lit("1976-2002"))
+            .when(
+                pl.col("construction_age_band").is_in(
+                    ["11_2003-2007", "12_2007 onwards"]
+                )
+            )
+            .then(pl.lit("2003 onwards"))
+            .otherwise(None)
```
sofiapinto, line 429: We need to think carefully about how we handle sampling from this group, thinking about

- Records without EPC. What are they biased towards?
- Records with EPC, but no building year information. What could they be biased towards?

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+        .join(domestic_uprns, how="left", on="UPRN")
+        .with_columns(pl.col("is_domestic").fill_null(False))
+    )
+    del domestic_uprns
+
+    # ------------------------------------ #
+    # ADD IMD DECILE DATA
+    # ------------------------------------ #
+    print("Add IMD decile data...")
+    scotland_dz_lookup_df = load_df_scotland_postcode_lookup()
```
sofiapinto, line 285: why do we need the scotland specific postcode lookup? i thought the lookup we typically use had UK coverage?

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+    local_authorities = [la.lower() for la in args.local_authorities]
+    local_authority_dict = local_authority.get_dict_la_data(local_authorities)
+    grid_squares = local_authority_dict["grid_squares"]
+
+    print("Map UPRNs to building footprints...")
+    buildings_gdf = load_geodata.load_gdf_os_openmap_layer(
+        layer="building", grid_squares=grid_squares
+    )
+
+    if args.map_uprns_to_building:
```
sofiapinto, line 217: why not move this below, where the `if not args.map_uprns_to_building:` is? would make it easier to read the code.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+    )
+    uprns_df = uprns_df.with_columns(
+        (pl.col("UPRN").is_in(flat_uprns) & pl.col("is_domestic")).alias("is_flat")
+    )
+    del flat_uprns
+
+    # ------------------------------------ #
+    # MAP UPRNS TO BUILDINGS
+    # ------------------------------------ #
+    # Load mapping if not generated
```
sofiapinto:
```suggestion
    # Load UPRN to building ID mapping (previously generated)
```

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+            # Add proportion of flats
+            (pl.col("n_flats") / pl.col("n_uprns")).alias("proportion_flats"),
+            # Group flat count
+            pl.when(pl.col("n_flats").is_between(2, 6, closed="both"))
+            .then(pl.lit("2-6_flats"))
+            .when(pl.col("n_flats").is_between(7, 15, closed="both"))
+            .then(pl.lit("7-15_flats"))
+            .when(pl.col("n_flats") > 15)
+            .then(pl.lit("16+_flats"))
+            .otherwise(None)
```
sofiapinto, line 395: We should revisit this once the archetypes are finalised.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+        pl.col(attributes).cast(pl.String).fill_null("unknown")
+    )
+    # Compute group sizes and initial per-group quota (capped at sample_n)
+    n_combinations = buildings_df.group_by(attributes).agg(pl.len()).height
+    sample_n = target_n // n_combinations
+    print(
+        f"There are {n_combinations} groups to sample from. Target of {sample_n} samples per group."
+    )
+
+    group_counts = (
```
sofiapinto, line 483:
```suggestion
    # Calculates the target sampling count (n_to_sample) for each attribute group, 
    # capped at sample_n or the group's total size (group_size), whichever is smaller.
    group_counts = (
```

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+        .with_columns(
+            pl.min_horizontal(pl.col("group_size"), pl.lit(sample_n)).alias(
+                "n_to_sample"
+            )
+        )
+    )
+
+    # Distribute shortfall (from underpopulated groups) evenly across groups with remaining capacity
+    shortfall = target_n - int(group_counts["n_to_sample"].sum())
+    if shortfall > 0:
```
sofiapinto, line 497: I believe that, in some cases, there might still be a small number left to reassign after this `if`. You might need a `while` loop and some changes to the code within it.

In other cases, you might actually get over the number you originally intended to sample, given the way you're using `remainder` below.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+        how="left",
+    )
+
+    sampled_ids = (
+        buildings_with_quota.with_columns(
+            pl.int_range(pl.len()).shuffle(seed=seed).over(attributes).alias("_rank")
+        )
+        .filter(pl.col("_rank") < pl.col("n_to_sample"))
+        .select("building_id")
+    )
```
sofiapinto, line 560:
```suggestion
    sampled_ids = (
        buildings_with_quota.with_columns(
            #Generates a sequential integer index from 0 to N - 1 for all rows in the DataFrame.
            pl.int_range(pl.len())
            # Randomly shuffles the row indices independently within each group of `attributes`
            .shuffle(seed=seed).over(attributes)
            # Assigns the randomised integer index to a temporary column called `_rank`
            .alias("_rank")
        )
        # Keeps only the rows where `_rank` is strictly less than `n_to_sample`
        .filter(pl.col("_rank") < pl.col("n_to_sample"))
        .select("building_id")
    )
```

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+        sample_df.to_pandas(), how="inner", left_on="ID", right_on="building_id"
+    )
+    del buildings_gdf
+
+    # ------------------------------------ #
+    # ADD GOOGLE MAPS URL TO EACH SAMPLE
+    # ------------------------------------ #
+    print("Enrich with Google Maps URL...")
+    # Convert to 4326 projection and create google maps URL
+    sample_gdf = sample_gdf.merge(
```
sofiapinto: we probably don't need a merge here? we could do something like

sample_gdf[["x", "y"]] = sample_gdf.centroid.get_coordinates()

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+        sample_gdf.centroid.get_coordinates(),
+        how="left",
+        left_index=True,
+        right_index=True,
+    ).to_crs(epsg=4326)
+    sample_gdf["url"] = sample_gdf.apply(
+        lambda row: f"https://www.google.com/maps/search/?api=1&query={row['y']},{row['x']}",
+        axis=1,
+    )
```
sofiapinto: If you want to avoid using apply:

```
sample_gdf["url"] = (
    "https://www.google.com/maps/search/?api=1&query="
    + sample_gdf["y"].astype(str)
    + ","
    + sample_gdf["x"].astype(str)
)
```

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+    BUCKET = "asf-local-heat-planning-tool"
+    kml = simplekml.Kml()
+    for idx, r in sample_gdf.iterrows():
+        pol = kml.newpolygon(
+            name="unlabelled",
+            description=f"Location: {r['url']}\nN flats: {r['n_flats']}\nN total: {r['n_total']}",
+            outerboundaryis=list(r["geometry"].exterior.coords),
+        )
+        pol.style.polystyle.color = "9939FF14"
+        pol.style.polystyle.outline = 1
```
sofiapinto: Depending on the size of the sample, using `iterrows()` might get really slow, so you could do:

```
for url, n_flats, n_total, geom in zip(
    sample_gdf["url"],
    sample_gdf["n_flats"],
    sample_gdf["n_total"],
    sample_gdf["geometry"],
):
    pol = kml.newpolygon(
        name="unlabelled",
        description=f"Location: {url}\nN flats: {n_flats}\nN total: {n_total}",
        outerboundaryis=list(geom.exterior.coords),
    )
    pol.style.polystyle.color = "9939FF14"
    pol.style.polystyle.outline = 1
```

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+    l = len(sample_gdf)
+    fpath = (
+        f"{today}_UNLABELLED_GB_buildings_containing_flats_sample_n{l}_seed{seed}.kml"
+    )
```
sofiapinto: will you assign the labeller in a separate file?

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -0,0 +1,596 @@
...
+    kml.save(fpath)
```
sofiapinto, line 614: This won't save to the local outputs folder, but prob should.

sofiapinto (commented):
Hey @crispy-wonton ,

Really great work! 🥇 

A few thoughts:
- I think we still need to think through the potential bias of the `unknown`'s, especially for construction age band.
- As per the chat yesterday, it may be good to allow for the script to generate proportionate and disproportionate samples. The disproportionate set could be used for training, covering ~60-70% of the samples we're annotating. The proportionate set could cover ~30-40%. Part of it would be used for validation, and the other part would be the unseen test.
- The script would benefit from some refactoring into functions for future understanding. 

P.S. Some of the questions I have left we're prior to our chat about the sampling strategy, so might be the same as the ones I asked during the call.

sofiapinto:
> These samples will be affected by the UPRN domestic labelling. Changes to the domestic labelling will change the count of domestic UPRNs in each building and potentially the population of buildings to sample from would change (mostly, improving the domestic identification model should reduce the population of buildings to sample from as fewer get labelled as containing domestic properties).

Question: did we apply the threshold for all areas? or only plymouth uses the threshold at the moment?

> The grouping of the attributes we are sampling for. We end up with over 1000 combinations of attributes so it would probably be useful to reduce the number of groups for each attribute. Are there any meaningful groupings that come to mind? I can also expand this to the wider team.

I haven't yet run the script. I will do and will think about the combinations.

sofiapinto:
On your question about the attributes used for stratification. Ideally we would reduce the strata to 50/100 groups max - over 1000 is definitely too much, and we probably end up with a sample that's not very meaningful. A few things we could do:

- Collapsing groups:
   - **Construction age band**: Pre-1950, 1950-2003, Post-2003. It could also be Pre-1930, 1930-2003, Post-2003. But ideally no more than 3 groups.  _Unknown_: we need to double check if the Unknowns might be biased to a specific category before we assess what to do about them. Depending on that, and also with a part our sample being proportionate stratified (or random?), then not including them here might become less of a problem? to discuss..
   - Over 80% of UPRNs in building are flats: we can probably get rid of this one and use _**Count of flats group**_ only.
   - **IMD**: Low and medium deprivation (<8), High deprivation (>=8)
   - **Rurality**: urban vs rural.
- We could also stop doing full cross-tabulation. Instead we can use primary and secondary stratification: Move to stratifying across 2 or 3 variables only (such as country/area, construction age band, and flat count) to form the main strata. Then we can use the remaining variables (Rurality and IMD) as secondary balancing constraints?

asf_heat_pump_suitability/getters/load_data.py
crispy-wonton: same! a new discovery for me!

asf_heat_pump_suitability/pipeline/impute/property_type.py
crispy-wonton: Yes, `uprns_gdf` should contain unique UPRNs so I used `count` instead of `nunique` here as it's more computationally efficient, especially when `uprns_gdf` is large. I can specify in the docstring that `uprns_gdf` must contain unique UPRN values

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
crispy-wonton: I will double check about the missing values but I believe rurality and construction age band. I need to check about the others.

On your second question: I'm not removing any buildings based on the proportion of flats, this is just a binary feature that buildings are grouped on because I imagine it would separate some of the archetypes out better for sampling. E.g. mixed terraces of some flats and some single properties, vs. a block of flats would have different proporitons.

crispy-wonton: Yes, IMD decile is at LSOA/DZ level. I included it because I think there will be a diversity of archetypes related to deprivation levels. E.g. tenements and smaller purpose-built social housing blocks versus mansion blocks for example.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
crispy-wonton: I know we discussed in the sprint planning but noting here for future reference:
- I expect to initially sample around 3000. This is of course a very large number to label so if you think we should start with a smaller number then do let me know. We can always add more later.
- My thinking was that we would use the previously labelled sample, just because of the high resource required for labelling. If we did use it, we would need to ensure the labelling is consistent with the updated archetypes. However, I can see there could be arguments for not including it - e.g. lower quality of labelling of some of the data; inconsistency of sampling technique; skewing the distribution of sampled data.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
crispy-wonton: To your questions here and below - it definitely could be moved to getters. The reason I kept it here for now is because the sampling is separate from the main pipeline so I thought I'd keep functions only used in the sampling separate. If we end up using these datasets to generate new features for the model, they can be moved to `getters`.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
crispy-wonton: As above

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
crispy-wonton: I put this at the top of the script so that if the arg is True, the UPRN to building mapping happens first. This is because the process takes up a lot of memory with very large datasets. The large datasets are then deleted immediately after the processing and the rest of the script then continues. I did this to avoid OutOfMemory errors from crashing the code, as the rest of the script also uses lots of memory. It's basically a memory-saving choice. Will add to annotation.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
crispy-wonton: Yes, I can update this to filter for LAs if the arg is passed.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
crispy-wonton: No this is a different dataset. I just used the same variable name of `uprns_df` because it is still a UPRNs-based dataset where UPRNs are the ID. Usually, we use the OS Open UPRN data which is just the UPRN plus the coordinates and lat/lon. Here, I am using the National Statistics UPRN Lookup dataset which contains UPRNs with census information about the area the UPRN is located in, as well as postcode & area codes.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
crispy-wonton: Thanks for this question - I will add to the annotation.

This is because the National Statistics UPRN Lookup (NSUL) (which has GB/ UK coverage, cant remember) has a IMD column but it has no data in it so I had to find separate sources of IMD data.

The IMD data available for Scotland is from 2019, but the Data Zones to join it to in the NSUL are from 2022. I therefore needed to find a lookup for postcodes to 2019 DZs in Scotland so that I can join UPRNs to their 2019 DZs and then join the 2019 IMD deciles onto them. This was required because the DZs changed between 2019 and 2022.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
crispy-wonton: Yes this was intentional as I was aiming to sample the approximate number required rather than add more complexity to the code but will revisit to see if I can improve this!

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
crispy-wonton: Yes not sure why I did this but I copied across from the older sampling script here - will revisit but I think you are right!

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
crispy-wonton: True!

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
crispy-wonton: I can assign it here or separately. At the time of writing this PR, I wasn't sure how we would distribute the labelling so I left it off for now.

crispy-wonton:
Thanks so much for your review and all your thoughts @sofiapinto !

I have responded to your comments and will now go through to action them. Regarding these suggestions:

> * I think we still need to think through the potential bias of the `unknown`'s, especially for construction age band.
> * As per the chat yesterday, it may be good to allow for the script to generate proportionate and disproportionate samples. The disproportionate set could be used for training, covering ~60-70% of the samples we're annotating. The proportionate set could cover ~30-40%. Part of it would be used for validation, and the other part would be the unseen test.
> * The script would benefit from some refactoring into functions for future understanding.

I agree with all of these. My suggestion would be: given that the PR is already very large, I refine it according to your in-line comments; merge if you're happy; and then conduct these further changes in a separate PR. Let me know if you prefer it in this PR - I just think that might be easier!

crispy-wonton:
> Question: did we apply the threshold for all areas? or only plymouth uses the threshold at the moment?

I looked at the `uprns.py` script again. Actually, the threshold is not applied to any area, including Plymouth, when running for GB because of the way the code applies the threshold. When running for Plymouth specifically, it works as expected. The GB-level version will need to be fixed and rerun. Given that the threshold is for UPRN density, I think it shouldn't be too much of a risk to apply it to all LAs for this purpose. My thinking is:
1. Applying the threshold to all LAs might remove true buildings containing flats archetypes. The risk is that we lose true buildings containing flats from our sampling pool. I think it is unllikely that there will be many buildings that truly contain domestic flats above the threshold. I suppose the one archetype that this might remove is a really large, mostly commercial building that then has a small block of flats on one part of it. The other case could be a super luxury apartment block where all the flats are massive. The threshold is ~260m2 so I think it would be very rare. Probably not worth polluting the dataset with lots of false positives just for a very rare archetype at this initial stage.
2. Not applying the threshold means that we will get more non-domestic buildings, but these should at least be obvious when we label the data. The risk is that we end up with non-usable data in our sample and waste labelling time - this is the more pressing issue I think. Therefore I'd suggest going with option 1.

crispy-wonton:
> * Collapsing groups:
>   
>   * **Construction age band**: Pre-1950, 1950-2003, Post-2003. It could also be Pre-1930, 1930-2003, Post-2003. But ideally no more than 3 groups.  _Unknown_: we need to double check if the Unknowns might be biased to a specific category before we assess what to do about them. Depending on that, and also with a part our sample being proportionate stratified (or random?), then not including them here might become less of a problem? to discuss..
>   * Over 80% of UPRNs in building are flats: we can probably get rid of this one and use _**Count of flats group**_ only.
>   * **IMD**: Low and medium deprivation (<8), High deprivation (>=8)
>   * **Rurality**: urban vs rural.
> * We could also stop doing full cross-tabulation. Instead we can use primary and secondary stratification: Move to stratifying across 2 or 3 variables only (such as country/area, construction age band, and flat count) to form the main strata. Then we can use the remaining variables (Rurality and IMD) as secondary balancing constraints?

- Agree with collapsing the construction age band. We could even do pre and post 1950 to represent pre and post war basically.
- On the proportions one - I do think it's important as I think it can separate out mixed type buildings (flats & other types) vs flats only buildings which the counts alone can't do. E.g. I added this in to ensure we collect TT; TE; and TF types in our sample.
- Agree on IMD - probably the middle group doesn't give too much of a meaningful distinction.
- Rurality - I think I agree. The reason I initially kept them separate was because I was thinking about villages (small rural) vs small towns (large rural) vs urban (towns and cities). I thought it's possible that small and large rural would have different archetype compositions - e.g. in small towns I can imagine more purpose built blocks of 4-20 flats, vs small rural I would expect to have more small blocks, maisonettes, and maybe some mixed terraces where some of the units are flats but others are individual units. Anyway, I'm not feeling super confident in that so happy to collapse into a binary but just wanted to lay out my thinking here.

~~I like your idea of primary and secondary stratified sampling. Stupid question incoming : wouldn't that result in the same number of groups though, just created in a different order?~~

ETA: Having discussed this in more detail in the stand up, I understand your point on the primary stratification and secondary stratification being to: sample using a primary smaller selection of attributes and then checking that there is good representation **across the whole sample** (rather than within each group) of the secondary attributes. Can I just confirm that is correct?

asf_heat_pump_suitability/config/base.yaml
crispy-wonton: will add.

On your second point - yes I'm not too fussy about sticking to the exact style guide for naming here - as long as the name is descriptive enough. That said, I think the scotland lookup needs updating.

asf_heat_pump_suitability/config/README.md
```diff
@@ -15,7 +15,7 @@ This table contains the source data citations used in the project, including lin
...
...               | UPRNs of GB with their latitude and longitude.                                                                                                                                                                                                                                                                                           | Contains public sector information licensed under the [Open Government Licence v.3.0](https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/)                                                                                                                                                                                                                                                                                                                                                                                                                                                              | Baseline data                                                                                                                        | data; geodata; uk_osopen_uprn                                                                              |
```
crispy-wonton: Note: I've just corrected a typo in the word 'Open' here - no other changes to this source.

asf_heat_pump_suitability/config/README.md
```diff
@@ -40,3 +40,7 @@ This table contains the source data citations used in the project, including lin
...
...               | Welsh Index of Multiple Deprivation for Lower layer Super Output Areas (LSOA)                                                                                                                                                                                                                                                            | TBD                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            | Block of flats classification                                                                                                        | data; imd_deciles; wales                                                                                   |
```
crispy-wonton: Note: I cannot find licensing information for this Welsh IMD dataset yet. I still need to locate it but I was wondering if you mind taking a quick look at the website linked here in case you can see it and I'm missing something obvious: https://stats.gov.wales/en-GB/9706edd9-73ad-4902-bb12-7ccd7038626e#data

sofiapinto: I believe it is OGL, because that's the case for all stats gov wales data:  https://www.gov.wales/welsh-government-publication-scheme

but have asked the legal team :)

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
sofiapinto: The caveat is that if they aren't in the getters and we only use them in a couple of months, we might forget we already have these getters. but happy for this to be moved later.

sofiapinto (approved):
have reviewed changes & your answers. Have replied where applicable. Looks good since we're already branching off this work!

asf_heat_pump_suitability/config/README.md
crispy-wonton: Thanks so much for finding this!

Issue #485, opened 2026-08-26
Separate communal clusters by assignment origin


helloaidank:
## Current behaviour

Buildings become "Communal solution" via two routes: a block of flats outside a heat-network zone / city centre (decision tree), or a Networked-heat-pump building within the anchor radius of an anchor load. (A third route — a building whose homes split between District heat network and Networked heat pump — existed on dev when this issue was drafted, but the #452/#459 branch chain removed the DHN decision-tree path and that mixing rule.) Clustering dissolves buildings by assigned tech alone, so touching communal buildings merge into one cluster regardless of route, and the cluster-level logic trace (PR #466) cannot state one truthful reason — it hedges with "There might also be one or multiple blocks of flats in the cluster."

## Desired behaviour

Every communal building carries a communal-origin label (block of flats / anchor proximity). The decision-tree origin takes precedence when a building is also near an anchor; anchor proximity stays visible via the existing within-radius flag. Clustering only merges communal buildings sharing the same origin. Each communal cluster's logic trace then states its single reason. Includes the logic-trace rewrite (also fixing the runtime bugs on the communal branches).

## Acceptance criteria

- [ ] Building-level tech output includes a communal-origin value for every communal building
- [ ] No cluster contains communal buildings with different origins; same-origin neighbours still merge
- [ ] A block of flats within the anchor radius keeps origin "block of flats"
- [ ] Communal trace text has no hedging sentence and the two TODOs are removed
- [ ] Logic-trace code runs without error (tuple-call and `&`-precedence bugs fixed)
- [ ] Tests cover origin assignment and origin-aware dissolve

## Out of scope

- Anchor buildings inside clusters (#457)
- Separate anchor map layer (#453)
- Making each block of flats its own single-building cluster
- A "mixed solutions within building" origin — the rule producing such buildings was removed upstream; revisit only if DHN mixing returns

Related: #459, PR #466. Asana: https://app.asana.com/1/5571817120120/project/1214222223606748/task/1217818496734712

helloaidank:
Amended during implementation: the mixed DHN/NHP → Communal origin was dropped. This work is stacked on `459_logic_trace`, whose base chain (#452/#459) already removed the DHN decision-tree path and the mixing rule, so that origin has nothing to attach to. Spec updated accordingly (`docs/specs/485_splitCommunalClustersByOrigin.md`, decision 4).

PR #490, opened 2026-08-27
485 Split communal clusters by assignment origin

helloaidank:
> **Stacked PR.** Base is `459_logic_trace` (#466), not `dev`. Review only the commits on top of `33aef82`.

# Description

Fixes #485. Spec: `docs/specs/485_splitCommunalClustersByOrigin.md` (on this branch).

**Problem.** A building can become "Communal solution" for two different reasons: it's a block of flats (decision tree), or it's a networked building within 50m of an anchor like a school or hospital. Clustering dissolved on technology alone, so touching communal buildings merged no matter *why* they were communal, and the cluster's `logic_trace` had no single truthful reason to state — hence the "There might also be one or multiple blocks of flats" hedge and the two TODOs on the base branch.

**Change.** Track the reason at building level and cluster on it.

- New building-level `communal_origin` column, set at the point each origin is decided: decision tree writes `"block of flats"`, anchor reassignment writes `"anchor proximity"` and only for buildings it actually flips (buildings already communal keep their decision-tree origin).
- Clusters now dissolve on `(assigned_tech, communal_origin)` with `dropna=False`, so non-communal clustering is unchanged.
- Communal `logic_trace` branches are keyed on origin; hedge wording and both TODOs removed.
- Fixes a live operator-precedence bug in the trace conditions (`&` binding tighter than `==`).
- `communal_origin` is written to the output geojson, with a description row in the metadata.

# Instructions for Reviewer

Start with the spec, then read `test_cluster_dissolve_splits_communal_origins` — it's the clearest statement of the behaviour change (9 clusters become 10 once origins are separated).

Please pay special attention to whether "only flip buildings that weren't already communal" is the right call in the anchor-reassignment step, and to the trace wording for each origin.

**Testing:** 13 new tests written red-first across `decision_tree`, `cluster`, and `compute_contextual_features`, including one cluster per trace branch. Full suite is 69 green; `pre-commit` green. No new setup or local files needed — `python -m pytest asf_heat_pump_suitability` from the repo root.

**@sofiapinto** — this rebases cleanly onto your latest `459_logic_trace` tip (`33aef82`), and your trace-function extraction and wording are adopted unchanged outside the communal branches. One thing worth flagging on your branch: `logic_trace` itself still has no row in the "Variable names and descriptions" metadata.

# Checklist:

- [x] I have tested the code


asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -133,15 +137,21 @@ def generate_gdf_clusters(
         radius=radius,
     )
 
+    reassigned_lookup = reassigned_gdf.set_index(id_col)
     cells_gdf["assigned_tech"] = cells_gdf[id_col].map(
-        reassigned_gdf.set_index(id_col).to_dict()["assigned_tech"]
+        reassigned_lookup["assigned_tech"]
```
sofiapinto, line 142: doesn't this need to be:
```suggestion
        reassigned_lookup.to_dict()["assigned_tech"]
```

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -133,15 +137,21 @@ def generate_gdf_clusters(
...
+        reassigned_lookup["communal_origin"]
     )
 
-    # Create cluster geometries
+    # Create cluster geometries. Dissolving on (assigned_tech, communal_origin) keeps
+    # communal clusters with different origins separate; dropna=False retains the
+    # non-communal cells, whose communal_origin is null.
     clusters_gdf = (
-        cells_gdf.dissolve(by="assigned_tech")
+        cells_gdf.dissolve(by=["assigned_tech", "communal_origin"], dropna=False)
```
sofiapinto, line 152: why are we doing dropna=False?

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -737,10 +717,11 @@ def reassign_gdf_near_anchor_properties(
     radius: float,
 ) -> gpd.GeoDataFrame:
     """
-    Reassign building tech type to communal if within a given radius of an anchor load property, if assigned N-GSHP by the decision tree
+    Reassign building tech type to communal if within a given radius of an anchor load property, if assigned N-GSHP by the decision tree.
+    Flipped buildings get the anchor-proximity `communal_origin`; buildings already communal keep their decision-tree origin.
```
sofiapinto, line 721: by flipped do you mean reassigned?


```suggestion
    Buildings getting their technology reassigned due to anchor-proximity get `communal_origin` updated to reflect that; buildings already communal keep their original value of `communal_origin`.
```

asf_heat_pump_suitability/pipeline/cluster/cluster.py
```diff
@@ -757,11 +738,21 @@ def reassign_gdf_near_anchor_properties(
 
     # distance_m column now reads a number (distance from anchor) for all buildings within radius of anchor, and NaN for all outside of that radius
     # if distance column is not NaN (i.e. building is within the radius of an anchor), reassign tech type according to the map
+    # Flag the buildings the reassignment flips to communal, before flipping them.
+    # Buildings that were already communal keep their decision-tree origin.
+    newly_communal = tech_gdf["distance_m"].notna() & (
+        tech_gdf["assigned_tech"] == NETWORKED
```
sofiapinto, line 744: is this enough to identify those close to an anchor load? `tech_gdf["distance_m"].notna()`

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
sofiapinto, line 417: Why are these being removed? They need to be mapped to True/False. Or is this because of the merge conflicts. It should be whatever is in dev :)

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -189,21 +196,37 @@ def identify_df_building_most_suitable_tech(
...
     buildings_with_multiple_solutions_df = assign_df_unique_solution(
         buildings_with_multiple_solutions_df
     )
     buildings_with_multiple_solutions_df = buildings_with_multiple_solutions_df.select(
-        [id_col, "assigned_tech"]
+        [id_col, "assigned_tech", "communal_origin"]
     )
 
     # For building footprints with only 1 solution, assign that solution as the unique solution for the building
+    # A single-solution communal building means all its UPRNs are in a block of flats
```
sofiapinto, line 218: I'd double check if this is the best place to make this change. 

We shouldn't just assign "blocks of flats" when there's a single solution. Even where there's multiple solutions, but communal is the selected one because it's the most collaborative, we should still tag as blocks of flats. Basically, at the end of the assignments in decision_tree.py, all communal solutions are considered blocks of fats. I understand that this may lead to some "mistakes" (some things won't effectively be blocks of flats), but if we don't do this, then these that don't get tagged as blocks of flats will later be tagged as "reassignment due to anchor load", and that is wrong.

In summary: 
- The reassignments because of anchor loads only happen later in the pipeline, in `cluster.py`. So all communal identified in `decision_tree.py` should be tagged as blocks of flats.
- What I want to avoid is "communal solutions" identified in `decision_tree.py` getting null `communal_origin`. Could you please double check this isn't happening?

asf_heat_pump_suitability/pipeline/transform/decision_tree.py
```diff
@@ -260,7 +286,14 @@ def assign_df_unique_solution(solutions_per_footprint_df: pl.DataFrame) -> pl.Da
...
             .otherwise(pl.lit("Unexpected combination"))
-        )
+        ),
+        # A set containing communal means at least one UPRN is in a block of flats,
+        # which is the only route to a communal UPRN in the decision tree
+        communal_origin=(
+            pl.when(pl.col("assigned_tech").list.contains(TECH_TYPES["communal"]))
+            .then(pl.lit(COMMUNAL_ORIGIN["block_of_flats"]))
+            .otherwise(pl.lit(None, dtype=pl.String))
+        ),
```
sofiapinto, line 296: Looking at this I believe it is probably fine to have the code above + this, but I wonder if this tagging with the communal original could be done in one go at the end of decision_tree.py just before saving?

sofiapinto (commented):
Hey @helloaidank , thanks for this, looks great! ✨ 

Taking care of merge conflicts:
There's a couple of merge conflicts with dev based on some latest changes i've made to branches (that this one was originally coming off from) that need to be carefully taken care of.

Having taken a look at it, this is what I would propose (but please double check):
- README.md: you should accept the changes from dev.
- load_boundaries.py: you should accept the incoming changes from dev.
- compute_contextual_features.py:
    - first conflict: stick with your branch
    - second conflict: accept both
    - remaining ones: stick with dev


Code review - On the actual code from this PR:
- I only had time to review the changes to the code required to separate the two communal cases. Have left some questions, asked you to double check a few things and left some suggestions too.
- I didn't have time to review the tests. I'd propose moving the tests to a separate PR, that branches of this one.

This will require a second review - happy to re-review when i'm back, but otherwise happy for you to ask @crispy-wonton if she has time.

PR #506, opened 2026-09-10
499 fix bug due to lack of default in cluster.py's map_df_uprns_to_clusters()

### LINKED ISSUE #499: bug due to lack of default in cluster.py's map_df_uprns_to_clusters()
```
Traceback (most recent call last):
    uprns_df = cluster.map_df_uprns_to_clusters(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
....
polars.exceptions.InvalidOperationError: incomplete mapping specified for `replace_strict`
```

sofiapinto:
---

# Description

Fix bug due to lack of default in cluster.py's `map_df_uprns_to_clusters()`

Fixes #499 

Question: is there a reason for this to be happening now and not have been a problem before? Is adding the default just masking an existing problem? The number of properties without mapping is low.

### DIFF
--- a/asf_heat_pump_suitability/pipeline/cluster/cluster.py
+++ b/asf_heat_pump_suitability/pipeline/cluster/cluster.py
@@ -796,11 +796,12 @@ def map_df_uprns_to_clusters(
         "cluster_id"
     ].to_dict()
     uprns_df = uprns_df.with_columns(
-        pl.col(building_id).replace_strict(building_cluster_mapping).alias("cluster_id")
+        pl.col(building_id)
+        .replace_strict(building_cluster_mapping, default=None)
+        .alias("cluster_id")
     )
 
     return uprns_df
-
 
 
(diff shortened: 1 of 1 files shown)


crispy-wonton:
> Question: is there a reason for this to be happening now and not have been a problem before? Is adding the default just masking an existing problem? The number of properties without mapping is low.

Do you mean you weren't getting this error on a specific LA and now you are on the same LA? If so, I think you would need to investigate further before merging because, as you've said, this could be masking an error. I'm guessing there's some upstream change we made that now causes there to be UPRNs with no cluster. Are these UPRNs matched to a building footprint?

sofiapinto:
hey @crispy-wonton - I have now added a TODO so that we can 
1) merge this to fix the issue temporarily,
2) investigate the issue separately

crispy-wonton (approved):

PR #507, opened 2026-09-10
500 Fix issue caused by areas with no spatial signatures

### LINKED ISSUE #500: bug in `compute_contextual_features.py`
`spatial_signatures_gdf["source_annotation"].iloc[0]` errors when there are no spatial signatures. We need to account for the cases where no spatial signature are found

sofiapinto:
---

# Description

`spatial_signatures_gdf["source_annotation"].iloc[0]` errors when there are no spatial signatures. We need to account for the cases where no spatial signature are found. Have done the same for HN zones too.

Fixes #500

### DIFF
--- a/asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
+++ b/asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
@@ -420,14 +420,21 @@ def create_gdf_contextual_features(
     geo_utils.verify_gdf_crs(spatial_signatures_gdf, target_crs=target_crs)
 
     # Add in_hn_zone and in_city_centre flags to clusters_gdf
-    clusters_with_contextual_features_gdf["in_hn_zone"] = (
-        clusters_with_contextual_features_gdf.intersects(hn_zones_gdf.union_all())
-    ).map({True: "Yes", False: "No"})
-    clusters_with_contextual_features_gdf["in_city_centre"] = (
-        clusters_with_contextual_features_gdf.intersects(
-            spatial_signatures_gdf.union_all()
-        )
-    ).map({True: "Yes", False: "No"})
+    if hn_zones_gdf is not None and not hn_zones_gdf.empty:
+        clusters_with_contextual_features_gdf["in_hn_zone"] = (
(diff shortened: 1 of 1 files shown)


asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -420,14 +420,21 @@ def create_gdf_contextual_features(
...
 
     # Add in_hn_zone and in_city_centre flags to clusters_gdf
-    clusters_with_contextual_features_gdf["in_hn_zone"] = (
-        clusters_with_contextual_features_gdf.intersects(hn_zones_gdf.union_all())
-    ).map({True: "Yes", False: "No"})
-    clusters_with_contextual_features_gdf["in_city_centre"] = (
-        clusters_with_contextual_features_gdf.intersects(
-            spatial_signatures_gdf.union_all()
-        )
-    ).map({True: "Yes", False: "No"})
```
sofiapinto, line 430: this part of the code wasn't erroring, but still made the additions below, as some safeguards.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
```diff
@@ -602,21 +609,28 @@ def create_json_contextual_features_metadata(
...
-                    "source_annotation": [
-                        spatial_signatures_gdf["source_annotation"].iloc[0]
-                    ],
-                    "geometry": [spatial_signatures_gdf.geometry.union_all()],
-                },
-                crs=spatial_signatures_gdf.crs,
-            ),
-        ]
-    ).to_crs(epsg=4326)
+    if len(spatial_signatures_gdf) > 0:
```
crispy-wonton, line 612: question: won't this break if `len(hn_zones_gdf) == 0`?

crispy-wonton (commented):
Think there's still a potential fragility in the new code.

asf_heat_pump_suitability/pipeline/run/compute_contextual_features.py
sofiapinto: It didn't break, but i have updated the code to have all possible cases.

Issue #503, opened 2026-09-10
Implement proportional + padding sampling methodology


crispy-wonton:
For training / validation set: take the proportional + padding forward, with a defined floor of ~50 samples per primary group (to get to 1200 samples), then sample proportionally for the remaining 1250 samples.

For the test set: sample purely proportionally from the whole population across the primary and secondary constraints.

PR #508, opened 2026-09-10
503 implement proportional padding

crispy-wonton:
Fixes #503 

# Description

Modified `sample_for_labelling.py` to create full sample for labelling according to the requirements outlined in #503 and refactored some parts of the script into functions. Also implemented optional saving which wasn't implemented correctly before (missing `if` statement to trigger it).

# Instructions for Reviewer

Run:
```
uv run asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py --release_date 20260820
```

- Lines 805-864 - these are the major changes to the methodology. The functions (except `_calculate_float_constraint_proportions` are basically just refactoring existing code - it would still be good to check the parts where I've added keyword arguments in them though).
- Is the final sample doing what we've laid out in #503 ?
- ~~I've removed buildings that we've already labelled - please can you check that you agree I've done this in the right place in the script?~~
- ETA: on second thought, I decided to keep the buildings that we've already labelled to include in the sample. I can then assign them back to their original labels which we can then review as we do the manual labelling. The reason I decided to keep them is because removing them changes the proportions of the groups in the whole population, and it affects the smaller groups the most. My proposal is therefore to retain them in the whole population and run the sampling. Then we conduct training on the new sample, but exclude the remaining labelled buildings from the first iteration so as not to over-index on Plymouth buildings. We can then think more carefully about how to combine the old labelled data in a way that doesn't skew the model. Let me know if you don't think that's a problem as I might be overthinking it!

# Checklist:

- [ ] I have tested the code


asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -295,6 +306,130 @@ def generate_df_sampling_cells(
...
+    """
+    if even:
+        return 0.5
+    else:
+        return round(
+            population_df.filter(pl.col(attribute)).height / population_df.height, 2
+        )
+
+
+def sample_df_by_quota(
```
crispy-wonton, line 332: Mostly the same as before but please double check the kwargs and how they are used in the function.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -295,6 +306,130 @@ def generate_df_sampling_cells(
...
+        )
+        # Keep rows where '_rank' is less than 'n_to_sample' - this is the method to identify the sample rows
+        .filter(pl.col("_rank") < pl.col(quota_col)).select(id_col)
+    )
+
+    # Filter population dataset to sample IDs
+    return population_df.filter(pl.col(id_col).is_in(sampled_ids[id_col].to_list()))
+
+
+def _enrich_gdf_google_maps_url(gdf: gpd.GeoDataFrame) -> gpd.GeoDataFrame:
```
crispy-wonton, line 370: Exactly the same as before but please double check the `gdf` kwargs and how they are used in the function.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -295,6 +306,130 @@ def generate_df_sampling_cells(
...
+    gdf["url"] = (
+        "https://www.google.com/maps/search/?api=1&query="
+        + gdf["y"].astype(str)
+        + ","
+        + gdf["x"].astype(str)
+    )
+    return gdf
+
+
+def save_building_sample_to_kml(
```
crispy-wonton, line 393: Mostly the same as before but please double check the kwargs and how they are used in the function.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -166,7 +173,8 @@ def calculate_array_sample_allocations(
     constraints.
 
     Args:
-        grouped_df (pl.DataFrame): sampling cells containing building counts for each unique combination of primary and
+        population_df (pl.DataFrame): whole population dataframe to sample from. Must contain secondary attribute columns.
```
sofiapinto, line 194: Is this correct?
```suggestion
        population_df (pl.DataFrame): whole population dataframe to sample from. Must contain primary and secondary attribute columns.
```

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -295,6 +306,130 @@ def generate_df_sampling_cells(
     )
 
 
+def _calculate_float_constraint_proportions(
```
sofiapinto, line 325: nice :)

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
sofiapinto: looks good :)

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
sofiapinto: looks good!

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -618,32 +749,30 @@ def parse_arguments() -> argparse.Namespace:
...
+        .with_columns((pl.col("proportion_flats") > 0.8).alias("over_80_pc_flats"))
     )
 
     del uprns_df
 
-    save_utils.save_to_s3(
-        df=buildings_df,
-        path="s3://asf-local-heat-planning-tool/outputs/models/block_of_flats_classifier/GB_enriched_buildings_with_flats.parquet",
-    )
+    if args.save:
```
sofiapinto, line 776: ah! nice :D i did wonder about this in the previous PR.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -618,32 +749,30 @@ def parse_arguments() -> argparse.Namespace:
...
-        df=buildings_gdf,
-        path="s3://asf-local-heat-planning-tool/outputs/models/block_of_flats_classifier/GB_enriched_buildings_with_flats_and_geometries.parquet",
-    )
-    del _save_buildings_gdf
+        _save_buildings_gdf = buildings_gdf[["ID", "geometry"]].merge(
+            buildings_df.to_pandas(), how="inner", left_on="ID", right_on="building_id"
+        )
+        save_utils.save_to_s3(
+            df=buildings_gdf,
+            path="s3://asf-local-heat-planning-tool/outputs/models/block_of_flats_classifier/GB_enriched_buildings_with_flats_and_geometries.parquet",
```
sofiapinto: do we want to move this to base.yaml?

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -671,88 +800,82 @@ def parse_arguments() -> argparse.Namespace:
...
-    ):
-        pol = kml.newpolygon(
-            name="unlabelled",
-            description=f"Location: {url} -------- N flats: {n_flats} -------- N total: {n_total}",
-            outerboundaryis=list(geom.exterior.coords),
+    if args.save:
+        fname = f"{release_date}_UNLABELLED_GB_buildings_containing_flats_sample_n{len(sample_gdf)}_seed{seed}"
+        save_utils.save_to_s3(
+            sample_df,
+            path=f"s3://asf-local-heat-planning-tool/outputs/models/block_of_flats_classifier/{fname}.parquet",
```
sofiapinto: do we want to move path and fname to base.yaml

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -671,88 +800,82 @@ def parse_arguments() -> argparse.Namespace:
...
+        )
+        s3 = boto3.resource("s3")
+        BUCKET = "asf-local-heat-planning-tool"
```
sofiapinto: this could live in base.yaml

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
sofiapinto: looks good!

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -671,88 +800,82 @@ def parse_arguments() -> argparse.Namespace:
         primary_attributes=primary_strata,
         secondary_attributes=secondary_constraints,
     )
+
+    # Split the training set into a padding portion (~50 samples per primary stratum) and a proportional group of the remainder
+    padding_n = 50 * group_counts_df["primary_strata"].n_unique()
+    remaining_n = training_n - padding_n
+    # Padding sample is even, remaining sample is proportional (True/False)
+    sampling_strategy = [(padding_n, True), (remaining_n, False)]
```
sofiapinto, line 831: this is true for the training set, but for the unseen test set we want the sample purely proportionally from the whole population across the primary and secondary constraints.

sofiapinto: oh I think I understand what you're doing now. you're using this script not to either take a proportional or even sample, depending on the params you pass. 

you're using this script to sample for the training and unseen test in one go.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -295,6 +306,130 @@ def generate_df_sampling_cells(
...
+    Args:
+        population_df (pl.DataFrame): whole population dataframe with binary attribute column
+        attribute (str): binary attribute to calculate proportions for
+        even (bool): set to `True` for even proportions (0.5) or `False` for whole-population proportions
+
+    Returns:
+        float: proportions of positive class
+    """
+    if even:
+        return 0.5
```
sofiapinto, line 341: Looking at this more closely, it seems like this function is only ready to do proportional sampling where the attribute is binary. So i'm guessing you're only thinking of applying this to secondary constraints, even for the creation of the test set? 

So... for the test set, the sample will still be even for the primary constraints, but proportional on the secondary constraints? was this intentional?

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
sofiapinto, line 1: Script docstring should be updated.

sofiapinto (commented):
Hey, this looks great @crispy-wonton thank you!

The main question I left was about how we're sampling for the test set: is it that we're still doing even sampling on the primary constraints, but proportional on the secondary? Or is it fully proportional for the unseen test set?

Because i'll be on leave when finish answering/actioning suggestions, perhaps ask for a second review from Seth to ensure the methodology is as discussed.

The script dosctring should be updated too.

Have left some minor suggestions: a few things could be moved to base.yaml. Perhaps some functions could also live in utils, rather than in this script. But all minor, happy for you to not be implemented in this PR.


And to answer the questions you left in your PR summary:

> Is the final sample doing what we've laid out in https://github.com/nestauk/asf_heat_pump_suitability/issues/503 ?

I'm not sure about the test set that we're _"sampling purely proportionally from the whole population across the primary and secondary constraints."_. Maybe I misunderstood the code, so could you please double check that we're the samples are not just proportional for the secondary constraints, but even for the primary ones?


> ETA: on second thought, I decided to keep the buildings that we've already labelled to include in the sample. I can then assign them back to their original labels which we can then review as we do the manual labelling. The reason I decided to keep them is because removing them changes the proportions of the groups in the whole population, and it affects the smaller groups the most. My proposal is therefore to retain them in the whole population and run the sampling. Then we conduct training on the new sample, but exclude the remaining labelled buildings from the first iteration so as not to over-index on Plymouth buildings. We can then think more carefully about how to combine the old labelled data in a way that doesn't skew the model. Let me know if you don't think that's a problem as I might be overthinking it!

I agree with this.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
crispy-wonton: No, we only need this df to calculate the proportions for the secondary constraints. We use the `grouped_df` to calculate the proportions for the primary strata for the whole population.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
crispy-wonton: This function is only required to calculate the proportions for secondary constraints. The primary strata proportions are calculated differently because they are not binary, but instead have multiple categories (at least some of them do). Primary strata proportions are calculated in lines 200-206. The calculation is nested in an `else` statement so will only trigger when `even=False` in the `calculate_array_sample_allocations` function.

So in conclusion: if `even=True`, primary strata are sampled evenly and secondary constraints are sampled to match 50% of the sample population. If `even=False`, primary strata are sampled in accordance with their population proportions, and so are secondary constraints.

In line 829, you can see I have set `even=False` for creating the test set sample, meaning both primary strata and secondary constraints are proportional to the whole population.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
crispy-wonton: Yes it's taken in one go :) much easier and faster this way.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -616,32 +768,34 @@ def parse_arguments() -> argparse.Namespace:
...
-    save_utils.save_to_s3(
-        df=_save_buildings_gdf,
-        path="s3://asf-local-heat-planning-tool/outputs/models/block_of_flats_classifier/GB_enriched_buildings_with_flats_and_geometries.parquet",
-    )
-    del _save_buildings_gdf
+        _save_buildings_gdf = buildings_gdf[["ID", "geometry"]].merge(
+            buildings_df.to_pandas(), how="inner", left_on="ID", right_on="building_id"
+        )
+        save_utils.save_to_s3(
+            df=buildings_gdf,
```
caldwellst: This should be saving `_save_buildings_gdf`?

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
caldwellst: I also find this confusing, as you need to check the implementation to understand exactly how this functions. Please could you update the docstring just so its clear this is only intended for secondary constraints?

caldwellst: Might consider just removing this function entirely? It's only used once, and the docstring is longer than the code, which would be clearer if we didn't have the function abstraction I think.

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
```diff
@@ -669,88 +823,96 @@ def parse_arguments() -> argparse.Namespace:
...
+        ),
+        quota_col="n_to_sample_test",
+        id_col="building_id",
+        attributes=attributes,
+        seed=seed,
+    ).with_columns(pl.lit("test").alias("split"))
+
+    # Join labels from first labelling round where label is confident
+    already_labelled = pl.read_parquet(
+        "s3://asf-local-heat-planning-tool/outputs/models/block_of_flats_classifier/first_round_confident_LABELLED_buildings_containing_flats_sample_n806.parquet"
```
caldwellst: This is hardcoded unlike other paths, maybe change.

caldwellst (changes requested):
I think there is an error in the saving function (wrong file saved?). Other than that, just minor suggestions and this should be good to go!

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
crispy-wonton: Have gone with the option to remove the function :)

asf_heat_pump_suitability/pipeline/model/block_of_flats/sample_for_labelling.py
crispy-wonton: Thank you! Have somehow messed this up about 3 times now 😩 should be good now

